PR #170, opened 2015-03-11
Glpk

macklin:
Adding an interface for GLPK (called through CVXOPT).

CVXOPT must be installed with GLPK support which can be done with:

```bash
CVXOPT_BUILD_GLPK=1 CVXOPT_GLPK_LIB_DIR=/path/to/glpk-4.55/lib CVXOPT_GLPK_INC_DIR=/path/to/glpk-4.55/include pip install cvxopt
```

Added a test for GLPK output and put it in `test_solvers.py`--wasn't sure if this was the right location

Had to modify `test_verbose()` in `test_problem.py` so that it would capture GLPK's output (which is not under the control of python's `sys.stdout`).  It's a bit of a hack, but it seems to work.

### DIFF
--- a/cvxpy/__init__.py
+++ b/cvxpy/__init__.py
@@ -24,7 +24,7 @@ from cvxpy.expressions.constants import Parameter, CallbackParam, Constant
 from cvxpy.problems.problem import Problem
 from cvxpy.problems.objective import Maximize, Minimize
 from cvxpy.error import SolverError
-from cvxpy.settings import (CVXOPT, ECOS, ECOS_BB, SCS,
+from cvxpy.settings import (CVXOPT, CVXOPT_GLPK, ECOS, ECOS_BB, SCS,
 OPTIMAL, UNBOUNDED, INFEASIBLE, SOLVER_ERROR,
 OPTIMAL_INACCURATE, UNBOUNDED_INACCURATE, INFEASIBLE_INACCURATE)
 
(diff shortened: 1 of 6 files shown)


SteveDiamond:
This is great! What's the best way to install GLPK on OS X and on ubuntu?

macklin:
I most recently [downloaded](https://www.gnu.org/software/glpk/) and tested it on our Centos server from a source install (the usual `configure`, `make`, `make install`).

While Ubuntu does have a `glpk` package that you can `apt-get install`/`aptitude install`, it's version 4.45, which, based on the dates [here](http://ftp.gnu.org/gnu/glpk/) was released in December 2010.  [Version 4.55](http://ftp.gnu.org/gnu/glpk/glpk-4.55.tar.gz) ([gpg signature](http://ftp.gnu.org/gnu/glpk/glpk-4.55.tar.gz.sig)) was released in August 2014.  That's why I do the installation from source.

Would you like step-by-step instructions on Ubuntu for the installation from source?

I'm not sure about OS X--I primarily use Linux.

SteveDiamond:
That's fine, no need for instructions. I was just wondering if there was a
packaged version I should know about.

SteveDiamond:
I merged the PR in by hand and made some changes. The user facing interface is the same as before, except the key is GLPK instead of CVXOPT_GLPK.

I've also updated the documentation at cvxpy.org.

GLPK doesn't seem to work for problems with redundant constraints, but that's alright for the moment since one can always try GLPK first then switch to a different solver.

macklin:
Great, thanks!

macklin:
There's no `glpk_intf.py` solver on master.  Did it not get added to the commit?

SteveDiamond:
Yeah, I forgot to add it before I committed. It's there now.

Issue #168, opened 2015-03-07
Feature request: Gurobi interface


macklin:
Would it be possible to add a Gurobi interface?  The problems that I work with (variations on flux balance analysis) can't be solved by most convex solvers due to (what I think are) numerical issues--the numerical values in real-life problems span 7-8 orders of magnitude and I always get "optimal inaccurate" solutions.  If I use Gurobi, though, I do get optimal solutions.  I would, however, like to be able to call Gurobi using cvxpy's interface if possible.  How feasible would it be to implement this?

SteveDiamond:
Is your problem an LP or a QP? cvxpy can't target QP solvers right now. It converts QPs to SOCPs.

Adding Gurobi as an LP solver would not be very hard. Adding closed source solvers creates a few complications though. One is testing. cvxpy heavily uses continuous integration with Travis CI. You can't test closed source solvers on Travis though. Does anyone have suggestions for testing closed source solvers?

If you have time, it would be great if you could add Gurobi to cvxpy yourself. All you need to do is create a "gurobi_intf.py" file in the style of 

https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/problems/solvers/ecos_intf.py

Even if you just make an attempt it would be helpful, and I can finish it up. Basically you just need code to take an LP or MILP stored as

min c^Tx
subject to Ax = b
                 Gx <= h

and export it to Gurobi.

macklin:
Right now my problem is an LP (minimizing an L1 norm subject to constraints).  All of the solvers that cvxpy currently targets failed to give an optimal accurate solution.

I'll start with a GLPK interface (through CVXOPT) since that should be a fairly easy modification from the current CVXOPT interface.  Then I'll try a Gurobi interface.

It'd be nice to be able to target a QP solver at some point--I will eventually play with using an L2 rather than L1 norm, and it's be great to do that within cvxpy if possible.  How hard would this be to incorporate?  Significantly more difficult than just another interface?

SteveDiamond:
The problem is that you can't convert an arbitrary SOCP into a QP or QCQP.
cvxpy would need to recognize problems that are QPs or QCQPs but not SOCPs.
This wouldn't be super hard, but I would prefer to avoid the complexity and
just add a mixed integer SOCP solver like MOSEK. Such a solver could handle
all mixed integer QPs or QCQPs.

SteveDiamond:
[...]

PR #171, opened 2015-03-18
GUROBI_LIN

macklin:
This partially addresses issue #168 by adding a Gurobi interface to solve LPs.  A couple of tests are included.

Still to do (ordering based on my priorities):

1. Add warm start capability.  After reading the `scs_intf.py` code, I have a good idea how to do this.  However, I have a couple questions:

    A. I see some tests in `test_solvers.py`, for example `test_ecos_options`, which set the `warm_start` option.  I don't see any sort of assertion that warm starting is actually working, though.  Is there a way to assert this?

    B. For performance reasons, I would prefer to sparsely update constraints which have changed from the previous run.  Is there a way to get that information through CVXPY somehow?  Or do I need to compute that myself?

2. Add capability for other classes of problems (this is a much lower priority for me than warm start).  If someone can describe and point me to (or write) tests that should pass with Gurobi for other classes of problems, I can try and write the code in the interface.  I'm definitely open to contributions from other people to get this working.

In terms of Travis integration, I'm not sure what the best thing to do is.  Gurobi does offer free academic licenses.


macklin:
Commit 38592c3 adds a test to catch a bug I had in 63876dc: Gurobi's default lower bound on a decision variable is zero.  The bug is fixed in c6e4f08.

macklin:
Commit e58b969 has warm starting capabilities where I assume that CVXPY does not provide me any information on constraints that have changed.  While I've included some tests for the output (more tests should definitely be written), none of them assert that Gurobi is actually warm starting internally (I'm not timing anything).

SteveDiamond:
This is phenomenal! I emailed the developers of JuMP and mathprogbase to see if they had any ideas about testing Gurobi in travis. I'll look into this myself also.

You're right that warm start isn't really being tested. One idea I had was to see if a solver converges in one iteration when it's warm started with the solution (or less robustly, whether the warm started solve is faster than the original solve). I didn't get too far with this because there's something wrong with how cvxpy is warm starting SCS.

It looks like you found a good solution for sparsely updating the Gurobi constraints. You could potentially use Parameters for this too. Could you give me an example snippet from your code where you're updating constraints at the CVXPY level?

It looks like Gurobi can solve SOCPs and mixed-integer SOCPs, which would also capture QPs and mixed integer QPs: 
https://github.com/JuliaOpt/gurobi.jl
That would be really useful!

macklin:
I have some tests in `test_gurobi_lin_warm_start()` in `test_solvers.py` that demonstrate updating constraints at the CVXPY level using Parameters.  Do Parameters provide information about values that have been updated (something that I could use in my code rather than calculating changes myself)?

SteveDiamond:
Right now what Parameters provide is a way for cvxpy to know which values it can cache and which might change across solves. So the assumption is that any values derived from a Parameter may have changed. There's no way right now to see which Parameters in particular have changed. This might be worth adding.

To see if a constraint contains parameters, you can use this snippet:
```
len(lu.get_expr_params(constr.expr)) > 0
```

Do you need the ability to change all parts of your problem efficiently? Or would it be enough to use Parameters to isolate a subset of values that might change and update them on every solve?

I'm not a huge fan of the "update_objective", "update_eq_constr", etc. solver options, since (if I understand correctly) they may cause cvxpy to solve a different problem then is currently stored in the Parameters.

I heard back from the JuMP/mathprogbase developers, and they didn't have any ideas for doing CI with Gurobi. I've emailed Gurobi support to see if we can get a special license.

macklin:
I could definitely construct my problems to only use Parameters for values which might change.  I agree that that'd be a cleaner approach than the `update_*` solver options that I currently have.

Using the snippet you provided, I can figure out if a constraint contains Parameters.  But then I need to know where that constraint gets mapped in `A`, `b`, `G`, or `h`.  Is there a way to do that?

SteveDiamond:
You can currently get the matrix entries associated with parameterized constraints. Here's code to do it:
```
# Inside the solver interface.
# Get the data relating to the problem matrices.
matrix_data = self.get_matrix_data(objective, constraints, cached_data)

# Check if the objective must change.
if len(lu.get_expr_params(objective)) > 0:
   Redo the objective because it's all parameterized.

# Get the parameterized part of the equality constraints.
eq_cache = matrix_data.get_param_cache(matrix_data.eq_cache)
eq_cache.coo_tup # A (V, I, J) triplet for the parameterized matrix.
eq_cache.const_vec # The parameterized vector offset.

# Get the parameterized part of the inequality constraints.
ineq_cache = matrix_data.get_param_cache(matrix_data.ineq_cache)
ineq_cache.coo_tup # A (V, I, J) triplet for the parameterized matrix.
ineq_cache.const_vec # The parameterized vector offset.
```
This gives V, I, J triplet for all the parameterized entries in the different matrices. It also gives the parameterized vector offset. The problem is that in the vector offset zero-valued entries could be parameters that evaluated to 0 or from the initialization of the vector. So you can't tell which part of the vector offset was updated by the parameters.

I tried storing the vector offset as a V, I, J triplet, but that made cvxpy substantially slower.

SteveDiamond:
I heard back from Gurobi, and they aren't willing to give a license for us to use on travis CI. Apparently all their licenses are either tied to a machine or a LAN. I'm alright with not testing the Gurobi integration in travis. At least we'll know the Gurobi code doesn't break anything else.

I'm away from Stanford right now, so I can't use Gurobi. I'll test this PR when I get back to campus.

macklin:
Haven't had a chance to incorporate the Parameters yet. Hopefully by the end of this week.

SteveDiamond:
I merged in the PR by hand. cvxpy now supports gurobi! Thanks for doing such a great job putting this together. 

The only thing I changed is that I removed the update_* keywords and instead had cvxpy assume that part of the problem had changed if and only if it contains parameters. Philosophically I'm against telling cvxpy to solve a different problem then you've specified.

I'll work on adding MI and SOCP support through gurobi.

macklin:
Ok, thanks!  Sorry I didn't get around to implementing those changes myself--haven't had a chance to work on it recently.

PR #174, opened 2015-04-25
get autodoc to work on readthedocs.org

ajfriend:
This should allow the docs to build with autodoc on readthedocs.org.
In the readthedocs configuration page, change the location of the requirements file to `doc/requirements.txt`. This seemed like a better place for it, since the requirements are specifically for the documentation.


SteveDiamond:
Great! Thanks for doing this!

PR #176, opened 2015-04-27
geo_mean

ajfriend:
Add geo_mean atom and documentation.


cvxpy/atoms/geo_mean.py
```diff
@@ -0,0 +1,842 @@
...
+        constraints = []
+
+        vars_ = [index.get_index(arg_objs[0], [], i, 0) for i in range(len(w))] + [d[w_dyad]]
+        for i, (p, v) in enumerate(zip(w_dyad, vars_)):
+            if p > 0:
+                tmp = [0]*len(w_dyad)
+                tmp[i] = 1
+                d[tuple(tmp)] = v
+
+        for elem, children in tree.iteritems():
```
SteveDiamond: Use tree.items(), not iteritems.

cvxpy/tests/test_constant_atoms.py
```diff
@@ -146,9 +146,11 @@
                   [8, 16, 2, 4],
                   [5, 2, 5, 2],
                   [2, 4, 2, 4]] ], Constant([7.7424])),
-        (geo_mean, (1, 1), [4,1], Constant([2])),
-        # (geo_mean, (1, 1), [0,7], Constant([0])),
-        (geo_mean, (3, 2), [ [[2,63,3], [1,1,2]], [[8,7,3],[10,4,2]] ], Constant([[4,21,3],[math.sqrt(10),2,2]])),
+        (geo_mean, (1, 1), [[4, 1]], Constant([2])),
```
SteveDiamond: Let's add some weighted geo_mean tests. Replace geo_mean with "lambda x: geo_mean(x, [weights])"

SteveDiamond:
This looks great! Python 3 compatibility is handled automatically by 2to3. Just fix the things that break travis, or test it in a python 3 environment.

I'll merge this once it passes travis.

ajfriend:
Weird. It doesn't even look like that latest travis failure is connected to the geo_mean code.

ajfriend:
Oh, I see where its coming from now. I'll fix it in a bit.

SteveDiamond:
Is this ready to merge?

ajfriend:
yup!

SteveDiamond:
Did you change something about how the function names in the functions table display? long names like 
"mul_elemwise" wrap now, whereas they didn't before.

ajfriend:
I did. In `cvxpy_alabaster.css`, I swapped in `wrap` for `nowrap`. Without it, the parameters section on the geo_mean page wouldn't wrap, and would go off the edge of the screen. I didn't notice the change
on the table until now, tho. Sorry about that. 

Is there an RST markup to wrap the function names in the table so that they don't get broken over lines?

SteveDiamond:
Not that I ever found. I spent a long time trying to format the table when
I first made the website. I don't remember the details of what I did, but
the main takeaway was that it's borderline impossible except through global
CSS changes. There just aren't adequate formatting options in RST.

ajfriend:
I just tried this out: http://stackoverflow.com/questions/11830242/non-breaking-space

Check out ajfriend/cvxpy@f94205282bab0c8fbc9fa3eeb2692361345d089f

It seems to work. What do you think?

Another option would be to use latex. something like
``:math:`\text{lambda_sum_smallest(X, k)}`` , which seems to avoid the line wrap.

SteveDiamond:
That worked, thanks!

PR #177, opened 2015-05-07
power atom

ajfriend:
Adds power.py atom. If everything looks good, I can start replacing atoms like `sqrt`, `square`, `inv`, etc with wrappers around `power`.


SteveDiamond:
This looks great!

PR #178, opened 2015-05-08
Add p-norm

ajfriend:
This adds `pnorm.py`, which computes the vector norm of a vector or matrix (effectively concatenating columns of a matrix to form a vector) for p >= 1 or p = 'inf'.

I also changed the code of `norm.py`, `norm1.py`, `norm2.py`, and `norm_inf.py` to use `pnorm` whenever possible.


SteveDiamond:
Looks good!

PR #179, opened 2015-05-08
replace square with power.py equivalent and fix huber

ajfriend:
On a personal note, I'm sad that my attempts to sneak relative imports into the codebase have been thwarted :(


SteveDiamond:
Relative imports are bad style right? That's what pylint tells me anyway.

ajfriend:
I think old style relative imports (`import submodule`) are bad style since there's ambiguity about whether you're actually doing a relative or absolute import, depending on the directory structure, but that functionality has been disabled in python3, so it wouldn't work anyway. New style, python 3, relative imports (`from . import submodule`) are considered good style, I think, since that ambiguity is removed.

And who could disagree with David Beazley? :)
He talks about the debate a bit in a recent PyCon talk:
https://youtu.be/0oTh1CXRaQ0?t=17m

In general, I think it just makes the imports less verbose.

PR #180, opened 2015-05-12
Add pnorm with p < 1

ajfriend:
Adds a concave (but not actually a norm) branch of `pnorm` for values of `p < 1`.


SteveDiamond:
Look great!

PR #182, opened 2015-05-21
Readme: syntax highlighting

scottsievert:
Just a quick PR to add syntax highlighting to the readme.

Instead of

    ```
    from cvxpy import *
    # ...
    ```

We have 

    ```python
    from cvxpy import *
    # ...
    ```


SteveDiamond:
Thanks!

PR #196, opened 2015-06-15
Support adding Problem instances

tridao:


cvxpy/problems/problem.py
```diff
@@ -404,3 +404,28 @@ def __str__(self):
     def __repr__(self):
         return "Problem(%s, %s)" % (repr(self.objective),
                                     repr(self.constraints))
+
+    def __add__(self, other):
+        if not isinstance(other, Problem):
+            raise TypeError("Can only add Problem instances.")
```
SteveDiamond: You should ``return NotImplemented`` instead of raising a TypeError here.

cvxpy/problems/problem.py
```diff
@@ -404,3 +404,28 @@ def __str__(self):
...
+                           list(set(self.constraints + other.constraints)))
+        else:
+            raise TypeError(
+                "Can only add Problems that are all Minimize or Maximize.")
+
+    def __radd__(self, other):
+        if other == 0:
+            return self
+        else:
+            raise TypeError("Can only add Problem instances.")
```
SteveDiamond: You should return NotImplemented instead of raising a TypeError here.

tridao: You mean for both Min + Max and __radd__(), or just __radd__()?

SteveDiamond: Not for Min + Max. Just the two places I listed. You could raise a DCP
error instead of a TypeError for Min + Max.

PR #200, opened 2015-06-30
Adding partial_optimize functionality, a NonNegative variable class, and a small change to the Variable class.

alnurali:


SteveDiamond:
Could you make this a pull request to the develop branch instead? Things have gotten too complicated (many pull requests and features in progress) to develop on the master branch.

SteveDiamond:
Also, it may take me a little while to merge this in. I may want to implement partial_optimize in a different way. I need to think about it.

I don't like changing deepcopy. It would be much better to just copy the LinOp tree manually. I'll help you with this on the cvxstoch repo.

PR #201, opened 2015-07-02
CVXCanon 

jpmpentwater:
Modifications to matrix stuffing code to use CVXcanon library
https://github.com/jacklzhu/CVXcanon

Currently, the modifications do not support the use of parameters. 

The CVXcanon library depends on SWIG for bindings to C++. Currently, this dependency is not automatically installed by CVXcanon. However, if this is an issue, we can add that to the setup.py script in CVXcanon.

### DIFF
--- a/cvxpy/atoms/affine/add_expr.py
+++ b/cvxpy/atoms/affine/add_expr.py
@@ -38,7 +38,6 @@ class AddExpression(AffAtom):
         self.args = []
         for group in arg_groups:
             self.args += self.expand_args(group)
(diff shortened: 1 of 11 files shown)


SteveDiamond:
You should add swig to setup.py. CVXPY needs to be installable via ``pip install``.

SteveDiamond:
Also, could you make this a pull request to the develop branch instead? Things have gotten too complicated (many pull requests and features in progress) to develop on the master branch.

PR #202, opened 2015-07-03
CVXCanon 

jpmpentwater:
Modifications needed to use CVXcanon for matrix stuffing.

Still need to fix the dependency on swig, but wanted to get the merge process started. 

Also, CVXcanon now also has continuous integration working with Travis-CI.


SteveDiamond:
I see you can't pip install swig. The key then is going to be to compile distributions of CVXcancon for different platforms and use pip wheels.

SteveDiamond:
I had time recently to try out the new version of cvxcanon. I almost have cvxpy integrated with cvxcanon, but I'm running into weird segfaults. Here's a code segment that segfaults for me on the current develop branch:
```
from cvxpy import *
C = Variable(3, 3)
obj = Maximize(C[0, 2])
constraints = [diag(C) == 1,
               C[0, 1] == 0.6,
               C[1, 2] == -0.3,
               C == C.T,
               C >> 0]
prob = Problem(obj, constraints)
result = prob.solve(solver=SCS)
```
Could you figure out why this is segfaulting in CVXcanon? The code calling cvxcanon is in the ``_lin_matrix`` function in ``matrix_data.py``.

The ``C >> 0`` notation means that C is positive semidefinite but not necessarily symmetric.

jpmpentwater:
I spent a while looking into this, and I'm still not 100% sure where the issue is. Is there another example where the code is segfaulting? The behavior is non-deterministic for me.

Also, the program you gave works without a problem if you use variable **C = Semidef(3,3)** instead of **C>>0** and **C = C.T**. Shouldn't the representation for both problems be identical?

Is it possible that there is some inconsistency in **semidefinite.py** or in the canonical form output for the **>>** operator?

SteveDiamond:
I don't have any other examples where it consistently segfaults. This example always segfaults for me.

``C = Semidef(3,3)`` is not exactly the same as ``C >> 0`` and ``C == C.T``. ``C = Semidef(3, 3)`` creates a variable in R^6 which is mapped onto a symmetric matrix in R^{3x3} by viewing the variable in R^6 as the matrix's upper triangular entries. ``C >> 0`` imposes the constraint that ``(C + C.T)/2`` is positive semidefinite.

Here's what I get when I run the example under lldb:

```
(develop)➜  cvxpy git:(develop) ✗ lldb python ../scripts/cvxcanon_segfault.py
(lldb) target create "python"
Current executable set to 'python' (x86_64).
(lldb) settings set -- target.run-args  "../scripts/cvxcanon_segfault.py"
(lldb) r
Process 20065 launched: '/Users/stevend2/anaconda/envs/develop/bin/python' (x86_64)
Process 20065 stopped
* thread #1: tid = 0xab481, 0x0000000104474132 _CVXcanon.so`Eigen::internal::conservative_sparse_sparse_product_selector<Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, 0, 0, 0>::run(Eigen::SparseMatrix<double, 0, int> const&, Eigen::SparseMatrix<double, 0, int> const&, Eigen::SparseMatrix<double, 0, int>&) + 3778, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10c36e040)
    frame #0: 0x0000000104474132 _CVXcanon.so`Eigen::internal::conservative_sparse_sparse_product_selector<Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, 0, 0, 0>::run(Eigen::SparseMatrix<double, 0, int> const&, Eigen::SparseMatrix<double, 0, int> const&, Eigen::SparseMatrix<double, 0, int>&) + 3778
_CVXcanon.so`Eigen::internal::conservative_sparse_sparse_product_selector<Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, 0, 0, 0>::run:
->  0x104474132 <+3778>: movq   (%r9,%rdi,8), %rsi
    0x104474136 <+3782>: movzbl %dl, %edx
    0x104474139 <+3785>: btq    %rdx, %rsi
    0x10447413d <+3789>: jae    0x104474160               ; <+3824>
```

SteveDiamond:
Actually, here's another script that fails:
```
from cvxpy import *
C = Symmetric(2, 2)
obj = Maximize(C[0, 0])
constraints = [C << [[2,0], [0, 2]]]
prob = Problem(obj, constraints)
result = prob.solve(solver=SCS)
```

jpmpentwater:
I successfully resolved the issue, which had to do with a problem in the MUL linop when the data matrix is a scalar. In this case, cvxpy (and now CVXcanon) relies on broadcasting rather than constructing a matrix of the stated sized.  I'm not sure why this issue only surfaced in conjunction with the ```>>''' operator. 

All of the tests pass now and the code successfully handles the cases you specified along with a few other ones I constructed to test this.

I just put a new source distribution and wheel on pypi to reflect the changes in the code.

SteveDiamond:
The new cvxcanon isn't installing for me. When I do pip install I get a compilation error:
```
(develop)➜  cvxpy git:(develop) ✗ pip install CVXcanon          
Collecting CVXcanon
  Using cached CVXcanon-0.0.13.tar.gz
Requirement already satisfied (use --upgrade to upgrade): numpy in /Users/stevend2/anaconda/envs/develop/lib/python2.7/site-packages (from CVXcanon)
Installing collected packages: CVXcanon
  Running setup.py install for CVXcanon
    Complete output from command /Users/stevend2/anaconda/envs/develop/bin/python -c "import setuptools, tokenize;__file__='/private/var/folders/gb/15j5cwts2lscjq9301sc9hx80000gn/T/pip-build-JDQb6P/CVXcanon/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /var/folders/gb/15j5cwts2lscjq9301sc9hx80000gn/T/pip-QNsQm0-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_py
    creating build
    creating build/lib.macosx-10.5-x86_64-2.7
    copying src/python/canonInterface.py -> build/lib.macosx-10.5-x86_64-2.7
    copying src/python/CVXcanon.py -> build/lib.macosx-10.5-x86_64-2.7
    running build_ext
    building '_CVXcanon' extension
    swigging src/python/CVXcanon.i to src/python/CVXcanon_wrap.cpp
    swig -python -c++ -outdir src/python -I./src/ -I./src/python/ -o src/python/CVXcanon_wrap.cpp src/python/CVXcanon.i
    creating build/temp.macosx-10.5-x86_64-2.7
    creating build/temp.macosx-10.5-x86_64-2.7/src
    creating build/temp.macosx-10.5-x86_64-2.7/src/python
    gcc -fno-strict-aliasing -I/Users/stevend2/anaconda/envs/develop/include -arch x86_64 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -Isrc/ -Isrc/python/ -Iinclude/Eigen -I/Users/stevend2/anaconda/envs/develop/include/python2.7 -c src/python/CVXcanon_wrap.cpp -o build/temp.macosx-10.5-x86_64-2.7/src/python/CVXcanon_wrap.o
    In file included from src/python/CVXcanon_wrap.cpp:3072:
    In file included from src/CVXcanon.hpp:20:
    In file included from src/LinOp.hpp:22:
    src/Utils.hpp:18:10: fatal error: '../include/Eigen/Sparse' file not found
    #include "../include/Eigen/Sparse"
             ^
    1 error generated.
    error: command 'gcc' failed with exit status 1
```

jpmpentwater:
Sorry for the delay. I was out of town for the past week. I fixed the compilation error, put a new distribution on pypi, and verified that installation works on both mac and my linux machine.

In general, it seems like the SWIG compilation step is pretty clunky. Do you think it would be better/more robust to just write a cython wrapper around the C++ portion of CVXcanon? As a bonus, this would also remove the potential CVXPY swig dependency on Linux.

SteveDiamond:
It seems like I need SWIG to install CVXcanon on my mac. Ideally I we wouldn't have SWIG as a dependency. Writing a cython wrapper would be a good way to eliminate this dependency.

Does the cvxpy develop branch pass its tests for you? When I updated CVXcanon a lot of the tests started failing. I know the problem is in CVXcanon because when I revert to the old code for matrix stuffing all the tests pass. It's possible I'm using the CVXcanon interface wrong in ``problems/problem_data/matrix_data.py``.

jpmpentwater:
I just pushed a minor change, and all of the tests on the develop branch are passing for me now. The pypi distributions have also been updated. 

I don't think SWIG should be needed to install CVXcanon on Mac since there is a Mac wheel on pypi. However, I agree that the SWIG dependency is suboptimal. I've started working on a cython wrapper, and I'll definitely update you when its finished or if I run into any major problems.

SteveDiamond:
Are you sure CVXcanon installs on travis? I get this CVXcanon installation error on travis when I listed CVXcanon as a cvxpy dependency:
```
AttributeError: 'dict' object has no attribute '__NUMPY_SETUP__'
```
See https://travis-ci.org/cvxgrp/cvxpy/jobs/80059387. I even installed swig.

I looked at the CVXcanon repo, and there haven't been any changes in 2 months. Could you try updating it and making sure it builds?

jpmpentwater:
CVXcanon installs on travis (and passes all of the cvxpy tests) when installed directly using ```pip install CVXcanon```. See https://travis-ci.org/jacklzhu/CVXcanon/jobs/80064759. 

However, it looks like there is some discrepancy when CVXcanon is listed in the ```install_requires``` list in cvxpy. 

I'm now trying to figure out what is generating the installation error when CVXcanon is a dependency instead of directly installed via ```pip```. I was previously under the impression that these were identical.

jpmpentwater:
The key to getting the install to work was switching to using binary eggs. This works on both Mac and Linux distributions and does not require swig to install. 

See https://travis-ci.org/jacklzhu/CVXcanon/builds/80068741. CVXcanon is installed as a dependency without swig, and all of the cvxpy unit tests pass.

SteveDiamond:
Thanks for figuring this out! Travis builds now for Python 2, but the Python 3 builds are still failing with the same error message (https://travis-ci.org/cvxgrp/cvxpy/jobs/80117780).

jpmpentwater:
Thanks for pointing that out! I just put binary eggs for python 3 (both mac and linux) on pypi. This seems to have fixed the issue for the python 3 builds (https://travis-ci.org/jacklzhu/CVXcanon/builds/80270610).

SteveDiamond:
CVXcanon is now a dependency for the main branch of cvxpy. Everyone who installs cvxpy from now on will be using it. Thanks so much for all your great work! 

I hadn't realized pip wheels are such a nice distribution mechanism. I'm glad you're working on a cython wrapper though.

PR #211, opened 2015-08-04
Initial Mosek interface (SOCP only). With tests.

enzobusseti:
I tested it and it works. 

Install Mosek https://www.mosek.com/resources/downloads and get a free academic license.

So far I only saw improved performance (over both CVXOPT and ECOS).


SteveDiamond:
looks great!

enzobusseti:
Thanks! 

next steps:

- integer programming (should be fairly straightforward)
- SDP (I still have to understand a couple of things, but I don't think it will be more than another 50 lines)
- solver options (I'm using mosek for work right now, will figure something out).

Enzo

PR #213, opened 2015-08-06
overloaded multiplication for problems and objectives

mseesquared:
Added unit tests.

### DIFF
--- a/cvxpy/tests/test_problem.py
+++ b/cvxpy/tests/test_problem.py
@@ -352,12 +352,33 @@ class TestProblem(BaseTest):
 
     # Test problem multiplication by scalar
     def test_mul_problems(self):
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_problem.py
```diff
@@ -350,6 +350,15 @@ def test_add_problems(self):
             prob_bad_sum = prob1 + prob3
         self.assertEqual(str(cm.exception), "Problem does not follow DCP rules.")
 
+    # Test problem multiplication by scalar
+    def test_mul_problems(self):
+        prob1 = Problem(Minimize(cvxpy.pow(self.a)), [self.a >= 2])
```
SteveDiamond: just write ``pow(self.a)`` not ``cvxpy.pow(self.a)``.

SteveDiamond:
There's a bug in your test.

PR #214, opened 2015-08-10
Fixed unit tests for linear combinations of problems

mseesquared:


SteveDiamond:
You need to add a ``__truediv__`` method for Python 3.

cvxpy/problems/objective.py
```diff
@@ -55,6 +60,39 @@ def __radd__(self, other):
...
+            return Maximize(self.args[0] * other)
+
+    __rmul__ = __mul__
+
+    def __div__(self, other):
+        if not isinstance(other, (int, float)):
+            return NotImplemented
+        return self * (1.0/other)
+
+    def __truediv__(self, other):
```
SteveDiamond: You should reuse ``__div__`` instead of copying the code, e.g., __truediv__ = __div__.

PR #215, opened 2015-08-11
Implement copy function for each atom

tridao:
This function allows one to create a (shallow) copy of an atom, with an optional ARGS parameters to change the args of the atom.

This commit includes a a generic `copy()` function for all atoms and 4 specific `copy()` functions for AddExpression, index, power, and geo_mean.

### DIFF
--- a/cvxpy/atoms/affine/add_expr.py
+++ b/cvxpy/atoms/affine/add_expr.py
@@ -59,6 +59,28 @@ class AddExpression(AffAtom):
     def numeric(self, values):
         return reduce(op.add, values)
 
+    # As __init__ takes in the arg_groups instead of args, we need a special
+    # copy() function.
+    def copy(self, args=None):
+        """Returns a shallow copy of the AddExpression atom.
+
+        Parameters
+        ----------
+        args : list, optional
+            The arguments to reconstruct the atom. If args=None, use the
+            current args of the atom.
+
+        Returns
+        -------
+        AddExpression atom
+        """
+        if args is None:
+            args = self.args
+        # Avoid calling __init__() directly as we do not have the args_groups.
+        copy = type(self).__new__(type(self))
+        super(type(copy), copy).__init__(*args)
+        return copy
+
     @staticmethod
     def graph_implementation(arg_objs, size, data=None):
         """Sum the linear expressions.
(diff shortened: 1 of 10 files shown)


tridao:
I've also added `copy()` for objective and eq/leq_constraint. 

However, not all objects in cvxpy can have `copy()`, as many objects don't have the `args` attribute (`BoolConstr`, for example).

cvxpy/atoms/affine/add_expr.py
```diff
@@ -59,6 +59,28 @@ def name(self):
     def numeric(self, values):
         return reduce(op.add, values)
 
+    # As __init__ takes in the arg_groups instead of args, we need a special
+    # copy() function.
+    def copy(self, args=None):
```
SteveDiamond: Is this special copy necessary? Why won't ``__init__(*args)`` work?

cvxpy/atoms/affine/index.py
```diff
@@ -50,6 +50,26 @@ def get_data(self):
         """
         return self.key
 
+    # As get_data(self) returns a tuple, the function atom.copy() is confused.
+    # We have to implement another copy() function here.
+    def copy(self, args=None):
```
SteveDiamond: Is this copy necessary? It looks like the default copy in atom will do the same thing.

cvxpy/problems/objective.py
```diff
@@ -75,6 +75,23 @@ def is_dcp(self):
...
+
+        Parameters
+        ----------
+        args : list, optional
+            The arguments to reconstruct the objective. If args=None, use the
+            current args of the objective.
+
+        Returns
+        -------
+        Constraint
```
SteveDiamond: Returns an objective.

cvxpy/atoms/affine/add_expr.py
tridao: The `__init__()` of `AddExpression` takes a list that's not unpacked, so we can't do `__init__(*args)__`.

cvxpy/atoms/affine/index.py
tridao: As `index.get_data(self)` returns a tuple `(row slice, column slice)`, the default copy in atom will convert it to a list to unpack it. It will then call `__init__()` with these as two arguments, but `index.__init__()` requires `key` to be passed as one argument (as a tuple).

cvxpy/problems/objective.py
tridao: Fixed.

SteveDiamond:
Looks good to me. Thanks for answering my questions.

PR #219, opened 2015-08-19
Change MatrixCache.F to allow pickling solved problems.

tridao:
Previously we couldn't pickle problems that are already solved, as `MatrixCache.F` is a function that cannot be pickled. I changed `MatrixCache.F` to be a function object (thanks to `functools.partial()`) to fix this. This should allow pickling and sending solved problems through `multiprocessing`.


SteveDiamond:
looks good.

PR #224, opened 2015-09-04
Implement solving separable problems in parallel

tridao:
If problem is separable, we split it into smaller problems and solve them in parallel.
To enable the option, run `problem.solve(parallel=True)`.

I cache the separable problems (as a list) by decorating `problem._separable_problems` with `lazyprop`. I'm still not sure how I should detect if the objective or constraint has changed to recompute these separable problems.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -20,20 +20,24 @@ along with CVXPY.  If not, see <http://www.gnu.org/licenses/>.
 import cvxpy.settings as s
 import cvxpy.utilities as u
 import cvxpy.interface as intf
-from cvxpy.expressions import types
 from cvxpy.error import SolverError, DCPError
 from cvxpy.constraints import EqConstraint, LeqConstraint, PSDConstraint
 from cvxpy.problems.objective import Minimize, Maximize
 from cvxpy.problems.solvers.solver import Solver
 from cvxpy.problems.solvers.utilities import SOLVERS
 from cvxpy.problems.problem_data.problem_data import ProblemData
-from cvxpy.utilities import performance_utils as pu
+# Only need to import cvxpy.transform.get_separable_problems, but this creates
+# a circular import (cvxpy.transforms imports Problem). Hence we need to import
+# cvxpy here.
+import cvxpy
 
-import itertools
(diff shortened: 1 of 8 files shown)


cvxpy/problems/problem.py
```diff
@@ -187,8 +197,75 @@ def get_problem_data(self, solver):
...
+        problem_list = []
+        for index in range(num_components):
+            terms = [obj_terms[i] for i in term_ids_per_subproblem[index]
+                     if i < num_obj_terms]
+            # If we just call sum, we'll have an extra 0 in the objective.
+            obj = sum(terms[1:], terms[0]) if terms else 0
+            constrs = [constraints[i - num_obj_terms]
+                       for i in term_ids_per_subproblem[index]
+                       if i >= num_obj_terms]
+            problem_list.append(Problem(type(self.objective)(obj), constrs))
```
SteveDiamond: You can now use ``self.objective.copy(obj)``, which I like better.

SteveDiamond:
This is really excellent code! Sorry for not getting to it sooner.

You should make sure the cache is invalidated if the user changes the objective or constraints. This is done currently in ``problems/solvers/solver.py`` in the ``validate_cache`` function.

You could factor some version of this function out into ``solvers/utilities`` and reuse it.

tridao:
Sorry for taking so long to fix this PR. I've added code to check if the problem has changed before solving in parallel. However, I didn't use the `validate_cache` function since it only works if the cache is of type `SymData`. I just keep a cache for the parallel solver consisting of the objective and constraints (as LinOps).

I ran into some problem with CVXcanon version 0.0.21, but upgrading to version 0.0.22 fixes it. So I'm also adding the requirement of CVXcanon >= 0.0.22.

After putting `get_separable_problems` into `cvxpy.transforms`, I ran into circular imports. `Problem` needs to import `cvxpy.transforms`, and `cvxpy.transforms` needs to import `Problem`. I resolved it by just importing `cvxpy` in `problem.py`. I'm not sure if this is the best way to solve it, but I couldn't think of anything else.

SteveDiamond:
This is beautiful code! Committing now, and sorry for the delay.

PR #229, opened 2015-09-27
add logistic() atom

mwytock:


SteveDiamond:
Looks good! Thanks for doing this.

argriffing:
```python
    @Elementwise.numpy_numeric
    def numeric(self, values):
        """Evaluates e^x elementwise, adds 1, and takes the log.
        """
        return np.log(1 + np.exp(values[0]))
```

Numerically this would be better as `return np.logaddexp(0, values[0])`.

```python
>>> x = np.array([-500, -10, 0, 10, 500, 2000000])
>>> np.log(1 + np.exp(x))
__main__:1: RuntimeWarning: overflow encountered in exp
array([  0.00000000e+00,   4.53988992e-05,   6.93147181e-01,
         1.00000454e+01,   5.00000000e+02,              inf])
>>> np.log1p(np.exp(x))
array([  7.12457641e-218,   4.53988992e-005,   6.93147181e-001,
         1.00000454e+001,   5.00000000e+002,               inf])
>>> np.logaddexp(0, x)
array([  7.12457641e-218,   4.53988992e-005,   6.93147181e-001,
         1.00000454e+001,   5.00000000e+002,   2.00000000e+006])
```

mwytock:
thanks, that makes sense

PR #240, opened 2015-10-17
fixed way SDP constraints were enforced in MOSEK + test

enzobusseti:
The nosetest fails on one test. I tried to replicate it but can't find where in the tests it is. Can you give me the code that generates the error? I'll figure out what's wrong.

I have preliminary performance tests :)
10x10 SDP 
  MOSEK 300ms
  CVXOPT 2sec
20x20 SDP
  MOSEK 2sec
  CVXOPT 40sec

### DIFF
--- a/cvxpy/constraints/semidefinite.py
+++ b/cvxpy/constraints/semidefinite.py
@@ -59,7 +59,7 @@ class SDP(Constraint):
         solver : str
             The solver being called.
         """
-        if solver.name() == s.CVXOPT:
+        if solver.name() in [s.CVXOPT, s.MOSEK]:
             new_eq_constr, new_leq_constr = self.__CVXOPT_format
         elif solver.name() == s.SCS:
             new_eq_constr, new_leq_constr = self.__SCS_format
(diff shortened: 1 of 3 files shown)


SteveDiamond:
The test is failing because of a "NEAR_OPTIMAL" status. It's this problem
```
from cvxpy import *
X = Variable(3,3)
z = Variable(3)
constr = [X == [[1, 0, 0],
                             [0, 1, 0],
                             [0, 0, 1]],
         z == [1, 2, 3]]
prob = Problem(Minimize(matrix_frac(z, X)[0,0]), constr)
prob.solve(solver=MOSEK)
```
but it succeeds when I run it outside of nosetests.

enzobusseti:
I tried to run it, I do get an "optimal_inaccurate" status. Just to be clear, what "optimal_inaccurate" means? I thought it means that the problem is solved but the solver is not sure whether the solution is accurate. If it is a failure status I'll simply output "optimal" whenever I'm outputting "optimal_inaccurate"

SteveDiamond:
optimal inaccurate means the solver solved the problem to a lower accuracy than it normally demands, but still a reasonable accuracy. Keep outputting optimal inaccurate. I'd just change the line
```
if problem.status == OPTIMAL:
```
to
```
if problem.status in [OPTIMAL, OPTIMAL_INACCURATE]:
```

enzobusseti:
Sounds good. You mean in test_constant_atoms?

SteveDiamond:
Yeah, in test_constant_atoms

SteveDiamond:
Everything passes for me! I'm merging it in. Thanks again for doing this!

PR #246, opened 2015-11-28
in parade route example, store solution as `xsol`

jwkvam:
The parade route example doesn't run the last cell correctly since `xsol` isn't defined. This simply fixes that. Now the notebook runs without error for me.

I suppose to verify, you can download and run the file.
https://raw.githubusercontent.com/jwkvam/cvxpy/fix-parade-route/examples/notebooks/WWW/parade_route.ipynb


SteveDiamond:
Thanks for fixing this!

PR #252, opened 2016-01-08
generalize matrix_frac to accept matrix in first arg

mwytock:
Hi Steven,

This form comes up from time to time (see e.g. http://www.cs.cmu.edu/~mwytock/papers/gcrf_full.pdf, eq. 3) and should generalize the column vector case without a problem. 

Cheers,
Matt


SteveDiamond:
Interesting! Looks good.

PR #254, opened 2016-01-12
GLPK_MI setting solver params

### LINKED ISSUE #251: CVXOPT GLPK params not passing through
When using the GLPK and GLPK_MI solvers with cvxpy, setting solver parameters such as
```
prob.solve(solver=cvx.GLPK_MI, verbose=False, tmlim=60)
```
do not get passed through to the solver. The simpliest example is setting verbose=False on any problem, you will see that all outputs are still printed to stdout. 

To fix this I went into the specific solver interface file: cvxpy/problems/solvers/glpk_mi_intf.py and set the glpk options directly instead of using the generic cvx solvers options

```python
if verbose:
    #cvxopt.solvers.options["msg_lev"] = "GLP_MSG_ON"
    cvxopt.glpk.options["msg_lev"] = "GLP_MSG_ON"
else:
    #cvxopt.solvers.options["msg_lev"] = "GLP_MSG_OFF"
    cvxopt.glpk.options["msg_lev"] = "GLP_MSG_OFF"
```
This is a workaround, but there should be a better fix to use the generic cvxopt solver options. 

For reference I am using cvxpy 0.3.4 and cvxopt 1.1.8

cmale:
Resolves #251 for GLPK_MI. Params need to be explicitly set to cvxopt.glpk.options rather than cvxopt.solvers.options since glpk.ilp is not part of cvxopt LP solvers.

### DIFF
--- a/cvxpy/problems/solvers/glpk_mi_intf.py
+++ b/cvxpy/problems/solvers/glpk_mi_intf.py
@@ -73,19 +73,19 @@ class GLPK_MI(GLPK):
         import cvxopt, cvxopt.glpk
         data = self.get_problem_data(objective, constraints, cached_data)
         # Save original cvxopt solver options.
-        old_options = cvxopt.solvers.options.copy()
+        old_options = cvxopt.glpk.options.copy()
         # Silence cvxopt if verbose is False.
         if verbose:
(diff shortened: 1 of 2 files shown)


SteveDiamond:
Looks good! Thanks!

PR #256, opened 2016-01-16
add axis parameter to pnorm(x, 2)

mwytock:
Hi Steven,

This makes it much easier to write sum-of-norms regularization e.g. without unrolling the sum (for example, network lasso paper you dont need |E| terms, http://web.stanford.edu/~hallac/Network_Lasso.pdf)

This is consistent with how linalg.norm() works, although i just added the code for p=2 for now.

Cheers,
Matt

### DIFF
--- a/cvxpy/atoms/pnorm.py
+++ b/cvxpy/atoms/pnorm.py
@@ -134,7 +134,17 @@ class pnorm(AxisAtom):
         if self.p < 0 and np.any(values == 0):
             return 0.0
 
-        return np.linalg.norm(values, float(self.p), axis=self.axis, keepdims=True)
+        retval = np.linalg.norm(values, float(self.p), axis=self.axis)
+
+        # NOTE: workaround for NumPy <=1.9 and no keepdims for norm()
+        if self.axis is not None:
+            if self.axis == 0:
+                retval = np.reshape(retval, (1, self.args[0].size[1]))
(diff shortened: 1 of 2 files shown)


SteveDiamond:
Looks good to me. I've been meaning to add this for a while. Does it require NumPy 1.10? There was a NumPy error in the Travis tests (which use NumPy 1.9).

At some point we should add a version of ``SOC_Elemwise`` that can take a matrix for ``X`` and an axis argument, so everything is vectorized.

mwytock:
looks like keepdims was added to linalg.norm in 1.10 so i think its needed
(i believe weird things happen if numeric() collapses its return value to a
1D array)

mwytock:
let me know how you want me to resolve this... if 1.9 support is important than i think we could do something manually reshape to the right dimension

SteveDiamond:
1.9 support isn't that important, but this seems like too small a thing to
require 1.10 for. Could you see if there's an easy fix for 1.9?

mwytock:
sure thing, thats fixed now ... fixing one more python 3 bug (i typed xrange for some reason...)

agree re: SOC_Elemwise taking matrix args that would be nice here

mwytock:
fyi - green now :)

SteveDiamond:
Great, merging now.

PR #259, opened 2016-02-05
fix log_sum_exp with axis and non-square matrix

mwytock:


SteveDiamond:
Oops. Glad you caught and fixed this.

PR #260, opened 2016-02-08
Cbc (Clp, Cgl) interface for LP/MIP (using cylp)

sschnug:
Hi Steven,

i implemented a basic Cbc-interface for cvxpy, which uses cylp to use Cbc/Clp/Cgl. I think cvxpy should support these solvers because they are probably the best non-commercial ones in their domain, especially regarding MIP-solving.
Thanks to cylp, users can add Cut-Generators too (like Gomory-Cuts; uses Cgl).

This commit includes:
- Cbc interface & basic changes to cvxpy so that Cbc is recognized
- A small MIP-example which solves the (symmetric-)TSP problem (where ecos_bb is struggling hard; GLPK_MI seems to be okay, but Cbc + some cuts should be much more powerful)
- Reflections in documentation regarding installing and usage of the Cbc-interface
- Tests

Missing:
- [x] Tests
- [ ] Additional solver-options like desired MIP-gap, Time-limit and co. (supported by cylp)
- [ ] Allow finer granularity regarding output
- [ ] Allow to add limits to cut-generation
- [ ] Warm-start capabilities
- [ ] Extend to support quadratic objectives or even SOCP; maybe too much changes in cvxpy needed; there is some crazy stuff happening in JuMP.jl
- [x] Optional importing of cut-generators (needed for making cbc available to cvxpy for now)

Things which should be checked:
- Status handling / solver-result implications

Rough edges in code:
- ~~Objective function needs CyLPArray in expression (which reflect some cylp examples). Not sure, why a numpy-array is not working (it is for sparse-matrix algebra in other parts of the code)~~
- ~~Very ugly handling of solver_options: maybe i'm doing something wrong here~~

Other things to think about:
- cvxpy tries to use ECOS_BB on the tsp_mip example (which is not working great) even if GLPK_MI is available -> maybe it would be better to change the solver-selection a bit (Cbc would be an even better choice; but the interface might not be mature enough)

Sascha


cvxpy/problems/solvers/cbc_intf.py
```diff
@@ -0,0 +1,285 @@
...
+                     'stopped due to errors': s.SOLVER_ERROR,
+                     'stopped by event handler (virtual int ' \
+                                    'ClpEventHandler::event())': s.SOLVER_ERROR}
+
+    def name(self):
+        """The name of the solver.
+        """
+        return s.CBC
+
+    def import_solver(self):
```
SteveDiamond: The purpose of import_solver is really just to test whether the solver exists. So you probably don't need to import so many things.

cvxpy/problems/solvers/cbc_intf.py
```diff
@@ -0,0 +1,285 @@
...
+        # Build model & solve
+        status = None
+        if self.is_mip(data):
+            cbcModel = model.getCbcModel()  # need to convert model
+            if not verbose:
+                cbcModel.logLevel = 0
+
+            # Add cuts if desired
+            if "solver_opts" in solver_opts:
+                if "GomoryCuts" in solver_opts["solver_opts"]:
```
SteveDiamond: solver_opts["solver_opts"] shouldn't be a dict. People should just call solve with extra keyword args. Like this:
```
prob.solve(solver=CBC, GomoryCuts=True)
```

cvxpy/problems/solvers/cbc_intf.py
```diff
@@ -0,0 +1,285 @@
...
+                    from cylp.cy.CyCgl import CyCglGomory
+                    gom = CyCglGomory()
+                    cbcModel.addCutGenerator(gom, name="Gomory")
+                if "MIRCuts" in solver_opts["solver_opts"]:
```
SteveDiamond: You should at least break out all these if statements into a separate function that adds the cut generators. Ideally it would be shorter, but I'm not sure how to do that.

doc/source/tutorial/advanced/index.rst
```diff
@@ -450,6 +457,66 @@ Here's the complete list of solver options.
     whether to initialize the solver with the previous solution (default: False).
     The use case for warm start is solving the same problem for multiple values of a parameter.
 
+`CBC`_ options:
+
+Cut-Generation through `CGL`_
+
+General remark: these Cut-Generators will generate noisy output even if ``'verbose=False'``
```
SteveDiamond: Is there any way to make the output silent?

doc/source/tutorial/advanced/index.rst
```diff
@@ -450,6 +457,66 @@ Here's the complete list of solver options.
     whether to initialize the solver with the previous solution (default: False).
     The use case for warm start is solving the same problem for multiple values of a parameter.
 
+`CBC`_ options:
+
+Cut-Generation through `CGL`_
+
+General remark: these Cut-Generators will generate noisy output even if ``'verbose=False'`` 
+
+``'GomoryCuts'``
```
SteveDiamond: Can you organize these into something more compact? Also it should be keyword args like ``GomoryCuts=True``.

SteveDiamond:
This is fantastic! I had no idea there was a Python Cbc interface.

The solver selection right now only considers the default solvers. It does make sense to consider the other solvers as well, like you suggest, though that could get complicated.

Does Cbc support QPs and SOCPs? The JuliaOpt website says it only supports MILPs.

The code overall looks really good. I made some comments. I'll commit this after you add some tests. You can see ``cvxpy/tests/test_elemental.py`` as an example of a test for an optional solver.

It would be good also to add Cbc installation in ``continuous_integration/install.sh`` for travis, so Travis can run the Cbc tests. Currently it runs two installations, one with only ECOS, CVXOPT, and SCS and the other with GLPK in addition. So you would add Cbc to the GLPK installation.

cvxpy/problems/solvers/cbc_intf.py
sschnug: I was just afraid of an cylp installation process without Cgl beeing shipped and later used within cvxpy. But i will think about some safe design regarding this.

doc/source/tutorial/advanced/index.rst
sschnug: I didn't find something in the code of cylp, but i will ask / post an issue there.

sschnug:
**QP/SOCP:**
Cbc/Clp seems to support quadratic objectives (and maybe quadratic constraints) but i don't know how mature/supported this functionality is. There is not much documentation.
In Cbc.jl (which is used in JuMP.jl) Cbc is defined as LinearQuadratic solver (a class from MathProg.jl) and the last commit in Cbc.jl allows Cbc.jl used for Conic problems (supporting some cones, they are transforming the problem though). Maybe there is a lot to learn from these JuliaOpt projects.

I will have a look at your comments and try to clean up the code, implement tests and look at the installation process for travis (where i don't have any experience at all). Thanks for your comments.

sschnug:
I think i now patched in all the request you had (solver_opts handling; docs shortening, tests, cut-gen refactoring), except for the following:
- Some cut-generators are still noisy (no progress on the linked issue @cylp; but there is a reaction of one active developer)
- Travis-CI integration is not working
 - (1) I'm not able to install cbc/cylp correctly; the error with the current state of the code is ```continuous_integration/install.sh: line 71: cd: cvxgrp/cvxpy: No such file or directory``` which is due to some directory changes i needed to compile cbc correctly (it looks ugly: if i don't cd before cbc, the cbc makefile will compile into cvxgrp/cvxpy folder which is messy). Maybe you got some ideas... I'm not familiar with the Travis-CI system-architecture.
 - (2) Additionally the tests with GLPK=False always fail, because CBC won't be installed then. I'm not sure if CBC-test is wrong (try...except) or if the GLPK-solver-check-tests are somehow disabled when GLPK=False.

(I'm using the six lib for pleasant iterating over dictionaries when py2 + py3 should be supported. This is reflected in setup.py dependencies. This could be approached differently)

cvxpy/tests/test_cbc.py
```diff
@@ -0,0 +1,145 @@
...
+            prob.solve(verbose=True, solver=CBC)
+            self.assertAlmostEqual(prob.value, z)  # objective
+        else:
+            with self.assertRaises(Exception) as cm:
+                prob = Problem(Minimize(norm(self.x, 1)), [self.x == 0])
+                prob.solve(solver=CBC, verbose=False)
+            self.assertEqual(str(cm.exception), "The solver %s is not installed." % CBC)
+
+
+    def test_options(self):
```
SteveDiamond: You need to check if CBC is installed here. The idea is none of the CBC tests will run if CBC isn't installed.

continuous_integration/install.sh
```diff
@@ -46,6 +46,29 @@ if [[ "$DISTRIB" == "conda" ]]; then
...
+
+        # sudo apt-get install coinor-libcbc-dev coinor-libcbc0 coinor-libcbc-doc
+
+        # Install cyLP -> which is needed for CBC-interface
+        cd /home/travis
+        git clone -b py3 https://github.com/jjhelmus/CyLP.git  # use custom-branch because of py3
+        cd CyLP
+        python setup.py install
+        
+        cd cvxgrp/cvxpy
```
SteveDiamond: Maybe this should be ``/home/travis/cvxgrp/cvxpy``?

SteveDiamond:
This all looks great! We're very close. I pointed out a few things to change that may cause the errors you mentioned. I debug travis by running the install script on a new linux VM/container, so I'd suggest doing that. It looks to me like a minor error with ``cd``.

I'm fine with adding six as a dependency, but you should check if 2to3 handles things correctly without it.

sschnug:
Continuous-Integration (Travis-CI and Appveyor) is working now. I still think that the Travis-CI behaviour made no sense at all (i tested at least 15 different configs; your proposition didn't work too) and the current install.sh is not as elegant as before. But it works...

Another puzzling thing is the setup.py script including 2to3. It should be able to convert my dict.iteritems(), but it doesn't. I even tried different `use_2to3_fixers` but i had no luck (the cli tool should be able to fix dict-access without explicit activation of 'dict' fixer; but well...; the documentation of 2to3 is quiet bad too) and switched back to `six`.

Feel free to post other changes you want to see.
Maybe there is a need to discuss some things about later changes (MIP-gap-param, Time-limit-params, Warm-starts) because some of them could need some changes in cvxpy (e.g.: more solver-states).

SteveDiamond:
Everything looks great to me! Thanks for doing such a great job with this! I'm going to merge it. 

We can add additional features later. What new solver states would you suggest?

PR #271, opened 2016-03-24
Initial Implementation of nosetests from Python Interpreter

halwai:


SteveDiamond:
Sorry for sitting on this for so long. I looked more into how NumPy and SciPy implement this, and they actually have deprecated this method of running nosetests. I don't think it's a good idea any more to add this feature to cvxpy. @halwai I really appreciate your work on this and I'm sorry about not merging it in.

PR #282, opened 2016-05-15
quad_form to allow indefinite and asymmetric matrices

jaehyunp:
Some changes to quad_form and related test scripts.

1. Allow asymmetric P: This change is not necessary, but since ``x^T P x`` is a scalar anyway, the specification is unambiguous even when P is asymmetric. This is handled by setting ``P = (P+P^T)/2`` internally and should not touch the original data.
2. Allow indefinite P, but with a warning message: Currently quad_form rejects P with both positive and negative eigenvalues. The behavior of quad_form doesn't change when P is positive or negative semidefinite. With the proposed change, users can use indefinite P, but will get a warning message when that happens. quad_form will then return a difference of convex function. More specifically, when ``P = M1*M1^T - M2*M2^T``, ``quad_form(x, P)`` returns ``sum_squares(M1^T * x) - sum_squares(M2^T * x)``.

### DIFF
--- a/cvxpy/atoms/quad_form.py
+++ b/cvxpy/atoms/quad_form.py
@@ -18,6 +18,7 @@ along with CVXPY.  If not, see <http://www.gnu.org/licenses/>.
 """
 
 import cvxpy.interface as intf
+import warnings
 from cvxpy.expressions.expression import Expression
 from cvxpy.expressions.constants import Constant
 from .sum_squares import sum_squares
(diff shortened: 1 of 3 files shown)


SteveDiamond:
Could you look into why the travis build failed? Travis basically follows the continuous_integration/install.sh script on a clean Ubuntu 12 system. It looks good to merge once we figure out the Travis issue.

jaehyunp:
I haven't check every single one of them yet, but most of them are CVXOPT errors that have nothing to do with these changes. These errors exist in the master branch too---could you check?

PR #284, opened 2016-06-18
addition of the LS solver

jaehyunp:
This change set is not big in terms of the code length, but can potentially affect a lot of applications, considering how widespread least-squares problems are. There are two main changes to the codes:

**(1) Addition of the LS solver**
This specialized solver uses (for now) the SciPy sparse linear algebra interface to solve linearly constrained least squares problems. It cannot handle any general cone problem. When a suitable problem is found, it completely bypasses the canonicalization step, forms the KKT system, and then invokes the SciPy linear algebra solve method. The solver interface was kept as similar to that of other solvers as possible.

**(2) Addition of the QuadCoeffExtractor utility class**
This utility class provides an interface for extracting the coefficients of any quadratic expression. It takes a variable-to-offset map and a CVXPY expression (quadratic in the variables it involves) as input. The returned value is a triple (Ps, Q, R), where Ps is a list of SciPy sparse matrices, Q is a SciPy sparse matrix, R is a NumPy array. See the code for the detailed input/output specification. This utility class is to be used in the QCQP extension package as well.


SteveDiamond:
This looks great! I'll merge it in.

PR #285, opened 2016-06-24
Add support for tuning MOSEK optimizer parameters

nkoep:
I recently ran into an issue where cvxpy would return infeasible solutions despite reporting an optimal solution status when using MOSEK. Since I wanted to play around with the optimizer parameters, I checked the cvxpy docs and realized there was currently no way to override MOSEK's default parameters, so I hacked up a quick patch to add support for doing so.

### DIFF
--- a/cvxpy/problems/solvers/mosek_intf.py
+++ b/cvxpy/problems/solvers/mosek_intf.py
@@ -130,7 +130,12 @@ class MOSEK(Solver):
         env = mosek.Env()
         task = env.Task(0, 0)
 
-        self._handle_mosek_params(task, solver_opts.get("mosek_params"))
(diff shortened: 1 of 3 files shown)


SteveDiamond:
This looks good! Did you fix the optimal -> infeasible issue too? Could you add a test with MOSEK parameters to cvxpy/tests/test_solvers.py?

nkoep:
I'll add a test case tomorrow. I haven't had time to look further into the infeasibility issue yet. Will hopefully do that tomorrow as well. In case I can't figure it out I'll post a minimal working example to see if you guys  have an idea what's going on with the problem.

SteveDiamond:
Alright, this looks good to me.

PR #287, opened 2016-07-19
Adding SizeMetrics object and constants() method for getting info about problems

nishi951:
Added a SizeMetrics object that serves as an attribute of Problems and contains information about the number of scalar variables, scalar eq and leq constraints, data values, and the maximum dimension of any data block. It is instantiated when the problem is instantiated.

Added a constants() method that functions mostly the same as the variables() and parameters() methods by recursively finding the Constant objects in the objective and constraints of the problem.

### DIFF
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -209,6 +209,22 @@ class Atom(Expression):
         # Remove duplicates.
         return list(set(param_list))
 
+    def constants(self):
+        """Returns all the constants present in the arguments.
+        """
+        const_list = []
+        const_dict = {}
+        for arg in self.args:
+            const_list += arg.constants()
+        # Remove duplicates:
+        try:
+            const_dict = {str(constant): constant for constant in const_list}
+        except:
+            print const_list
+        return const_dict.values()
+
+
+
     @property
     def value(self):
         # Catch the case when the expression is known to be
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -217,13 +217,8 @@ class Atom(Expression):
         for arg in self.args:
             const_list += arg.constants()
         # Remove duplicates:
-        try:
-            const_dict = {str(constant): constant for constant in const_list}
-        except:
-            print const_list
+        const_dict = {id(constant): constant for constant in const_list}
         return const_dict.values()
-
-
 
     @property
     def value(self):
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -218,7 +218,7 @@ class Atom(Expression):
             const_list += arg.constants()
         # Remove duplicates:
         const_dict = {id(constant): constant for constant in const_list}
-        return const_dict.values()
+        return list(const_dict.values())
 
     @property
     def value(self):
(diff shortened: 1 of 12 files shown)


cvxpy/atoms/atom.py
```diff
@@ -209,6 +209,22 @@ def parameters(self):
...
+        """
+        const_list = []
+        const_dict = {}
+        for arg in self.args:
+            const_list += arg.constants()
+        # Remove duplicates:
+        try:
+            const_dict = {str(constant): constant for constant in const_list}
+        except:
+            print const_list
```
SteveDiamond: Why is this a try catch?

cvxpy/problems/problem.py
```diff
@@ -143,6 +148,29 @@ def parameters(self):
...
+    def constants(self):
+        """Returns a list of the constants in the problem.
+        """
+        const_dict = {}
+        constants_ = self.objective.constants()
+        for constr in self.constraints:
+            constants_ += constr.constants()
+        # Remove duplicates.
+        # Note that numpy matrices are not hashable, so we use the buildin function id
+        try:
```
SteveDiamond: Why is this a try catch?

cvxpy/problems/problem.py
```diff
@@ -288,7 +317,7 @@ def _solve(self,
             solver.validate_solver(constraints)
         else:
             raise SolverError("Unknown solver.")
-
+
```
SteveDiamond: What's up with the spacing here? Is this a tab? No tabs allowed! :)

cvxpy/problems/problem.py
```diff
@@ -574,3 +602,91 @@ def __div__(self, other):
...
+            if constraint.__class__.__name__ is "EqConstraint":
+                self._num_scalar_eq_constr += np.prod(constraint._expr.size)
+        
+        # num_scalar_leq_constr
+        self._num_scalar_leq_constr = 0
+        for constraint in problem.constraints:
+            if constraint.__class__.__name__ is "LeqConstraint":
+                self._num_scalar_leq_constr += np.prod(constraint._expr.size)
+
+    @property
```
SteveDiamond: You don't need these functions. num_scalar_variables etc can just be raw fields.

SteveDiamond:
Looks good overall. I had a few comments. Also it breaks on python 3. The failure on windows in appveyor also happens for the main cvxpy branch, though if you could figure that one out it would be a huge help.

cvxpy/atoms/atom.py
nishi951: Whoops, forgot to get rid of all of my debugging stuff. Numpy matrices aren't hashable so I tried to work around that for a while.

Incidentally, it's also written this way in problem.py so I'm changing it there too.

cvxpy/problems/problem.py
nishi951: Same as above, Numpy matrices aren't hashable so I had some errors to debug.

cvxpy/problems/problem.py
nishi951: Ah, sorry my sublime was configured wrong for a while! I must have missed that one.

SteveDiamond:
There's an issue in python 3 on travis. You should test it also in Python 3 locally. This is easy, you just make a new conda environment with Python 3.

PR #291, opened 2016-08-03
use "with" for mosek task and env to prevent memory leaks

mwytock:
current version leaks memory, see
http://docs.mosek.com/7.0/pythonapi/Efficiency_considerations.html
for details.


mwytock:
Note this actually makes the solving with mosek somewhat slower presumably due to clean up. This could likely be improved by reusing env and task, something like env as global/class/threadlocal variable and task a problem-specific variable... but I'm not 100% sure what is appropriate here.

SteveDiamond:
Thanks for figuring this out!

PR #300, opened 2016-10-11
I've added is_pwl() for expressions and is_qp() for problems

takapoui:


SteveDiamond:
All the code looks good, but you're not catching the case where the objective is quadratic + pwl.

SteveDiamond:
I'm merging this in and adding support for mixed quadratic/PWL objectives later.

PR #308, opened 2016-11-04
Qp parsing compiling

enzobusseti:
it just demonstrative now (since we don't have a QP solver yet) but it's already functional. it checks if a problem is a valid QP based on me and nicholas' criterion, and compiles it in QP form

### DIFF
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -19,6 +19,7 @@ along with CVXPY.  If not, see <http://www.gnu.org/licenses/>.
 
 
 from .. import utilities as u
+from ..error import QPError
 from .. import interface as intf
 from ..expressions.constants import Constant, CallbackParam
 from ..expressions.expression import Expression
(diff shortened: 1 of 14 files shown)


cvxpy/atoms/atom.py
```diff
@@ -172,6 +173,29 @@ def canonicalize(self):
                                                                 data)
             return (graph_obj, constraints + graph_constr)
 
+    def QP_canonicalize(self):
```
SteveDiamond: I would segregate the QP canonicalization into a file in the utilities folder, like @jaehyunp did with LS. This is a hack and everything will be improved during the hackathon rewrite, but that's what I'd prefer for now. The same goes for all the QP_canonicalize functions in the atoms.

SteveDiamond:
This is a great start! The CI is failing so you should look into that. Also I would prefer that the QP canonicalization be separated from the atoms, like the quadratic canonicalization.

enzobusseti:
@SteveDiamond the CI failing is not my code's fault, look at the commit history. i haven't looked too deep into the logs but it seems to be an ECOS error

cvxpy/atoms/atom.py
enzobusseti: we should talk in person about it. i'm not sure @jaehyunp approach of segregating the code for LS canonicalization is optimal. if you look at my code, the canonicalization is already functional with minimal code changes (~100 lines), reusing the cone canonicalization code. it's really really similar (conic and qp canonicalization), with only two atoms needing a specialized function (quad_over_lin and pow)

SteveDiamond:
The current version of master passes the CI (https://travis-ci.org/cvxgrp/cvxpy/builds/172748571), so somehow merging in your PR is breaking the build. If you look at the actual error message (https://travis-ci.org/cvxgrp/cvxpy/jobs/173148993) it's clearly because some atoms are missing a QP_canonicalize function.

I guess the QP_canonicalize functions can stay in the atoms. I'm very concerned about how adding more types of canonicalization is straining the design. I guess it's fine for now but do consider that all problem transformations have to be rewritten as reductions for cvxpy 1.0.

enzobusseti:
ops you're right i hadn't seen the correct CI stuff. will fix it. and yeah, i guess my point is about minimizing code replication. QP canonicalization is the same as conic canonicalization, with the only exception that we catch quadratic forms and stuff them in the objective. all the rest of the atoms do epigraph transformations in exactly the same way

SteveDiamond:
This is all in cvxpy 1.0.

PR #317, opened 2017-01-06
callback params now store the atom they represent, not just cb function

moehle:
Doesn't change much, but is helpful for codegen.


SteveDiamond:
This fails in CI but it's not due to anything in the commit.

PR #324, opened 2017-02-08
example: consensus ADMM: add missing variables

dplarson:
The consensus ADMM example was missing two variables: gamma (for
weighting the L_1 norm) and NUM_PROCS (the number of threads of the
multiprocessing pool).

Related: since I've noticed multiple examples with similar issues (i.e. missing problem variables), would the maintainers prefer a bulk PR that attempts to update all the example scripts? Or one PR per example?

### DIFF
--- a/examples/consensus_admm.py
+++ b/examples/consensus_admm.py
@@ -5,6 +5,8 @@ from multiprocessing import Pool
 # Problem data.
 m = 100
 n = 75
+gamma = 1.0
+NUM_PROCS = 2
 np.random.seed(1)
 A = np.random.randn(m, n)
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Good catch! Thanks!

dplarson:
@SteveDiamond Since there appear to be similar issues (i.e. missing variables) in other examples, would you prefer a single PR that fixes all the issues or one PR per example?

SteveDiamond:
A single PR would be best. Thanks for fixing these mistakes!

dplarson:
Happy to help!

PR #328, opened 2017-02-14
fix simplex in mosek

goriri:
WARNING: code from python novice. Trying to fix format_results exception when using mosek with simplex method.

### DIFF
--- a/cvxpy/problems/solvers/mosek_intf.py
+++ b/cvxpy/problems/solvers/mosek_intf.py
@@ -157,7 +157,7 @@ class MOSEK(Solver):
                 # size of problem
                 numvar = len(c) + sum(dims[s.SOC_DIM])
                 numcon = len(b) + dims[s.LEQ_DIM] + sum(dims[s.SOC_DIM]) + \
-                         sum([el ** 2 for el in dims[s.SDP_DIM]])
+                    sum([el ** 2 for el in dims[s.SDP_DIM]])
 
                 # otherwise it crashes on empty probl.
                 if numvar == 0:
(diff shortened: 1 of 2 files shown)


cvxpy/problems/solvers/mosek_intf.py
```diff
@@ -273,8 +274,20 @@ def format_results(self, task, data, cached_data):
                       mosek.solsta.near_dual_infeas_cer: s.UNBOUNDED_INACCURATE,
                       mosek.solsta.unknown: s.SOLVER_ERROR}
 
-        task.getprosta(mosek.soltype.itr)  # unused
-        solsta = task.getsolsta(mosek.soltype.itr)
+        soltype = None
+        for sol in [mosek.soltype.itr, mosek.soltype.bas]:
```
SteveDiamond: What is this doing? Can you add comments?

SteveDiamond:
The code looks reasonable but I need more explanatory comments to be sure. Travis CI failed because of style errors. Run flake8 on your code and fix mosek_intf.py.

I'm having contributors sign a CLA now via this license/cla thing. It should be easy. Let me know if you have any problems.

cvxpy/problems/solvers/mosek_intf.py
goriri: For simplex solution, the mosek.soltype.itr is empty but the actual solution lies in mosek.soltype.bas. I cannot find a way to deduce the soltype here so trying to fall back to mosek.soltype.bas if mosek.soltype.itr is not working.

RhysU: This is not unlike https://github.com/cvxgrp/cvxpy/issues/335.

SteveDiamond:
Sorry for not responding for so long. #335 seems quite related, and I prefer the suggestion there to use  "task.solutiondef" rather than a try catch block. Could you update the pull request along those lines?

SteveDiamond:
This was resolved by PR #347

Issue #335, opened 2017-03-06
mosek.soltype.itr -> unknown but mosek.soltype.bas giving optimal


RhysU:
I've a case where, for `verbose=True`, Mosek 8.0 completes with
```
Interior-point solution summary
  Problem status  : UNKNOWN
  Solution status : UNKNOWN
  Primal.  obj: 5.0643588790e+05    nrm: 2e+08    Viol.  con: 3e-01    var: 0e+00  
  Dual.    obj: 5.0415624857e+05    nrm: 2e+05    Viol.  con: 5e-02    var: 4e-04  

Basic solution summary
  Problem status  : PRIMAL_AND_DUAL_FEASIBLE
  Solution status : OPTIMAL
  Primal.  obj: 4.9845334246e+05    nrm: 2e+08    Viol.  con: 2e-07    var: 0e+00  
  Dual.    obj: 4.9845334246e+05    nrm: 8e-01    Viol.  con: 3e-05    var: 6e-17  
```

Per [the Mosek interface](https://github.com/cvxgrp/cvxpy/blob/be172685c1e7198e14ded8361f0ac99c85b18a52/cvxpy/problems/solvers/mosek_intf.py#L277) this causes `task.getprosta(mosek.soltype.itr)` to return `unknown` while `task.getsolsta(mosek.soltype.bas)` reports `optimal`.  I see that [Mosek 7 has the same `soltype` enum](http://docs.mosek.com/7.0/pythonapi/Solution_types.html)-- I've not confirmed this outcome is version 8-specific but I doubt it given that documentation.

Is `mosek.soltype.bas` irrelevant or is not interrogating `soltype.bas` a shortcoming of the cvxpy interface?  Monkeying around a bit in the Python debugger suggests that I can walk through much of the remainder of that interface logic using `bas`.  That said, a simple `s/itr/bas/g` in `mosek_intf.py` does not permit running the front-page example at http://www.cvxpy.org/en/latest/.

Gross hackery in mosek_intf.py like...
```        # Retrieve basic solution if possible, otherwise fallback to interior
        try:
            task.getprosta(mosek.soltype.bas)  # unused
            solsta = task.getsolsta(mosek.soltype.bas)
            soltype = mosek.soltype.bas
        except:
            task.getprosta(mosek.soltype.itr)  # unused
            solsta = task.getsolsta(mosek.soltype.itr)
            soltype = mosek.soltype.itr
```
[...]

Issue #327, opened 2017-02-13
Using simplex method in mosek raises exception


goriri:
Format result raises an exception when using

`prob.solve(solver=MOSEK,mosek_params={mosek.iparam.optimizer: mosek.optimizertype.free_simplex},verbose=True)`
I guess the problem is with the line `task.getprosta(mosek.soltype.itr)` in mosek_intf.py line 279 where only mosek.soltype.bas is defined.
Using cvxpy version 0.4.8

SteveDiamond:
You seem to have a much better knowledge of the MOSEK interface and features than I do. Why don't you make a pull request to fix this? Remember to include a test in ``test_solvers.py``.

RhysU:
Has this been resolved?  Seems like the fallback in https://github.com/cvxgrp/cvxpy/blob/4bf33893a5d3eeaf1ae9bd0346f76f04c60d913e/cvxpy/problems/solvers/mosek_intf.py#L246 may have it covered.

PR #347, opened 2017-03-29
Basic solution in Mosek

aszekMosek:
An update to the Mosek interface which takes into account the basic solution ``mosek.soltype.bas`` in addition to ``mosek.soltype.itr``. Hopefully resolves issues #335, #327. Serves the same purpose as proposed PR #328.


SteveDiamond:
Looks good!

PR #362, opened 2017-06-12
adds missing connect method from MultiEdge

jwgwalton:
Example doesn't work without the connect method being overwritten to take into account the directed flow attributes of MultiEdge.

Example error

  File "./commodity_flow.py", line 70, in <module>
    edges[-1].connect(nodes[n1], nodes[n2])
  File "/home/jwgwalton/Documents/ConvexOptimisation/examples/cvxpy_graph_examples/max_flow.py", line 15, in connect
    in_node.edge_flows.append(-self.flow)
AttributeError: 'MultiEdge' object has no attribute 'flow'


SteveDiamond:
Looks good! Thanks for fixing this.

Issue #364, opened 2017-06-26
ECOS_BB mixed integer problem: ValueError: Invalid sign for Bool value


vodp:
I encounter the ValueError when trying to solve a mixed integer quadratic problem using ECOS_BB solver. Basically my problem is declared as follow

```
Alpha = cvx.Bool(len(alpha))
obj = cvx.Minimize(.5*cvx.norm(WX*Alpha-y,2) + theta*cvx.sum_entries(Alpha))
prob = cvx.Problem(obj)
prob.solve(solver=cvx.ECOS_BB, verbose=True)
```

After running a while, the errors show up. Any way to prevent this with some parameter settings?

```
ECOS 2.0.4 - (C) embotech GmbH, Zurich Switzerland, 2012-15. Web: www.embotech.com/ECOS

It     pcost       dcost      gap   pres   dres    k/t    mu     step   sigma     IR    |   BT
 0  +8.038e-01  +7.577e-01  +6e+02  8e-01  5e-03  1e+00  3e+00    ---    ---    1  1  - |  -  - 
 1  +4.411e+00  +4.418e+00  +1e+02  4e-01  8e-04  2e-01  5e-01  0.8213  6e-03   1  2  2 |  0  0
 2  +6.515e+00  +6.519e+00  +1e+01  5e-02  8e-05  2e-02  4e-02  0.9155  5e-03   2  2  2 |  0  0
 3  +8.496e+00  +8.563e+00  +7e+00  1e-01  2e-04  1e-01  3e-02  0.7385  6e-01   2  2  2 |  0  0
 4  +8.927e+00  +8.927e+00  +1e-01  3e-03  3e-06  1e-03  5e-04  0.9823  2e-04   2  1  2 |  0  0
 5  +8.953e+00  +8.953e+00  +2e-02  3e-04  4e-07  1e-04  7e-05  0.8704  6e-03   2  2  2 |  0  0
 6  +8.964e+00  +8.964e+00  +1e-03  3e-05  4e-08  4e-05  6e-06  0.9890  7e-02   2  2  2 |  0  0
 7  +8.965e+00  +8.965e+00  +2e-05  5e-07  6e-10  7e-07  9e-08  0.9835  1e-04   2  2  2 |  0  0
 8  +8.965e+00  +8.965e+00  +4e-07  1e-08  1e-11  2e-08  2e-09  0.9788  1e-04   2  2  2 |  0  0
[...]

PR #369, opened 2017-07-12
fixed issue 364

RafalZajac:
After upgrading cvxpy from version 0.3.4 to 0.4.10 the mixed integer optimisation using ECOS_BB seems to be failing with an error "ValueError: Invalid sign for Bool value". The problem is caused by the method is_positive() which now returns True. Bool is neither positive nor negative though. 

The method is_negative() override returns the same value as the base class version (Variable class) so it's redundant. 

Original issue:  https://github.com/cvxgrp/cvxpy/issues/364

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ env_*
 
 # pycharm
 .idea
+*.pyproj
--- a/cvxpy/expressions/variables/bool_var.py
+++ b/cvxpy/expressions/variables/bool_var.py
@@ -30,13 +30,3 @@ class Bool(Variable):
         """String to recreate the object.
         """
         return "Bool(%d, %d)" % self.size
-
-    def is_positive(self):
-        """Is the expression positive?
-        """
-        return True
-
-    def is_negative(self):
-        """Is the expression negative?
-        """
-        return False


RafalZajac:
The AppVeyor checks fail with the message "Install blas+lapack and re-compile SCS with blas+lapack libray locations". I'm not sure how to fix it :) 

I apologise if this pull request is causing any problems - it's my first ever pull request and I don't even use git as my primary source control...

SteveDiamond:
Appveyor has been failing forever. It's nothing you did! 

Thanks for this fix! The issue is actually a little more complicated, but this is a good fix for now. Basically the ideal fix would keep the fact that boolean variables are positive but round to {0,1} when they're assigned an optimal value.

RafalZajac:
I'm glad I could help! I'm not sure what the proper fix should be - i don't really know cvxpy code - but if the boolean variable method is_positive() returns True it suggests that it is "positive" in mathematical sense. That wouldn't be aligned with the mathematical nomenclature because a variable which takes values 0 or 1 wouldn't be called "positive" in maths, it would be called "non-negative".

Anyway - thanks for merging in my pull request. I look forward to new version version of cvxpy! :)

bsperisen:
Hi all, I just actually had the same issue, I think (i.e. Issue #364), since my code outputs "ValueError: Invalid sign for Bool value." When I comment out the is_positive() method in bool_var.py, as @RafalZajac suggested, the problem goes away (or at least the error doesn't come up anymore!).

I'm using the conda package, version 0.4.9-py27_0. Looking at the source code, the method simply returns True, so I'm guessing this fix is in the version I have but a problem is appearing in some other possibly subtle way?

Issue #368, opened 2017-07-09
Include name of variables in the repr.


musically-ut:
Currently, the `__repr__` implementation for `Variable`, `Bool`, `Int` classes, etc. only show the variable sizes and not the names:

```python
>> CVX.Variable(rows=2, cols=3, name='A')
Variable(2, 3)
```

Personally, I would like to see the `name` of the variable (if provided) as well. I would find that more informative and it can be very handy during debugging sessions.

This will also be compatible with the current documentation of `repr` string and I expect this not to be a breaking change.

I can work on a PR if this sounds like an enhancement `cvxpy` would be open to.

SteveDiamond:
I would be fine with adding the name to the repr for Variable. Feel free to make a PR.

musically-ut:
Was closed by #371.

PR #371, opened 2017-07-17
EHN Show names in repr, if provided.

musically-ut:
Addresses #368.

I've chosen to show the name of the variables only if it was provided while variable initialization; otherwise there is no benefit from seeing it.

Also, I've not update the `repr` for `Symmetric` and `Semidef` variables because their `repr` do not conform to the default pattern currently.

I'll be happy to make any stylistic changes.

### DIFF
--- a/cvxpy/expressions/variables/bool_var.py
+++ b/cvxpy/expressions/variables/bool_var.py
@@ -26,14 +26,6 @@ class Bool(Variable):
         obj, constr = super(Bool, self).canonicalize()
         return (obj, constr + [BoolConstr(obj)])
 
-    def __repr__(self):
-        """String to recreate the object.
-        """
-        if self._name_given:
-            return "Bool(%d, %d, '%s')" % (self._rows, self._cols, self._name)
-        else:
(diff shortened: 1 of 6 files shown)


musically-ut:
Umm... I'm not sure if the failures of the CI are related to this PR.

cvxpy/expressions/variables/bool_var.py
```diff
@@ -29,7 +29,10 @@ def canonicalize(self):
     def __repr__(self):
         """String to recreate the object.
         """
-        return "Bool(%d, %d)" % self.size
+        if self._name_given:
```
SteveDiamond: These ``__repr__`` methods are kind of redundant. Can you make the Variable ``__repr__`` be something like 
```
if self._name_given:
  return "%s(%d, %d, '%s')" % (type(self), self.size[0], self.size[1], self._name)
else:
  return "%s(%d, %d)" % (type(self), self.size[0], self.size[1])
```

musically-ut: Used `self.__class__.__name__` and have made the change.

PR #390, opened 2017-08-19
Added SuperSCS interface + changes to SizeMetrics

nishi951:


cvxpy/problems/solvers/superscs_intf.py
```diff
@@ -0,0 +1,183 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import cvxpy.settings as s
+from cvxpy.problems.solvers.ecos_intf import ECOS
+import numpy as np
+
+
+class SUPERSCS(ECOS):
```
SteveDiamond: Can't you use the SCS interface as a base class and have less replication?

nishi951:
Oh yeah that's clearly better.

Unrelated, but I also changed the SizeMetrics object to be a collections.namedtuple instead of a full-fledged object - this should make it more compact in memory and also a little easier to maintain, while essentially preserving all the original syntax.

cvxpy/problems/solvers/superscs_intf.py
nishi951: Yes definitely, didn't think of that...

SteveDiamond:
Looks good. Merged now.

PR #408, opened 2017-11-02
Added inaccurate statuses for OSQP

bstellato:
I have added "inaccurate" statuses for OSQP. They will be used in its next release.


SteveDiamond:
Thanks!

SteveDiamond:
Can you fix the flake8 issues?

bstellato:
It should be fixed now with flake8 3.5.0.

PR #431, opened 2018-01-22
Performance improvements for Gurobi

gglockner:
Based on a support case from a user, we at Gurobi are proposing these changes.  They seem to make a big improvement in Gurobi performance with cvxpy. This replaces a prior pull request, where the email address was not set correctly in my fork.


gglockner:
These CI tests seem unrelated to the pull request - can you recheck my pull request independently of these tests?

SteveDiamond:
Agreed, this failure has nothing to do with your commit. Thanks for adding this, and my apologies for the slow response!

PR #435, opened 2018-02-07
1.0

rileyjmurray:
# Highlights

Wrote a cvxpy 1.0 interface for mosek. Improvements over the existing interface include 
- support for the exponential cone (if the user has mosek 9 or higher installed)
- recovery of semidefinite dual variables
- mixed integer programming

All changes to executable code are confined to 
- mosek_conif.py
- test_mosek_conif.py
- "Building models with fast compile times.ipynb"

The ipython notebook is unrelated to the work done for the mosek interface.

Documentation was added to conic_solver.py. I very much think these changes should be kept.

Some comments were added to mosek_qp.py. They should probably be removed and listed under "issues" instead.

# Further work

Mosek has very sophisticated status codes and termination criteria. The current handling of status codes is more-or-less copy-pasted from the cvxpy 0.4 interface. I have reason to believe that the current handling of status codes can result in unexpected behavior (e.g. reporting that an optimal solution was found even though it violates constraints). I would like to look into handling these status codes in a way that prevents such behavior.


SteveDiamond:
This looks good to me. Perhaps you're right about removing the MOSEK QP interface. I'll have to think about it more.

PR #436, opened 2018-02-09
Fixed and improved Cbc-interface

sschnug:
## STATUS
Not ready to merge!

### Bug

I think commit ef5f95477 broke Cbc (mip_tsp.py breaks after changes needed for 1.0).

The problem is rooted in:

    model.addConstraint(x[idxs] >= 0)

where *lhs* will not have the same dimension as the problem, needed for this overloaded-usage.

I still feel somewhat uneasy with cylp (and it's nearly non-happening development), but i made use of the ```A * x >= 0``` approach in the past (with success), while explicit usage of bound-setters crashed (maybe i should recheck that). In regards to this usage, the sparse-matrix approach i used here should be quite efficient.

### Other changes

I also changed the calls in regards to solving for Cbc and Clp, as i realized a few weeks ago, that preprocessing never seemed to be used (Cbc was slower than GPLK and i wondered why). One should probably check this if it's doing the expected (i'm pretty sure Clp-usage does):

-  for **Cbc**: [this usage](https://github.com/sschnug/kemeny_ranking/blob/master/kemeny.py#L167) worked great for me, but it will probably use all Cbc-CLI defaults including cutting-planes
    - maybe there is some merit just to use this as a good default for MIPs (cutting-planes pre-selected)
    - only question then: still allow manual selection?
    - values returned also need re-mapping (i think those are the original Cbc-lib status codes)
- for **Clp**: [this usage / like commited](https://github.com/sschnug/footrule_ranking/blob/master/solver.py#L103) will use preprocessing, advanced crashing-stuff and co. as observed (and i think heuristics to choose primal/dual)

### DIFF
--- a/cvxpy/problems/solvers/cbc_intf.py
+++ b/cvxpy/problems/solvers/cbc_intf.py
@@ -17,13 +17,11 @@ You should have received a copy of the GNU General Public License
 along with CVXPY.  If not, see <http://www.gnu.org/licenses/>.
 """
 
-import importlib
-import six
 import cvxpy.interface as intf
 import cvxpy.settings as s
 from cvxpy.problems.solvers.solver import Solver
 import numpy as np
-import scipy.sparse as sp
+
 
 class CBC(Solver):
     """ An interface to the CBC solver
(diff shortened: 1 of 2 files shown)


SteveDiamond:
It looks good. I haven't really used Cbc so whatever defaults seem best to you are fine with me. Could you fix the flake8 issues and other problems that are failing in the CI?

sschnug:
I will have a look at it tomorrow and re-check direct-access to variable-bounds (the whole constraint-thing is not that exploiting in terms of Simplex).

I thought CI-errors came from something else and were observed for your commit too? This part is not something i'm very familiar with and manually calling nosetests on the cbc-part works, but i will check again.

SteveDiamond:
It's probably just the linter. The CI runs flake8 to check syntax after running nosetests. But Travis passes on 1.0 after a ton of work and I'm trying to maintain that.

sschnug:
@SteveDiamond 

That was a fast merge. I wanted to modify the docs to reflect those changes at least.

Part of that is warning (missing currently) about the need for py3-branches: what fork/pull-request should be linked to?

Is yours needed (sidenote: the original ```isRange-thingy```, which i interpreted as: use bounds instead of constraints for range-like expressions did not work for me when i tried it; so i'm using a more low-level approach now)? Or should i just link to the ```jjhelmus``` pull-request?

PR #438, opened 2018-02-10
1.0: New (open-source) MIQP solver based on CoinOR's Bonmin

sschnug:
Hi Steven, let me know if there is any chance in merging this into cvxpy!

### Status
**Not ready to merge yet!**

### What is this
It's a new MIQP-solver using a small wrapper of mine: [pyMIQP](https://github.com/sschnug/pyMIQP), which uses the general (potentially non-convex) MINLP-solver [CoinOR Bonmin](https://projects.coin-or.org/Bonmin).

The wrapper basically is using all those assumptions of QPs to calculate the necessary jacobian/hessian information internally, exploiting sparsity. It somewhat resembles the idea used in ```nonlinear_to_lpqp``` from JuliaOpt's [MathProgBase.jl](https://github.com/JuliaOpt/MathProgBase.jl/blob/master/src/SolverInterface/nonlinear_to_lpqp.jl) (which is more general supporting QCQP i think; it was not ported -> no license-trouble!).

### Why am i interested in adding this
To improve the open-source situation in regards to discrete-problems within cvxpy.

### Anything interesting for people having access to commercial-solvers?
```limited-memory hessian-approximation``` is supported. Maybe this can help for somewhat dense Hessians.

#### Documentation
- None yet

#### Tests
The tests, as added, all work. All tests from test_qp.py were re-used, but duals are not available.

Tests are completely self-contained in ```test_bonmin_qp.py``` (which might fragment this directory a bit).

- Repair / Analyze *broken* tests: it seems marking a variable as boolean does not introduce 0/1 bounds for the relaxation -> Bonmin-research needed
    - This feels bad in the test-example ```class TestStatus```
- Add non-small tests
    - Problem: i don't have access to *gold-standard* solvers (Gurobi, Mosek, ...) for comparison
- Add some real-world test

#### Support more options
- ```initial_point``` already supported in external lib, but not yet in cvxpy-interface

#### Not that important to me right now
- duals

### Questions
- Are there any MIQP-tests in cvxpy? I did not find any yet. I suppose Gurobi's, Mosek's, ... interfaces are now ready to solve these problems, right?
    - If so: this would be a good source of additional test-problems (like mentioned above: no access to commercials to compare)

### CODE REVIEW

File: continuous_integration/install.sh [hunk 1]
```diff
@@ -78,6 +78,45 @@ if [[ "$DISTRIB" == "conda" ]]; then
         cd "$oldpath"
     fi
 
+    # Install BONMIN_QP requirements: Bonmin + pyMIQP
```
SteveDiamond: I have two different type of CIs. One with the base solvers installed and one with all the extra solvers. Extra solvers need to be put under the "$INSTALL_GLPK" == "true"

doc/source/tutorial/advanced/index.rst
```diff
@@ -689,17 +701,46 @@ Here is the complete list of solver options.
 
 `CBC`_ options:
 
-Cut-generation through `CGL`_
+Cbc currently supports no option despite ``'verbose'`` supported by all solvers.
```
SteveDiamond: supports no option except for ``'verbose'``, which is supported by all solvers.

cvxpy/tests/test_bonmin_qp.py
```diff
@@ -0,0 +1,578 @@
...
+from cvxpy.reductions.qp2quad_form.qp_matrix_stuffing import QpMatrixStuffing
+from cvxpy.reductions.qp2quad_form.qp2symbolic_qp import Qp2SymbolicQp
+from cvxpy.tests.base_test import BaseTest
+
+# Exact-hessian: check by first-order + second-order by finite-differences
+# Hessian-approximation: check first-order only
+# Key in this dict: boolean option: hessian_approximation_
+deriv_check_map = {False: 'second-order', True: 'first-order'}
+
+class TestQp(BaseTest):
```
SteveDiamond: Change the class name to something new.

cvxpy/tests/test_bonmin_qp.py
```diff
@@ -0,0 +1,578 @@
+"""
```
SteveDiamond: Are these tests copied from the other QP tests? That's not bad necessarily, just clarifying. Also, do these tests succeed if BONMIN QP isn't installed? I would assume not. They need to check if it's installed before running the tests.

sschnug:
I will have a look into your suggestions and questions, probably within the next week. (First i need to redo my VM-setup). Thanks for getting back!

SteveDiamond:
@sschnug I'm pinging this in case you're still interested. The code base has changed a fair amount, but it should be fairly easy to merge in your interface. I wouldn't add it to the Travis build.

akshayka:
@sschnug, are you still interested in submitting this PR? If not, we'll close it for now, due to inactivity.

rbf22:
It would be great if this could be added, the planned removal of ECOS_BB will have the effect of leaving cvxpy without a MIQP solver.

rileyjmurray:
@rbf22 you're very right that removal of ECOS_BB means this PR has potential for significant benefit to cvxpy users. However I should clarify that cvxpy does have access to several MIQP solvers after removing ECOS_BB-- it's just that none of these is open-source. For small MIQP problems (e.g. 1000 variables or constraints) you can use the free edition of CPLEX.

PR #439, opened 2018-02-14
Fix gurobi for python 3

befelix:
PR #431 broke the gurobi interface for python 3 by using `iteritems` on dictionaries. This PR uses `iteritems` from `six` instead.


sschnug:
Should this not be automatically done by ```setup.py```'s ```use_2to3=True```? ([docs](https://docs.python.org/3/library/2to3.html#2to3fixer-dict)). It also looks like the tests, after modifying other failing ones did still work out for this PR.

I'm not sure how much six-usage is really needed. I think it was introduced as dependency, when i submitted the Cbc-interface in the past (i'm not sure if i had problems or i had not much luck with ```2to3```).

befelix:
Huh, I didn't know that setup.py runs 2to3 automatically. Then I don't
understand why I got an error on Python 3 without the patch. Maybe I'll
have some time to look into this next week.

PR #447, opened 2018-03-02
Fix for log_sum_exp

aseyboldt:
`log_sum_exp` is erroneously marked as non-increasing and positive.

This seems to have been introduced in
https://github.com/cvxgrp/cvxpy/commit/9366ccb54440ae0354f52cfbd14f30cba371b4db#diff-e6d9effeb355db619cf5bf6d59da6685


SteveDiamond:
Woops good catch. Can you please sign the CLA?

aseyboldt:
Done.
But don't you think you are maybe overthinking that copyright stuff a bit? :-)

Issue #440, opened 2018-02-15
Mosek interface in 1.0


aszekMosek:
First of all huge thanks to @rileyjmurray for the tremendous work. Two things I wanted to discuss:

1. For an integer problem the dual solution does not exist. In fact (and the only reason you don't see it with current Mosek 8.1 is because of a bug on our side) accessing the dual values raises an exception:

```
mosek.Error: (2950) No dual information is available for the integer solution.
```

2. This is more of an open question in the spirit of #434. In version 0.4 we had a selector between interior point and basic solution depending on quality. The issue #405 even asks to make it configurable. Let me know what you think.

rileyjmurray:
I'll be working on these things in the coming days / weeks. 

I'll also be looking into mapping cvxpy standard options (e.g. max_iters) to corresponding mosek parameter names. Right now if a user wants to change a mosek setting they have to do so by providing a dictionary from parameter names (as strings) to values. This difference in parameter specification can make it difficult for end-uses to use cvxpy in a solver-agnostic way.

rileyjmurray:
When the user asks for mosek to solve an LP, mosek_qpif.py is called, instead of mosek_conif.py.

If we want the user to be able to select the basic solution from the simplex optimizer, then either (1) this change needs to happen in mosek_qpif.py, or (2) mosek_qpif.py needs to be removed, so that mosek_conif.py is called for all problem classes.

cvxpy is capable of reformulating QP's as conic programs, and the backend of the mosek API actually does this conversion as well. If mosek's QP solver *didn't* perform this conversion, and could somehow exploit sparsity in the quadratic-form problem data, then it would make sense to keep the QP interface. However as it stands, mosek_qpif.py is redundant, and increases the complexity of maintaining cvxpy (particularly in writing test code, and handling solver parameters).

What are your thoughts @aszekMosek and @SteveDiamond ?

SteveDiamond:
Removing mosek_qpif.py seems best.

bstellato:
The only reason I see to keep the QP interface is that the SOCP conversion in Mosek, i.e., the matrix factorization, might happen more quickly (probably in plain C (?)) than in CVXPY (in numpy/scipy). Otherwise it is just more work to maintain it.

rileyjmurray:
[...]

PR #448, opened 2018-03-05
updates to mosek interface

rileyjmurray:
This pull request contains updates to resolve issue #440 . The effective changes are (1) dual variables are not recovered for discrete problems, and (2) the user can specify if they want a basic feasible solution, if the model is in fact an LP. As a result of these changes, mosek_qpif was removed from cvxpy, and mosek_conif was added to the set of modules covered by test_qp. 

Mosek is currently not passing some of the tests in test_constant_atoms that involve the exponential cone; this is likely because my version of mosek 9 is only an alpha release. For the time being, I have added the offending atoms to a list of exclusions in test_constant_atoms, so that I could run the entire set of nosetests (note: no one will see the effects of this change unless they have mosek 9 installed on their machine). With these exclusions, all nosetests are passing.


cvxpy/tests/test_constant_atoms.py
```diff
@@ -286,7 +290,7 @@ def run_atom(atom, problem, obj_val, solver, verbose=False):
         print(problem.objective)
         print(problem.constraints)
         print("solver", solver)
-    if check_solver(problem, solver):
+    if check_solver(problem, solver) and not (atom, solver) in KNOWN_SOLVER_ERRORS:
```
SteveDiamond: why did you change this? It might get lucky and work :)

rileyjmurray: There was already a check like 

'try:
   problem.solve(solver=solver)
catch:
    assert( (atom, solver) in KNOWN_SOLVER_ERRORS )'

This works when a solver's error is catch-able by Python. But for the alpha release of MOSEK 9 I currently have, some error codes kill the entire Python thread (and so are not catch-able). The only way for me to run all tests in test_constant_atoms.py was to preemptively avoid creating such fatal errors.

In principle, the changes to this file could be reverted. They only apply to people with installations of mosek 9, and hopefully these issues will be fixed well before mosek 9 is released to the public.

SteveDiamond:
Travis CI is failing. You need to have MOSEK only called if it's installed. Fix that and it looks good to merge.

rileyjmurray:
@SteveDiamond 

Can you check the Travis CI logs? Three of the six configurations pass. Of those that fail, none of them reference files that I changed (and some of the issues are timing errors, when warm starting fails to speed up the solver).

SteveDiamond:
Ok when I looked before there were errors were due to your changes.

Issue #449, opened 2018-03-13
one unit test about mosek fails in 1.0 branch


mickeydonkey:
My mosek version is 8.1.41, installed by conda. This seems a compatibility issue, and I am wondering which version of mosek should I try.
here is the log infomation:

```
======================================================================
ERROR: test_mosek_LP_solution_selection (cvxpy.tests.test_mosek_conif.TestMosek)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tao.luo/Projects/cvxpy/cvxpy/tests/test_mosek_conif.py", line 192, in test_mosek_LP_solution_selection
    prob.solve(solver=cvx.MOSEK, bfs=True)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/problems/problem.py", line 246, in solve
    return solve_func(self, *args, **kwargs)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/problems/problem.py", line 361, in _solve
    self.unpack_results(solution, self._solving_chain, inverse_data)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/problems/problem.py", line 453, in unpack_results
    solution = chain.invert(solution, inverse_data)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/reductions/chain.py", line 66, in invert
    solution = r.invert(solution, inv)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py", line 475, in invert
    dual_vars = MOSEK.recover_dual_variables(task, sol, inverse_data)
  File "/home/tao.luo/Projects/cvxpy/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py", line 536, in recover_dual_variables
    task.getsnxslice(sol, inverse_data['n0'], inverse_data['n0'] + snx_len, snx)
  File "/home/tao.luo/miniconda2/envs/cvxpfl/lib/python2.7/site-packages/mosek/__init__.py", line 5146, in getsnxslice
    raise Error(rescode(res),msg)
Error: (2953) snx is not available the basis solution.
```

rileyjmurray:
[...]

PR #451, opened 2018-03-16
make 1.0 pass unit tests for mosek 8.0

mickeydonkey:
This solves issue #449, I tested with two mosek versions, 8.1.34 and 8.1.41


mickeydonkey:
As suggested by @rileyjmurray in #449 , check basic feasible solution with 10 decimal places for higher precision.

SteveDiamond:
Ok looks good.

PR #452, opened 2018-03-20
Fixes to the tests

pietrodn:
1. `test_sign.py` failed because the name of unittest's method setUpClass() was wrong. This is now fixed.
2. I renamed `assertAlmostEquals()` to `assertAlmostEqual()` and `assertEquals()` to `assertEqual()`. Those are [deprecated aliases](https://docs.python.org/3/library/unittest.html#deprecated-aliases).


SteveDiamond:
Looks great, thanks! I'm not sure why it's failing on Python 3.6 but it's obviously not a problem you caused.

PR #453, opened 2018-03-20
Efficiency: use generator expressions for any(), all(), sum()

pietrodn:
When using the `any()` and `all()` Python built-in functions, it's more efficient to pass the parameters as [generator expressions](https://www.python.org/dev/peps/pep-0289/) than as list comprehensions.

We can *[take out the brackets](https://youtu.be/OSGv2VnC0go?t=47m49s)* as Raymond Hettinger suggests and enjoy the lazy evaluation of logic conditions. :-)


pietrodn:
Superseded by #458.

Issue #442, opened 2018-02-22
PSD matrices rejected by the value check in expressions/leaf.py


akisato-:
(This problem is closely related to https://github.com/cvxgrp/cvxpy/issues/415)

An attempt to store a positive semi-definite (PSD) matrix into cvxpy.Parameter((n,m), PSD=True) can be rejected, due to the value check in [expressions/leaf.py:344](https://github.com/cvxgrp/cvxpy/blob/1.0/cvxpy/expressions/leaf.py#L344).

```
344:            elif np.sum(self.project(val) != val):
```

Since self.project contains eigenvalue decomposition (see [expressions/leaf.py:290](https://github.com/cvxgrp/cvxpy/blob/1.0/cvxpy/expressions/leaf.py#L290)), the result might include small computation errors.

```
290:       elif any([self.attributes[key] for
291:                  key in ['symmetric', 'PSD', 'NSD']]):
292:            val = (val + val.T)/2
293:            if self.attributes['symmetric']:
294:                return val
295:            w, V = LA.eig(val)
296:            if self.attributes['PSD']:
297:                w = np.maximum(w, 0)
298:            else:  # NSD
299:                w = np.minimum(w, 0)
300:            return V.dot(np.diag(w)).dot(V.T)
```

I think that expressions/leaf.py:344 should be
```
elif np.sum(not np.allclose(self.project(val), val, rtol=rtol)):
```
after defining an appropriate `rtol` value.

You can reproduce this problem by the following test code.

```
import cvxpy as cvx
import numpy as np
np.random.seed(2)

a = np.random.normal(size=(100,100))
a2 = a.dot(a.T)    # This must be a PSD matrix.

b = (a2 + a2.T) / 2    # fixed
w, V = np.linalg.eig(b)
c = V.dot(np.diag(w)).dot(V.T)    # simulating self.project function
# You will see that np.max(c-a2) != 0 (1.22e-12 in this case)

d = cvx.Parameter((100,100), PSD=True)
d.value = a2
```

--> ValueError: Parameter value must be positive semidefinite.

rileyjmurray:
For the test code you provided, I think "b" should be "b = (a2 + a2.T) / 2". Perhaps I'm misunderstanding?

Also (although not terribly important), the "np.max(c-a2)" calculation should probably be replaced by "np.max(np.abs(c-a2))".

akisato-:
@rileyjmurray Thanks for the comments. 2. is what I meant exactly.

[...]

Issue #443, opened 2018-02-23
Infinity norm for matrices different than convention


sbarratt:
According to the documentation, the function `norm(X, 'inf')` when X is a matrix computes \sum_{i,j}|X_{i,j}|, which is different than the convention, \max_i \sum_j |a_{i,j}|. This could lead to some confusion.

source: http://mathworld.wolfram.com/MatrixNorm.html

rileyjmurray:
Hmm. Yes there is a discrepancy here.

In cvxpy 0.4, I think it makes sense to keep the current implementation. A large number of users may have run into this same discrepancy and simply changed their code accordingly.

The situation is a little different with cvxpy 1.0. Version 1.0 is designed to be as numpy compatible as possible, and numpy uses the same convention that you've pointed out. The result is that in version 1.0, the cvxpy Expression objects

expr1 = cvxpy.norm(X, 'inf')
and
expr2 = np.linalg.norm(X, np.inf) 

have different dimensions, even though the user's intent is indistinguishable.

@SteveDiamond , since version 1.0 is yet to be fully released, do you think it would be reasonable to have cvxpy's norm function behave in the same way as numpy's norm function? In addition to the matrix infinity-norm, there's also the matrix 1-norm.

SteveDiamond:
Yes, this behavior should be changed for cvxpy 1.0. Thanks for pointing this out!

rileyjmurray:
I've made the changes necessary to resolve this issue. It will be included in my next pull request.

rileyjmurray:
This can be marked as resolved.

PR #454, opened 2018-03-26
Resolving #442, #443, and #449. 

rileyjmurray:
This merge request affects a large number of files. The number of affected files is high because references to the undocumented file "norm2.py" were replaced by references to "pnorm.py" as part of resolving #443.

cvxpy/expressions/leaf.py was changed to use np.allclose( ... ) in the comparison during the project function, rather than "==" tests for equality. This resolves issue #442 . When adding unit tests for this change I noticed that although there were tests for Parameter objects, those tests were only for malformed inputs. I have split the tests for Parameter objects into two functions: one for malformed inputs, and one for well-formed inputs.

 test_constant_atoms.py has been changed to remove tuples "(mosek, exponential cone atom)" from the list of expected failures. I made this change because the most recent version of mosek 9 (although still in alpha release) passes these tests.

If this pull request is accepted, then CVXPY documentation should change as follows:
* cvxpy.norm( X, p ) defaults to the matrix p-norm when X.ndim == 2. Note that when X is an n-by-1 column vector, the matrix 1-norm, 2-norm, and infinity-norm all agree with the vector 1-norm, 2-norm, and infinity-norm. The only edge case is when X is a 1-by-n row vector. In this case the matrix 1-norm is equivalent to the vector infinity-norm, and the matrix infinity-norm is equivalent to the vector 1-norm.
* cvxpy.pnorm(X, p ) behaves as the documentation currently states. In particular, the equation cvxpy.pnorm(X, p) == cvxpy.pnorm(cvxpy.vec(X), p) trivially holds.

The changes to mosek_conif.py are to wrap up my suggestions from #449 .


cvxpy/expressions/leaf.py
```diff
@@ -341,7 +343,12 @@ def _validate_value(self, val):
                     "Invalid dimensions %s for %s value." %
                     (intf.shape(val), self.__class__.__name__)
                 )
-            elif np.sum(self.project(val) != val):
+            proj = self.project(val)
```
SteveDiamond: Can you explain this more?

SteveDiamond: Do we really need to make sparse matrices dense for the comparison? The current approach is very naive, but I was hoping to compare sparse to sparse.

SteveDiamond:
Also, any idea what happened to the Python 3.6 branch? It looks like NumPy updated and broke some assumption in the tests. I can fix this myself but curious if you had any thoughts. The timing based warm start test clearly is a mess and needs to go.

cvxpy/expressions/leaf.py
rileyjmurray: numpy's allclose function requires numpy arrays (rather than scipy matrices) and numpy doesn't have a sparse "array" type. 

Thinking about it more, the following route would be more efficient:
```
delta = np.abs(val - proj)  # might be a numpy array, scipy matrix, or sparse scipy matrix.
if intf.is_sparse(delta):   # is a scipy sparse matrix
    is_close_enough = np.allclose(delta.data, 0)   # only check for near-equality on nonzero values.
else:
    delta = np.array( delta )  # make sure we have a numpy array.
    is_close_enough = np.allclose(delta, 0)
if not is_close_enough:
    ... throw appropriate error ...
```

SteveDiamond: I like it!

rileyjmurray:
I'm not sure what happened on the Python 3.6 branch. Although I've looked back and it seems that the issue came up in my first commits for the mosek interface. Relevant builds are:

* The last time I can find where Python 3.6 doesn't generate that error:
https://travis-ci.org/cvxgrp/cvxpy/jobs/348206613
* The very next build (https://travis-ci.org/cvxgrp/cvxpy/builds/349391496) failed everything because of widespread flake8 issues.
* The build that was merged is https://travis-ci.org/cvxgrp/cvxpy/builds/349885887.

I googled "api version 0xc vs 0xb" and get quite a few hits for numpy with python 3.6. Unfortunately there isn't a universal solution here (or at least one that makes sense in our context).

SteveDiamond:
Ok fair enough we can deal with that later. It's clearly not related to this commit. I'm happy to merge after you update the project comparison so sparsity is preserved.

rileyjmurray:
We should be good to go!

PR #458, opened 2018-03-31
Efficiency: use generator expressions, cache expression properties

pietrodn:
This PR is an evolution of #453.

1. When using the `any()`, `all()`, `sum()`, ... Python built-in functions, it's more efficient to pass the parameters as [generator expressions](https://www.python.org/dev/peps/pep-0289/) than as list comprehensions. For example, if `any()` gets any `True` value for the generator, it can avoid to compute all the other elements. I systematically replaced all unneeded list comprehensions with generator expressions.
2. The return values of some functions like `is_constant()`, `is_convex()`, `is_zero()` [don't change](https://gitter.im/cvxgrp/cvxpy?at=5abc2ed57c3a01610d85ec1a) between invocations. They can be cached for any expression, in order to avoid useless recomputations. This should mitigate some of the problems about the [slowness](https://gitter.im/cvxgrp/cvxpy?at=5ab043f5e4d1c636041b8295) of CVXPY 1.0 in the problem generation phase.

Are there any benchmark I can run?

### DIFF
--- a/cvxpy/CVXcanon/tests/python/364A_scripts/material_blend.py
+++ b/cvxpy/CVXcanon/tests/python/364A_scripts/material_blend.py
@@ -44,7 +44,7 @@ for i in range(n):
 
 to_add = [-f[j,i] * p[j] for i in range(n) for j in range(m)]
 obj = sum(to_add)
-obj += sum([f_tilde[i] * pTilde[i] for i in range(n)])
+obj += sum(f_tilde[i] * pTilde[i] for i in range(n))
 
 objective = Maximize(obj)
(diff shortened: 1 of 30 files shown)


SteveDiamond:
This is an awesome commit, thanks so much! We do have some benchmarks that I can tell you about in the dev meeting.

SteveDiamond:
I'm going to go ahead and merge this. The tests all pass, except for the victims of my eternal battle against the entropic forces of package updates :)

pietrodn:
Thank you! :D

PR #469, opened 2018-04-14
Resolving #466, #467, #468. Cleaned up TravisCI build.

### LINKED ISSUE #466: canonicalizing complex quadratic forms
I'm trying to perform least squares on complex data with quadratic constraints.  The simple example is as follows:

```python
from cvxpy import *
import numpy as np

m = 10
n = 5
np.random.seed(1)
A = np.random.randn(m, n) + 1j*np.random.randn(m, n)
b = np.random.randn(m) + 1j*np.random.randn(m)

tmp = np.random.randn(m, n)+1j*np.random.randn(m, n)
Q1 = np.dot(tmp.transpose(1,0).conj(),tmp)

tmp = np.random.randn(m, n)
Q2 = np.dot(tmp.transpose(1,0),tmp)

# Construct the problem.
x = Variable(n,complex=True)
objective = Minimize(sum_squares(A*x - b))
constraints = [quad_form(x,Q1) <= 3,
               quad_form(x,Q2) <= 3]

prob = Problem(objective, constraints)

print "Optimal value", prob.solve()
print "Optimal var"
print x.value # A numpy matrix.
```

I receive the following error:

```
  File "/usr/local/lib/python2.7/dist-packages/spyder/utils/site/sitecustomize.py", line 880, in runfile
    execfile(filename, namespace)
  File "/usr/local/lib/python2.7/dist-packages/spyder/utils/site/sitecustomize.py", line 94, in execfile
    builtins.execfile(filename, *where)
  File "/home/akuurstr/cvxpyComplexTest.py", line 29, in <module>
    objective = Minimize(sum_squares(A*x - b))
  File "/softdev/akuurstr/python/thirdparty/cvxpy_top/build/lib.linux-x86_64-2.7/cvxpy/atoms/sum_squares.py", line 36, in sum_squares
    return quad_over_lin(expr, 1)
  File "/softdev/akuurstr/python/thirdparty/cvxpy_top/build/lib.linux-x86_64-2.7/cvxpy/atoms/quad_over_lin.py", line 34, in __init__
    super(quad_over_lin, self).__init__(x, y)
  File "/softdev/akuurstr/python/thirdparty/cvxpy_top/build/lib.linux-x86_64-2.7/cvxpy/atoms/atom.py", line 44, in __init__
    self.validate_arguments()
  File "/softdev/akuurstr/python/thirdparty/cvxpy_top/build/lib.linux-x86_64-2.7/cvxpy/atoms/quad_over_lin.py", line 108, in validate_arguments
    super(quad_over_lin, self).validate_arguments()
  File "/softdev/akuurstr/python/thirdparty/cvxpy_top/build/lib.linux-x86_64-2.7/cvxpy/atoms/atom.py", line 64, in validate_arguments
    "Arguments to %s cannot be complex." % self.__class__.__name__
ValueError: Arguments to quad_over_lin cannot be complex.
```

Is it possible to perform complex least squares?  Should I use something other than sum_squares?

### LINKED ISSUE #467: projection for PSD matrices can fail when eigenvalues not distinct
The numpy function `w, V = np.linalg.eig(A)` does not guarantee that `V` is an orthogonal matrix, even when `A` is orthogonally diagonalizable.

The issue arises when `A` has repeating eigenvalues. In this case there can be column-blocks of `V` that span the appropriate eigenspaces, but are not orthogonal.

Here is an example:
```
import numpy as np
np.random.seed(1)
m = 10
n = 5
A = np.random.randn(m, n) + 1j*np.random.randn(m, n)  # a random complex matrix
A = np.dot(A.T.conj(), A)  # a random Hermitian positive definite matrix
lifted_A = np.bmat([[np.real(A), -np.imag(A)],
                    [np.imag(A), np.real(A)]])
print( np.linalg.norm( lifted_A - lifted_A.T ) )  # lifted_A is symmetric.
w, V = np.linalg.eig(lifted_A)
print( np.sort( w ))  # all eigenvalues occur twice
print( np.linalg.det(V) )  # V doesn't have unit determinant, so it can't be orthogonal
```
The result is that if `w, V = np.linalg.eig(A)`, the matrix `A - np.dot(V, np.diag(w)).dot(V.T)` can have *very* large norm even when `A` is orthogonally diagonalizable.

Luckily, the fix is not hard. We just need to switch from `np.linalg.eig` to `np.linalg.eigh`. The latter function assumes that the matrix is Hermitian (symmetric in the real case) and [claims to be designed](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.eigh.html#numpy.linalg.eigh) so that the matrix `V` is orthogonal.

Note: this did not show up in the tests we have right now, because the PSD matrices in those tests had all distinct eigenvalues with high probability.

### LINKED ISSUE #468: CVXPY expression become constant zero unexpectedly
I am using CVXPY (version 1.0). I do not understand what causes the difference between a ZERO expression and a NONNEGATIVE expression as demonstrated by the following code:

```
import cvxpy as cv
a = cv.Variable(1)
b = cv.Variable(1)
print((cv.square(a-b)*1/(1e-6)).__repr__()) 
# Expression(CONSTANT, ZERO, (1,))
print((cv.square(a-b)*float(1/1e-6)).__repr__()) 
# Expression(CONVEX, NONNEGATIVE, (1,))
```

rileyjmurray:
Changes to matrix_utilities.py resolve #468.
Changes to leaf.py resolve #467.
Changes to matrix_canon.py resolve #466.

Changes to .travis.yml are to prepare for OSX tests.
Changes to install.sh were to clean things up, and add SCS to the set of minimal tests (ensures SDP tests).


cvxpy/reductions/complex2real/atom_canonicalizers/matrix_canon.py
```diff
@@ -88,6 +89,7 @@ def quad_canon(expr, real_args, imag_args, real2imag):
             imag_args[1] = np.zeros(real_args[1].shape)
         matrix = bmat([[real_args[1], -imag_args[1]],
                        [imag_args[1], real_args[1]]])
+        matrix = Constant(matrix.value)
```
SteveDiamond: No we are not doing this. The correct solution is to make a new atom to represent the lifting of a complex matrix into the reals.

SteveDiamond:
This is fantastic work! Thanks so much for doing this. I have a small critique about the solution to #466 that we can discuss.

SteveDiamond:
I need this code to add the Mac builds to Travis so I'm going to merge it, but we need to fix the complex lifting via a new atom.

PR #479, opened 2018-04-26
Gurobi status codes

gglockner:
A few minor changes to handle Gurobi status codes more gracefully. For example, if a model has a solution but it is not proven optimal, reset the status to OPTIMAL_INACCURATE.


gglockner:
Ignore this while I fix the CLA

PR #495, opened 2018-05-18
Added OSQP references and settings in docs

bstellato:
Minor changes to the docs to mention OSQP and its settings.

Slightly unrelated, I think many of the cvxpy 1.0 examples in the documentation are actually QPs and they would now solve with OSQP.


SteveDiamond:
Thanks! This looks good.

Issue #500, opened 2018-05-25
AttributeError: 'Task' object has no attribute '_Task__library'


shokirovnozir:
Hi, 

After solving an optimal power flow problem I get the following error, 
Could help me in fixing this, (I am using CVXPY + MOSEK). Thanks.
```
Problem
  Name                   :                 
  Objective sense        : min             
  Type                   : CONIC (conic optimization problem)
  Constraints            : 984             
  Cones                  : 0               
  Scalar variables       : 141             
  Matrix variables       : 6               
  Integer variables      : 0               

Optimizer started.
GP based matrix reordering started.
GP based matrix reordering terminated.
Problem
  Name                   :                 
  Objective sense        : min             
  Type                   : CONIC (conic optimization problem)
  Constraints            : 984             
  Cones                  : 0               
  Scalar variables       : 141             
  Matrix variables       : 6               
  Integer variables      : 0               

Optimizer  - threads                : 4               
Optimizer  - solved problem         : the primal      
Optimizer  - Constraints            : 876
Optimizer  - Cones                  : 1
Optimizer  - Scalar variables       : 40                conic                  : 31              
Optimizer  - Semi-definite variables: 6                 scalarized             : 468             
Factor     - setup time             : 0.05              dense det. time        : 0.01            
Factor     - ML order time          : 0.01              GP order time          : 0.02            
Factor     - nonzeros before factor : 7.30e+04          after factor           : 2.41e+05        
Factor     - dense dim.             : 2                 flops                  : 8.53e+07        
Factor     - GP saved nzs           : 2608              GP saved flops         : 1.35e+06        
ITE PFEAS    DFEAS    GFEAS    PRSTATUS   POBJ              DOBJ              MU       TIME  
0   1.8e+00  6.4e+00  1.0e+00  0.00e+00   9.529630469e+00   9.529630469e+00   1.0e+00  0.06  
1   1.8e-01  6.4e-01  9.2e-02  1.45e+00   1.237231359e+00   5.351471555e+00   1.0e-01  0.16  
2   3.6e-02  1.3e-01  7.0e-02  6.35e-01   4.210277235e-01   6.378552349e-01   2.0e-02  0.19  
3   7.5e-03  2.7e-02  8.1e-02  2.00e+00   5.839043985e-01   5.815552570e-01   4.3e-03  0.22  
[...]

Issue #497, opened 2018-05-22
Mosek _Task_library error with cvxpy 1.0.3 and Mosek 8.1.53


eckara:
Running the tsp_mip.py [example](https://github.com/cvxgrp/cvxpy/examples/tsp_mip.py ) suggests MOSEK interface errors.

`  
Integer solution solution summary
  Problem status  : PRIMAL_FEASIBLE
  Solution status : INTEGER_OPTIMAL
  Primal.  obj: 1.6386400651e+00    nrm: 5e+00    Viol.  con: 0e+00    var: 0e+00    itg: 0e+00  
Traceback (most recent call last):
  File "tsp_mip.py", line 51, in <module>
    prob.solve(verbose=True)
  File "/usr/local/lib/python2.7/dist-packages/cvxpy/problems/problem.py", line 246, in solve
    return solve_func(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/cvxpy/problems/problem.py", line 361, in _solve
    self.unpack_results(solution, self._solving_chain, inverse_data)
  File "/usr/local/lib/python2.7/dist-packages/cvxpy/problems/problem.py", line 453, in unpack_results
    solution = chain.invert(solution, inverse_data)
  File "/usr/local/lib/python2.7/dist-packages/cvxpy/reductions/chain.py", line 66, in invert
    solution = r.invert(solution, inv)
  File "/usr/local/lib/python2.7/dist-packages/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py", line 460, in invert
    task.getprosta(sol)  # mosek "problem status"; unused.
  File "/root/.local/lib/python2.7/site-packages/mosek/__init__.py", line 142, in accept
    return fun(*[ t(a) for (t,a) in zip(argtlst,args) ])
  File "/root/.local/lib/python2.7/site-packages/mosek/__init__.py", line 37, in sync
    return f(self,*args,**kwds)
  File "/root/.local/lib/python2.7/site-packages/mosek/__init__.py", line 5310, in getprosta
    _,msg = self.__getlasterror(res)
  File "/root/.local/lib/python2.7/site-packages/mosek/__init__.py", line 37, in sync
    return f(self,*args,**kwds)
  File "/root/.local/lib/python2.7/site-packages/mosek/__init__.py", line 2237, in __getlasterror
[...]

PR #502, opened 2018-05-25
this should fix issue with early deletion of mosek Task objects

rileyjmurray:
Changes should resolve issues in #500 and #497 . I have tested with mosek 8 and 9, with both nosetests and running `cvxpy/examples/tsp_mip.py`.


SteveDiamond:
Hi Riley, what is the current plan for this PR? It's not the end of the world if we need to put out a simple patch to fix the Mosek issue and deal with other challenges later.

rileyjmurray:
I'm working on it today. Goal is to get automated pypi deployment to work (but to do this only when the version number updates).

rileyjmurray:
Should be ready to merge. As long as appropriate environment variables are set in the Travis CI website, a single commit that bumps version should result in automatic deployment to anaconda and pypi (for mac and linux).

SteveDiamond:
Awesome!

PR #518, opened 2018-06-13
include COPYING in source distributions

dougalsutherland:
Makes sure that the `COPYING` file is included in `setup.py sdist` files.


SteveDiamond:
Thanks!

PR #528, opened 2018-06-20
simplify and update conda recipe

mcg1969:
Updated to use the Anaconda 5.0 [compiler tools](https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html).

I've also taken the liberty to remove the solvers as `conda`-specific dependencies. I would really like to encourage you to adopt this. For one thing, because the free version of ECOS is GPLv3, you're effectively preventing anyone who is GPL-averse from using cvxpy through conda. But more fundamentally, it reduces the install requirements for people who know exactly which solver they want.

To play devil's advocate, this may confuse users who do not appreciate that at least one solver must be present. A remedy here is to update the documentation but also to print an informative error message when the solver resolution is attempted. I can add that code if you like.

While Anaconda isn't yet putting `cvxpy` into the defaults channel, I am actually building it for a couple of customers, and this is the recipe (or a slight modification of it) is the one that I will be using.

### DIFF
--- a/conda-recipe/conda_build_config.yaml
+++ b/conda-recipe/conda_build_config.yaml
@@ -1,0 +1,19 @@
+python:
+  - 2.7
+  - 3.5
+  - 3.6
+numpy:
+  - 1.11
+c_compiler:
+  - vs2008                     # [win]
+  - vs2015                     # [win]
+  - vs2015                     # [win]
+cxx_compiler:
+  - vs2008                     # [win]
+  - vs2015                     # [win]
+  - vs2015                     # [win]
+zip_keys:
+  -                             # [win]
+    - c_compiler                # [win]
+    - cxx_compiler              # [win]
+    - python                    # [win]
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,17 +1,20 @@
+{% set version = "1.0.6" %}
+
 package:
-  name: cvxpy
-  version: "1.0.6"
+  name: cvxpy-base
+  version: {{ version }}
 
 source:
-  path: ..
+  git_url: https://github.com/cvxgrp/cvxpy.git
+  git_tag: v{{ version }}
 
 build:
   number: 0
-  script: python setup.py install --single-version-externally-managed --record record.txt
+  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record record.txt
 
 requirements:
   host:
-    - python
+    - python {{ python }}
     - setuptools
     - numpy >=1.9
   build:
@@ -61,8 +64,18 @@ test:
     - cvxpy.utilities
     - cvxpy.cvxcore.python
 
+outputs:
+  - name: cvxpy-base
+  - name: cvxpy
+    requirements:
+      - {{ pin_subpackage('cvxpy-base', exact=True) }}
+      - osqp
+      - ecos >=2
+      - scs >=1.1.3
+    build:
+      noarch: generic
+
 about:
   home: http://github.com/cvxgrp/cvxpy/
   license: Apache License, Version 2.0
-  license_file: LICENSE
   summary: 'A domain-specific language for modeling convex optimization problems in Python.'
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,34 +1,29 @@
+{% set name = "cvxpy" %}
 {% set version = "1.0.6" %}
 
 package:
-  name: cvxpy-base
+  name: {{ name|lower }}
   version: {{ version }}
 
 source:
-  git_url: https://github.com/cvxgrp/cvxpy.git
-  git_tag: v{{ version }}
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: a1b5b3ac31f6ed02649196c34e9b2d23320f170a7b05d6796029d9e7d4fc78d1
 
 build:
-  number: 0
-  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record record.txt
+  number: 1
 
 requirements:
   host:
     - python {{ python }}
-    - numpy {{ numpy }}
-    - setuptools
-  build:
-    - {{ compiler('cxx') }}
   run:
-    - python
-    - {{ pin_compatible('numpy') }}
-    - scipy >=0.15
-    - multiprocess
-    - fastcache
-    - toolz
+    - cvxpy-base
+    - osqp
+    - ecos >=2
+    - scs >=1.1.3
 
 test:
-  # Python imports
+  requires:
+    - nose
   imports:
     - cvxpy
     - cvxpy.atoms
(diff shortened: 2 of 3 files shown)


SteveDiamond:
I can see your point about the advantages of removing the solvers as dependencies. My perspective is that I advocate conda installation because it's a braindead simple way of getting everything set up. If I require people to install solvers separately, that may seem like a tiny barrier (and it is to sophisticated users), but even understanding what a solver is and which solver the user needs is quite a lot to ask of a complete beginner.

That's why my philosophy is "batteries included" and I strongly favor including a reasonable list of default solvers as dependencies. One option for getting the best of both worlds is to have both a ``cvxpy`` package that includes the solvers and a ``cvxpy-dev`` package that doesn't include any solvers, for people with the use cases you described.

mcg1969:
I think a dual package approach has promise. However: I still am concerned about the tie to GPL'ed ecos, even in the "braindead" case. We work with a lot of customers who are careful with their license choices, and implement systems to specifically exclude GPL packages. If ecos is not available to them, the braindead approach will fail and produce confusion. I grant that's a smaller subset of folks. But at minimum, some prominent documentation documenting the differences between a `cvxpy` and `cvxpy-dev` installation, a sort of "which should I install?" section, would be helpful.

Also: if one goes down this route, the right way to do it will be for `cvxpy` to become a metapackage that includes `cvxpy-dev` and the solvers. That's easily constructed. In fact I'm happy to work on that for you.

bstellato:
Just two comments....

- What about `cvxpy-suite` for the whole bundle and just `cvxpy` for the version without the solvers?

- There is always the option to include only the solvers with permissive licenses as strict conda requirements.

rileyjmurray:
@mcg1969 without including ECOS by default, non-expert cvxpy users won't have any decent way to solve problems involving the exponential cone (imo neither SCS nor CVXOPT are adequate substitutes).

If GPL is a problem, perhaps we can ask [Embotech](https://www.embotech.com/ECOS) to release ECOS under a different license. It would be a big request, but having ECOS as part of the default cvxpy distribution seems like great marketing for them.

mcg1969:
I do understand their position. In some cases the GPL gets commercial clients to the table so they can pay for alternative license. Alas, in other cases, it simply drives them away, no matter how loudly you shout that they can pay for said alternative... @echu @adomahidi thoughts?

echu:
Well, relicensing ECOS is not trivial as it depends on SuiteSparse, which
is GPLv3.

If anything, it might be worth rewriting it (it’s due for a rewrite)
properly under an Apache license along with SDP support. It’d be a (very)
big undertaking, but not impossible.

@adomahidi, thoughts?

rileyjmurray:
>Well, relicensing ECOS is not trivial as it depends on SuiteSparse, which ...

@echu I have been interested in making a version of ECOS for workstations for a while now. I would be happy to contribute to such a “rewrite,” and I would have pretty substantial time to dedicate to it as early as late July.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

echu:
Sure. You should go for it. Add me as a reviewer if you’d like.

I believe @adomahidi started from the CVXOPT source and changed some extra
stuff on the factorization side (typical IPMs reduce the KKT system and use
a dense solver, Alex lifted the system and used a sparse ldl solver with
iterative refinement, but this really only works up to second order cones
where you have block arrow structures).

I did some equilibration stuff for ECOS which helps make it a little more
robust.

@tiagoakle figured out how to do non symmetric cones with that setup, so
duplicating that might be a bit tricky. I’m not sure I still have his
references around.

And the branch and bound stuff is a straightforward wrapper—and should be
done in Python, I feel, and not as part of the original solver package (as
in ECOS).

It’s likely pretty easy to get a first cut going, which is what you should
aim for. Try to avoid solving research problems in the process :)

I’m happy to help too, but I think my commitment will be limited to code
reviews and maybe writing test cases. Still, having multiple eyes on the
code will only make the software better (and more maintainable), so I
highly encourage it.

I’m not being too serious: but I did want to try and write the thing in
Rust and use cffi to load it. From the bits of things I’ve read though,
that seems to perform poorly.

Anyway, you’re likely stuck with a C/C++ interface. If going C++ or C++11,
I might recommend using Eigen for the linear algebra.

Do let me know how you get on.

mcg1969:
> Well, relicensing ECOS is not trivial as it depends on SuiteSparse, which
is GPLv3

It's my understanding that it depends only on AMD and sparseLDL. The former has now been relicensed as BSD-3. The latter is GPLv2 still, but it's also short. Seems to me it would be a worthwhile and limited-scope project to 1) update AMD to the latest, free version, and 2) create a new LDL from scratch, drawn from non-GPL sources, say from the Elemental sources.

echu:
I think that would be up to Alex. @adomahidi?

adomahidi:
[Embotech](www.embotech.com) has no plans of changing the ECOS license. As @mcg1969 mentioned, we have a dual licensing strategy: We are happy to provide non-GPL licenses on a per project/product basis against a fee. Most people who are interested in this want to ship a product with ECOS inside, so they generate some form of revenue with it. We think it's fair to participate in that revenue. Everyone else, but in particular academics and researchers are free to use ECOS, review its source etc. Our intention is not to block people who do research but to participate if someone makes money from ECOS. (This excludes all hedge funds, btw., as they're not distributing a product) Unfortunately our experience is that most people don't even realize that ECOS (or other solvers for that matter) is running under the hood when solving problems with CVXPY, so if packaged together, there is zero value for people who create solvers.

As for sparseLDL, if someone writes a good sparse LDL solver that'd be useful for many people. I am not very familiar with Elemental but I am sure you can find a good LDL solver there. Again, this is not on our current roadmap. And I don't think it's a trivial task, besides the fact that sparseLDL is very small code - while you're at it, you may want to do the full-blown version with 1x1 and 2x2 blocks and online (e.g. Bunch-Kaufman) pivoting etc., which makes the sparse case pretty complicated.

mcg1969:
@adohmahidi Understood. I assume therefore that you’ve purchased an alternate license for SuiteSparse on behalf of your customers as well; otherwise the GPL is still in effect for the entire package.

In any case this is why we need to make sure a non-GPL installation option is available here for CVXPY.

echu:
@mcg1969, what do you think the approach should be? I feel we should write a brand new solver from scratch, but that’s quite a bit of work to package a suitably licensed solver with CVXPY.

Also, is this work that you think Continuum could support? We could try and have a student (like GSoC, but for the python scientific community) work on it along with @rileyjmurray.

mcg1969:
OK, I've modified the conda recipe now. I'm now using conda build 3.0's `outputs` functionality to generate two packages with the same recipe: `cvxpy` and `cvxpy-base`. `cvxpy` is actually just a metapackage that depends on `cvxpy-base`, `osqp`, `ecos`, and `scs`. `cvxpy-base` actually contains the core code.

mcg1969:
@echu, sorry, I didn't see your comment above.

To be clear: I totally understand Embotech's position. I didn't know for sure if they were still relying on an ECOS revenue stream; they've been doing other cool things lately! But given Alex's statement above, it's clear stripping GPL virality is not an option. 

On the other hand, some companies _actively filter_ GPL software even if they would never have used it in a non-compliant manner. They just choose not to take the risk. So that means they can't install `cvxpy` via conda as currently constituted.

And of course, `cvxpy` users have multiple commercial solver options available to them, not just ECOS.

It is outside of Anaconda's scope to sponsor the development of a liberally licensed solver, I'm afraid.

For now, what I've done is make it possible to install _just_ cvxpy, without any solvers, by doing `conda install -c cvxgrp cvxpy-base`. Of course, you'll need a solver at that point; e.g., `conda install -c cvxgrp cvxpy-base -c mosek mosek`.

echu:
Alright, thanks. That all sounds reasonable. Now, we just need to document the two stage install step somewhere. :)

mcg1969:
Sure, although since `conda install cvxpy` hasn't changed, it's not super-urgent.

mcg1969:
Don't merge this just yet—even though the test failure above looks to be just an intermittent server issue. I'm going to add a `conda_build_config.yaml` file to define a matrix of versions to build.

bstellato:
Pssssst! There something very exciting out there: [qdldl](https://github.com/oxfordcontrol/qdldl)!
New LDL solver: concise, library free, and of course Apache License 2.0. 

@gbanjac @goulart-paul

rileyjmurray:
@mcg1969 is this ready for final review by Steven?

h-vetinari:
The conda-forge [feedstock](https://github.com/conda-forge/cvxpy-feedstock) for cvxpy has been publishing for all OSes (as well as the AArches) for a while, so this can probably be closed?

rileyjmurray:
@h-vetinari thanks for pointing that out. The conda-forge recipe isn’t maintained by us and is slightly out of sync (e.g. it doesn’t ship with OSQP, which is a default solver in CVXPY).
Regardless, it’s clear that the existing conda-forge feedstock obviates this PR.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

h-vetinari:
@rileyjmurray 
Actually it does ship with [osqp](https://github.com/conda-forge/cvxpy-feedstock/blob/3fe24c6d893e616cad144bd1db89e5db5ae37f2e/recipe/meta.yaml#L35). :)

It's maybe slightly confusing that the last state of the recipe points to 1.0.21 (I don't know why @scopatz needed to trigger a rebuild for that), but in any case, the latest version on conda-forge is 1.0.25 as you can see on the [landing page](https://github.com/conda-forge/cvxpy-feedstock/#current-release-info).

rileyjmurray:
Ah yes it looks like is misread something there. Thanks for the clarification!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #531, opened 2018-06-22
Fix entropy atom with scalars

bstellato:
Looks like the entropy atom returns infinity with scalars independently from them being or not in the domain. I made a small fix.

Problem was raised [here](https://stackoverflow.com/questions/50991294/infinity-objective-value-given-by-cvxpy-on-a-convex-program).


SteveDiamond:
Thanks!

PR #539, opened 2018-06-29
Fix MIP solution statuses in Mosek

aszekMosek:
A fix for the Mosek interface. It corrects two situations when solving a problem with integer variables in Mosek resulted in ``SolverError``:

1. When the problem is infeasible - now returns ``s.INFEASIBLE``.
2. When an integer solution was found, but is not optimal - now returns ``s.OPTIMAL_INACCURATE``. I'm not sure this is the best status here, but I don't see a better one. It seems consistent with #480. This happens for example when a time limit was set for the solver.


SteveDiamond:
Thanks!

Issue #553, opened 2018-08-02
is_psd() and is_nsd() depending on eigvals


milandotlic:
In the file expressions/constants.py functions psd() and is_nsd() should be changed

if one of the eigenvalues is mathematically zero, numerically it doesn't have to be. For example if the largest eigenvalue is 10^3, eigenvalue that is mathematically zero can be anything in range (-10^-13,10-13). 

Therefore lines 
197 return all(self._eigvals.real >= 0)
and 
218 return all(self._eigvals.real <= 0)

should be changed to something like 
(>=) - 10^-(16-exponent_of(np.max(self._eigvals.real)))
(<=)

rileyjmurray:
Constant objects are used to cast numeric types (including numpy n-d arrays) to cvxpy Expressions. They are meant to have rigid tests for properties like sign, and PSD/NSD-ness.

If you have a term in your model that is PSD in exact arithmetic, you can use more generous tests for PSD-ness by declaring ``P = cvxpy.Parameter(P_array, PSD=True) ``.  When you declare ``P`` in this way, the constructor checks if the problem data ``P_array`` has sufficiently small projection onto the PSD cone after an eigen-decomposition (this happens in ``_validate_value`` of ``Leaf.py``, which in turn relies on ``project`` in ``Leaf.py``). We have verified that this allows for PSD matrices with nontrivial kernel.

Issue #407, opened 2017-11-01
Positive semidefinite quad_form => Problem does not follow DCP rules (cvxpy 1.0)


bstellato:
I have noticed that cvxpy 1.0 complains about the convexity of QPs with semidefinite Hessian. Here is the MWE:

```python
import cvxpy
import scipy.sparse as sparse

# Define problem data
P = sparse.csc_matrix([[4., -4., 0.],
                       [-4., 8., -4.],
                       [0., -4., 4.]])
#  P += 1e-10 * sparse.eye(3)  # Uncomment to make it work


# Define CVXPY problem
x = cvxpy.Variable(P.shape[0])

# Standard objective
objective = cvxpy.quad_form(x, P)

problem = cvxpy.Problem(cvxpy.Minimize(objective), [x <= 1., x >= 0])
problem.solve()
```

This code returns the error
```python
DCPError: Problem does not follow DCP rules.
```
I suspect this is related to the eigenvalues of `P` which are (12, 4, 0) but the last one appears to be `-1.34708082e-16` if computed via
```python
lam, _ = np.linalg.eig(P.todense())
```

By uncommenting one line in the first code snippet everything works. Would it make sense to add a tolerance on the negativity of the eigenvalues of `P`?

rileyjmurray:
In practice, if you face this issue you should reformulate the problem.

Suppose P is PSD with eigendecomposition P = U * D * U'. Then in your case, you would solve the problem:

min y' * D * y :  0 <= x <= 1, y = U' * x

By reformulating things this way it should be clear to the user that sufficiently small negative values in D should be set to zero.

rileyjmurray:
Follow-up: this has some overlap with #442 . In that situation a projection operation for PSD matrices is performed in standard arithmetic, but the result of that projection is tested against hard equality constraints (rather than np.allclose( ... )). As a result, even positive _definite_ parameter matrices can fail tests for being PSD.

The issue in #442 really does seem to be a matter of correctness, rather than convention. Assuming it gets fixed, there's a chance that cvxpy would allow you to declare your matrix P as a PSD cvxpy Parameter (since the projection of P onto the PSD cone could have sufficiently small residual). Such behavior is undesirable because it creates inconsistency with (1) recognizing P as a PSD Parameter, but (2) refusing to allow P.value to appear in a quadratic form.

rileyjmurray:
Here is a possible solution (one that I've recommended in a few other issues):

[...]

PR #555, opened 2018-08-06
Make Parameters and Constants behave similarly for possibly-singular PSD/NSD matrices.

### LINKED ISSUE #442: PSD matrices rejected by the value check in expressions/leaf.py

rileyjmurray:
### Motivation ###

Changing ``Leaf._validate_value`` as was done in #454 had two effects. 
* First, it resolved #442: incorrect rejection of positive definite matrices due to rounding errors in computing the residual ``A - proj_A``. 
* Second (as a side-effect) it allowed singular PSD and NSD matrices to be used in cvxpy. All the user needed to do was define those matrices as Parameter objects with the ``PSD=True`` or ``NSD=True`` keywords.

I want to follow-up on this side-effect. 

Although the side-effect is useful in that it provides some resolution to issues like #407 and #553, it creates subtle differences in when cvxpy will accept PSD/NSD matrices. Specifically, if ``P_arr`` is a numpy array representing a singular PSD matrix, then
* ``P = cvxpy.Parameter(shape=P_arr.shape, value=P_arr, PSD=True)`` can be constructed without error, and will have ``P.is_psd() == True``.
* ``P = cvxpy.Constant(P_arr)`` may have ``P.is_psd() == False``. 

The behavior of Constant objects is important. Even if a user doesn't *declare* something as a cvxpy Constant, it will automatically be cast to a Constant once it participates in a cvxpy Expression or Atom (e.g. ``quad_form``). 

### Modifications ###

This pull request modifies ``Constant.is_psd()`` and ``Constant.is_nsd()`` to allow for a 1e-8 tolerance on the eigenvalues. This is largely the same as tolerance that is *already* allowed in ``Leaf._validate_value`` for Parameters with ``PSD=True`` or ``NSD=True``. The only distinction is that on the official cvxpy master branch, ``Leaf._validate_value`` currently does an element-wise check on the residual ``A - proj_A``, while the proposed change checks the largest singular value of ``A - proj_A``.

### Considerations for approving or denying ###
* It would allow users to provide a value like ``-1e-9  * np.eye(n)`` and have the resulting cvxpy Constant be identified as a PSD matrix. I should remark that this is *already possible* with Parameter objects right now.
* It would make Constant and Parameter objects behave in almost identical ways.
* The calculation of eigenvalues in Constant objects can be made *much* more efficient than in Parameter objects. (The latter needs the eigenvectors, while the former does not.)


rileyjmurray:
I needed to sync with cvxgrp/cvxpy/master in order for the commit logs to show up right. I've made a new request at https://github.com/cvxgrp/cvxpy/pull/ 556.

PR #558, opened 2018-08-08
lazy import numpy for header injection

shariat:
This is not still enough, since `scs` and `ecos` have the same problem. Also used `find_packages` instead of listing all the packages manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,15 @@ os:
 env:
   matrix:
     - DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_ATLAS="true"
-      NUMPY_VERSION="1.13" SCIPY_VERSION="0.19"
+      NUMPY_VERSION="1.14" SCIPY_VERSION="0.19"
       COVERAGE="true"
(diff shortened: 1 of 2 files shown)


SteveDiamond:
This is interesting, but all the CI tests failed on Travis.

shariat:
doesn't seem to be related to this change, is `cvxpy` tested with `numpy==1.15.0` ?

shariat:
numpy version was fixed to 1.13 in travis.

rileyjmurray:
@shariat , just for my education, can you explain this PR? Specifically, what is meant by "lazy import [X] for header injection", and what are the changes made to setup.py?

Also, if this is ready for merging you should probably re-ping Steven and let him know.

shariat:
ping @SteveDiamond

shariat:
@rileyjmurray the goal is to run `setup.py` without the need for `numpy` being already installed.

rileyjmurray:
@SteveDiamond it looks like this is ready, from what I can tell.

SteveDiamond:
Yeah, looks good!

Issue #562, opened 2018-08-12
SuperSCS support!


Cerebrock:
To my understanding cvxpy should support superscs ( https://kul-forbes.github.io/scs/page_doc.html#seccvxpython ), but after installing in a conda env and being able to import, it doesn't appear in installed_solvers()

SteveDiamond:
Sorry, we must have lost the interface in the upgrade to 1.0. We'll take a look and put it back!

Cerebrock:
Please ping this thread then!

rileyjmurray:
PR #565 addresses this.

Issue #336, opened 2017-03-08
str does not parenthesize subexpressions


RhysU:
I expect `10 * (var0 + 5)` below but see something different:
```
In [1]: import cvxpy

In [2]: a = cvxpy.Variable()

In [3]: b = a + 5

In [4]: c = 10 * b

In [5]: c
Out[5]: Expression(AFFINE, UNKNOWN, (1, 1))

In [6]: str(c)  # Seemingly wrong
Out[6]: '10 * var0 + 5'

In [7]: a.value = 1

In [8]: c.value  # Value as expected
Out[8]: 60

In [9]: cvxpy.__version__
Out[9]: '0.4.8'
```

SteveDiamond:
Yes, this has been a known bug for quite a while. I don't find the ``str(expression)`` outputs useful, so I haven't bothered to fix it. You could fix it with a PR by checking if a subexpression should be parenthesized in the ``__str__`` method for the arithmetic operators in https://github.com/cvxgrp/cvxpy/tree/master/cvxpy/atoms/affine

rileyjmurray:
This is resolved in #565.

PR #565, opened 2018-08-20
SuperSCS (resolving #562) and expression printing (resolving #336)

rileyjmurray:
### SuperSCS stuff

Wrote a simple wrapper extending ``scs_conif.py`` to ``super_scs_conif.py``.

Key pieces of info:

1. Users with SuperSCS installed can now invoke in cvxpy by setting ``solver='SUPER_SCS'`` (note the underscore, in keeping with the style of solvers like ``ECOS_BB``).
2. The default precision is ``eps=1e-8``. This is in keeping with the default settings for SuperSCS when invoked in CVX + Matlab.
3. The default linear system behavior is ``use_indirect=False``.
4. If SuperSCS is installed, it's the preferred solver over SCS.

Tests passing on Python 2.7, 3.5, 3.6 with my mac.

### Expression printing stuff

Changed implementation of ``.name()`` in the abstract class ``BinaryOperator`` defined in ``binary_operators.py``. 
* The "name" function is what defines the string representation of an Expression.
* I changed the "name" function so that it parenthesizes addition or division operations. To the best of my knowledge, those are the only operations that are (1) distributive / non-associative, and (2) not represented as ``operator_name(arg1, arg2, ...)`` when printed.

As future work, It seems reasonable to add LaTeX printing support. If properly implemented it could go a long way in helping users debug their models.


SteveDiamond:
This looks good to me! Thanks so much for doing this!

Issue #560, opened 2018-08-10
qp test case can't pass with CPLEX as solver


mickeydonkey:
This is the problem I encounter with commit 51c6a3becf9456aab8be8798e9128a57219a756f
CPLEX I use is community edition, the version is 12.8, installed from anaconda channel ibmdecisionoptimization

`
$ conda list  |grep cplex
cplex                     12.8                     py27_0    ibmdecisionoptimization
`
I run it on my linux workstation.

here is the log, It says Q matrix is not symetric, So I guess it is some accuracy issue?

```
CPLEX Error  5012: Q is not symmetric.

Error
Traceback (most recent call last):
  File "/home/tao/miniconda2/envs/cvx/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/tao/Projects/cvxpy/cvxpy/tests/test_qp.py", line 101, in test_all_solvers
    self.equivalent_forms_3(solver)
  File "/home/tao/Projects/cvxpy/cvxpy/tests/test_qp.py", line 381, in equivalent_forms_3
    s = self.solve_QP(p3, solver)
  File "/home/tao/Projects/cvxpy/cvxpy/tests/test_qp.py", line 73, in solve_QP
    return problem.solve(solver=solver_name, verbose=True)
  File "/home/tao/Projects/cvxpy/cvxpy/problems/problem.py", line 246, in solve
    return solve_func(self, *args, **kwargs)
  File "/home/tao/Projects/cvxpy/cvxpy/problems/problem.py", line 360, in _solve
    kwargs)
  File "/home/tao/Projects/cvxpy/cvxpy/reductions/solvers/solving_chain.py", line 233, in solve_via_data
    solver_opts, problem._solver_cache)
  File "/home/tao/Projects/cvxpy/cvxpy/reductions/solvers/qp_solvers/cplex_qpif.py", line 152, in solve_via_data
    model.objective.set_quadratic(qmat)
  File "/home/tao/miniconda2/envs/cvx/lib/python2.7/site-packages/cplex/_internal/_subinterfaces.py", line 4118, in set_quadratic
    self.set_quadratic(_HBMatrix(args[0]))
  File "/home/tao/miniconda2/envs/cvx/lib/python2.7/site-packages/cplex/_internal/_subinterfaces.py", line 4114, in set_quadratic
    args[0].matval)
[...]

PR #568, opened 2018-08-23
FIX: asymmetric Q matrix in quadratic form

mickeydonkey:
This PR tries to solve an issue #560 I found when this project introduced CPLEX solver recently. 
it implements the first method @rileyjmurray proposed.


SteveDiamond:
Alright, this looks fine.

PR #577, opened 2018-09-06
more efficient project() for PSD/NSD

dougalsutherland:
I passed in a 102x102 PSD Parameter whose smallest eigenvalue was 1e-4, but the numerical error caused by the reconstruction in `project` made it think the matrix wasn't PSD. This changes `project` to just return the (symmetrized) array if it doesn't actually need to change.

Also be a little more efficient by dividiing by 2 inplace, and using broadcasting instead of constructing a diagonal matrix and calling `dot` with it.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -290,11 +290,13 @@ class Leaf(expression.Expression):
                 val = np.diag(val)
             return sp.diags([val], [0])
         elif self.attributes['hermitian']:
-            return (val + np.conj(val).T)/2
+            return (val + np.conj(val).T)/2.
         elif any([self.attributes[key] for
                   key in ['symmetric', 'PSD', 'NSD']]):
+            if val.dtype.kind in 'ib':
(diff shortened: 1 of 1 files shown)


rileyjmurray:
Personally I like the changes you propose, but the continuous-integration builds are failing for Python 3.5 and 3.6. Please update your solution so it's Python 3.x compatible.

dougalsutherland:
The error was `val /= 2` where `val` might be an integer type (as it is in one of the tests). This errors in python 3, or does silently slightly weird things in python 2 without `from __future__ import division`: it's right if you pass in an already-symmetric array, but if it wasn't symmetric it might round the off-diagonals strangely.

There was a similar problem in the hermitian case, which is worked around by doing `/ 2.` instead of `/ 2`.

Anyway, the old code that went through eigendecomposition and back returned floats, so I added an explicit cast to float if necessary.

rileyjmurray:
@SteveDiamond imo this is ready to merge.

PR #597, opened 2018-10-10
Fix indicator test error

gbanjac:
Fix indicator test error.

Problem described in https://github.com/cvxgrp/cvxpy/issues/570


gbanjac:
Fixed a typo in canonicalization of the quad_over_lin atom.

PR #602, opened 2018-10-13
handle builtin when it is a dict

### LINKED ISSUE #594: 1.0.9 fails setup.py when installing over pip (__NUMPY_SETUP__ attribute)
The newest built could not be installed correctly over pip.
The new numpy header injection method seems problematic.(?)

`AttributeError: 'dict' object has no attribute '__NUMPY_SETUP__'`
Detailed log on travis at the installation step can be found [here](https://travis-ci.org/lpfann/fri/jobs/437299138).

shariat:
closes #594.
`__builtins__` can be a dict. see https://docs.python.org/2/reference/executionmodel.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ class build_ext_cvxpy(build_ext):
             __builtins__['__NUMPY_SETUP__'] = False
         else:
             __builtins__.__NUMPY_SETUP__ = False
-        
+
         import numpy
         self.include_dirs.append(numpy.get_include())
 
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Can you explain what happened in #594? Why it's a dict?

akshayka:
@SteveDiamond apparently whether or not .__builtin__ is a dict depends on particular Python implementation. From the linked documentation:

> CPython implementation detail: Most modules have the name __builtins__ (note the 's') made available as part of their globals. The value of __builtins__ is normally either this module or the value of this modules’s __dict__ attribute. Since this is an implementation detail, it may not be used by alternate implementations of Python.

bstellato:
Looks like a problem with the GLPK optimizer precision: https://travis-ci.org/cvxgrp/cvxpy/jobs/441951455#L3487.

SteveDiamond:
ok, I'll change the tolerance.

PR #604, opened 2018-10-13
Huber QP canonicalization

### LINKED ISSUE #603: QP representation of the Huber regression
It is known that the Huber regression problem can be cast as a QP (see page 190 in [Boyd's CVX book](http://stanford.edu/~boyd/cvxbook/bv_cvxbook.pdf)). However, the current canonicalizer cannot do this.

Here is a numerical example:
```
from cvxpy import *
import numpy as np
import scipy as sp
import scipy.sparse as sparse

# Generate problem data
sp.random.seed(1)
n = 10
m = 100
A = sparse.random(m, n, density=0.5, format='csc')
x_true = np.random.randn(n) / np.sqrt(n)
ind95 = (np.random.rand(m) < 0.95).astype(float)
b = A.dot(x_true) + np.multiply(0.5*np.random.randn(m), ind95) \
	+ np.multiply(10.*np.random.rand(m), 1. - ind95)

# Solve the Huber regression problem
x = Variable(n)
objective = sum(huber(A*x-b))
Problem(Minimize(objective)).solve(solver=OSQP)
```
I get the following error:
```
SolverError: Problem could not be reduced to a QP, and no conic solvers exist among candidate solvers (['OSQP']).
```
If I specify a conic solver (or do not specify any), then the problem is reduced to an SOCP and solved correctly.

Maybe it would be good to add another atom `sum_huber` which is QP representable. What do you think?

bstellato:
Now the Huber function is QP compatible.

I have to say the cvxpy 1.0 canonicalization system very intuitive! I did not imagine it to be so easy.
I copied most of what was already in the conic canonicalizer.

It gives the same results as the conic Huber canonicalization so it seems correct to me. I have also added a unit test.

FIxes #603

### DIFF
--- a/cvxpy/atoms/elementwise/huber.py
+++ b/cvxpy/atoms/elementwise/huber.py
@@ -76,6 +76,11 @@ class huber(Elementwise):
         """
         return self.args[idx].is_nonpos()
 
+    def is_quadratic(self):
+        """Quadratic if x is affine.
+        """
+        return self.args[0].is_affine()
+
     def get_data(self):
         """Returns the parameter M.
         """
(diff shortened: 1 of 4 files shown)


gbanjac:
Great! Thanks for making the PR!

SteveDiamond:
Thanks! That's nice to hear some validation of the 1.0 system :)

PR #605, opened 2018-10-16
Option to dump task file from MOSEK

aszekMosek:
This adds an optional filename where MOSEK will dump the task data, for example

```
prob.solve(solver=MOSEK, save_file="dump.opf")
```

It is useful if the users can dump the problem for debugging esp. if the optimizer failed or crashed. Occasionally can be used for checking how cvxpy transformed the problem.

Online docs updated accordingly.

Tested in python 2.7 and 3.5.


SteveDiamond:
Looks good! Thanks!

PR #611, opened 2018-10-22
Fixed a bug in QP interfaces

gbanjac:
When solving a QP with CPLEX or GUROBI in Python3, CVXPY is throwing the following error:
```
TypeError: 'dict_keys' object does not support indexing
```
This is happening because the `keys()` method of a `dict` object does not return a list in Python3 (see e.g. [here](https://stackoverflow.com/questions/16819222/how-to-return-dictionary-keys-as-a-list-in-python)).

Here is a numerical example to reproduce the error:
```
from cvxpy import *
import numpy as np
import scipy as sp
import scipy.sparse as sparse

# Generate problem data
sp.random.seed(1)
n = 10
m = 100
A = sparse.random(m, n, density=0.5, format='csc')
x_true = np.random.randn(n) / np.sqrt(n)
ind95 = (np.random.rand(m) < 0.95).astype(float)
b = A.dot(x_true) + np.multiply(0.5*np.random.randn(m), ind95) \
    + np.multiply(10.*np.random.rand(m), 1. - ind95)

# Define problem
x = Variable(n)
objective = sum(huber(A*x-b))

# Solve with OSQP
Problem(Minimize(objective)).solve(solver=CPLEX, verbose=True)
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.9
+current_version = 1.0.10
 files = setup.py doc/source/conf.py conda-recipe/meta.yaml cvxpy/__init__.py
 commit = True
 tag = True
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: cvxpy
-  version: "1.0.9"
+  version: "1.0.10"
 
 source:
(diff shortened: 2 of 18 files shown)


gbanjac:
Fixes https://github.com/cvxgrp/cvxpy/issues/610

gbanjac:
Fixes https://github.com/cvxgrp/cvxpy/issues/610

PR #612, opened 2018-10-23
fix issue 608

Hennich:
It seems like just the string representation is wrong.
```
import cvxpy as cvx
x = cvx.Variable()
expr = -(1+x)
prob = cvx.Problem(cvx.Minimize(x), [expr == 0])
prob.solve()
```
gives the correct value `-1.0`.
The easy fix would be to just use brackets, in particular since this method is meant for all unary operators. So we get `-(var0 + 1.0)`

### DIFF
--- a/cvxpy/atoms/affine/unary_operators.py
+++ b/cvxpy/atoms/affine/unary_operators.py
@@ -28,7 +28,7 @@ class UnaryOperator(AffAtom):
         super(UnaryOperator, self).__init__(expr)
 
     def name(self):
-        return self.OP_NAME + self.args[0].name()
+        return self.OP_NAME + "(" + self.args[0].name() + ")"
 
     # Applies the unary operator to the value.
     def numeric(self, values):


SteveDiamond:
Yeah the string representation is broken. This fix would help. You should update the tests so they still pass.

rileyjmurray:
This caught my attention since #565 was supposed to address expression printing. In that PR I made the statement 

> I changed the "name" function so that it parenthesizes addition or division operations. To the best of my knowledge, those are the only operations that are (1) distributive / non-associative, and (2) not represented as operator_name(arg1, arg2, ...) when printed.

That was wrong. The main issue is that __neg__ isn't represented as "operator_name(arg1, arg2,...)", but it is of course distributive. -- This PR will fix that oversight.

rileyjmurray:
@Hennich do you want to finish the changes (those needed for the tests to pass), or would you rather that I do that?

PR #616, opened 2018-10-24
Fixed Gurobi statuses

### LINKED ISSUE #521: GUROBI interface: SolverError returned for infeasible model
An example code of CVXPY with GUROBI:
```python
import numpy as np
import numpy.random as nprand
import cvxpy as cvx

nprand.seed(0)

Q = nprand.randn(2,2) # np.array([[ 1.01092346, -1.246334  ], [-1.246334  , 11.24925237]])
q = np.array([-0.79305302, -0.67245977])
C = np.array([[ 1.74951088, -0.86604018],
       [ 1.83903552,  0.28968686],
       [ 1.85200233,  1.34964664],
       [-0.62858037,  0.15304167]])
c = np.array([ 0.58550308,  0.96431765,  0.40713688, -0.79539809])
z = np.array([ 1.71458648, -1.54797607])

x = cvx.Variable(2)
# obj = cvx.quad_form(x, Q.T @ Q) / 2 + q.T*x + cvx.norm(x-z,2)**2 / 20
obj = cvx.norm(Q*x,2)**2 / 2 + q.T*x + cvx.norm(x-z,2)**2 / 20
constr = [C*x <= c]
prob = cvx.Problem(cvx.Minimize(obj), constr)
prob.is_dcp()
prob.solve(solver=cvx.GUROBI, verbose=True)
```

When I run this code (Python 3.6.5, CVXPY 1.0.5, GUROBI 8.0), GUROBI returns that the model is infeasible:

> Barrier performed 25 iterations in 0.00 seconds
Infeasible model

but CVXPY throws a `SolverError` exception instead of reporting an "Infeasible model" status.
> Traceback (most recent call last):
  File "/Users/tn396/Downloads/testcvx.py", line 22, in <module>
    prob.solve(solver=cvx.GUROBI, verbose=True)
  File "/Users/tn396/anaconda3/lib/python3.6/site-packages/cvxpy/problems/problem.py", line 246, in solve
  File "/Users/tn396/anaconda3/lib/python3.6/site-packages/cvxpy/problems/problem.py", line 361, in _solve
    self.unpack_results(solution, self._solving_chain, inverse_data)
  File "/Users/tn396/anaconda3/lib/python3.6/site-packages/cvxpy/problems/problem.py", line 471, in unpack_results
    "to reduce the dynamic range."
cvxpy.error.SolverError: Solver 'GUROBI' failed. Try another solver or solve with verbose=True for more information. Try recentering the problem data around 0 and rescaling to reduce the dynamic range.
Freed default Gurobi environment

Furthermore, if I use `quad_form` instead of `norm` in the objective function in the above code (see the commented line), which I suppose to be equivalent to the `norm` code, GUROBI has numerical problems in addition to the model being infeasible, while CVXPY still throws the same exception:
> Barrier performed 45 iterations in 0.00 seconds
Numerical trouble encountered

bstellato:
Fixes #521. I have used the changes in #480 and added then to the QP interface as well.

### DIFF
--- a/cvxpy/problems/solvers/gurobi_intf.py
+++ b/cvxpy/problems/solvers/gurobi_intf.py
@@ -292,7 +292,7 @@ class GUROBI(Solver):
             results_dict["status"] = s.SOLVER_ERROR
 
         if results_dict["status"] == s.SOLVER_ERROR and model.SolCount:
-        	results_dict["status"] = s.OPTIMAL_INACCURATE
+            results_dict["status"] = s.OPTIMAL_INACCURATE
         results_dict["model"] = model
         results_dict["variables"] = variables
         results_dict["gur_constrs"] = gur_constrs
@@ -448,7 +448,7 @@ class GUROBI(Solver):
             try:
                 new_results[s.EQ_DUAL] = results_dict["y"][0:dims[s.EQ_DIM]]
                 new_results[s.INEQ_DUAL] = results_dict["y"][dims[s.EQ_DIM]:]
-            except:
+            except KeyError:
                 if not self.is_mip(data):
                     new_results[s.STATUS] = s.OPTIMAL_INACCURATE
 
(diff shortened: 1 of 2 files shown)


SteveDiamond:
Thanks! Please fix the flake8 errors though.

bstellato:
Should be fixed now. We should also close #480

Issue #578, opened 2018-09-12
Incorrect solution to simple ILP problem


SeanRBurton:
The following Python3 program:

```
import numpy as np
import cvxpy

v = np.array([-1, -1])
m = np.zeros((2, 2))
selection = cvxpy.Variable(2, boolean=True)
constraint = m @ selection <= v
problem = cvxpy.Problem(cvxpy.Minimize(0), [constraint])
print(problem.solve())                        
```

should surely not find a feasible solution, but it prints `0.0`. Maybe I'm missing something?

SteveDiamond:
Yes, it seems best to not have ECOS_BB as a default. It might even be best to not have any default integer solver at all, and if the user tries to solve an integer program cvxpy tells them to install one of various options.

rileyjmurray:
Okay cool. Sometime soon I'll look into things a bit more just to be sure that the issue with ECOS_BB isn't because of some crazy canonicalization thing happening within cvxpy. 

I don't really like the idea of not having a default integer programming solver. Are there licensing issues to making GLPK a dependency of cvxpy?

rileyjmurray:
I have confirmed that the error is due to ECOS_BB, and not something on cvxpy's end; I've opened up a ticket https://github.com/embotech/ecos/issues/166 on the ECOS page.

PR #619, opened 2018-10-29
Fix #599: Simplify installation of GLPK support

### LINKED ISSUE #599: GLPK installation
CVXOPT now [bundles GLPK](http://cvxopt.org/install/index.html#installing-a-pre-built-package) as part of its standard Wheel distribution binaries for Linux, Windows, and macOS on PyPI. Thus all that's needed to make GLPK work out of the box with cvxpy on all but the strangest setups is `pip install cvxopt`. Rather than just updating the [now out-of-date instructions](http://www.cvxpy.org/install/index.html#install-with-glpk-support) in the documentation, cvxpy could simply add `"cvxopt >= 1.2.1"` as one of the `install_requires` items in `setup.py`. Then GLPK would become a first-class pre-bundled solver, just like SCS, ECOS, and OSQP (meaning updating those docs too).

If you don't want CVXOPT to be installed by default, you can always put it in `setup.py`'s `extra_requires` so that people can install GLPK support with `pip install cvxpy [glpk]`.

wkschwartz:
Since CVXOPT version 1.2, CVXOPT's built Wheels on PyPI include GLPK bundled for Windows, Linux, and macOS. (Windows support is actually limited to 64-bit platforms running Python 3.5 or greater; I have included a PEP 508 dependency marker for the Python version, but there isn't one as far as I can tell for 64-bitness.)

I have also updated documentation to reflect the fact that installation of CVXPY on Linux and macOS is as simple as ``pip install cvxpy[glpk]``. With dependencies' Wheels including everything CVXPY needs, Anaconda is no longer necessary to easy installation.

# Note on finishing this PR
I have selected "Allow edits from maintainers". I cannot promise I'll make timely changes to this PR, especially since I have access to few of the systems CVXPY is to be installed on: I only have Python 3.7 on macOS 10.13, and I only use Pip, not Anaconda. I encourage the maintainers to check that `setup.py` and the new installation instructions work satisfactorily on other platforms. Please make any changes you deem necessary.

Even if I can't be sending code fixes, I'll try to answer questions if you post them here.

### DIFF
--- a/doc/source/install/index.rst
+++ b/doc/source/install/index.rst
@@ -10,28 +10,14 @@ compatible with 1.0.
 Mac OS X and Linux
 ------------------
 
-CVXPY supports both Python 2 and Python 3 on OS X and Linux. We recommend using
-Anaconda for installation, as we find that most users prefer to let Anaconda
-manage dependencies and environments for them. If you are comfortable with
-managing your own environment, you can instead install CVXPY with
-:ref:`pip <pip-installation>`.
-
-1. Install `Anaconda`_.
-
-2. Install ``cvxpy`` with ``conda``.
+CVXPY supports both Python 2 and Python 3 on OS X and Linux. You can
+install CVXPY as described in :ref:`pip <pip-installation>`, or, if you prefer
+Anaconda, with
 
    ::
 
       conda install -c conda-forge lapack
       conda install -c cvxgrp cvxpy
-
-3. Test the installation with ``nose``.
-
-  ::
-
-       conda install nose
-       nosetests cvxpy
-
 
 Windows
 -------
@@ -68,13 +54,22 @@ The CVXPY installation process on other platforms is less automated and less wel
 Pip
 ----------------
 
-CVXPY can be installed on all platforms with `pip`_. We recommend isolating
-your installation in a `virtualenv <https://virtualenv.pypa.io/en/stable/>`_.
+CVXPY can be installed on all platforms with `pip`_. We recommend isolating your
+installation in a virtual environment, either with `virtualenv
+<https://virtualenv.pypa.io/en/stable/>`_ on Python 2 or with :mod:`venv` on
+Python 3.3 and later.
 After activating the environment, simply execute:
 
   ::
 
       pip install cvxpy
+
+To add optional support for `GLPK`_ on :ref:`most platforms <glpk-installation>`,
+execute:
+
+  ::
+
+      pip install cvxpy[glpk]
 
 
 Install from source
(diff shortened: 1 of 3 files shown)


SteveDiamond:
Hmm I'm not sure I like adding to the setup.py. It would be great to simplify the installation instructions, but does it need to involve the cvxpy setup.py? Every solver in setup.py implies that we are responsible for making sure that installation works.

rileyjmurray:
@SteveDiamond it wouldn't be so bad to take responsibility for GLPK working with cvxpy. GLPK has an integer programming solver, and per #578 we need a reliable default ILP solver.

wkschwartz:
The biggest obstacle to making GLPK work consistently is Windows. See cvxopt/cvxopt#127, but I have a patch I'll submit for that soon.

However, until SCS makes a release on PyPI of its current master branch, installation on Windows has other problems too. Namely, master branch of SCS-Python fixes the problem where Numpy has to be installed before installing SCS. Unless you want to point your `setup.py` at a specific SCS commit, installing CVXPY on Windows will be a multistep process regardless of GLPK working or not.

doc/source/install/index.rst
```diff
@@ -10,29 +10,15 @@ compatible with 1.0.
...
-CVXPY supports both Python 2 and Python 3 on OS X and Linux. We recommend using
-Anaconda for installation, as we find that most users prefer to let Anaconda
-manage dependencies and environments for them. If you are comfortable with
-managing your own environment, you can instead install CVXPY with
-:ref:`pip <pip-installation>`.
-
-1. Install `Anaconda`_.
-
-2. Install ``cvxpy`` with ``conda``.
+CVXPY supports both Python 2 and Python 3 on OS X and Linux. You can
```
wkschwartz: Windows is part of normal testing via CI. Should the docs commit CVXPY to supporting Windows? Maybe this PR is the wrong place for it.

```suggestion
CVXPY supports both Python 2 and Python 3 on Windows, macOS, and Linux. You can
```

I leave this question totally up to the maintainers. Just something I noticed.

wkschwartz:
The remaining Travis failures are from flake8 errors in files not touched by this PR.

wkschwartz:
@rsemenoff This PR simplifies GLPK installation. It has nothing to do with CPLEX. I’ve never used Conda, so I can’t speak to that. I’ve had no problems installing cvxpy with Pip.

SteveDiamond:
I saw you bumped this in another thread. I appreciate the work, but I'm honestly not sure about adding another solver. Every installation method that we take responsibility for becomes a huge pain.

rileyjmurray:
@SteveDiamond in this particular case, it's worth considering how cvxpy has integer programming functionality, but the default solver (ECOS_BB) has correctness issues. I was thinking that GLPK would replace ECOS_BB in cvxpy.

SteveDiamond:
I've implemented this PR in master.

wkschwartz:
You’re referring to 081af1c1648e0c68ca24be64e469faacaae06466?

SteveDiamond:
@wkschwartz yes that's right. If there's anything important I missed, let me know. Some of the changes related to installation, and those instructions have change a lot.

Issue #617, opened 2018-10-24
Cplex and Gurobi python interfaces very slow


bstellato:
Gurobi calls seem much slower than Mosek or OSQP. For example, in this script solving multiple QPs

```python
import numpy as np
import scipy as sp
import scipy.sparse as sparse
from cvxpy import *
import time


def solve_problem(solver):
    start_time = time.time()
    x = Variable(n)
    gamma = Parameter(nonneg=True)
    objective = 0.5*sum_squares(A*x - b) + gamma*norm1(x)
    prob = Problem(Minimize(objective))

    # Solve problem for different values of gamma parameter
    avg_solve_time = 0
    for gamma_val in gammas:
        gamma.value = gamma_val
        prob.solve(solver=solver, warm_start=True)
        avg_solve_time += prob.solver_stats.solve_time/len(gammas)
    end_time = time.time()

    return end_time - start_time, avg_solve_time


if __name__ == "__main__":
    sp.random.seed(1)
    n = 100
    m = 1000
    A = sparse.random(m, n, density=0.5)
    x_true = np.multiply((np.random.rand(n) > 0.8).astype(float),
                         np.random.randn(n)) / np.sqrt(n)
    b = A.dot(x_true) + 0.5*np.random.randn(m)
    gammas = np.linspace(1, 10, 50)

    total_gurobi, solve_gurobi = solve_problem('GUROBI')
    total_mosek, solve_mosek = solve_problem('MOSEK')
    total_osqp, solve_osqp = solve_problem('OSQP')

    print("GUROBI: total: %.2e, solve: %.2e" % (total_gurobi, solve_gurobi))
    print("MOSEK: total: %.2e, solve: %.2e" % (total_mosek, solve_mosek))
    print("OSQP: total: %.2e, solve: %.2e" % (total_osqp, solve_osqp))
```

I get:
```
GUROBI: total: 1.48e+01, solve: 1.02e-01
MOSEK: total: 5.63e+00, solve: 5.29e-02
OSQP: total: 3.05e+00, solve: 5.66e-03
```

You can find the profiling results here 
[profiling.prof.zip](https://github.com/cvxgrp/cvxpy/files/2511099/profiling.prof.zip)

I looks like the function [solve_via_data](https://github.com/cvxgrp/cvxpy/blob/b53c13a79324993ec090743f9394bca115efc96b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py#L93) is very slow.

[...]

PR #623, opened 2018-11-06
Removed list comprehensions in Gurobi QP interface

bstellato:
This should speedup the Gurobi interface by using numpy indexing instead of list comprehensions.

Related to #617


bstellato:
The build is failing for some flake8 issues not related to this PR. It seems that flake8 no longer likes some math symbols in the comments.

SteveDiamond:
Hmm we'll have to look into this. Obviously not related to your change.

PR #629, opened 2018-11-17
Fixed subgradient calculation for absolute values

miggy2129:
Hi,

I noticed that the absolute value atom has a few mistakes especially if the shape of the argument is (). This turns up in cases such as defining x = cvx.Variable(), setting y = cvx.abs(x), and querying for y.grad.

I've fixed this issue by setting row and col to 1 if either the inner or outer shapes are ().

### DIFF
--- a/cvxpy/atoms/elementwise/abs.py
+++ b/cvxpy/atoms/elementwise/abs.py
@@ -74,8 +74,14 @@ class abs(Elementwise):
             A list of SciPy CSC sparse matrices or None.
         """
         # Grad: +1 if positive, -1 if negative.
-        rows = self.args[0].shape[0]*self.args[0].shape[1]
-        cols = self.shape[0]*self.shape[1]
+        if self.args[0].shape != ():
+            rows = self.args[0].shape[0]*self.args[0].shape[1]
+        else:
+            rows = 1
+        if self.shape != ():
+            cols = self.shape[0]*self.shape[1]
+        else:
+            cols = 1
         D = np.matrix(np.zeros(self.args[0].shape))
         D += (values[0] > 0)
         D -= (values[0] < 0)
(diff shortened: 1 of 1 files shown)


cvxpy/atoms/elementwise/abs.py
```diff
@@ -74,8 +74,14 @@ def _grad(self, values):
             A list of SciPy CSC sparse matrices or None.
         """
         # Grad: +1 if positive, -1 if negative.
-        rows = self.args[0].shape[0]*self.args[0].shape[1]
-        cols = self.shape[0]*self.shape[1]
+        if self.args[0].shape != ():
```
SteveDiamond: Use self.args[0].size and self.size instead of the current approach.

SteveDiamond:
This is a good catch! The logic you added is already in the ``expr.size`` attribute, so you should use ``expr.size``.

cvxpy/atoms/elementwise/abs.py
```diff
@@ -74,9 +74,15 @@ def _grad(self, values):
             A list of SciPy CSC sparse matrices or None.
         """
         # Grad: +1 if positive, -1 if negative.
-        rows = self.args[0].shape[0]*self.args[0].shape[1]
-        cols = self.shape[0]*self.shape[1]
-        D = np.matrix(np.zeros(self.args[0].shape))
+        if self.expr.size != 1:
+            rows = self.expr.shape[0]*self.expr.shape[1]
```
SteveDiamond: ``self.expr.shape[0]*self.expr.shape[1]`` is the same as ``self.expr.size``. What I meant was that you should simply write ``rows = self.expr.size`` and ``cols = self.size`` instead of all the logic with shapes.

miggy2129:
My bad. Got it! thanks for the feedback.

SteveDiamond:
Thanks!

PR #634, opened 2018-11-29
Update jupyter notebook examples

arlenk:
update most jupyter notebook examples for cvxpy 1.0 and python 3 (notebooks should still work in python 2.7).  Most changes were minor and related to:
1) using numpy arrays instead of matrices
2) avoiding "from cvxpy import *"
3) updating print statements to print functions for python 3
4) using 1d cvxpy objects where possible

I did not update the nba example as I don't know where the required data file is (it is not in the repository)


akshayka:
Thanks for doing this! This is excellent. I haven't looked at the change in detail yet, but one super small nit: can you import cvxpy as cp instead of as cvx?

arlenk:
> Thanks for doing this! This is excellent. I haven't looked at the change in detail yet, but one super small nit: can you import cvxpy as cp instead of as cvx?

sure, no problem, updated all the examples to use cp instead of cvx.  Happy to make any other changes too.

SteveDiamond:
@akshayka does this look good to you? It looks good to me!

akshayka:
Looks good to me. Thanks!

akshayka:
@SteveDiamond I've updated the .rst files for these notebooks (67197ec). Can you update the website when you get a chance?

Issue #608, opened 2018-10-19
__neg__ gives incorrect expression


jdark-tremor:
Negating expressions of the following form give incorrect results:

    from cvxpy import Variable
    x = Variable()
    c = 1.0
    print(str(-(x + c)))
    print(str(-(c + x)))

Output:

    -var0 + 1.0
    -1.0 + var0

Expected:

    -var0 - 1.0

PR #635, opened 2018-11-30
fix for issue #608

bogiebro:
The bug reported in issue #608 is just an issue with printing unary operators. This fix puts parens around unary operator arguments.

### DIFF
--- a/cvxpy/atoms/affine/unary_operators.py
+++ b/cvxpy/atoms/affine/unary_operators.py
@@ -28,7 +28,7 @@ class UnaryOperator(AffAtom):
         super(UnaryOperator, self).__init__(expr)
 
     def name(self):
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Great, thanks!

rileyjmurray:
@SteveDiamond the Travis and AppVeyor builds fail with this PR because the tests were not updated. As it stands, this PR is an almost exact duplicate of #612. I recommend reverting this merge until tests are updated.

SteveDiamond:
Right, that was dumb. I'll revert it.

SteveDiamond:
Thanks! I fixed those, and I'll keep looking now that I know the pattern. I fixed the flake8 version in Travis but for some reason that didn't make the builds consistent.

PR #643, opened 2018-12-11
Disciplined geometric programming

akshayka:
This pull request contains an implementation of [disciplined geometric programming](https://web.stanford.edu/~boyd/papers/dgp.html). It adds ...

1. Log-log curvature, which is the analog of (convex) curvature.
2. Rules for determining whether an expression is DGP.
3. A reduction that takes DGP problems to DCP problems (Dgp2Dcp).
4. A few new DGP atoms (prod, pf_eigenvalue, one_minus_pos, diff_pos)
5. The ability to solve DGPs using `problem.solve(gp=True)`.
6. Website documentation for DGP.
7. Example notebooks for DGP.


SteveDiamond:
It looks good to me based on our review yesterday, and looking over additions since.

PR #648, opened 2018-12-23
Found a minor issue on log_det

miggy2129:
I found a minor issue in the log_det atom as it doesn't consider raising the same TypeError when the shape is only 1 dimensional.

I had edited this so that it would throw the proper exception when someone accidentally inputs a vector, instead of a matrix, into the function.


SteveDiamond:
thanks!

PR #658, opened 2019-01-18
Vectorize scaled_lower_tri, tri_to_full

pietrodn:
I ran the test [`cvxpy/performance_tests/test_robustness.py`](https://github.com/cvxgrp/cvxpy/blob/941b1caf04dc16824e7d49fb2343eb70803c2c7e/cvxpy/performance_tests/test_robustness.py) with the [`yappi`](https://github.com/sumerc/yappi) performance profiler, and saw two functions — both in [`cvxpy/reductions/solvers/conic_solvers/scs_conif.py`](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/scs_conif.py) — taking a lot of resources:
* [`scaled_lower_tri`](https://github.com/cvxgrp/cvxpy/blob/941b1caf04dc16824e7d49fb2343eb70803c2c7e/cvxpy/reductions/solvers/conic_solvers/scs_conif.py#L48-L86)
* [`tri_to_full`](https://github.com/cvxgrp/cvxpy/blob/941b1caf04dc16824e7d49fb2343eb70803c2c7e/cvxpy/reductions/solvers/conic_solvers/scs_conif.py#L48-L86)

I profiled the `test_robustness.py` suite with yappi, exporting the results as callgrind files.
Then, I examined those files with QCacheGrind, sorting the called functions by the "self" column, i.e., how much time was spent inside the function's own code, not taking into account the functions called from there.
schermata 2019-01-18 alle 22 53 41

Those functions are inefficient because they operate on matrices using `for` loops instead of NumPy vectorized operations. I vectorized the code; after that, those two functions do not appear at the top of the list:

schermata 2019-01-18 alle 23 02 30

In the current `master` branch, executing `test_robustness.py` takes 23.144s.
In the optimized version, it takes 21.592s.

I think that the compute time is dominated by the solver, which doesn't show up in the profiler because it isn't Python code).

I'm sharing the full callgrind files, if you want to examine them.
[callgrind.not_optimized.txt](https://github.com/cvxgrp/cvxpy/files/2774669/callgrind.not_optimized.txt)
[callgrind.optimized.txt](https://github.com/cvxgrp/cvxpy/files/2774670/callgrind.optimized.txt)

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scs_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scs_conif.py
@@ -64,24 +64,20 @@ def scaled_lower_tri(matrix):
     """
     rows = cols = matrix.shape[0]
     entries = rows * (cols + 1)//2
-    val_arr = []
-    row_arr = []
-    col_arr = []
-    count = 0
-    for j in range(cols):
-        for i in range(rows):
-            if j <= i:
-                # Index in the original matrix.
-                col_arr.append(j*rows + i)
-                # Index in the extracted vector.
-                row_arr.append(count)
-                if j == i:
-                    val_arr.append(1.0)
-                else:
-                    val_arr.append(np.sqrt(2))
-                count += 1
+
+    row_arr = np.arange(0, entries)
+
+    lower_diag_indices = np.tril_indices(rows)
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Thanks!

PR #659, opened 2019-01-19
Avoid useless COO -> CSC sparse matrix convertions

pietrodn:
It's better to directly create a CSC matrix, than to create a COO only to convert it to a CSC after.


akshayka:
Thanks!

Issue #624, opened 2018-11-07
Randomness in constraints/variables order (bad for parametric problems)


bstellato:
The order of variables and constraints passed to the solver is not always preserved.  

Every time a problem is built constraints/variables can appear in different orders which correspond to data matrices with permuted rows/columns. This makes it harder to detect if OSQP (or any other solver) problem updates can be applied correctly.

_One idea could be to force the variables and constraints IDs to be in increasing/decreasing order so that the problem rows/columns are not permuted._ What do you think?

As @gbanjac pointed out, removing randomness from the data building process would make warm starting much more efficient. At the moment, OSQP needs to refactor the KKT matrix whenever the ordering of variables or constraints is not preserved.

Related to #620.

SteveDiamond:
This should definitely be fixed. It came up in the old cvxpy, and was fixed there. I took exactly the approach you propose, where I sorted the variables and constraints. Probably we're getting the issue because cvxpy 1.0 creates a lot of new variables as it rewrites problems. Another solution would be to have some sort of hash based on the actual content of the expressions, which could be used for consistent ordering.

dave31415:
Is this the cause of issue #614 Non-deterministic output. I've asked for that issue to be reopened as it is definitely a CVXPY issue not just an ECOS issue. See updates.

dave31415:
It would be useful I think to guarantee that the output is deterministic. Perhaps we could add some test cases to ensure that. Otherwise it's hard to use this in production because, in the enterprise, it's often required that you can reproduce any issue exactly. And it makes testing everything else far easier.

nepslor:
Experienced the same problem. It took me ages to find out it was a non deterministic solver behavior, since I would expect no source of randomness in a solver.

bstellato:
Can this be relevant for Python 3.6+? The insertion order in dictionaries seems to be preserved https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6

SteveDiamond:
It does seem to happen in Python 3.6 somehow. We've had reports from the class. #664 should remove any non-determinism in solving the same problem repeatedly within a piece of code, but there is still non-determinism between code runs.

bstellato:
Could it be fixed simply by adjusting the order of what is returned by `problem.variables()` and `problem.constraints`? Maybe I am missing something.

bstellato:
Are there any updates on this? Several people are having the same problem. 

[...]

Issue #410, opened 2017-11-14
CVXPY 1.0 parameters confer no performance benefit


farrellm:
In 0.4.11, using parameters and resolving a problem for different parameter values gives a significant performance improvement. Equivalent code with 1.0 runs in the same time whether only parameter values are change or if the whole problem is recreated from scratch.

test code:

```python
import time
import random
import statistics

import cvxpy as cvx
import numpy as np


def problem():
    x = cvx.Variable(shape=(8,))
    mu = cvx.Parameter(shape=(8,))

    u = mu.T * x - cvx.sum_squares(x) - 0.1 * cvx.sum(cvx.abs(x) ** (3/2))
    c = [-0.2 <= x, x <= 0.2]

    return {"problem": cvx.Problem(cvx.Maximize(u), c),
            "variable": x,
            "forecast": mu}

def with_compile():
    p = problem()
    p["forecast"].value = np.random.normal(size=8)
    p["problem"].solve()
    return p


p0 = problem()
def without_compile():
    p0["forecast"].value = np.random.normal(size=8)
    p0["problem"].solve()
    return p0


functions = with_compile, without_compile
times = {f.__name__: [] for f in functions}

for i in range(100):  # adjust accordingly so whole thing takes a few sec
    func = random.choice(functions)
    t0 = time.time()
    func()
    t1 = time.time()
    times[func.__name__].append((t1 - t0) * 1000)

for name, numbers in times.items():
    print('FUNCTION:', name, 'Used', len(numbers), 'times')
    print('\tMEDIAN', statistics.median(numbers))
    print('\tMEAN  ', statistics.mean(numbers))
    print('\tSTDEV ', statistics.stdev(numbers))
```

wkschwartz:
I can confirm this report. Here's my output from the script in the ticket. Ran this on Python 3.7, macOS 10.13.5, Core i7 1.7GHz. cvxpy version 1.0.6.

```
FUNCTION: with_compile Used 59 times
	MEDIAN 27.675867080688477
	MEAN   30.742257328356725
	STDEV  7.337865665104942
FUNCTION: without_compile Used 41 times
	MEDIAN 29.330015182495117
	MEAN   31.004120663898746
	STDEV  7.93215427430337
```

I also ran the script replacing `p0["problem"].solve()` with `p0["problem"].solve(warm_start=True)` in the `without_compile` function, and I got the same result.

[...]

Issue #409, opened 2017-11-14
For small problems, CVXPY 1.0 is slower than 0.4.11


farrellm:
For an equivalent problem, cvxpy 1.0 is take 3x of cvxpy 0.4.11 to compile and solve.

0.4.11

```python
def problem():
    x = cvx.Variable(8)
    mu = cvx.Parameter(8)

    u = mu.T * x - cvx.sum_squares(x) - 0.1 * cvx.sum_entries(cvx.abs(x) ** (3/2))
    c = [-0.2 <= x, x <= 0.2]

    return {"problem": cvx.Problem(cvx.Maximize(u), c),
            "variable": x,
            "forecast": mu}
```

1.0

```python
def problem():
    x = cvx.Variable(shape=(8,))
    mu = cvx.Parameter(shape=(8,))

    u = mu.T * x - cvx.sum_squares(x) - 0.1 * cvx.sum(cvx.abs(x) ** (3/2))
    c = [-0.2 <= x, x <= 0.2]

    return {"problem": cvx.Problem(cvx.Maximize(u), c),
            "variable": x,
            "forecast": mu}
```

common

```python
import time
import random
import statistics

import cvxpy as cvx
import numpy as np

def with_compile():
    p = problem()
    p["forecast"].value = np.random.normal(size=8)
    p["problem"].solve()
    return p


p0 = problem()
def without_compile():
    p0["forecast"].value = np.random.normal(size=8)
    p0["problem"].solve()
    return p0


functions = with_compile, without_compile
times = {f.__name__: [] for f in functions}

for i in range(100):  # adjust accordingly so whole thing takes a few sec
    func = random.choice(functions)
    t0 = time.time()
    func()
    t1 = time.time()
    times[func.__name__].append((t1 - t0) * 1000)

for name, numbers in times.items():
    print('FUNCTION:', name, 'Used', len(numbers), 'times')
    print('\tMEDIAN', statistics.median(numbers))
    print('\tMEAN  ', statistics.mean(numbers))
    print('\tSTDEV ', statistics.stdev(numbers))
```

SteveDiamond:
Thanks so much for trying out cvxpy 1.0! This is exactly the kind of feedback I was hoping for. I will look into this and try and speed things up.

rileyjmurray:
Following up on this. After running some pretty thorough tests on problems of various sizes, I've found that 
[...]

PR #664, opened 2019-02-01
Splitting the solving chain into intermediate + solving chains to speedup repeated solves

bstellato:
I thought about **splitting the SolvingChain in two parts** to allow canonicalization caching and faster solution with parametric problems. A good approach could be to split the SolvingChain into:

- A intermediate chain (Chain object) performing the canonicalization
- A solving chain (SolvingChain object) performing: `[EvalParams, MatrixStuffing, solver]`

What do you think?

At the moment the split works for QPs if we always redefine the chains at each problem solution, i.e., by ignoring the `if chain_key != self._cached_chain_key:` condition. However, I get two errors:

*  conic solvers complain about a `NotImplemented` `graph_implementation` method when I apply the `_solving_chain` 
* QP solvers return an infeasible status when we use the cached symbolic problem within multiple solves.

Could you have a look at this as well? I believe it would solve many issues such as #410 (speed!) and #624 (different order or constraints in parametric problems).


SteveDiamond:
It sounds reasonable! I'll take a look as soon as I can.

SteveDiamond:
I did some work here: https://github.com/cvxgrp/cvxpy/tree/pr664
I fixed almost all the errors with canonicalizing problems. The remaining errors are either from ``get_problem_data`` and ``unpack_results``, which are methods that give the user the solving chain, from the new messages you added for non-DCP/DGP problems, or a few other more complicated cases.

SteveDiamond:
Ok I will look into the two issues later. Would you mind evaluating the changes on some kind of benchmark to see how much it speeds things up with parameters? Something really simple is fine. I know there are other benefits like the fixed variable ordering, but I'm curious about the speed up.

bstellato:
Thanks! I wanted to add a couple of examples in the `cvxpy/performance_tests` folder but the files there seem deprecated. Are there any other places where you would like to store performance tests?

I have used the example in #409 and #410 with ECOS using [this script](https://pastebin.com/B1AFQxyg). I obtained:

- master:  [ecos_master.txt](https://github.com/cvxgrp/cvxpy/files/2837454/ecos_master.txt)

- this pr:  [ecos_pr.txt](https://github.com/cvxgrp/cvxpy/files/2837455/ecos_pr.txt)

The benefits seem more significant for smaller problems. For larger ones, I believe the matrix stuffing is still the bottleneck.

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
-        except Exception as e:
-            raise e
-        data, inv_data = solving_chain.apply(self)
-        return data, solving_chain, inv_data
+        # Check if chain already exists
+        chain_key = (solver, gp)
+
+        if chain_key != self._cached_chain_key:
+            try:
+                # Find candidate solvers
```
akshayka: nit: Delete this comment. Comments should not reiterate what is clear from the code.

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+                candidate_solvers = self._find_candidate_solvers(solver=solver,
+                                                                 gp=gp)
+
+                # Create symbolic chain and canonicalize problem
```
akshayka: nit: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+                # into symbolic one
+                self._symbolic_chain = \
+                    construct_symbolic_chain(self, candidate_solvers, gp=gp)
+                self._symbolic_problem, self._symbolic_inverse_data = \
+                    self._symbolic_chain.apply(self)
+
+                # Create solving chain
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+                self._solving_chain = \
+                    construct_solving_chain(self._symbolic_problem,
+                                            candidate_solvers)
+            except Exception as e:
+                raise e
+
+        self._cached_chain_key = chain_key
+
+        # Apply solving chain to get data
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+        data, solving_inverse_data = \
+            self._solving_chain.apply(self._symbolic_problem)
+
+        # Construct combined chain from the two
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+                                solver=None,
+                                gp=False):
+        """
+        Find candiate solvers for the current problem. If solver
+        is not None, it checks if the specified solved is compatible
+        with the problem passed.
+
+        Parameters
+        ----------
+        gp : bool
```
akshayka: nit: `solver` should be listed as the first argument, to match the function's argspec

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+            and `conic_solvers`.
+
+        Raises
+        ------
+        SolverError
+            Raised if the problem is not DCP and `gp` is False.
+        DGPError
+            Raised if the problem is not DGP and `gp` is True.
+        """
+        # Initialize candidates dictionary
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+                s for s in candidates['conic_solvers']
+                if slv_def.SOLVER_MAP_CONIC[s].MIP_CAPABLE]
+            if not candidates['conic_solvers'] and \
+                    not candidates['qp_solvers']:
+                raise SolverError("Problem is mixed-integer, but candidate "
+                                  "QP/Conic solvers (%s) are not MIP-capable."
+                                  % [candidates['qp_solvers'],
+                                     candidates['conic_solvers']])
+
+        # Return candidate qp and conic solvers
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,119 @@ def get_problem_data(self, solver):
...
+            if solver is not None and solver not in slv_def.CONIC_SOLVERS:
+                raise SolverError(
+                  "When `gp=True`, `solver` must be a conic solver "
+                  "(received '%s'); try calling " % solver +
+                  " `solve()` with `solver=cvxpy.ECOS`."
+                  )
+            elif solver is None:
+                candidates['qp_solvers'] = []  # No QP solvers allowed
+
+        # Check if we found MIP compatible solvers
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -368,20 +482,43 @@ def _solve(self,
...
 
+        # Check if chain already exists
         chain_key = (solver, gp)
+
         if chain_key != self._cached_chain_key:
             try:
-                self._solving_chain = construct_solving_chain(self,
-                                                              solver=solver,
-                                                              gp=gp)
+                # Find candidate solvers
```
akshayka: Since this block of code is identical to the code in get_problem_data, it deserves its own private method; i.e., add a method to Problem with signature `_construct_chains(self, solver, gp)` that finds candidates, constructs the symbolic chain, applies the symbolic chain, and constructs the solving chain

cvxpy/problems/problem.py
```diff
@@ -368,20 +482,43 @@ def _solve(self,
...
             except Exception as e:
                 raise e
+
         self._cached_chain_key = chain_key
 
-        data, inverse_data = self._solving_chain.apply(self)
-        solver_output = self._solving_chain.solve_via_data(
-          self, data, warm_start, verbose, kwargs)
-        self.unpack_results(solver_output, self._solving_chain, inverse_data)
+        # Apply solving chain to get data
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -368,20 +482,43 @@ def _solve(self,
...
+        data, inverse_data = self._solving_chain.apply(self._symbolic_problem)
+
+        # Solve via data
```
akshayka: delete comment

cvxpy/problems/problem.py
```diff
@@ -368,20 +482,43 @@ def _solve(self,
...
+        solution = self._solving_chain.solve_via_data(self, data,
+                                                      warm_start,
+                                                      verbose, kwargs)
+
+        # Unpack results
```
akshayka: delete comment

cvxpy/reductions/canonicalization.py
```diff
@@ -79,16 +79,7 @@ def canonicalize_tree(self, expr):
         return canon_expr, constrs
 
     def canonicalize_expr(self, expr, args):
-        if isinstance(expr, Expression) and not expr.variables():
```
akshayka: nice, this block of code was weird, glad it's gone.

cvxpy/reductions/solvers/symbolic_chain.py
```diff
@@ -0,0 +1,79 @@
...
+
+    reductions = []
+    if len(problem.variables()) == 0:
+        return Chain(reductions=reductions)
+    if complex2real.accepts(problem):
+        reductions += [Complex2Real()]
+    if gp:
+        reductions += [Dgp2Dcp()]
+
+    # Check DCP
```
akshayka: delete this comment and the other redundant ones

cvxpy/problems/problem.py
```diff
@@ -494,7 +631,10 @@ def unpack(self, solution):
         self._status = solution.status
         self._solution = solution
 
-    def unpack_results(self, solution, chain, inverse_data):
+    def unpack_results(self, solution,
+                       chain, inverse_data,
+                       symbolic_chain=None,
```
akshayka: Instead of adding new parameters to this function, can you instead concatenate the solving_chain and the symbolic_chain at the call-site (as you do in get_problem_data)?

cvxpy/reductions/solvers/symbolic_chain.py
```diff
@@ -0,0 +1,79 @@
...
+from cvxpy.problems.objective import Maximize
+from cvxpy.reductions import (Chain, Dcp2Cone,
+                              FlipObjective, Dgp2Dcp, Qp2SymbolicQp,
+                              CvxAttr2Constr, Complex2Real)
+from cvxpy.reductions.complex2real import complex2real
+from cvxpy.reductions.qp2quad_form import qp2symbolic_qp
+#  from cvxpy.reductions.solvers import defines as slv_def
+
+
+def construct_symbolic_chain(problem, candidates, gp=False):
```
akshayka: The names "symbolic" chain and "symbolic" problem are confusing, since all reductions and problems are by definition symbolic. (We used to have a confusing notion of symbolic data pre-1.0, and we still have a confusing notion of symbolic QPs :P ) 

To bring this terminology closer to software compilers, let's rename this to "intermediate_chain" , and let's say that this function "Builds a chain that rewrites a problem into an intermediate representation suitable for numeric reductions."

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -185,6 +118,11 @@ def __init__(self, problem=None, reductions=[]):
             raise ValueError("Solving chains must terminate with a Solver.")
         self.solver = self.reductions[-1]
 
+    def concatenate_with(self, chain):
```
akshayka: nit: rename to "prepend" to make the ordering of the concatenation clear

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -185,6 +118,11 @@ def __init__(self, problem=None, reductions=[]):
...
+        """Create and return a new Solving Chain by concatenating `chain` with this instance."""
```
akshayka: nit: Solving Chain --> SolvingChain

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -185,6 +118,11 @@ def __init__(self, problem=None, reductions=[]):
...
+        # Construct combined chain from the two
```
akshayka: nit: delete comment

cvxpy/reductions/solvers/symbolic_chain.py
```diff
@@ -0,0 +1,79 @@
+from cvxpy.error import DCPError, DGPError, SolverError
+from cvxpy.problems.objective import Maximize
+from cvxpy.reductions import (Chain, Dcp2Cone,
+                              FlipObjective, Dgp2Dcp, Qp2SymbolicQp,
+                              CvxAttr2Constr, Complex2Real)
+from cvxpy.reductions.complex2real import complex2real
+from cvxpy.reductions.qp2quad_form import qp2symbolic_qp
+#  from cvxpy.reductions.solvers import defines as slv_def
```
akshayka: delete

bstellato:
Thanks for the comments @akshayka. I made the edits you suggested. 

I totally agree on using `intermediate` instead of `symbolic`.  I did not really think about the name.

All the tests pass except the ones I described below. Do you have any idea why?

> * `test_mult_constant_atoms`: cvxpy complains that the `Product of two non-constant expressions is not DCP` even though one of the factors is constant.
> * I have added a test `test_parametric` in `test_qp.py` where I solve a parametric problem with and without parameters. When I use the parameters I get weird data passed to the solver. For example, that specific univariate problem gets additional variables from the construction and it becomes infeasible. EDIT: This happens only for QPs apparently.
>

SteveDiamond:
I fixed the first test and possibly the second. It's a small fix, on the pr664 branch.

cvxpy/problems/problem.py
```diff
@@ -315,12 +323,131 @@ def get_problem_data(self, solver):
...
+                self._intermediate_problem, self._intermediate_inverse_data = \
+                    self._intermediate_chain.apply(self)
+
+                self._solving_chain = \
+                    construct_solving_chain(self._intermediate_problem,
+                                            candidate_solvers)
+            except Exception as e:
+                raise e
+
+        self._cached_chain_key = chain_key
```
akshayka: Can you move this to the end of the try block? No need to overwrite _cached_chain_key if the key hasn't changed.

cvxpy/problems/problem.py
```diff
@@ -368,20 +495,21 @@ def _solve(self,
...
-                                                              gp=gp)
-            except Exception as e:
-                raise e
-        self._cached_chain_key = chain_key
+        self._construct_chains(solver=solver, gp=gp)
+
+        data, solving_inverse_data = \
+            self._solving_chain.apply(self._intermediate_problem)
+
+        solution = self._solving_chain.solve_via_data(self, data,
```
akshayka: I ran `TestQp.test_parametric` and dropped into pdb. I noticed that solve_via_data modifies `self._intermediate_problem` (!!). That's a bug, and that's the reason `test_parametric` is failing. We'll need to figure out which reduction is mutating its input --- reductions aren't supposed to modify their input problem.

bstellato: Wow you were faster than me in commenting :) That's exactly what I thought (see my comment below).

cvxpy/problems/problem.py
bstellato: Sure! Just fixed it in 8f56102

akshayka:
Hey @bstellato, yes, you're right! There's a bug in QpMatrixStuffing. The expression should be copied before the objective is stuffed, since apparently the coefficient extractor modifies the expression in place.

So making the following change to qp_matrix_stuffing.py should (hopefully) fix the problem. 

```
def stuffed_objective(self, problem, extractor):
   # extract to x.T * P * x + q.T * x + r
   expr = problem.objective.expr.tree_copy()
   P, q, r = extractor.quad_form(expr)

   ... [remainder of implementation stays the same as before]
```

SteveDiamond:
@akshayka is right. I did my own fix that undoes the extractor's changes (on the pr664 branch).

bstellato:
Thank you guys! It seems to be working now.

Before merging I noticed a local Mosek error that does not happen on the CI. Do you also get [this output](https://gist.github.com/bstellato/f31b10b885b24a5e4201ffe79ce5941a) if you have mosek locally installed?

cvxpy/problems/problem.py
```diff
@@ -315,12 +322,132 @@ def get_problem_data(self, solver):
...
+        inverse_data = self._intermediate_inverse_data + solving_inverse_data
+
+        return data, full_chain, inverse_data
+
+    def _find_candidate_solvers(self,
+                                solver=None,
+                                gp=False):
+        """
+        Find candiate solvers for the current problem. If solver
+        is not None, it checks if the specified solved is compatible
```
akshayka: nit: solved --> solver

cvxpy/problems/problem.py
```diff
@@ -494,7 +622,8 @@ def unpack(self, solution):
         self._status = solution.status
         self._solution = solution
 
-    def unpack_results(self, solution, chain, inverse_data):
+    def unpack_results(self, solution,
```
akshayka: super small nit: The signature of this function should (I think) fit on one line.

akshayka:
@bstellato , I can reproduce that error locally, at master. So it's not related to your changes.

cvxpy/problems/problem.py
bstellato: Fixed!

cvxpy/problems/problem.py
bstellato: Fixed!

PR #666, opened 2019-02-11
Correctly handle negative passband angles in beamforming example.

ebezzam:
In the beamformer design example, if the passband goes into negative angles (i.e. `(theta_tar - halfbeam)%360 > (theta + halfbeam)%360)`, these angles are incorrectly added in the stopband.

I remember this same bug in the original [Matlab code](http://web.cvxr.com/cvx/examples/antenna_array_design/html/ant_array_min_sidelobe.html) as well.


SteveDiamond:
Thanks!

PR #669, opened 2019-02-11
Update sdp example to the latest interface.

bamos:
Just sending in a quick fix for this example because I was looking at it.


SteveDiamond:
Thanks!

PR #688, opened 2019-03-13
Additional CPLEX status mapping.

tarpalsus:
Additional status mapping for CPLEX solver that allows optimal within tolerance solutions (code 102) to be considered valid and not raise errors.


SteveDiamond:
Thanks! Sorry, I hadn't realized this was so simple. I've merged in the change.

PR #707, opened 2019-04-18
Fasten problem transformation

rostyboost:
This PR:

1.  makes `matrix_stuffing` faster by amortizing the cost of turning constant matrices into expressions during iterations over the constraints

2. makes the Gurobi path faster when there is a large number of booleans/integer variables by avoiding lookups into large arrays.

These two changes combined lead to a 40% reduction of overall problem solve time for a (proprietary) dataset of 200 MIP problems:
image

The MIP problems have the following characteristics (first 3 rows are obtained by instrumenting `gurobi_qpif.py` around the spot modified, the other others are obtained through [SizeMetrics](https://www.cvxpy.org/api_reference/cvxpy.problems.html#sizemetrics)):
```
                  name |    mean |     std |     q25 |  median |     q75
        # boolean vars |   997.7 |   594.7 |   480.0 |  1056.0 |  1376.0
            # int vars |    29.9 |    18.6 |    14.0 |    30.0 |    40.5
           # cont vars |   991.1 |   600.8 |   484.0 |   996.0 |  1332.0
  num_scalar_variables |  1029.6 |   613.3 |   496.0 |  1088.0 |  1418.5
       num_scalar_data |  4072.9 |  2434.3 |  1892.0 |  4232.0 |  5792.0
  num_scalar_eq_constr |    14.9 |     9.3 |     6.8 |    15.0 |    20.2
 num_scalar_leq_constr |   162.2 |    43.3 |   126.0 |   167.0 |   192.0
    max_data_dimension |    63.2 |     5.0 |    64.0 |    64.0 |    64.0
 max_big_small_squared | 19916.8 | 20635.3 |  3136.0 | 14400.0 | 26256.0
```


akshayka:
@rostyboost , it looks like several tests are failing at this change.

cvxpy/reductions/matrix_stuffing.py
```diff
@@ -78,6 +79,9 @@ def apply(self, problem):
         # Batch expressions together, then split apart.
         expr_list = [arg for c in cons for arg in c.args]
         Afull, bfull = extractor.affine(expr_list)
+        Afull = cvxtypes.constant()(Afull)
```
akshayka: @rostyboost This doesn't work when the shape of Afull or bfull has a '0' in it, because Leaf's constructor requires each dimension to be > 0. You should guard these casts with something like:

`if 0 not in Afull.shape and 0 not in bfull.shape:`

rostyboost: Thanks @akshayka . What's the best way to run the unit tests locally without having all the backends installed? I only have the default open-source ones + Gurobi. But by default `nosetests .` runs some tests that are against other proprietary backends.

akshayka:
Are you sure you don't have a stale license for the proprietary solvers lying around? I just built CVXPY in a fresh virtual environment and ran `nosetests cvxpy/tests/` (with only the default solvers), and the tests ran without invoking any proprietary solvers.

akshayka:
@rostyboost , let me know if I can help! What do you see when you run the unit tests?

rostyboost:
Sorry @akshayka I was busy with other things... Will try to get the PR in a green state either today or tomorrow. I think most of the perf gains are coming from 2) for my dataset. I found 1) while running the python profiler on the calls and verified that with the PR the time spent in this part shrunk, but I'd like to generate some example problem demonstrating the gains for 1) alone.

rostyboost:
Ok @akshayka all the tests pass (CI jobs are queued but should be green once they finish), and I added a`test_stuffing_perf_many_constraints` in your recent `test_benchmark`. When running it with `iters=10` on the committed problem (2k constraints), I get with this PR:
```
$ python -m nose -s cvxpy/tests/test_benchmarks.py:TestBenchmarks.test_stuffing_perf_many_constraints
<lambda>: avg=1.894e+00 s , std=4.144e-02 s (10 iterations)
```
VS `(master)`:
```
$ python -m nose -s cvxpy/tests/test_benchmarks.py:TestBenchmarks.test_stuffing_perf_many_constraints
<lambda>: avg=2.349e+00 s , std=6.432e-02 s (10 iterations)
```
So the change to `matrix_stuffing` itself is about 20% faster for this example problem.
I verified that on my MIP problems though its impact is negligible, and so the 40% gains basically come from the Gurobi improvements.

I also added a tiny optim to only compute the problem size metrics if needed.

cvxpy/problems/problem.py
```diff
@@ -827,7 +826,7 @@ def __init__(self, problem):
             if self.max_data_dimension < big:
                 self.max_data_dimension = big
 
-            max_big_small_squared = np.int64(big)*np.int64(small)**2
+            max_big_small_squared = big*(small**2)
```
rostyboost: This is to avoid a `RuntimeWarning` overflow in large problems. Default to python multiplication instead which can't overflow.

SteveDiamond: The Python multiplication is what raised a warning.

rostyboost: @SteveDiamond which versions of python/numpy were you using? With my commit the warning doesn't show up in any of the CI logs: https://travis-ci.org/cvxgrp/cvxpy/builds/524531778?utm_source=github_status&utm_medium=notification

rostyboost: I noticed the warning because my change to `matrix_stuffing` changes the order of operations for `SizeMetrics` computation, and it triggered the overflow for [this test](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/tests/test_benchmarks.py#L45). However I also removed the computation of the `SizeMetrics` by default, so the warning isn't visible anymore regardless, since `SizeMetrics` isn't computed for this test anymore.

rostyboost: Without the lazy compute of `SizeMetrics` and without the change to the multiplication computation, you can see the warning here: https://travis-ci.org/cvxgrp/cvxpy/jobs/524479516

akshayka: I think this change makes sense. Integer overflow can't happen in Python 3, but it can happen in Python 2, when using ints instead of longs (3 got rid of long). We do support Python 2, so perhaps big and small should be cast to `long`s if the user is running Python 2 (which you can check with `sys`).

cvxpy/problems/problem.py
```diff
@@ -842,3 +843,6 @@ def __init__(self, problem):
         for constraint in problem.constraints:
             if isinstance(constraint, (Inequality, NonPos)):
                 self.num_scalar_leq_constr += constraint.expr.size
+
+    def __str__(self):
+        return ', '.join('%s=%s' % (k, v) for k, v in vars(self).items())
```
akshayka: This will also print private members, which we probably want to avoid. I think the current behavior for __repr__ / __str__ is okay. Can you revert this?

cvxpy/problems/problem.py
SteveDiamond: To comment further, the problems came up in Windows, where somehow 32 bit integers are used instead of 64 bit. Casting to long doesn't actually fix it. What's wrong with the way it is now?

rostyboost: @SteveDiamond the way it is now with the changes to [this file](https://github.com/cvxgrp/cvxpy/pull/707/files#diff-fe776d1461882082a19b60c402dc1cfd) from my PR triggers a warning in [this test](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/tests/test_benchmarks.py#L45) as can be seen in the logs of [this green build](https://travis-ci.org/cvxgrp/cvxpy/builds/524479506), both in 2.7 and 3.6 Linux and OSX. But interestingly it's not there in the associated Windows CI [build](https://ci.appveyor.com/project/StevenDiamond/cvxpy/builds/24095817) neither for 2.7 nor 3.6. 
Knowing that with my latest commit...
- all the`(linux, osx, windows) x (2.7, 3.6.X)` builds are green and none shows the warning anymore (see [here](https://travis-ci.org/cvxgrp/cvxpy/builds/524534780) and [here](https://ci.appveyor.com/project/StevenDiamond/cvxpy/builds/24099279))
- the value of `max_big_small_squared` is not really actionable in any of these cases
- [python2 countdown](https://pythonclock.org/)

... Should I spend more time on this? To be clear again, without this mitigation the warning shows on Linux and OSX.

cvxpy/problems/problem.py
rostyboost: done with https://github.com/cvxgrp/cvxpy/pull/707/commits/6c885565f54a6352daff73797a6446b474585f6c

cvxpy/problems/problem.py
SteveDiamond: No, thanks for clarifying. I misunderstood. It all looks good to me, thanks for making this PR!

PR #709, opened 2019-04-20
fixed typo in "changing the problem"

BartMassey:
Source referred to "prob2" where "prob3" was needed.


BartMassey:
Won't be signing the CLA. I'll add this to the issue tracker instead.

PR #715, opened 2019-04-26
OSQP takes max_iter instead of max_iters as an argument

r-barnes:
Fixes a syntax error in the OSQP solver's argument list


r-barnes:
Note that since some solvers require `max_iters` and OSQP requires `max_iter`, a cleaner interface might be to standardize some of the arguments sent to solvers so that, e.g., cvxpy internally translates `max_iters` to `max_iter` as appropriate.

akshayka:
Thanks!! :)

PR #721, opened 2019-05-03
DQCP

akshayka:
Implementation of DQCP (http://web.stanford.edu/~boyd/papers/dqcp.html)

Adds an `is_dqcp()` method for expressions. problem.solve(qcp=True) will parse a problem as a DQCP and solve it with bisection. Adds a reduction Dqcp2Dcp that takes a DQCP problem and emits a parameterized DCP problem.

### DIFF
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -246,6 +246,7 @@ class Atom(Expression):
                         (arg.is_convex() and self.is_incr(idx)) or
                         (arg.is_concave() and self.is_decr(idx))):
                     return False
+            return True
         else:
             return False
 
@@ -266,6 +267,7 @@ class Atom(Expression):
                         (arg.is_concave() and self.is_incr(idx)) or
                         (arg.is_convex() and self.is_decr(idx))):
                     return False
+            return True
         else:
             return False
 
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -17,6 +17,7 @@ limitations under the License.
 
 from cvxpy import utilities as u
 from cvxpy import interface as intf
+from cvxpy.expressions import cvxtypes
 from cvxpy.expressions.constants import Constant, CallbackParam
 from cvxpy.expressions.expression import Expression
 import cvxpy.lin_ops.lin_utils as lu
@@ -133,6 +134,16 @@ class Atom(Expression):
         """
         return False
 
+    def is_atom_quasiconvex(self):
+        """Is the atom quasiconvex?
+        """
+        return self.is_atom_convex()
+
+    def is_atom_quasiconcave(self):
+        """Is the atom quasiconcave?
+        """
+        return self.is_atom_concave()
+
     def is_atom_log_log_affine(self):
         """Is the atom log-log affine?
         """
@@ -215,6 +226,46 @@ class Atom(Expression):
                         (arg.is_log_log_convex() and self.is_decr(idx))):
                     return False
             return True
+        else:
+            return False
+
+    @perf.compute_once
+    def is_quasiconvex(self):
+        """Is the expression quaisconvex?
+        """
+        # Verifies the DQCP composition rule.
+        if self.is_convex():
+            return True
+        elif type(self) == cvxtypes.maximum():
+            return all(arg.is_quasiconvex() for arg in self.args)
+        elif self.is_scalar() and len(self.args) == 1 and self.is_incr(0):
+            return self.args[0].is_quasiconvex()
+        elif self.is_atom_quasiconvex():
+            for idx, arg in enumerate(self.args):
+                if not (arg.is_affine() or
+                        (arg.is_convex() and self.is_incr(idx)) or
+                        (arg.is_concave() and self.is_decr(idx))):
+                    return False
+        else:
+            return False
+
+    @perf.compute_once
+    def is_quasiconcave(self):
+        """Is the expression quasiconcave?
+        """
+        # Verifies the DQCP composition rule.
+        if self.is_concave():
+            return True
+        elif type(self) == cvxtypes.minimum():
+            return all(arg.is_quasiconcave() for arg in self.args)
+        elif self.is_scalar() and len(self.args) == 1 and self.is_decr(0):
+            return self.args[0].is_quasiconcave()
+        elif self.is_atom_quasiconcave():
+            for idx, arg in enumerate(self.args):
+                if not (arg.is_affine() or
+                        (arg.is_concave() and self.is_incr(idx)) or
+                        (arg.is_convex() and self.is_decr(idx))):
+                    return False
         else:
             return False
 
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -230,28 +230,32 @@ class Atom(Expression):
             return False
 
     @perf.compute_once
+    def _non_const_idx(self):
+        return [i for i, arg in enumerate(self.args) if not arg.is_constant()]
+
+    @perf.compute_once
     def is_quasiconvex(self):
         """Is the expression quaisconvex?
         """
         # Verifies the DQCP composition rule.
         if self.is_convex():
             return True
-        elif type(self) == cvxtypes.maximum():
+        if type(self) == cvxtypes.maximum():
             return all(arg.is_quasiconvex() for arg in self.args)
-        elif self.is_scalar() and len(self.args) == 1 and self.is_incr(0):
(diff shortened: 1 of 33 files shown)


cvxpy/atoms/atom.py
```diff
@@ -218,6 +229,57 @@ def is_log_log_concave(self):
         else:
             return False
 
+    @perf.compute_once
```
SteveDiamond: How does this compute_once work? Can we use it more places?

cvxpy/atoms/dist_ratio.py
```diff
@@ -0,0 +1,83 @@
...
+        if not self.args[2].is_constant():
+            raise ValueError("`b` must be a constant.")
+        self.a = self.args[1].value
+        self.b = self.args[2].value
+
+    @Atom.numpy_numeric
+    def numeric(self, values):
+        """Returns the distance ratio.
+        """
+        return np.linalg.norm(
```
SteveDiamond: This spacing looks kind of weird.

cvxpy/atoms/atom.py
akshayka: It's basically the same thing as lazy_prop, but not for properties. I've actually already sprinkled @perf.compute_once in various places at master: https://github.com/cvxgrp/cvxpy/search?utf8=%E2%9C%93&q=compute_once&type=

cvxpy/atoms/elementwise/ceil.py
```diff
@@ -0,0 +1,170 @@
...
+import scipy.sparse as sp
+
+
+class ceil(Elementwise):
+    """Elementwise ceiling."""
+
+    def __init__(self, x):
+        return super(ceil, self).__init__(x)
+
+    @Elementwise.numpy_numeric
```
SteveDiamond: Are we still doing numpy_numeric? I thought I had gotten rid of that.

cvxpy/reductions/dqcp2dcp/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+Copyright 2017 Robin Verschueren
```
SteveDiamond: Is this intended?

cvxpy/reductions/dqcp2dcp/dqcp2dcp.py
```diff
@@ -0,0 +1,201 @@
...
+    def _canonicalize_tree(self, expr):
+        canon_args, constrs = self._canon_args(expr)
+        canon_expr, c = self.canonicalize_expr(expr, canon_args)
+        constrs += c
+        return canon_expr, constrs
+
+    def _canon_args(self, expr):
+        """Canonicalize arguments of an expression.
+
+        Like Canonicalization.canonicalize_tree, but preserves signs.
```
SteveDiamond: Why doesn't canonicalize tree preserve signs? Aren't the variables the same?

cvxpy/reductions/solvers/bisection.py
```diff
@@ -0,0 +1,189 @@
...
+def _find_bisection_interval(problem, t, solver=None, low=None, high=None):
+    """Finds an interval for bisection."""
+    if low is None:
+        low = 0 if t.is_nonneg() else -1
+    if high is None:
+        high = 0 if t.is_nonpos() else 1
+
+    infeasible_low = t.is_nonneg()
+    feasible_high = t.is_nonpos()
+    for _ in range(100):
```
SteveDiamond: Should this 100 be a parameter?

cvxpy/reductions/solvers/intermediate_chain.py
```diff
@@ -49,13 +49,19 @@ def construct_intermediate_chain(problem, candidates, gp=False):
...
+            append = (" However, the problem does follow DQCP rules. "
+                      "Consider calling solve() with `qcp=True`.")
         raise DCPError("Problem does not follow DCP rules." + append)
     elif gp and not problem.is_dgp():
         append = ""
         if problem.is_dcp():
             append = (" However, the problem does follow DCP rules. "
-                      "Consider calling this function with `gp=False`.")
+                      "Consider calling solve() with `gp=False`.")
+        if problem.is_dqcp():
```
SteveDiamond: elif?

SteveDiamond:
I did a quick look over and it seems pretty good. I made some random comments. I don't totally know what's happening in the code. You said you added lazy constraints. I would have imagined that the quasiconvex atoms would actually take a parameter when they were canonicalized, rather than making some kind of constraint and then canonicalizing that. Maybe I misunderstood your approach. It seems kind of indirect. 

What's an example of an atom that can't be canonicalized directly into a constraint with a parameter? Just trying to understand. Would it be like ``length(x) <= negative number``?

cvxpy/atoms/elementwise/ceil.py
akshayka: I copied it from some another atom's code. Deleted.

cvxpy/reductions/dqcp2dcp/__init__.py
akshayka: fixed, thanks

cvxpy/reductions/dqcp2dcp/dqcp2dcp.py
akshayka: In regular canonicalization, when an expression like

nonlinear_atom(my_variable(pos=True))

is canonicalized, `nonlinear_atom` is replaced with a new slack variable, which doesn't have any attributes associated with it. This doesn't really matter for Dcp2Cone, Dgp2Dcp, or QP2QuadForm, but it does matter for Dqcp2Dcp. Should I move the logic in _canon_args into Canonicalization?

cvxpy/reductions/solvers/bisection.py
akshayka: done

cvxpy/reductions/solvers/intermediate_chain.py
akshayka: done, nice catch

cvxpy/atoms/dist_ratio.py
akshayka: I think it's standard? I can't think of a cleaner way to fit it that doesn't require `\` or wrapping the entire expression in (), both of which imo would make it harder to read.

akshayka:
Yes, `length(arg) <= t` is a good example. You can take a look at the `length_sub` function in dqcp2dcp/sets.py, which constructs the sublevel set for length.

The natural way to write the t-sublevel set of len(arg) <= t is something like arg[floor(t):] == 0. But if t < 0, the constraint is infeasible (and can't be written in the aforementioned natural way), and if t > arg.size, then arg[floor(t):] is invalid syntax and will throw an error. Because t might be the bisection parameter, we don't know its value during canonicalization. When / if the bisection algorithm passes in t < 0 or t > arg.size in this example, we need a way to communicate that the constraint is infeasible or that the sublevel set is all of R^n, respectively. This lazy evaluation was the simplest way I could think of to handle this.

akshayka:
Not sure what's going on with the windows build ... will take a look. @SteveDiamond , I'll let you know when the tests are passing.

cvxpy/atoms/dist_ratio.py
SteveDiamond: ok thanks, these are just random comments.

cvxpy/reductions/dqcp2dcp/dqcp2dcp.py
SteveDiamond: No, don't change it. The variable attributes would produce constraints, which we don't want.

SteveDiamond:
Ok sounds good. My guess would be something related to integer overflow (Windows has 32 bit integers for Python int, I think). I looked at the code more, and I understand your point better. @bstellato was actually suggesting lazy constraints for MIP problems, so it might be a concept worth introducing more broadly.

akshayka:
Yes, it's probably under/overflow ... but in the solvers (I think). The problems that are failing have constraints like x >= inv_pos(y) * -t, y >= 0, and the LHS might be blowing up ... I'll try with ECOS instead of SCS and see what happens

akshayka:
> @bstellato was actually suggesting lazy constraints for MIP problems, so it might be a concept worth introducing more broadly.

Yeah definitely! I don't yet have a good feeling for what the proper abstraction for lazy constraints would be, which is why I implemented them as a one-off thing for now. Once we collect more use cases, we can look at the requirements and design a good abstraction.

akshayka:
@SteveDiamond , the tests are now passing.

SteveDiamond:
Great! Merging in.

Issue #724, opened 2019-05-06
pip install fails with python 2.7


EnricoBeltramo:
Laungching the command:
pip install cvxpy

the installation fails with follow log:

~/VideoPopup$ pip install cvxpy
Collecting cvxpy
  Using cached https://files.pythonhosted.org/packages/82/d5/0b787694c85297e36d3afd4c8574973809a7cc52aa2c602413012da8999d/cvxpy-1.0.21.tar.gz
Collecting ecos>=2 (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/69/f1/f46f338769b79a0d98093d0911744e1aba7595ca70d4119c0d22e38205ae/ecos-2.0.7.post1-cp27-cp27mu-manylinux1_x86_64.whl
Collecting fastcache (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/5f/a3/b280cba4b4abfe5f5bdc643e6c9d81bf3b9dc2148a11e5df06b6ba85a560/fastcache-1.1.0.tar.gz
Collecting multiprocess (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/31/60/6d74caa02b54ca43092e745640c7d98f367f07160441682a01602ce00bc5/multiprocess-0.70.7.tar.gz
Requirement already satisfied: numpy>=1.15 in /home/ulix/.local/lib/python2.7/site-packages (from cvxpy) (1.16.3)
Collecting osqp (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/cd/97/94c24a224560f043e3a2ba5ca199b59d6427a02b22b13f38fb4b3d36a623/osqp-0.5.0-cp27-cp27mu-manylinux1_x86_64.whl
Collecting scipy>=1.1.0 (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/81/39/f1457091d0a45a84a2bd7815e2cf6bd45d4fe240728e9ed567cbb17c8abe/scipy-1.2.1-cp27-cp27mu-manylinux1_x86_64.whl
Collecting scs>=1.1.3 (from cvxpy)
  Using cached https://files.pythonhosted.org/packages/07/a7/0b19c8f9476a4762d296c6c5fa860f2fe580a4f579fa53aaa8515f4ca217/scs-2.1.0.tar.gz
    Complete output from command python setup.py egg_info:
    Namespace(blas64=False, extraverbose=False, float32=False, gpu=False, int32=False, scs=False)
[...]

PR #732, opened 2019-05-22
add pyproject.toml

ali5h:
since `pip 10`, `pyproject.yaml` is supported. It can be used to have `pip`
install `numpy` before running `setup.py`. So we can remove custom `numpy`
injection logic.


SteveDiamond:
Thanks! What will happen for people with older pip? Can you add a pip version as a requirement?

ali5h:
they will have to install `numpy` first or upgrade `pip`

ali5h:
you cannot add pip min version, same as `setuptools`

SteveDiamond:
@akshayka what do you think? I really appreciate the contribution, and it is cleaner, but I'm hesitant to break the installation for old versions of pip.

akshayka:
Is the NumPy injection a big issue? I suspect most people using CVXPY will already have NumPy installed, whereas many people might not have pip 10, which is only a year old. So the lesser of two evils, at least right now, might be to keep things as is, even if adding a pyproject.toml file is the correct thing to do.

ali5h:
if you `git blame` I added this header injection myself, but later found this better solution. So this was failing already for those who don't have `numpy` installed.

akshayka:
Oh, I wasn't aware of the history of the injection logic --- thanks for your contributions! :)

pip 10 seems like it came out just over a year ago (https://pip.pypa.io/en/stable/news/#id119). I wouldn't be surprised if many folks haven't upgraded yet, but I have no data on this. Have pip 10 / pyproject.toml's been adopted widely / how can we find out if they've been adopted widely?

SteveDiamond:
Could we leave the numpy injection and also have the pyproject.toml file? Maybe that would help with #724.

ali5h:
yes, i think it does help with #724, at least in my local tests

SteveDiamond:
Ok, I added a pyproject.toml. I'll leave the numpy injection as it is in case people don't have pip 10. Thanks for your contribution!

Issue #627, opened 2018-11-11
have AppVeyor deploy Windows builds to conda and pypi


rileyjmurray:
### Problem (and current "solution")
Right now cvxpy's documentation recommends that people use conda to get the latest version of cvxpy. This is fine advice for linux and mac users, but it doesn't work well for Windows users. The only way for a Windows user to get the latest version of cvxpy is to install from a source distribution hosted on [pypi](https://pypi.org/project/cvxpy/).

If a Windows user doesn't already use conda, then they simply install cvxpy by running ``pip install cvxpy`` in the command line. However, if a Windows users needs the most up-to-date version of cvxpy in a conda virtual environment, the installation process is more complicated. Specifically, if their desired conda virtual environment is named "myEnv", then they should run two pairs of commands:
```
activate myEnv
conda install pip
```
and
```
activate myEnv
pip install cvxpy
```
The first line in the second block of commands isn't really necessary, but it helps make the procedure fool-proof (in the event that the commands are not executed in the same command line window).

### Proposed changes

Although the above approach should get the job done, it is unnecessarily complicated. Our AppVeyor scripts should be updated so that Windows binaries are uploaded to the ``cvxgrp'' conda channel whenever AppVeyor builds succeed. 

On a related note, @bstellato suggests in https://github.com/cvxgrp/cvxpy/issues/615#issuecomment-434050947 that we upload *binaries* (in addition to source distributions) to pypi. Pypi allows Windows and OSX binaries natively, and also allows a certain type of Linux build ("[manylinux](https://www.python.org/dev/peps/pep-0513/#pypi-support)"). A full resolution of this issue should include pushing Windows binaries to pypi upon successful AppVeyor builds.

The main obstacle in making these changes is that I don't know how AppVeyor handles censoring sensitive environment variables (like conda usernames and passwords) that we will need for the upload process. (Right now our AppVeyor continuous integration scripts do not upload any distribution of cvxpy upon a successful build; the source distribution which Windows users currently rely on is in fact uploaded after successful TravisCI builds.)

bstellato:
Appveyor can encrypt variables just as travis. Check out the [secure variables](https://www.appveyor.com/docs/build-configuration/#secure-variables) section of their docs. That's what we have for the [osqp recipes](https://github.com/oxfordcontrol/osqp-recipes/blob/master/appveyor.yml).

[...]

PR #734, opened 2019-05-24
Update continuous integration scripts (mostly deployment related)

rileyjmurray:
This PR contains changes to how we use continuous integration services (TravisCI and AppVeyor). There are three types of changes here:
1. Simple reorganization of CI files (to improve readability / maintainability).
2. Correction of an error in meta.yaml, which caused conda builds to fail after [this commit](https://github.com/cvxgrp/cvxpy/commit/b42d5382a3e86047ad9fc0958fa2fb063520316a#diff-ca1ae5e1c32bb3b74558c6766de4f0c2).
3. New deployment features.

The new deployment features are as follows:
 - Windows wheels are uploaded to pypi. These work 100%.
 - Windows builds are pushed to Anaconda. These install correctly, but contain a poorly linked version of SCS, which does not have support for semidefinite constraints of order > 2. As a result, a Windows user can run ``conda install cvxpy`` and get to most up-to-date version of cvxpy, but they will need to separately install SCS (``pip install scs<=2.0``) in order to run tests. From extensive discussion in #627, there is no way around this poorly functioning Windows build without someone updating an SCS conda package.

Separately, I'll note that we are not far off from pushing "manylinux" wheels. However that will require changes to how the Travis "deploy_script.sh" is setup, and I'd like to make those changes in a separate PR.

In order to properly merge this PR, I need @SteveDiamond to setup the following environment variables from the TravisCI and AppVeyor websites, with their personal accounts (most of these should already be defined, at least for TravisCI):
 - CONDA_UPLOAD_TOKEN
 - CONDA_USER
 - PYPI_USER
 - PYPI_PASSWORD
 - PYPI_SERVER

The last of the above environment variables should be ``PYPI_SERVER=https://upload.pypi.org/legacy/``. If a cvxpy contributor needs to test deployments themselves (i.e. using their own fork of cvxpy), then they also need to set the above environment variables. However, they should set ``PYPI_SERVER= https://test.pypi.org/legacy/``. It may be good to update [CONTRIBUTING.md](https://github.com/cvxgrp/cvxpy/blob/master/CONTRIBUTING.md) to include a section on how to help with deployment and CI related issues. Such a section should (at the very least) mention the need to use the appropriate value for ``PYPI_SERVER``.


SteveDiamond:
Sorry for the slow reply! I was out of town. This all looks great! I added the environment variables you mentioned to Travis and Appveyor.

akshayka:
@SteveDiamond Is this ready to be merged?

@rileyjmurray , adding a deployment/CI section to our contributing guide is an excellent idea! Would you (or Steven) like to write a first draft, whenever you have time? I'm actually not very familiar with our CI process, so I'm looking forward to reading the draft to get myself up-to-speed :)

SteveDiamond:
@akshayka I believe it's ready to be merged. @rileyjmurray is that right? I've set the environment variables.

rileyjmurray:
It is indeed ready to be merged.

@akshayka I'll update the CI-related contributing guidelines in my next PR (which will be focused on adding deployment of manylinux wheels to pypi).

akshayka:
Merged, but the Python 2.7 travis tests are failing: https://travis-ci.org/cvxgrp/cvxpy/jobs/539536434

akshayka:
@rileyjmurray , do you know what might be wrong with the 2.7 Travis builds?

rileyjmurray:
The error messages are pretty awful, so I cannot say for sure. This only happens in the deployment step (install and tests succeed). There seems to be a difference where conda-build is using numpy 1.16 and scipy 1.2 (although the install and tests are with numpy 1.14 and scipy 1.1). The higher versions of these packages during conda-build may have inherent issues with python 2.7. If nothing else, the conda-recipe should be changed to have the same package version requirements as in setup.py.

akshayka:
> If nothing else, the conda-recipe should be changed to have the same package version requirements as in setup.py.

Interesting -- I can try to fix this ...

I'm unsure why the source deployment to PyPi failed at master, the builds with DEPLOY_PYPI=True succeeded ... I'll create a separate issue for that.

rileyjmurray:
It would be helpful if you looked into it a bit. I can take a look later this next week but I’m pretty slammed for the next few days.

PR #736, opened 2019-05-29
MIQP example

jonathantuck:
adding an example for MIQPs

### DIFF
--- a/examples/notebooks/WWW/MIQP.ipynb
+++ b/examples/notebooks/WWW/MIQP.ipynb
@@ -1,0 +1,117 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


akshayka:
Looks mostly good!

Very minor comments:
- Change the title to "Mixed-integer quadratic program" to match our other intro examples
- Define Z^n the first time you use it
- You define script I, but never use it
- "The solution" --> "A solution" , "the vector" --> "a vector", to emphasize that solutions are not necessarily unique
- Add a subheading "Example" and a sentence saying that you show how to specify and solve an MIQP using CVXPY (as in https://www.cvxpy.org/examples/basic/least_squares.html)
- (m,n) --> (m, n)
- do not specify the constraints (i.e., write prob = cp.Problem(objective))

akshayka:
Looks great, thanks @jonathantuck !!

PR #743, opened 2019-06-05
Missing status code in cvxopt STATUS_MAP

sjschneider:
Missing status code for "LP relaxation is dual infeasible" from CVXOPT: https://github.com/cvxopt/cvxopt/blob/master/src/C/glpk.c#L462


SteveDiamond:
Thanks!

Issue #747, opened 2019-06-13
Calling get_num_barrier_iterations() without checking whether model is a MIP


rkersh:
**Describe the bug**

In the `cplex_qpif.py` file, the `get_num_barrier_iterations()` method is called without first checking whether the model is a MIP or not. If `verbose=True` and the model is a MIP, an innocuous error message will be printed in the output. This can confuse the user.

**To Reproduce**
```
import numpy as  np
import cvxpy as cp

n = 20

ret = np.array(
    [0.633847, 0.0224574, 0.787235, -1.45919, -0.858576,
     -0.90331, 0.513264, -0.595323, 0.452494, 0.475518,
     -1.02247, 0.0798508, -0.120809, 0.6326, -0.274057,
     -0.450731, 0.709198, -0.196911, -0.101243, 0.178535]).reshape(20,1)

lamb = 5

Shorts = cp.Variable(n)
Longs = cp.Variable(n)

constraints=[]

constraints += [sum(Shorts) == -1]
constraints += [sum(Longs) == 1]

constraints += [Shorts >= -1]
constraints += [Shorts <= 0]

constraints += [Longs >= 0]
constraints += [Longs <= 1]

# Big M
M = 10
I = cp.Variable(n, boolean=True)

constraints += [Longs <= M * I]
constraints += [-Shorts <= M * (1 -I)]

obj = cp.Minimize(lamb * cp.sum_squares(Shorts + Longs) -
                  ret.T * Shorts - ret.T * Longs)
prob = cp.Problem(obj, constraints)

obj11 = prob.solve(solver=cp.CPLEX, verbose=True)

if prob.status == 'optimal':
    ww = np.vstack((Shorts.value,Longs.value)).T
    wt = np.sum(ww,axis=1)
    summ = [prob.solver_stats.solve_time, sum(wt), sum(abs(wt)), obj11]
    print("SUCCESS")
else:
    print(prob.status)
```

**Expected behavior**

Expecting the the model solves without any issues and no confusing error messages are shown in the output.

**Output**

With `verbose=True`, we get the following output:
```
...
CPLEX Error  1217: No solution exists.
SUCCESS
```
The error message makes the user think that something bad happened.

**Version**
 - OS: Any 
 - CVXPY Version:  1.0.24

**Additional context**

This was reported on the CPLEX forum here:
https://www.ibm.com/developerworks/community/forums/html/topic?id=97b21616-8835-4574-8cc9-d09aae0ed7f4

[...]

PR #748, opened 2019-06-13
Fix issue #747

rkersh:
In `cplex_qpif.py`, we should check that `inverse_data.is_mip` is `False`
before calling `get_num_barrier_iterations()`.


SteveDiamond:
Looks good, thanks!

PR #758, opened 2019-07-01
Fix AppVeyor builds (failing due to conda update)

rileyjmurray:
The AppVeyor builds are failing for some anaconda-associated reason. Specifically, the ``conda update conda -y`` command is failing in a nasty way. I have commented out that line (presumably it's a bug on Continuum Analytics end, so it will be fixed eventually). I have also switched the miniconda version from 2 to 3. Switching version from 2 to 3 makes sense, because all Windows builds are using Python 3.

You can refer to my fork's AppVeyor build status here: https://ci.appveyor.com/project/rileyjmurray/cvxpy.


rileyjmurray:
Oh I should probably explain: it used to be that ``conda update conda`` was necessary, because the AppVeyor scripts assume we're working with Anaconda version >= 4.6. When the AppVeyor scripts were first written, the file ``Miniconda2-latest-Windows-x86_64.exe`` provided Anaconda version 4.5. Now that ``Miniconda3-latest-Windows-x86_64.exe`` provides Anaconda 4.6, the update command is not strictly necessary.

PR #779, opened 2019-08-07
Minor optimization for Gurobi

rostyboost:
We can avoid the `tocoo()` conversion on the quadratic part in the Gurobi path if there are no quadratic terms. Avoids an unnecessary allocation of a potentially big matrix of zeros.


cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -172,7 +174,6 @@ def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=No
 
         # Set verbosity and other parameters
         model.setParam("OutputFlag", verbose)
-        # TODO user option to not compute duals.
```
rostyboost: passing `QCPDual=False` to `problem.solve` already works, as it's passed in through `solver_opts` args.

SteveDiamond:
Ok, nice catch!

PR #789, opened 2019-08-20
tutorial/advanced: use both "data" and "probdata" variables

link2xt:
Previously, "probdata" variable was constructed
but not passed to SCS. Instead, "data" was passed,
which contained more elements than required.

Now get_problem_data result is renamed to "probdata",
and "data" is constructed, so scs.solve(data, cones)
actually uses constructed dictionary.


SteveDiamond:
Thanks for fixing this! get_problem_data will be reworked in cvxpy 1.1. It's not in a great state right now.

PR #801, opened 2019-09-03
Copy swig import helper from branch 1.1

EQt:
this avoids the warning that `imp` has been deprecated in Python 3.4


SteveDiamond:
This may have broken the build. Not clear to be honest if it's because of your change.

EQt:
Seems like the issues with installing a working version of OSQP on Travis have been resolved.
The problems on AppVeyor seem to be rather fundamental and not related to the changes.

By the way, you can switch deprecation warnings on by setting the environment variable `PYTHONWARNINGS=d`, e.g.
```bash
PYTHONWARNINGS=d python -c 'import imp'
```

SteveDiamond:
Great, thanks!

PR #802, opened 2019-09-04
cbc mip status extension

BastiaanvG:
Added some CBC exit status messages,
I will also update the documentation to include the solver settings we can actually set with cvxpy


SteveDiamond:
Looks good! Sorry I missed this.

PR #832, opened 2019-10-08
Migrate examples from python2 to python3

tataudat:
Most of examples are in python2 (print, imap, izip, xrange) and quite outdated with the current state of `cvxpy` (still use `BoolVar()` instead of `Boolean()`, `vstack` with 2 variables instead of a list).

The goal of this PR is to update the examples to python3 and the latest version of `cvxpy`.

### DIFF
--- a/examples/admm_example.py
+++ b/examples/admm_example.py
@@ -75,10 +75,10 @@ def main():
 
 		iters = iters + 1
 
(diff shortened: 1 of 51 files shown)


akshayka:
This looks great. Thank you!

PR #833, opened 2019-10-09
Sbarratt/perspective

sbarratt:
Can I get some initial feedback on this? Not sure if I'm doing everything correctly, but it seems to work...


akshayka:
The approach looks good. For the copyright text, attribute the copyright to "the CVXPY authors" instead of yourself, since the code might be touched by others in the future: https://github.com/cvxgrp/cvxpy/blob/master/copyright.txt

sbarratt:
@moehle Can you take a look at this?

sbarratt:
@akshayka what do I need to do in terms of documentation?

cvxpy/atoms/perspective.py
```diff
@@ -0,0 +1,140 @@
...
+limitations under the License.
+"""
+
+from cvxpy.atoms.atom import Atom
+import numpy as np
+import warnings
+
+
+class perspective(Atom):
+    """ :math:`\text{perspective}(f)(x, t) = tf(x/t)` """
```
akshayka: Example usage?

cvxpy/atoms/perspective.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, *args, atom=None):
```
akshayka: Docstring explaining what `*args, atom` should be

cvxpy/reductions/dcp2cone/atom_canonicalizers/perspective_canon.py
```diff
@@ -0,0 +1,83 @@
...
+    t = args[-1].flatten()
+
+    try:
+        underlying_canonicalizer = atom_canonicalizers.CANON_METHODS[type(expr._atom_initialized)]
+    except KeyError:
+        raise ValueError(f"Cannot take perspective of {expr._atom}. "
+                         f"Must be able to canonicalize {expr.atom}.")
+    s, constraints_underlying = underlying_canonicalizer(
+        expr._atom_initialized,
+        expr._atom_initialized.args
```
akshayka: The canonicalizers assume that the arguments have been canonicalized, i.e. that they are affine. What happens if the expressions `expr._atom_initialized.args` aren't affine? (Maybe if they're not affine, then DCP verification will fail and you'll never reach this code ...?)

Add a unit test for that case.

cvxpy/reductions/dcp2cone/atom_canonicalizers/perspective_canon.py
```diff
@@ -0,0 +1,83 @@
...
+    If we have:
+        f(x) <= s <==> Ax + bs + c \in \mathcal K
+    Then:
+        perspective(f)(x, t) <= s <==> Ax + bs + c + c(t-1) \in \mathcal K
+    See https://web.stanford.edu/~boyd/papers/pdf/sw_aff_ctrl.pdf
+    """
+    t = args[-1].flatten()
+
+    try:
+        underlying_canonicalizer = atom_canonicalizers.CANON_METHODS[type(expr._atom_initialized)]
```
akshayka: 80 cols

cvxpy/tests/test_atoms.py
```diff
@@ -976,3 +976,32 @@ def test_indicator(self):
...
+        n = 5
+        P = np.random.randn(n, n)
+        P = P @ P.T + 1e-3*np.eye(n)
+        x = cp.Variable(n,)
+        t = cp.Variable(1,)
+
+        x.value = np.ones(n)
+        t.value = np.ones(1)*1
+
+        def quad_over_lin(x, t):
```
akshayka: Test other atoms as well, since the perspective can be applied to basically any atom.

cvxpy/atoms/perspective.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        args = values[:-1]
+        t = values[-1].flatten()[0]
+
+        if t > 0:
+            args = [a / t for a in args]
+            return t * self._atom(*args).value
+        elif t == 0:
+            # TODO(sbarratt): make this correct for every atom
+            warnings.warn(f"Assuming {self._atom} is superlinear. If not, \
```
akshayka: Instead of allowing users to take the perspective of any atom, how do you feel about creating a small whitelist of supported atoms and growing that whitelist over time? That way you could be confident that you didn't return incorrect values. Otherwise, if you allow every atom, that sort of puts an obligation on anyone who adds an atom to make sure that this method works for their atom.

sbarratt:
Due to the limited scope of this atom if we need a whitelist of atoms, I’ve
decided to put this PR on hold. It might however, become interesting, if we
can take the perspective of general Expressions.

cvxpy/tests/test_atoms.py
rileyjmurray: IMO a good test profile would hit the following points

* Does perspective of x^2 recover the rotated second order cone? This can be tested by comparison to quad_over_lin.
* Does perspective of exp(x) suitably recover the exponential cone? This can be tested by comparison to a small modification of kl_div.
* Does taking a perspective of a perspective behave correctly? Testing with a single case seems reasonable here.
* Does perspective play nice with existing atoms that are technically *already* perspectives? The main atoms I have in mind are quad_over_lin and kl_div.

As a secondary consideration, one can consider homogeneous atoms. I think this isn't as important, because people who use the perspective function should know to not bother doing this with homogeneous functions. If you wanted to write tests for this case, you could consider homogeneous functions which rely on different underlying cones. So you could write one test for an LP-representable atom (like max), one for an SOCP representable atom (like the 2 norm), and one for an SDP-representable atom (like the nuclear norm).

Issue #834, opened 2019-10-10
cvcore.i SWIG interface is licensed under GPLv3


alexdyck:
Hello,

as I am really happy to use cvxpy, I was checking the commercial usability in terms of licensing. 

I found that the [SWIG interface file for cvxcore](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/cvxcore/python/cvxcore.i) is licensed under GPLv3. 

As the corresponding __init__.py and all other files are licensed under the Apache 2.0 license, I wanted to ask whether this is on purpose or was not updated when switching cvxpy to the Apache 2.0 license.

If this file was overlooked, it would be great, to change the license paragraph of the file, so it becomes commercially usable without having to license my software under GPLv3.

Kind Regards,
Alex

akshayka:
Interesting catch ... my guess is that it's a mistake. cvxcore is a fork of cvxcanon: https://github.com/cvxgrp/CVXcanon

@SteveDiamond , do you know whether the license is supposed to be Apache 2.0?

SteveDiamond:
The license is supposed to be Apache 2.0. Let's change the file.

PR #836, opened 2019-10-11
Fix license in cvxcore.i (#834)

alexdyck:
This should be the correct license paragraph as it is taken from the corresponding [cvxcore init file](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/cvxcore/python/__init__.py)


akshayka:
I believe SWIG uses C-style comments; please prefix each line with  `//` and delete the quotes. Once that's done, we'll merge it in.

PR #838, opened 2019-10-15
Remove extra verb from the operators subsection at the atomic functions section

viniciusd:
At https://www.cvxpy.org/tutorial/functions/index.html#operators, there is an extra `are`.


SteveDiamond:
Thanks! Nice catch.

viniciusd:
Great!
Thanks for the comprehensive docs

PR #839, opened 2019-10-16
setup.py should be executable

marcusmueller:
Without this, the standard ways of installing setup.py-bearing Python projects
doesn't work.

This is a file permissions-only fix.


akshayka:
I’ve looked at setup.py files for a few other popular projects, and none of them make their script executable. Can you provide evidence that this is standard / what kind of workflows require your change?

PR #841, opened 2019-10-18
1.1 alpha

akshayka:


akshayka:
Travis is passing (with the exception of Python 2.7, which is fine since we're dropping 2.7 support in 1.1)

PR #845, opened 2019-10-22
Update for Gurobi v9 API

rluce:
This PR implements the transition to the public, stable matrix friendly API of Gurobi's gurobipy.  This is a backwards compatible change, but we may consider to drop the usage of the temporary API functions starting with `_v811_` for reasons of code hygiene once Gurobi version 8.1.1 is no longer the widespread version.

I verified locally through `nosetests` that the change does not increase the number of failed tests (some were failing already in master).


SteveDiamond:
Looks good! Thanks!

PR #846, opened 2019-10-22
Differentiation

akshayka:
Add methods `backward`, `derivative` to `problem` for native differentiation. Add `diffcp` as an optional solver.


akshayka:
@SteveDiamond The build is passing; the failure is spurious (a Travis timeout).

PR #856, opened 2019-10-29
Keep zeros

akshayka:
When differentiating, keep zero entries in problem data that are affected by parameters

### DIFF
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -58,8 +58,35 @@ def get_parameter_vector(param_size,
     return param_vec
 
 
+def nonzero_csc_matrix(A):
+    # this function returns (rows, cols) corresponding to nonzero entries in
+    # A; an entry that is explicitly set to zero is treated as nonzero
+    assert np.nan not in A.data
+    zero_indices = (A.data == 0)
+    # scipy drops rows, cols with explicit zeros; use nan as a sentinel
+    # to prevent them from being dropped
+    A.data[zero_indices] = np.nan
+    A_rows, A_cols = A.nonzero()
+    ind = np.argsort(A_cols, kind='mergesort')
+    A_rows = A_rows[ind]
+    A_cols = A_cols[ind]
+    A.data[zero_indices] = 0
+    return A_rows, A_cols
+
+
+def A_mapping_nonzero_rows(problem_data_tensor, var_length):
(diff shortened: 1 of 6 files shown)


cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -58,8 +58,37 @@ def get_parameter_vector(param_size,
     return param_vec
 
 
+def nonzero_csc_matrix(A):
+    # this function returns (rows, cols) corresponding to nonzero entries in
+    # A; an entry that is explicitly set to zero is treated as nonzero
+    assert np.nan not in A.data
```
sbarratt: I think you need to change this line.
```python
import numpy as np                                                     
x = np.array([1., np.nan, 0.])                                         
np.nan in x # False
np.isnan(x).any() # True
```

cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -58,8 +58,37 @@ def get_parameter_vector(param_size,
...
+    zero_indices = (A.data == 0)
+    # scipy drops rows, cols with explicit zeros; use nan as a sentinel
+    # to prevent them from being dropped
+    A.data[zero_indices] = np.nan
+    A_rows, A_cols = A.nonzero()
+    ind = np.argsort(A_cols, kind='mergesort')
```
sbarratt: maybe comment why we're doing this

cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -58,8 +58,37 @@ def get_parameter_vector(param_size,
...
+def A_mapping_nonzero_rows(problem_data_tensor, var_length):
+    # get the rows in the map from parameters to problem data that
+    # have any nonzeros
+    problem_data_tensor_csc = problem_data_tensor.tocsc()
+    A_nrows = problem_data_tensor.shape[0] // (var_length + 1)
+    A_ncols = var_length
+    A_mapping = problem_data_tensor_csc[:A_nrows*A_ncols, :-1]
+    # don't call nonzero_csc_matrix, because here we don't want to
+    # count explicit zeros
+    A_mapping_nonzero_rows, _ = A_mapping.nonzero()
```
sbarratt: Can we take the set of A_mapping_nonzero_rows? Because duplicates don't matter. Or maybe this is just better addressed in the sparse matrix creation.

PR #860, opened 2019-11-03
support functions (transform, and atom)

rileyjmurray:
This PR focuses on the content of my accidental merge from earlier today. It also contains fixes to the MOSEK interface.

Users can get a handle on a ``SuppFunc`` object by calling ``sigma = cvxpy.suppfunc(x, constrs)``. From there, ``SuppFuncAtom`` objects are generated by evaluating the ``SuppFunc`` object. So a user could write ``sigma(y) <= 1``, and this generates a ``SuppFuncAtom`` object and a cvxpy inequality constraint in the same line of code.

It may seem odd to have two separate classes for this feature. This is necessary because cvxpy doesn't consider convex sets as first-class objects. After quite a bit of thinking, I think the best way to represent a convex set in cvxpy is with a pair ``(expr, constrs)``, where this pair stands in for ``S = {val : constraints in constrs can be satisfied, when expr.value = val}``.

From an implementation perspective, I also need to map an ``(expr, constrs)`` pair into a conic form ``S = {x : A @ [x, y] + b in K }``. I did this by having the ``SuppFunc`` class execute a call to ``get_problem_data(solver='SCS')``, and then reaching slightly back into the solver reduction chain. Due to the nature of this approach, I can't *cleanly* construct all the convex sets I'd really like to. The current implementation only allows sets ``(x, constrs)`` where ``x`` is a cvxpy Variable with no special attributes (i.e. no declared symmetry, nonnegativity, etc...). This isn't a huge limitation, but it's certainly worth mentioning.

The one thing I don't like about this implementation is how ``SuppFuncAtom`` objects maintain some of their metadata. When one of these objects is constructed, I stash an extra dictionary *inside* the ``__dict__`` field of an existing ``cvxpy.Constant`` object. I do this because the ``suppfunc_canon`` function defined within the ``dcp2cone`` folder doesn't have access to the ``SuppFuncAtom`` object in question; it only has access to the ``args`` field of that ``SuppFuncAtom`` object! I tried storing all necessary data in the ``args`` field of ``SuppFuncAtom`` objects, but it seems that strong assumptions are placed on the elements of ``args`` elsewhere in the compilation process. I don't know how to pass the necessary data through objects that satisfy the (unwritten?) assumptions on ``args``, so for now I have this hacky solution.

### DIFF
--- a/cvxpy/atoms/suppfunc.py
+++ b/cvxpy/atoms/suppfunc.py
@@ -10,14 +10,14 @@ class SuppFuncAtom(Atom):
     def __init__(self, y, A, b, K_sels, x, cons):
         self.id = lu.get_id()
         eta = Variable(shape=(b.size,))
-        self.args = [Atom.cast_to_const(y), Atom.cast_to_const(A), Atom.cast_to_const(b),
-                     Atom.cast_to_const(eta)]
-        horrible = self.args[2]
-        horrible.__dict__['K_sels'] = K_sels
+        self.args = [Atom.cast_to_const(y), Atom.cast_to_const(eta)]
+        self._A = A
+        self._b = b
+        self._K_sels = K_sels
         self._x = x
         self._xcons = cons
-        self.validate_arguments()
         self._shape = tuple()
+        self.validate_arguments()
         pass
 
     def validate_arguments(self):
@@ -27,14 +27,14 @@ class SuppFuncAtom(Atom):
             raise ValueError("Arguments to SuppFuncAtom must be affine.")
 
     def variables(self):
-        varlist = self.args[0].variables() + self.args[3].variables()
+        varlist = self.args[0].variables() + self.args[1].variables()
         return varlist
 
     def parameters(self):
         return []
 
     def constants(self):
-        return [self.args[1], self.args[2]]
+        return []
 
     def shape_from_args(self):
         return self._shape
--- a/cvxpy/atoms/suppfunc.py
+++ b/cvxpy/atoms/suppfunc.py
@@ -23,7 +23,6 @@ class SuppFuncAtom(Atom):
         self._eta = None  # store for debugging purposes
         self._shape = tuple()
         self.validate_arguments()
-        pass
 
     def validate_arguments(self):
         if self.args[0].is_complex():
--- a/cvxpy/atoms/suppfunc.py
+++ b/cvxpy/atoms/suppfunc.py
@@ -1,0 +1,132 @@
+import cvxpy.lin_ops.lin_utils as lu
+from cvxpy.atoms.atom import Atom
+from cvxpy.expressions.variable import Variable
+import scipy.sparse as sp
+import numpy as np
+
+
+class SuppFuncAtom(Atom):
+
+    def __init__(self, y, A, b, K_sels, x, cons):
+        self.id = lu.get_id()
+        eta = Variable(shape=(b.size,))
+        self.args = [Atom.cast_to_const(y), Atom.cast_to_const(A), Atom.cast_to_const(b),
+                     Atom.cast_to_const(eta)]
+        horrible = self.args[2]
+        horrible.__dict__['K_sels'] = K_sels
+        self._x = x
+        self._xcons = cons
+        self.validate_arguments()
+        self._shape = tuple()
+        pass
+
+    def validate_arguments(self):
+        if self.args[0].is_complex():
+            raise ValueError("Arguments to SuppFuncAtom cannot be complex.")
+        if not self.args[0].is_affine():
+            raise ValueError("Arguments to SuppFuncAtom must be affine.")
+
+    def variables(self):
+        varlist = self.args[0].variables() + self.args[3].variables()
+        return varlist
+
+    def parameters(self):
+        return []
+
+    def constants(self):
+        return [self.args[1], self.args[2]]
+
+    def shape_from_args(self):
+        return self._shape
+
+    def sign_from_args(self):
+        return (False, False)
+
+    def is_nonneg(self):
+        return False
+
+    def is_nonpos(self):
+        return False
+
+    def is_imag(self):
+        return False
+
+    def is_complex(self):
+        return False
+
+    def is_atom_convex(self):
+        return True
+
+    def is_atom_concave(self):
+        return False
+
+    def is_atom_log_log_convex(self):
+        return False
+
+    def is_atom_log_log_concave(self):
+        return False
+
+    def is_atom_quasiconvex(self):
+        return True
+
+    def is_atom_quasiconcave(self):
+        return False
+
+    def is_incr(self, idx):
+        return False
+
+    def is_decr(self, idx):
+        return False
+
+    def is_convex(self):
+        # The argument "y" is restricted to being affine.
+        return True
+
+    def is_concave(self):
+        return False
+
+    def is_log_log_convex(self):
+        return False
+
+    def is_log_log_concave(self):
+        return False
+
+    def is_quasiconvex(self):
+        # The argument "y" is restricted to being affine.
+        return True
+
+    def is_quasiconcave(self):
+        return False
+
+    def _value_impl(self):
+        from cvxpy.problems.problem import Problem
+        from cvxpy.problems.objective import Maximize
+        y_val = self.args[0].value.round(decimals=9).ravel(order='F')
+        x_flat = self._x.flatten()
+        cons = self._xcons
+        prob = Problem(Maximize(y_val @ x_flat), cons)
+        val = prob.solve(solver='SCS', eps=1e-6)
+        return val
+
+    def _grad(self, values):
+        # the implementation of _grad from log_det.py was used
+        # as a reference for this implementation.
+        y0 = self.args[0].value  # save for later
+        y = values[0]  # ignore all other values
+        self.args[0].value = y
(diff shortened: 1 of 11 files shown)


akshayka:
Very cool! Will take a closer soon. (Also, thanks for fixing the MOSEK interface.)

Just one comment for now.

> I do this because the suppfunc_canon function defined within the dcp2cone folder doesn't have access to the SuppFuncAtom object in question; it only has access to the args field of that SuppFuncAtom object! 

Actually, the first argument to `suppfunc_canon` (`expr`) should be an instance of `SuppFuncAtom`.

rileyjmurray:
> Actually, the first argument to `suppfunc_canon` (`expr`) should be an instance of `SuppFuncAtom`

Wonderful news! I will change the implementation to reflect that.

Although, this now raises a question for me: what is the purpose of the second argument to ``suppfunc_canon``, or any other canon method for that matter? If the first argument (``expr``) is the atom in question, then we should be able to retrieve args directly from that object.

rileyjmurray:
Okay, now that the hack with suppfunc_canon is fixed, I think this PR is in much better shape. 

Here are some points to consider in the review:

1. ``SuppFuncAtom`` overrides essentially all ``Atom`` and ``Expression`` functions. It seems to me that support functions are too general for us to automatically determine properties like ``is_incr`` or ``is_nonpos``, and so it is best to return pessimistic results for those queries.

2. The convex set described by a ``SuppFunc`` object cannot contain Parameters (a-la DPP programming). At least, not in the current implementation.

3. If this PR is merged, the features should be described (in some detail) on the website. Using these features, you can construct conjugate functions, explicitly solve dual problems, and do some standard operations in robust convex optimization.

One consistent thought I had throughout this process, is that it would be very useful if convex sets were first-class objects within cvxpy. If you'd like to discuss this more, let me know.

akshayka:
> Although, this now raises a question for me: what is the purpose of the second argument to suppfunc_canon, or any other canon method for that matter? If the first argument (expr) is the atom in question, then we should be able to retrieve args directly from that object.

The args passed to a canon method are canonicalized args, ie, they are affine expressions representing the epigraph variables for the args. You’re right that the semantics of these arguments should be documented. You can look at the Canonicalization reduction to see what exactly is passed to canonicalizers.

This is a really cool change! I agree with / am okay with all of your three points.

SteveDiamond:
@rileyjmurray can you say more about what you mean by convex sets are not first class objects? I added an Indicator transform at one point to convert constraints to atoms of the form ``f(x) = 1 if x satisfies constraints, +\infty otherwise``. Would that be helpful?

rileyjmurray:
@SteveDiamond Constraint objects completely handle the problem of *membership* in a set. However, in convex analysis, we often want to do more than just enforce membership. We want to transform sets, and build more complicated sets from simpler sets. Common operations here include applying a linear operator, taking set-intersection, or computing a Cartesian project. Less-common (but still fundamental) operations include taking Minkowski sums, polar bodies, conic hulls, or dual cones. Most of these operations have some implementation in terms of Constraint objects (or using the ``indicator`` transform), but these implementations are often clunky, unnatural, or result in cvxpy Problem instances with very slow compile times.

There are different ways that a ConvexSet class could be introduced. In one scenario, ConvexSet objects are only used in the background, as a way to facilitate more complicated transforms I've described above. In another scenario, ConvexSet objects could be exposed to the user. I think there are compelling mathematical and performance reasons to expose a ConvexSet class to users. I'll speak to the performance reasons below.

### A possible (end-user) use-case of a ConvexSet class.

Consider a scenario where a user has a repeated constraint pattern they want to apply to every expression in a long ``expr_list``. In particular, suppose the user conceptualizes their set as ``S``, and they want to require ``expr in S`` for every ``expr`` in ``expr_list``.

Right now, a user can do this in a clean way defining a function ``cvxpy_cons_for_S``, which takes in an Expression ``expr``, and returns a list of cvxpy Constraints which are satisfiable iff ``expr.value`` belongs to their conceptualization of ``S``. The user would then write:
```
constrs = []
for expr in expr_list
    constrs += cvxpy_cons_for_S(expr)
```
The downside to that approach, is that cvxpy has no idea that so many of the constraint objects are just slight variations on one another.

Now consider the same scenario where the user has a ConvexSet class at their disposal. They would simply construct the convex set ``S`` as a cvxpy object, and then write something like 
```
constrs = [expr in S for expr in expr_list]
```
If each of the Constraint objects ``con = (expr in S)`` referenced the ConvexSet object ``S`` (much like how ``SuppFuncAtom`` objects in this PR reference a parent ``SuppFunc`` object), then the canonicalizer can now see the underlying structure in the problem.

rileyjmurray:
@SteveDiamond as to examples, let me know if I need to add some in order to get this PR merged. There are several tests which could be converted into examples without too much trouble.

SteveDiamond:
@rileyjmurray no need, it looks good. Your comments on convex sets as a first class object were very clarifying.

cvxpy/atoms/suppfunc.py
```diff
@@ -0,0 +1,140 @@
...
+        parent : cvxpy.transforms.suppfunc.SuppFunc
+            The object containing data for the convex set associated with this atom.
+        """
+        self.id = lu.get_id()
+        self.args = [Atom.cast_to_const(y)]
+        self._parent = parent
+        self._eta = None  # store for debugging purposes
+        self._shape = tuple()
+        self.validate_arguments()
+        pass
```
SteveDiamond: Delete the pass.

cvxpy/reductions/dcp2cone/atom_canonicalizers/suppfunc_canon.py
```diff
@@ -0,0 +1,118 @@
...
+        else:
+            mat[r, c] = 1 / np.sqrt(2)
+            mat[c, r] = 1 / np.sqrt(2)
+        mat = vec[idx] * mat
+        mats.append(mat)
+    V = sum(mats)
+    return V
+
+
+def selector_matrix(selector, inshape):
```
SteveDiamond: I fixed special_index on master.

cvxpy/reductions/dcp2cone/atom_canonicalizers/suppfunc_canon.py
```diff
@@ -0,0 +1,118 @@
+from cvxpy import hstack, SOC, sum, Variable
+from cvxpy.constraints.exponential import ExpCone
+import numpy as np
+from scipy.sparse import csc_matrix
+
+
+def scs_psdvec_to_psdmat(vec, indices):
```
SteveDiamond: Should this be in the SCS cone solver file?

cvxpy/utilities/coeff_extractor.py
```diff
@@ -177,6 +177,6 @@ def quad_form(self, expr):
            q.shape[0] != self.x_length:
             raise RuntimeError("Resulting quadratic form does not have "
                                "appropriate dimensions")
-        if not np.isscalar(constant):
+        if not np.isscalar(constant) and constant.size > 1:
```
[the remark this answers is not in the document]
rileyjmurray: The original form of line caused a ton of local tests to fail, for reasons otherwise unrelated to this PR. It seems canonicalization can produce values for ``constant`` where ``np.isscalar(constant) == False``, but ``constant`` is still a numeric type as we would understand it (i.e. a numpy array with a single element). I can undo this change, but then local tests will fail, and tests on Travis might also fail.

cvxpy/reductions/dcp2cone/atom_canonicalizers/suppfunc_canon.py
rileyjmurray: I think that's a good idea; I'll move it there.

cvxpy/utilities/coeff_extractor.py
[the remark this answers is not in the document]
SteveDiamond: Ok, that's fine!

cvxpy/reductions/dcp2cone/atom_canonicalizers/suppfunc_canon.py
rileyjmurray: For the other thread where we discussed this, the issue at hand was with the QP codepath. Since I work purely within the conic_solver codepath I think I still can't use logical indexing at this point in the compilation process. I think I will need to either call ``special_index_canon`` manually, or stick with this function. Would you prefer that I call ``special_index_canon`` manually, or is this function good?

cvxpy/atoms/suppfunc.py
rileyjmurray: Will do.

cvxpy/reductions/dcp2cone/atom_canonicalizers/suppfunc_canon.py
SteveDiamond: I don't like introducing this function. I changed special_index so ``graph_implementation`` works. Will that help?

rileyjmurray: That should do it! I'll give it a shot and update tomorrow.

rileyjmurray:
@SteveDiamond I've resolved your comments. Should be ready to merge, assuming @akshayka doesn't want to make additional comments.

SteveDiamond:
Looks good! I merged it in.

PR #862, opened 2019-11-06
Fix exponential cone for MOSEK

akshayka:
Checked out from rileyjmurray's PR (#860)


akshayka:
@rileyjmurray, I copied this change from your PR. If it looks good to you, I'll merge it in.

rileyjmurray:
It's fine to merge as-is, but you might also want to specifically try the logistic regression example mentioned in the other thread. MOSEK is now passing its ``mosek_conif.py`` unittests (and it wasn't before), but I haven't run the logistic regression example. I did *try* to run the logistic regression example, but it's a Jupyter notebook and I can't get jupyter to work at the moment.

rileyjmurray:
Actually I just copied the ipython notebook into a python script, and that did run correctly on my laptop. So yes, I would say this is good to merge.

akshayka:
Thanks!

PR #869, opened 2019-11-15
Fix cplex 201911

rkersh:
This pull request contains various bug fixes, cleanup, and preparation for the CPLEX 12.10 release.

I'd appreciate feedback from someone who knows more about CVXPY internals.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
@@ -34,6 +34,34 @@ _LIN, _QUAD = 0, 1
 _CpxConstr = namedtuple("_CpxConstr", ["constr_type", "index"])
 
 
+def set_parameters(model, solver_opts):
+    """Sets CPLEX parameters."""
+    # TODO: Parameter support is functional, but perhaps not ideal.
+    # The user must pass parameter names as used in the CPLEX Python
+    # API, and raw values (i.e., no enum support).
+    kwargs = sorted(solver_opts.keys())
+    if "cplex_params" in kwargs:
+        for param, value in solver_opts["cplex_params"].items():
+            try:
+                eval("model.parameters.{0}.set({1})".format(param, value))
+            except AttributeError:
+                raise ValueError(
+                    "invalid CPLEX parameter, value pair ({0}, {1})".format(
+                        param, value))
+        kwargs.remove("cplex_params")
+    if "cplex_filename" in kwargs:
+        filename = solver_opts["cplex_filename"]
+        if filename:
+            model.write(filename)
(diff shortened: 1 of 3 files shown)


akshayka:
Thanks! @SteveDiamond , want to take a look at this?

cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
```diff
@@ -34,6 +34,176 @@
...
+                raise ValueError(
+                    "invalid CPLEX parameter, value pair ({0}, {1})".format(
+                        param, value))
+        kwargs.remove("cplex_params")
+    if "cplex_filename" in kwargs:
+        filename = solver_opts["cplex_filename"]
+        if filename:
+            model.write(filename)
+        kwargs.remove("cplex_filename")
+    if s.BOOL_IDX in kwargs:
```
SteveDiamond: I don't think this would be in the kwards.

cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
```diff
@@ -83,25 +253,30 @@ def apply(self, problem):
...
+                       inverse_data[s.OFFSET])
+
+            x = np.array(model.solution.get_values())
+            primal_vars = {inverse_data[CPLEX.VAR_ID]: x}
+
             if not inverse_data['is_mip']:
-                eq_dual = utilities.get_dual_values(
-                    solution['eq_dual'],
+                y = -np.array(model.solution.get_dual_values())
+                dual_vars = utilities.get_dual_values(
```
SteveDiamond: Are you sure all the dual variables are in the right order when they are merged in this way? Just checking.

SteveDiamond:
Looks good! I made random comments/questions. I guess this is the only solver interface not using the STATUS_MAP construct, but it's fine given how complex the logic is for CPLEX.

cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
rkersh: That last line was already there in cplex_conif.py. I just left it there because I didn't know what it was. I will go ahead and remove it.

cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
rkersh: They should be in the right order, yes, but I will double check this and add a comment at the least.

rkersh:
@SteveDiamond: Thanks for the feedback. I just pushed a few commits to address your concerns. Assuming all of the checks pass, I think this pull request is ready to go.

SteveDiamond:
Great!

PR #871, opened 2019-11-16
Default to conic solvers for LPs

akshayka:
QP solvers like OSQP can take many iterations for LPs. For example,
for the l1-regularized SVM example

  https://www.cvxpy.org/examples/machine_learning/svm.html,

OSQP can take > 20,000 iterations to converge (so a solver error is thrown),
whereas an IPM like ECOS is much faster.

### DIFF
--- a/cvxpy/reductions/solvers/solving_chain.py
+++ b/cvxpy/reductions/solvers/solving_chain.py
@@ -19,7 +19,23 @@ from cvxpy.reductions.solvers import defines as slv_def
 from cvxpy.utilities.debug_tools import build_non_disciplined_error_msg
 
 
+def _is_lp(self):
+    """Is problem a linear program?
+    """
+    for c in self.constraints:
+        if not (isinstance(c, (Equality, Zero)) or c.args[0].is_pwl()):
+            return False
+    for var in self.variables():
+        if var.is_psd() or var.is_nsd():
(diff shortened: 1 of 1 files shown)


bstellato:
These are actually good examples to try to improve OSQP! 

By the way, also SCS has issues since it returns "Solved/Inaccurate" for most of the problems in the regularization path. I believe it does not throw an error because it does not complain when the maximum number of iterations is reached. 

Adding a small quadratic regularization to the cost function makes both of them converge but it might not be ideal as a general approach.

akshayka:
It’d definitely be interesting to see how to improve OSQP for LPs! Yeah, SCS also struggles, which is why ECOS is our default conic solver.

This PR is totally open for discussion. As you said, using ECOS by default seemed like the easiest way to ensure things worked out-of-the-box, so to speak.

rileyjmurray:
I very much support using ECOS as the default solver for LPs.

akshayka:
@bstellato , @rileyjmurray , thanks for your input! Since we're all in agreement, I'll move forward with this PR.

rostyboost:
FYI, the behavior for Gurobi is the exact opposite unfortunately, because `gurobi_qpif.py` is faster to handshake to Gurobi backend than `gurobi_conif.py`

PR #878, opened 2019-11-21
Update supported problems for MOSEK in documentation

aszekMosek:
Updates problem typed supported by MOSEK.

From Cvxpy 1.0 Cvxpy+Mosek can be used for exponential cone problems.

MIP was supported even earlier. A slight issue is that mixed-integer SDP is not supported in MOSEK, but that is not a major use case and I would think that with MIP box ticked the information is closer to the truth than without. A footnote about it would also be an option.


SteveDiamond:
Could you add a footnote about how mixed integer SDPs are not supported? People always find little issues like that, so we'll need to add a note eventually.

PR #887, opened 2019-11-28
Drop zero arguments in addition

### LINKED ISSUE #883: `is_log_log_convex` not preserved under Python built-in sum
The sum of two log-log convex variables should be log log convex. However, this does not work as expected with the `sum` operator. MWE:

```python
import cvxpy as cp
q = cp.Variable(2, pos=True)
print(sum(q).is_log_log_convex())       # prints False
print((q[0]+q[1]).is_log_log_convex())  # prints True
```

**Version**
 - OS: OS X, Python 3.6
 - CVXPY Version: 1.0.25

akshayka:
Modify __add__ such that expr + 0 returns expr (similarly for 0 + expr);
this makes it possible to use Python's builtin sum when constructing LLCPs.

Fixes #883

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -441,12 +441,16 @@ class Expression(u.Canonical):
     def __add__(self, other):
         """Expression : Sum two expressions.
         """
+        if isinstance(other, cvxtypes.constant()) and other.is_zero():
+            return self
         return cvxtypes.add_expr()([self, other])
 
     @_cast_other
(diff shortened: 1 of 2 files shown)


akshayka:
@SteveDiamond , see if you like this change. I'll merge it in if you approve. It feels a just a little hacky, but my intuition is that the hack is worth it since it makes the code more usable.

SteveDiamond:
I guess it's fine. I think a more principled change would be overkill.

PR #893, opened 2019-12-04
Don't get extract dual from gurobi if it's not available

rmcgibbo:
Passing `QCPDual=False` to Problem.solve(solver="GUROBI") instructs gurobi not to calculate the dual values, but cvxpy will still try to extract they from the solution leading to a KeyError. At least with my models, they don't actually seem to be used at all, and just skipping their extraction here appears to work fine. This can save a couple minutes of compute.


SteveDiamond:
Great, thanks!

PR #895, opened 2019-12-04
Updated max transform function in scalarize function

mostafaelaraby:
when i used cp.max i encountered a bug so i tried to fix it by just using atoms.maximum instead of atoms.max_elemwise


SteveDiamond:
Thanks!

PR #905, opened 2019-12-17
Basis identification and basic solution retrieval from MOSEK

aszekMosek:
The bfs kwarg for MOSEK was not documented.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -268,6 +268,7 @@ class MOSEK(ConicSolver):
         # parameters if applicable).
         kwargs = sorted(solver_opts.keys())
         save_file = None
+        bfs = False
         if 'mosek_params' in kwargs:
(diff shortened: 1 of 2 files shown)


aszekMosek:
By the way for linear problems: The basis identification time in MOSEK is wasted time if CVXPY by default retrieves only the interior point solution. A sensible solution would be to disable basis identification if bfs=False.

I will be happy to expand this PR if you think it makes sense.

SteveDiamond:
Yes, please add the proposed PR. I wasn't aware that the basis identification was being computed.

akshayka:
@rileyjmurray , since you have familiarity with the MOSEK interface, would you like to review this change?

rileyjmurray:
I'll review (and merge, as appropriate) in the next couple days.

rileyjmurray:
It occured to me that @aszekMosek couldn't have run the new test suite (from #910) before submitting this PR. After running local tests on my machine, I can confirm that MOSEK passes the now-standard test-suite from #910.

PR #908, opened 2019-12-21
Fix typos

bwignall:
Should be non-semantic. I will highlight upfront that this touches the Eigen subdirectory; let me know if this should be reverted.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos, with https://github.com/bwignall/typochecker to help automate the checking.

`flake8 cvxpy` shows no errors/warnings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,7 +121,7 @@ nosetests test_dgp.py:TestDgp.test_product
 
 Please make sure that your change doesn't cause any of the unit tests to fail.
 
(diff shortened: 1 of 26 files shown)


akshayka:
Thanks, @bwignall. Perhaps revert the changes to Eigen? We'll merge it in once that's done.

bwignall:
@akshayka , I believe this is done. (However, I am not a Git expert, please let me know if I have made a mistake and need to try again.)

akshayka:
Looks great! Thanks, @bwignall!

Issue #892, opened 2019-12-04
SCS dual variables to PSD constraints dont work


rileyjmurray:
With cvxpy 1.1 (alpha), and SCS 2.1.0.

If a user supplies a constraint ``con = X >> 0``, the optimal dual variable ``Y = con.dual_value`` should satisfy ``np.trace(X @ Y) <= eps`` for a small value of ``eps`` (e.g. ``1e-4`` or ``1e-8``). It should also be PSD.

Right now, neither of these properties hold. It seems SCS can return dual variables ``Y`` which are not symmetric, and where the symmetric part ``Y_sym = (Y + Y.T)/2.0`` is not PSD (the latter by a huge margin). 

I ran into this issue when refactoring tests in preparation for better dual variable support. Below is the problem data where I first encountered the error.
```
"""
Solve "Example 8.3" from Convex Optimization by Boyd & Vandenberghe.

Verify (1) optimal objective values, (2) that the dual variable to the PSD constraint
belongs to the correct cone (i.e. the dual variable is itself PSD), and (3) that
complementary slackness holds with the PSD primal variable and its dual variable.
"""
rho = cvx.Variable(shape=(4, 4), symmetric=True)
constraints = [0.6 <= rho[0, 1], rho[0, 1] <= 0.9,
               0.8 <= rho[0, 2], rho[0, 2] <= 0.9,
               0.5 <= rho[1, 3], rho[1, 3] <= 0.7,
                -0.8 <= rho[2, 3], rho[2, 3] <= -0.4,
               rho[0, 0] == 1, rho[1, 1] == 1,
               rho[2, 2] == 1, rho[3, 3] == 1,
               rho >> 0]
obj = cvx.Minimize(rho[0, 3])
prob = cvx.Problem(obj, constraints)
prob.solve(solver='SCS', verbose=True,eps=1e-8)
# ^ solved to high precision
mat = self.prob.constraints[-1].dual_value
sym_mat = (mat + mat.T) / 2.0
eigs = np.linalg.eigvalsh(sym_mat)
print(np.min(eigs))  # the smallest value is around -0.9.
```
If you see SCS acceleration errors, you can solve again with ``acceleration_lookback=0`` (possibly include ``warm_start=False``). You should still find that the recovered dual variable is not symmetric, and its symmetric part is very far from being PSD.

It's worth noting that from what I can tell, there are no tests in cvxpy (official repo, master branch) which check complementarity of SCS solutions for SDP.

[...]

PR #910, opened 2019-12-23
improve conic dual variable recovery; redesign conic solver test framework

### LINKED ISSUE #830: [1.1] Extract all dual variables for SOC, Exponential Cone constraints
We don't recover all the dual variables associated with SOC and Exponential Cone constraints. We should fix this in 1.1 at some point (this issue is not blocking for the initial release).

cc @SteveDiamond , since you were close to fixing this at some point.

cc @angeris, since you want all the dual variables,

rileyjmurray:
This PR was originally intended to resolve #830; issue #892 was uncovered along the way, and is also fixed by this PR. The discussion in #830 mentions that to test dual variable recovery, I wanted to significantly refactor tests for conic solvers.

## Dual-variable-related changes

There are a large number of changes here, but each change on its own is pretty small. There wasn't much "design work" in these changes. Hopefully they should be uncontroversial.

* Introduced a ``scalar_product`` atom.
  * This was done to evaluate conic dual variable complementary slackness.
  * This atom isn't user-facing, but it could be if you want. Having it be user-facing could help when people are writing ``cvxpy.trace(A.T @ X)``, which compiles very inefficiently compared to using ``scalar_product(A, X)``.
* Introduced an atom called ``deep_flatten``.
  * The purpose of ``deep_flatten`` is to accept expressions (or lists thereof, or lists-of-lists thereof!), and return a vectorized representation of the input. This is used in the implementation of ``scalar_product``.
  * This atom is not user facing (and it should probably stay that way).
* Renamed ``save_value`` for Constraint objects to ``save_dual_value``.
  * There was a TODO for this.
* Added a TODO to compute dual variable domain violations (e.g. checking that the dual variable to an LMI constraint is actually a PSD matrix, or checking that the dual variable to an exponential cone constraint actually belongs to the dual exponential cone).
* Fixed ExpCone and SOC objects to allow for proper dual variable recovery.
  * This involved overriding ``ExpCone.shape`` and ``SOC.size``, as well as ``ExpCone.save_dual_value`` and ``SOC.save_dual_value``.
  * The implementations of ``save_dual_value`` for these functions need *additional tests* to make sure they handle vectorization and/or axis arguments correctly.
* Created a function called ``expcone_permutor``, in ``reductions/solvers/utilities.py``.
  * This is used in ``ecos_conif.py`` and ``mosek_conif.py``.
  * It's not used in ``scs_conif.py``, because ``SCS.EXP_CONE_ORDER = [0, 1, 2]``, and a call to ``expcone_permutor`` in that case would return the trivial permutation.
* Fixed a bug in ``tri_to_full`` from ``scs_conif.py``.

## Conic solver test changes

The changes here are substantial. The following files have been introduced:
* ``solver_test_helpers.py`` and
* ``test_conic_solvers.py``.

The files above consolidate all content from existing conic solver test files (six such files), and expand on those tests by a significant margin. 

I am open to having some design discussions, or revisions to the content of ``solver_test_helpers.py`` and ``test_conic_solvers.py``. In particular, I believe documentation should be added to ``solver_test_helpers.py`` before merging this PR. I don't have that documentation yet, because I wanted to get your input first.

### The file ``solver_test_helpers.py``
 * You should examine the class ``SolverTestHelper`` in detail.
 * There are a large number of pre-defined functions (e.g. ``lp_0``, ``socp_2``, ``sdp_1``, ``mi_lp2``) to construct certain ``SolverTestHelper`` objects.
 * There are also a few static classes (e.g. ``StandardTestLPs``, ``StandardTestSDPs``) for performing standard operations on a given set of SolverTestHelper objects.

### The file ``test_conic_solvers.py``
``test_conic_solvers.py`` defines a class for each conic solver (e.g. ``TestCVXOPT``).
  * If the solver had tests which didn't fit into the ``SolverTestHelper`` framework, then we subclass ``BaseTest``, and those tests are executed in their original form. See ``TestECOS(BaseTest)`` for an example.
  * If all existing solver tests could fit into the ``SolverTestHelper`` framework, then we subclass ``unittest.TestCase``. See ``TestMosek(unittest.TestCase)`` for an example.

## Misc remarks
* I removed two functions from ``mosek_conif.py``. Those functions were unused, as of cvxpy 1.1alpha.
* Added documentation to two existing functions in ``scs_conif.py``.
* Renamed ``test_qp.py`` to ``test_qp_solvers.py``.
* I have not run the new tests with GUROBI, XPRESS, CPLEX, SuperSCS, CBC, or NAG. The NAG interface is known to be broken, and SuperSCS seems to not have a python 3 version.


akshayka:
@rileyjmurray , thanks so much for contributing this PR. I'll take a look at it later today.

Once this PR is merged, we should consider releasing a new alpha version, 1.1.0a2. If that goes smoothly, we should make 1.1.0 stable shortly (within two weeks) thereafter. In fact, I'll start an issue for making 1.1.0 stable.

rileyjmurray:
Update: If I fix the GUROBI issue in #912, most tests for GUROBI pass. One of the SOC dual variable recovery tests for GUROBI still fails; the existing the logic in ``gurobi_conif.py`` doesn't recover SOC dual variables properly. I can make some minor modifications to that file which allow this PR's test suite to pass with GUROBI. I do think someone will need to take a closer look at the GUROBI conic interface though.

Let me know how you'd like to proceed with this PR and GUROBI.

akshayka:
> The NAG interface is known to be broken, and SuperSCS seems to not have a python 3 version.

Let's delete SuperSCS.

>  One of the SOC dual variable recovery tests for GUROBI still fails; the existing the logic in gurobi_conif.py doesn't recover SOC dual variables properly.

For the failing Gurobi tests, let's just let them fail. We don't run Gurobi on CI anyway. Letting them fail keeps us (and the Gurobi users & developers) honest.

cvxpy/atoms/affine/reshape.py
```diff
@@ -98,3 +100,24 @@ def graph_implementation(self, arg_objs, shape, data=None):
...
+        return x.flatten()
+    # recursion
+    if isinstance(x, list):
+        y = []
+        for x0 in x:
+            x1 = deep_flatten(x0)
+            y.append(x1)
+        y = hstack(y)
+        return y
+    raise RuntimeError('Unknown argument to deep_flatten.')
```
akshayka: This should be a ValueError, the semantics of which are 'The user passed in an incorrect input.'

ValueError('The input to deep_flatten must be an Expression, a NumPy array, an int or float, or a list thereof, but received %s' % type(x)')

cvxpy/atoms/affine/binary_operators.py
```diff
@@ -411,3 +414,11 @@ def graph_implementation(self, arg_objs, shape, data=None):
             (LinOp for objective, list of constraints)
         """
         return (lu.div_expr(arg_objs[0], arg_objs[1]), [])
+
+
+def scalar_product(x, y):
```
akshayka: I like this. If you write a bit of documentation for this function, we can expose it.

cvxpy/atoms/affine/binary_operators.py
```diff
@@ -411,3 +414,11 @@ def graph_implementation(self, arg_objs, shape, data=None):
...
+    x = deep_flatten(x)
+    y = deep_flatten(y)
+    prod = multiply(conj(x), y)
+    res = cvxpy_sum(prod)
```
akshayka: nit: replace with

return cvxpy_sum(prod)

cvxpy/constraints/second_order.py
```diff
@@ -80,14 +82,15 @@ def get_data(self):
...
+        return self.args[0].size
 
     @property
     def size(self):
         """The number of entries in the combined cones.
         """
-        # TODO use size of dual variable(s) instead.
-        return sum(self.cone_sizes())
+        cone_size = 1 + self.args[1].shape[self.axis]
+        cs = cone_size * self.num_cones()
```
akshayka: nit: return this value instead of saving it as an intermediary variable

cvxpy/constraints/second_order.py
```diff
@@ -97,10 +100,8 @@ def cone_sizes(self):
         list
             A list of the sizes of the elementwise cones.
         """
-        cones = []
         cone_size = 1 + self.args[1].shape[self.axis]
-        for i in range(self.num_cones()):
-            cones.append(cone_size)
+        cones = [cone_size] * self.num_cones()
```
akshayka: nit: return this value instead of saving it as an intermediary variable

cvxpy/tests/test_conic_solvers.py
```diff
@@ -0,0 +1,952 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import math
+import numpy as np
+import scipy.linalg as la
+import cvxpy as cvx
```
akshayka: nit: here and everywhere else, let's import cvxpy as cp, not cvx

cvxpy/atoms/affine/binary_operators.py
rileyjmurray: Great; I’ll add a doc string.

cvxpy/reductions/solvers/conic_solvers/scs_conif.py
```diff
@@ -224,7 +236,8 @@ def extract_dual_value(self, result_vec, offset, constraint):
             dim = constraint.shape[0]
             lower_tri_dim = dim * (dim + 1) // 2
             new_offset = offset + lower_tri_dim
-            lower_tri = result_vec[offset:new_offset]
+            indices = np.arange(offset, offset + lower_tri_dim)
```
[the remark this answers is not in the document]
rileyjmurray: Looking back, I think I made the change while debugging. I’ll revert this one.

cvxpy/tests/solver_test_helpers.py
```diff
@@ -0,0 +1,515 @@
...
+        expect = self.prob.value
+        assert abs(actual - expect) <= 10**(-places)
+
+    def check_primal_feasibility(self, places):
+        for con in self.constraints:
+            viol = con.violation()
+            if isinstance(viol, np.ndarray):
+                viol = np.linalg.norm(viol, ord=2)
+            assert viol <= 10**(-places)
+        pass
```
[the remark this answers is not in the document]
rileyjmurray: Just for symmetry. Will delete.

cvxpy/tests/solver_test_helpers.py
```diff
@@ -0,0 +1,515 @@
...
+        self.expect_dual_vars = [dv for _, dv in con_pairs]
+        self.expect_prim_vars = [pv for _, pv in var_pairs]
+
+    def solve(self, solver, **kwargs):
+        self.prob.solve(solver=solver, **kwargs)
+
+    def check_objective(self, places=4):
+        actual = self.prob.objective.value
+        expect = self.prob.value
+        assert abs(actual - expect) <= 10**(-places)
```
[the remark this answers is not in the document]
rileyjmurray: I agree that unittest methods are easier to read upon failure. I’ll update the PR to use such a style.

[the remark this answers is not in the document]
rileyjmurray: I've pushed a set of changes for this. The current implementation now creates ``test_case = BaseTest()`` at the start of functions like ``check_objective`` or ``verify_primal_values``, and calls assert methods against that object (like in your example).

This approach didn't involve changing any function signatures. I went with this approach partly because it was simple, but also because it allows us to call the core methods of a ``SolverTestHelper`` object without being in the middle of a larger unittest-based workflow.

rileyjmurray:
@akshayka there's currently a merge conflict with ``test_scs.py`` and ``test_super_scs.py``. Neither Travis nor AppVeyor can build as a result. The conflict stems from the recent "fix typos" PR (#908). Ultimately, the files in question should be removed, but I'm not sure how you'd like to go about doing that.

The safe thing would be to revert ``test_scs.py`` and ``test_super_scs.py`` (on master) to a state just before the merge of #908. Then GitHub should see the automatic merge with this PR, and tests on Travis and AppVeyor can run as usual.

akshayka:
No need to do a revert, @rileyjmurray. Instead, you can rebase your branch (locally) on top of master, which will rewrite your git history to be compatible with master. You can then update this PR by force pushing the rewritten history to your feature branch.

The command you’d use is ‘git rebase’.

cvxpy/reductions/solvers/conic_solvers/scs_conif.py
```diff
@@ -57,14 +57,19 @@ def tri_to_full(lower_tri, n):
...
+    SCS tracks "lower triangular" indices in a way that corresponds to numpy's
+    "upper triangular" indices. So the function call below uses ``np.triu_indices``
+    in a way that looks weird, but is nevertheless correct.
     """
     full = np.zeros((n, n))
     full[np.triu_indices(n)] = lower_tri
-    full[np.tril_indices(n)] = lower_tri
-
+    full += full.T
+    full[np.diag_indices(n)] /= 2
```
SteveDiamond: Should we be doing anything with the primal solution? Or is that already a full matrix?

cvxpy/tests/solver_test_helpers.py
```diff
@@ -0,0 +1,525 @@
+"""
```
SteveDiamond: Very nice!

cvxpy/reductions/solvers/conic_solvers/scs_conif.py
rileyjmurray: If a variable is declared as symmetric (or PSD/NSD), then the de-vectorization is handled in the following function:https://github.com/cvxgrp/cvxpy/blob/0eb2b20dab92407e4b45f13b6cc124ce96859515/cvxpy/reductions/cvx_attr2constr.py#L61
That function doesn't apply any off-diagonal scaling, however.

Although SCS works with a scaled + vectorized PSD cone, it imposes that scaling on the affine map ``(A, b)`` in the standard form ``A @ x + b in K``, rather than imposing the scaling on the primal variable. The fact that the affine map has to be scaled is reflected in ``psd_format_mat`` within ``scs_conif.py``. Dual variables (by contrast) require both de-vectorization and de-scaling.

rileyjmurray:
I realized I haven't added dual variable recovery tests for GLPK. I've added those tests, but they're failing at the moment. I'll work on fixing the merge conflict once GLPK is fixed.

cvxpy/tests/solver_test_helpers.py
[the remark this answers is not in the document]
akshayka: Hmm ... Does the test output look sensible on failure?

Say an assertion on one of the BaseTests created during the execution of `test_ecos_lp_0` fails. Will the output indicate that TestECOS.test_ecos_lp_0 failed, or will it say that an unnamed BaseTest failed, or both? 

The reason I ask is because sometimes, when we make large changes, many tests will fail (e.g., all of the solver tests might fail). It'll be important that we can easily see which solver tests are failing.

akshayka:
> No need to do a revert, @rileyjmurray. Instead, you can rebase your branch (locally) on top of master, which will rewrite your git history to be compatible with master. You can then update this PR by force pushing the rewritten history to your feature branch.

To do the rebase, on your local machine:

git checkout <cvxpy-master>
git pull
git checkout <your-branch>
git rebase <cvxpy-master>

Replace <cvxpy-master> with the name of a branch tracking the official master branch, and replace <your-branch> with the name of your local feature branch

cvxpy/tests/solver_test_helpers.py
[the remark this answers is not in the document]
rileyjmurray: The Gurobi tests are failing due to an uncaught exception. So output there won't be informative.

However, I temporarily modified an SCS test, to ensure failure, and then ran ``nosetests cvxpy/tests/test_conic_solvers.py``. Here's the output:
```
======================================================================
FAIL: test_scs_sdp_1min (cvxpy.tests.test_conic_solvers.TestSCS)
----------------------------------------------------------------------
  File "/Users/RJMurray/Documents/Research/cvxpydev/cvxpy/cvxpy/tests/test_conic_solvers.py", line 321, in test_scs_sdp_1min
    StandardTestSDPs.test_sdp_1min(solver='SCS', places=10)
  File "/Users/RJMurray/Documents/Research/cvxpydev/cvxpy/cvxpy/tests/solver_test_helpers.py", line 506, in test_sdp_1min
    sth.check_primal_feasibility(places)
  File "/Users/RJMurray/Documents/Research/cvxpydev/cvxpy/cvxpy/tests/solver_test_helpers.py", line 48, in check_primal_feasibility
    test_case.assertAlmostEqual(viol, 0, places)
  File "/Users/RJMurray/Documents/Research/cvxpydev/cvxpy/cvxpy/tests/base_test.py", line 38, in assertAlmostEqual
    super(BaseTest, self).assertAlmostEqual(a, b, places=places)
AssertionError: 1.1135423166885872e-07 != 0 within 10 places
```
Which looks good to me. Also, there are no unnamed BaseTests which fail.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -0,0 +1,972 @@
...
+        if cp.SUPER_SCS in cp.installed_solvers():
+            x = cp.Variable(10)
+            obj = cp.Minimize(cp.sum(cp.exp(x)))
+            prob = cp.Problem(obj, [cp.sum(x) == 1])
+            result = prob.solve(solver='SUPER_SCS', eps=1e-4)
+            result2 = prob.solve(solver='SUPER_SCS', warm_start=True, eps=1e-4)
+            self.assertAlmostEqual(result2, result, places=2)
+
+
+@unittest.skipUnless('CVXOPT' in cp.installed_solvers(), 'CVXOPT is not installed.')
```
akshayka: nit: use `cvxpy.reductions.solvers.defines.INSTALLED_SOLVERS` instead of `cp.installed_solvers()`, as in PR #915

cvxpy/tests/solver_test_helpers.py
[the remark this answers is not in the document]
akshayka: Great, thanks -- that looks good to me as well.

cvxpy/tests/test_conic_solvers.py
rileyjmurray: I'll make these changes.

cvxpy/tests/solver_test_helpers.py
```diff
@@ -0,0 +1,528 @@
...
+
+        self.expect_val = obj_pair[1]
+        self.expect_dual_vars = [dv for _, dv in con_pairs]
+        self.expect_prim_vars = [pv for _, pv in var_pairs]
+
+    def solve(self, solver, **kwargs):
+        self.prob.solve(solver=solver, **kwargs)
+
+    def check_objective(self, places):
+        test_case = BaseTest()
```
akshayka: Is it possible to create just one BaseTest instance, saving it as an attribute of the SolverTestHelper instance, instead of creating a new one each time a method is called?

rileyjmurray: Good idea; I’ll make that change.

rileyjmurray:
@akshayka @SteveDiamond ready to merge!

I'll leave it to one of you to push the big green button. Of course, if either of you have additional comments, we can still address those before merge.

SteveDiamond:
Great job @rileyjmurray ! Thanks for being so accommodating in merging in all the different PRs and changes. It's fine with me, I'll let Akshay approve it.

akshayka:
Just merged! Meant to squash the commits, but accidentally pulled them all in; not a big deal. Thanks for making this PR, @rileyjmurray!

PR #926, opened 2020-01-19
Deprecate ``*`` for matrix multiplication; drop Python 2.7 and 3.4.

### LINKED ISSUE #885: scalar multiplication expands to matrix
Hello

This is probably not a real bug, I just wanted to ask why multiplication using * is behaving like it does. Given this very small example:

```
X = cp.Variable(2)
constrains = [X <= 2.]
objective = cp.Maximize(cp.sum( X * np.array([1,1]) + np.array([1.,1.]) ) )
prob = cp.Problem(objective, constrains)
result = prob.solve()
print(result)
```

The result is 10 and not 6 as expected. Even though
`np.sum(X.value * np.array([1,1]) + np.array([1.,1.]))
`
is 6.

If the objective is changed to 
`objective = cp.Maximize(cp.sum( cp.multiply(X,np.array([1,1])) + np.array([1.,1.]) ) )`
the result is 6.

So why does .multiply behave differently to * ?

rileyjmurray:
The main motivation for this PR is to resolve #885. The basics of that change can be confined to ``cvxpy/expressions/expression.py:Expression``. However, once the deprecation warning was introduced, I thought it best to bring the rest of cvxpy code into compliance the the warning's recommendation. There were two ways to do this:
1. find offending instances of ``a * b``, and replace them with ``matmul(a, b)``, or
2. find offending instances of ``a * b``, and replace them with ``a @ b``.

I opted for the second of these. The most important thing to note about this decision is that it means cvxpy can't support python 3.4. At first I thought that was a big jump (we had previously only discussed dropping python 2.7), but the more I thought about it, it seemed like a better idea. Here are some key reasons that come to mind:

1. The Python Foundation dropped support for [python 3.4](https://www.python.org/downloads/release/python-3410/) in March of last year. Note: this happened 9 months *before* the Python Foundation dropped support for [python 2.7](https://www.python.org/doc/sunset-python-2/).
2. We don't run continuous integration tests with python 3.4, so we haven't been testing against it for a long time.
3. I don't think we've ever built a python 3.4 wheel or a conda distribution for cvxpy 1.x. The only way python 3.4 users get a hold of cvxpy 1.x is by installing from a source distribution.
4. Several unittests already used the ``@``operator. So we were already on a course to *effectively* drop support for python 3.4. This just makes it explicit.
5. Perhaps most importantly: I think it would be best to mark a big change like this as we shift from cvxpy 1.0 to cvxpy 1.1.

To help with this, I've done a pretty thorough job of changing not only cvxpy code, but also key documentation, and build files. Merging this PR (perhaps with minor modifications) would go a long way in setting us up for a clean release of version 1.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ b = numpy.random.randn(m)
 
 # Construct the problem.
 x = cp.Variable(n)
-objective = cp.Minimize(cp.sum_squares(A*x - b))
+objective = cp.Minimize(cp.sum_squares(A @ x - b))
 constraints = [0 <= x, x <= 1]
 prob = cp.Problem(objective, constraints)
 
@@ -63,7 +63,7 @@ pip install cvxpy
 
 CVXPY has the following dependencies:
 
-- Python 2.7, 3.4, 3.5, 3.6, or 3.7.
+- Python 3.5, 3.6, or 3.7.
 - multiprocess
 - OSQP
 - ECOS >= 2
@@ -74,7 +74,7 @@ CVXPY has the following dependencies:
 For detailed instructions, see the [installation
 guide](https://www.cvxpy.org/install/index.html).
 
-**Python 2.7 end of life**: The CVXPY development team plans to stop supporting Python 2 later this year, starting with the release of CVXPY 1.1.
+**Python 2.7 end of life**: The CVXPY development team will stop supporting python 2.7 and python 3.4, starting when CVXPY version 1.1 is released.
 
 ## Getting started
 To get started with CVXPY, check out the following:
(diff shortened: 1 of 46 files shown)


rileyjmurray:
The CI builds still show plenty of deprecation warnings. I think most of them are related to diffcp, which I neglected to test against. I'll fix those shortly.

rileyjmurray:
Pinging @akshayka and @SteveDiamond, in case you missed the initial notification for this PR.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SteveDiamond:
This looks great! Thanks for being so thorough about replacing "*" with "@" in the tests.

SteveDiamond:
Akshay must be busy. I'm approving this.

PR #930, opened 2020-01-30
Enable shift-invert mode for finding small eigenvalues

aadamson:
Follow up from https://github.com/cvxgrp/cvxpy/issues/929#issuecomment-579972424

For eigenvalues with very small magnitude, scipy recommends using shift-invert mode to solve an equivalent eigenvalue problem with large magnitude eigenvalues: https://docs.scipy.org/doc/scipy/reference/tutorial/arpack.html#shift-invert-mode.

The motivation for this pull request is that I had run into PSD matrices in quadratic forms that had eigenvalues that were (according to eigsh) ~-5e-9, just outside the tolerance. Enabling shift-invert mode fixed the issue.

### DIFF
--- a/cvxpy/expressions/constants/constant.py
+++ b/cvxpy/expressions/constants/constant.py
@@ -20,6 +20,7 @@ import cvxpy.interface as intf
 from cvxpy.settings import EIGVAL_TOL
 from cvxpy.utilities import performance_utils as perf
 from scipy.sparse.linalg import eigsh
+from scipy.sparse.linalg.eigen.arpack.arpack import ArpackError
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


aadamson:
The change has the complication that if you have a constant matrix with an eigenvalue that is exactly `+/- EIGVAL_TOL`, then eigsh in shift-invert mode will return NaN as an eigenvalue because we're trying to invert `A + EIGVAL_TOL * I`, which is singular. A way to work around this would be check for `NaN` and then rerun with `sigma=-EIGVAL_TOL + epsilon`. It's slightly less elegant looking, but it would work and give accurate eigenvalues.

SteveDiamond:
@aadamson I guess you should add that check for if it's NaN.

SteveDiamond:
Great, thanks! It looks fine to me.

Issue #927, opened 2020-01-19
update docs: Entropy Maximization Example error


notesofdabbler:
I tried running the .ipynb associated with the [Entropy Maximization example](https://colab.research.google.com/github/cvxgrp/cvxpy/blob/master/examples/notebooks/WWW/max_entropy.ipynb) at cvxpy website. The notebook uses CVXOPT as the solver. When I run the same notebook, I am getting the error that ends with the message:
```
SolverError: Either candidate conic solvers (['CVXOPT']) do not support the cones output by the problem (ExpCone), or there are not enough constraints in the problem.
```
But changing the solver to SCS solves fine. I saw in the website that SCS supports exp cones but CVXOPT doesn't which may be the reason for the error now. But since the notebook in the cvxpy site has been able to solve the problem with CVXOPT, I wanted to understand if I am missing something.

Thanks for any help and suggestions.

rileyjmurray:
CVXOPT nominally supports general nonlinear convex programs. In the past, CVXPY used that functionality to try and get CVXOPT to solve exponential cone programs (of which entropy maximization is a special case).

However the performance of CVXOPT on those exponential-cone problems was quite poor (see #424). CVXPY eventually dropped support for that use of CVXOPT. I don't remember exactly when this happened, but it was likely during the upgrade from cvxpy 0.4 to 1.0.

The example you describe (both the python file and the web documentation) should be updated to not use the CVXOPT solver.

PR #939, opened 2020-02-09
Fix entropy maximization example

tuelwer:
The problem definition was not rendered properly on the website. (https://www.cvxpy.org/examples/applications/max_entropy.html)

This pull request also fixes issue #927 by calling the ECOS solver instead of the CVXPY solver. The optimal value is the same up to eight decimal places.


SteveDiamond:
Thanks!

PR #953, opened 2020-02-25
Add aerospace shape design example to dqcp examples section

dberkens:
Python notebook developed with Akshay to highlight constrained minimization of a quasilinear objective function for an aerospace shape design example.


akshayka:
Thanks @dberkens! It'll be up on the website soon.

Issue #948, opened 2020-02-23
GUROBI failing SOC dual-feasibility check


rileyjmurray:
@SteveDiamond mentioned that GUROBI failed a dual variable complementarity check on his local machine. I've verified that now on my machine. The failing test is [socp_1](https://github.com/cvxgrp/cvxpy/blob/fec749b77af897ceba84a57a40960c905be418df/cvxpy/tests/solver_test_helpers.py#L222). Although the failure happens at a complementarity check, there is a larger issue that the dual value
```
dual_var = [np.array([0.28656026]), np.array([2.86560261, 2.22062537, 1.22062538])]
```
is very far from dual-feasible. (The dual-feasibility issue wasn't reported because we don't check for dual domains with ExpCone or SOC constraints.)

Paging contributors to the GUROBI interface who might be able to help: @rostyboost @rmcgibbo @rluce @gglockner.

**Version**
 - OS: macOS 10.13.6
 - CVXPY Version: 1.1.0a3 (master; slightly ahead of public 1.1.0a3 release).
 - GUROBI Version: 9.0.1 (find via ``gurobipy.gurobi.version()``).

rluce:
Thanks for pinging, I'll check what's happening here.

rileyjmurray:
Thanks!

I realize that CVXPY doesn't have a decent guide for maintaining solver interfaces. I just wrote a long comment about this (with emphasis on dual variables) on another thread. You can find that comment [here](https://github.com/cvxgrp/cvxpy/issues/923#issuecomment-590516011).

rluce:
@rileyjmurray The post on the dual values was in intersting read, thanks.  In this context I would like to understand what exactly `SOC.dual_value` actually represents.  If I query it, I see that a nested array is returned, like for the `socp_1` test,
```
>>> constraints[0].dual_value
[array([2.86560261]), array([[ 2.22062537],
       [ 1.22062538],
       [-1.33812376]])]
```

From what I see, these values are the dual values of some auxiliary linear constraints related to the SOC, but what is their interpretation?

Ultimately the reason for which the test socp_1 (and _2, _3) fail is related to these dual values:  Upon construction of the gurobipy problem a different convention for ordering the constraints is used than later to retrieve the dual values.

[...]

Issue #900, opened 2019-12-09
Incorrect solution when using Gurobi as the solver


rizalzaf:
**Describe the bug**
Incorrect solution returned when using Gurobi solver

**To Reproduce**
```
import cvxpy as cp
import numpy as np

S = cp.Variable((2,2))
v = cp.Variable()       
D = cp.Variable((2,2))  

X = np.array([[1, 1], [1, 1]])
M = np.array([[1, 2/3], [2/3, 0.5]])

k = np.array([1, 2])
l = np.array([1, 0.5])
K = np.tile(k, (2,1))
L = np.tile(l, (2,1))

E = np.ones((2,2)) @ D @ np.diag(l) 

constraints = [S >= 0, D >= 0, v >= 0]
constraints += [S <= np.ones((2,2)) @ S @ np.diag(l)]
constraints += [cp.sum(S) <= 1]

Q = S @ np.diag(k)

qs = 1 - cp.sum(cp.multiply(Q, L))
q0 = cp.multiply(qs, np.ones((2,1)))

dP = Q @ M.T

dp0 = q0 / 2
dP -= cp.multiply(cp.sum(dp0), L)

w = cp.sum(dp0)

G = dP @ np.diag(k)

constraints += [v >= G - D + E]
obj = v - cp.sum(cp.multiply(cp.multiply(S, K), X)) + w

lp_prob = cp.Problem(cp.Minimize(obj), constraints)
```

**Expected behavior**
Correct solution: `-1.0`
Outputs using other solvers:
```
In [32]: lp_prob.solve(solver=cp.ECOS)
Out[32]: -1.0000000000021085

In [33]: lp_prob.solve(solver=cp.OSQP)
Out[33]: -0.9999999999999996

In [34]: lp_prob.solve(solver=cp.MOSEK)
Out[34]: -1.0000000004011982

In [35]: lp_prob.solve(solver=cp.CVXOPT)
Out[35]: -0.9999999947383706
```

**Output**
Output using Gurobi solver
```
In [31]: lp_prob.solve(solver=cp.GUROBI)
Warning for adding constraints: zero or small (< 1e-13) coefficients, ignored
Out[31]: -1.3333333333333335
```

```
In [36]: lp_prob.solve(solver=cp.GUROBI, verbose = True)
Warning for adding constraints: zero or small (< 1e-13) coefficients, ignored
Parameter OutputFlag unchanged
   Value: 1  Min: 0  Max: 1  Default: 1
[...]

PR #955, opened 2020-02-29
Minor polish

rluce:
This PR has some tweaks for dual value computation (#948), a workaround for a bug in gurobipy 8.1.1, and better verbosity control (both #900).  nosetests ran through clean on my machine.


rileyjmurray:
@rluce can you resolve the following flake8 issues? They're causing the TravisCI builds to fail.
```
flake8 .
./cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py:133:36: E261 at least two spaces before inline comment
./cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py:156:36: E261 at least two spaces before inline comment
```

rluce:
Ooops... fixed.

SteveDiamond:
Great, thanks!

Issue #920, opened 2020-01-12
CPLEX dual variable issue


rileyjmurray:
@SteveDiamond recently informed me that a trial version of CPLEX is available from pip. I ran the current suite of tests against that version of CPLEX, and some of the dual variable tests failed. Most of the failures were fixed by changing precision of the tests, however one test failed by a huge margin. Specifically,
```
StandardTestSOCPs.test_socp_1(solver='CPLEX', places=0)
```
fails (follow [this link](https://github.com/cvxgrp/cvxpy/blob/f35960c65ee9dd5687929f0c95104f901d920db8/cvxpy/tests/solver_test_helpers.py#L222) for the problem definition).

The cause of the failure is that the dual variable to the final constraint (``y <= 5``) is reported as zero, when it should be about ``2.86560``. I checked to see if the test was over-restrictive (i.e. maybe some dual-optimal solution has zero dual variable to ``y <= 5``), but this is not the case. The current CPLEX solution fails a stationary Lagrangian check, and so it is not dual-feasible.

rkersh:
@rileyjmurray: I'm able to reproduce the issue, and I'm looking into it.

rkersh:
Here is the test script (github doesn't accept .py files, so I zipped it up).
[test.zip](https://github.com/cvxgrp/cvxpy/files/4073548/test.zip)

rileyjmurray:
@rkersh is there progress on this?

rkersh:
> @rkersh is there progress on this?

@rileyjmurray: Yes, an issue has been filed with the CPLEX team. There appears to be a problem when mapping the solution to the presolved model back to the original model. Can you mark the test as skipped or ignored for the time being? I'll let you know when a solid workaround or fix is available for this.

rileyjmurray:
I see. Thanks for the reply!

As of yesterday evening, we now have a couple new tests for SOCPs. The CPLEX interface is failing ``socp_1``, ``socp_2``, and ``socp_3`` because of dual variable issues.

Moving forward I can either disable all SOCP dual variable tests, or just the one that sparked this thread. Take a look at the new tests and let me know which you'd prefer.

rkersh:
@rileyjmurray: I think we should skip all three for the short term. For example:
```
diff --git a/cvxpy/tests/test_conic_solvers.py b/cvxpy/tests/test_conic_solvers.py
index 01812b58..74ffba35 100644
--- a/cvxpy/tests/test_conic_solvers.py
[...]

PR #956, opened 2020-03-03
Fix bug with RHS values in cplex_conif.py:add_model_soc_constr

rkersh:
If the supporting linear equality constraint was empty, we were
setting the RHS to 0, but we should have been using the RHS given in
vec[i].

With this change, the test_cplex_socp_2 and test_cplex_socp_3 tests
should now pass. This is for #920.


SteveDiamond:
Thanks!

PR #960, opened 2020-03-06
[WIP] Add ParamQuadProg

bstellato:
Work in progress to add parametric QPs.

Things needed:
- [x] Fix `CoeffExtractor.quad_form` and `CoeffExtractor.extract_quadratic_coeffs` from [coeff_extractor.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/utilities/coeff_extractor.py) (the most important step)
- [x] Add `ParamQuadProg` in [qp_matrix_stuffing.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py) to mimic `ConeQuadProg` in [cone_matrix_stuffing.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py) while using `Q` matrix as well
- [x] Adapt `QpMatrixStuffing` class from [qp_matrix_stuffing.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py) to return `ParamQuadProg`. Always mimic [cone_matrix_stuffing.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py)

I think the easiest way to define quad forms in DPP is to keep them as `sum_squares(S @ x)` where `S` is a matrix in `R`. This might make the first step way easier.

### DIFF
--- a/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
+++ b/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
@@ -99,8 +99,9 @@ class ParamQuadProg(object):
             tensor_application = self.P
         else:
             if sp.issparse(self.P):
-                param_vec = sp.csc_matrix(param_vec[:, None])
-            tensor_application = self.P @ param_vec
+                tensor_application = self.P @ sp.csc_matrix(param_vec[:, None])
+            else:
+                tensor_application = self.P @ param_vec
         P = tensor_application.reshape(
             (self.x.size, self.x.size), order='F').tocsc()
 
--- a/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
+++ b/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
@@ -18,7 +18,7 @@ from cvxpy.cvxcore.python import canonInterface
 from cvxpy.constraints import NonPos, Zero, Equality, Inequality
 from cvxpy.expressions.variable import Variable
 from cvxpy.problems.objective import Minimize
-from cvxpy.reductions import Solution, InverseData, cvx_attr2constr
+from cvxpy.reductions import Solution, InverseData
 from cvxpy.reductions.cvx_attr2constr import convex_attributes
 from cvxpy.reductions.matrix_stuffing import extract_mip_idx, MatrixStuffing
 from cvxpy.reductions.utilities import (are_args_affine,
@@ -31,7 +31,6 @@ import numpy as np
 import scipy.sparse as sp
 
 
-# TODO: Implement parametric quadratic program
 class ParamQuadProg(object):
     """Represents a parameterized quadratic program.
 
@@ -124,61 +123,17 @@ class ParamQuadProg(object):
         Returns:
             A dictionary param.id -> dparam
         """
-        if active_params is None:
-            active_params = {p.id for p in self.parameters}
-
-        del_param_vec = delq @ self.q[:-1]
-        flatdelA = delA.reshape((np.prod(delA.shape), 1), order='F')
-        delAb = sp.vstack([flatdelA, sp.csc_matrix(delb[:, None])])
-        del_param_vec += np.squeeze((delAb.T @ self.A).A)
-        del_param_vec = np.squeeze(del_param_vec)
-
-        param_id_to_delta_param = {}
-        for param_id, col in self.param_id_to_col.items():
-            if param_id in active_params:
-                param = self.id_to_param[param_id]
-                delta = del_param_vec[col:col + param.size]
-                param_id_to_delta_param[param_id] = np.reshape(
-                    delta, param.shape, order='F')
-        return param_id_to_delta_param
+        raise NotImplementedError
 
     def split_solution(self, sltn, active_vars=None):
         """Splits the solution into individual variables.
         """
-        if active_vars is None:
-            active_vars = [v.id for v in self.variables]
-        # var id to solution.
-        sltn_dict = {}
-        for var_id, col in self.var_id_to_col.items():
-            if var_id in active_vars:
-                var = self.id_to_var[var_id]
-                value = sltn[col:var.size+col]
-                if var.attributes_were_lowered():
-                    orig_var = var.variable_of_provenance()
-                    value = cvx_attr2constr.recover_value_for_variable(
-                        orig_var, value, project=False)
-                    sltn_dict[orig_var.id] = np.reshape(
-                        value, orig_var.shape, order='F')
-                else:
-                    sltn_dict[var_id] = np.reshape(
-                        value, var.shape, order='F')
-        return sltn_dict
+        raise NotImplementedError
 
     def split_adjoint(self, del_vars=None):
         """Adjoint of split_solution.
         """
-        var_vec = np.zeros(self.x.size)
-        for var_id, delta in del_vars.items():
-            var = self.id_to_var[var_id]
-            col = self.var_id_to_col[var_id]
-            if var.attributes_were_lowered():
-                orig_var = var.variable_of_provenance()
-                if cvx_attr2constr.attributes_present(
-                        [orig_var], cvx_attr2constr.SYMMETRIC_ATTRIBUTES):
-                    delta = delta + delta.T - np.diag(np.diag(delta))
-                delta = cvx_attr2constr.lower_value(orig_var, delta)
-            var_vec[col:col + var.size] = delta.flatten(order='F')
-        return var_vec
+        raise NotImplementedError
 
 
 class QpMatrixStuffing(MatrixStuffing):
@@ -205,15 +160,15 @@ class QpMatrixStuffing(MatrixStuffing):
     def stuffed_objective(self, problem, extractor):
         # extract to 0.5 * x.T * P * x + q.T * x + r
         expr = problem.objective.expr.copy()
-        P, q = extractor.quad_form(expr)
+        params_to_P, params_to_q = extractor.quad_form(expr)
         # Handle 0.5 factor.
-        P = 2*P
+        params_to_P = 2*params_to_P
 
         # concatenate all variables in one vector
         boolean, integer = extract_mip_idx(problem.variables())
         x = Variable(extractor.x_length, boolean=boolean, integer=integer)
 
-        return P, q, x
+        return params_to_P, params_to_q, x
 
     def apply(self, problem):
         """See docstring for MatrixStuffing.apply"""
--- a/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
+++ b/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
@@ -31,6 +31,15 @@ from cvxpy.utilities.coeff_extractor import CoeffExtractor
 import numpy as np
 
 
+
+# TODO: Implement parametric quadratic program
+class ParamQuadProg(object):
+    """Represents a parameterized quadratic program"""
+    pass
+
+
+
+
 class QpMatrixStuffing(MatrixStuffing):
     """Fills in numeric values for this problem instance.
 
--- a/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
+++ b/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
@@ -236,7 +236,7 @@ class QpMatrixStuffing(MatrixStuffing):
 
         inverse_data.constraints = ordered_cons
         # Batch expressions together, then split apart.
-        expr_list = [arg for c in cons for arg in c.args]
+        expr_list = [arg for c in ordered_cons for arg in c.args]
         params_to_Ab = extractor.affine(expr_list)
 
         inverse_data.minimize = type(problem.objective) == Minimize
--- a/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
+++ b/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
@@ -14,30 +14,159 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from cvxpy import problems
-from cvxpy.atoms import QuadForm, reshape
+from cvxpy.cvxcore.python import canonInterface
 from cvxpy.constraints import NonPos, Zero, Equality, Inequality
-from cvxpy.cvxcore.python import canonInterface as canon
-from cvxpy.expressions import cvxtypes
 from cvxpy.expressions.variable import Variable
 from cvxpy.problems.objective import Minimize
-from cvxpy.reductions import Solution, InverseData
+from cvxpy.reductions import Solution, InverseData, cvx_attr2constr
 from cvxpy.reductions.cvx_attr2constr import convex_attributes
 from cvxpy.reductions.matrix_stuffing import extract_mip_idx, MatrixStuffing
-from cvxpy.reductions.utilities import are_args_affine, lower_equality, lower_inequality
-import cvxpy.settings as s
+from cvxpy.reductions.utilities import (are_args_affine,
+                                        lower_equality,
+                                        lower_inequality,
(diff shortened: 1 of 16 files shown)


akshayka:
@bstellato : What's the status of derivatives in OSQP? 

Once you have something working, we should pull it into this branch.

Specifically, we'll need to do the following:
- [ ] Update `problem.solve` to accept `OSQP` as a solver when `requires_grad=True` (see [here](https://sourcegraph.com/github.com/cvxgrp/cvxpy@cc6445a3a92d2b65c49359a113f3427987dadbc1/-/blob/cvxpy/problems/problem.py#L580))
- [ ] Update `problem.derivative` and [`problem.backward`](https://sourcegraph.com/github.com/cvxgrp/cvxpy@cc6445a3a92d2b65c49359a113f3427987dadbc1/-/blob/cvxpy/problems/problem.py#L645) to work with OSQP / QPs 
- [ ] Make `osqp_qpif.py` capable of computing the derivative. If `solver_opts['requires_grad']` is True, then OSQP could additionally return methods to evaluate the derivative and its adjoint, similar to diffcp; these methods (or more generally whatever is needed to evaluate the derivative and its adjoint) would be stored in `solver_cache`.

Separately, once OSQP derivatives are ready, Shane and I can work on pulling them into CVXPY Layers; but I guess that discussion belongs in another repo.

bstellato:
There is a working version for adjoint derivatives [here](https://github.com/oxfordcontrol/osqp-python/pull/38). We use QDLDL (available via pip) and a symmetrized version of the derivatives matrix (no need for LSQR). Once we are happy with it I can make an OSQP release. It is not the final C/C++ derivatives computation but it might be good option for now.

SteveDiamond:
@akshayka if you have time it would be super helpful to do a pass on at least the first two points. I started tracing the derivative code and found a fair number of assumptions that need to change when we add QPs.

akshayka:
For sure! I can do that once the derivative implementation gets merged into osqp-python

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -14,21 +14,171 @@
...
+                                        lower_inequality,
+                                        group_constraints)
 from cvxpy.utilities.coeff_extractor import CoeffExtractor
-
+import cvxpy.settings as s
 import numpy as np
+import scipy.sparse as sp
+
+
+# TODO: Implement parametric quadratic program
```
akshayka: Remove TODO :)

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -14,21 +14,171 @@
...
+        del_param_vec = np.squeeze(del_param_vec)
+
+        param_id_to_delta_param = {}
+        for param_id, col in self.param_id_to_col.items():
+            if param_id in active_params:
+                param = self.id_to_param[param_id]
+                delta = del_param_vec[col:col + param.size]
+                param_id_to_delta_param[param_id] = np.reshape(
+                    delta, param.shape, order='F')
+        return param_id_to_delta_param
```
akshayka: Is this code correct & complete? For example, I don't see `delP` used anywhere.

If this code isn't correct, can we delete it (since it seems to be basically the same as what's in `ParamConeProg`) and replace it with

```
raise NotImplementedError
```
?

Otherwise we (or I) will forget that this isn't the correct code.

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -27,16 +27,40 @@ def import_solver(self):
...
+
+        """
+        problem, data, inv_data = self._prepare_data_and_inv_data(problem)
+
+        P, q, d, A, b = problem.apply_parameters()
+        inv_data[s.OFFSET] = d
+        data['n_eq'] = data[QpSolver.DIMS].zero
+        data['n_ineq'] = data[QpSolver.DIMS].nonpos
+
+        data[s.P] = sp.csc_matrix(P)
```
akshayka: Shouldn't `P` already be a CSC matrix (since `apply_parameters` makes sure of this)? I think this line will result in an unnecessary copy.

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -27,16 +27,40 @@ def import_solver(self):
...
+        data[s.Q] = q
+        data[s.A] = sp.csc_matrix(A)
```
akshayka: Ditto for this line.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -95,49 +93,59 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+
+        # TODO vectorize this code.
+        coeff_list = []
+        constant = param_coeffs[-1, :]
+        for p in range(param_coeffs.shape[1]):
+            c = param_coeffs[:-1, p].A.flatten()
+
+            # Combine affine data with quadforms.
+            coeffs = {}
+            for var in affine_expr.variables():
```
akshayka: It's very difficult for me to understand what this for loop does. Could you document it, to explain what the algorithm is doing, what it produces, and what the different branches correspond to? That would be a huge help for other developers and maintainers who might work on this code path in the future.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -95,49 +93,59 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                if var.id in quad_forms:
+                    var_id = var.id
+                    orig_id = quad_forms[var_id][2].args[0].id
+                    var_offset = affine_id_map[var_id][0]
+                    var_size = affine_id_map[var_id][1]
+                    if quad_forms[var_id][2].P.value is not None:
+                        c_part = c[var_offset:var_offset+var_size]
+                        P = quad_forms[var_id][2].P.value
+                        if sp.issparse(P):
+                            P = P.toarray()
```
akshayka: I know this code was here before your revision, but I get scared whenever a sparse matrix is converted to a dense one. Why is this happening, & is it necessary? Is `P` large here, or is it always small (in which case the `.toarray()` wouldn't be so bad)? Documentation would be very much appreciated if you know what 's going on.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -95,49 +93,59 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+        # TODO vectorize this code.
+        coeff_list = []
+        constant = param_coeffs[-1, :]
+        for p in range(param_coeffs.shape[1]):
+            c = param_coeffs[:-1, p].A.flatten()
+
+            # Combine affine data with quadforms.
+            coeffs = {}
+            for var in affine_expr.variables():
+                if var.id in quad_forms:
```
akshayka: Can you document what `quad_forms` is? It seems to be some sort of map from variable ids to ... something else?

cvxpy/utilities/coeff_extractor.py
```diff
@@ -95,49 +93,59 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                                                         affine_expr.size)
+
+        # TODO vectorize this code.
+        coeff_list = []
+        constant = param_coeffs[-1, :]
+        for p in range(param_coeffs.shape[1]):
+            c = param_coeffs[:-1, p].A.flatten()
+
+            # Combine affine data with quadforms.
+            coeffs = {}
```
akshayka: Can you include a sentence of documentation explaining what `coeffs` is, as a data structure?

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -53,24 +203,25 @@ def accepts(problem):
                 and problem.is_dpp())
 
     def stuffed_objective(self, problem, extractor):
-        # extract to x.T * P * x + q.T * x + r
+        # extract to 0.5 * x.T * P * x + q.T * x + r
         expr = problem.objective.expr.copy()
-        P, q, r = extractor.quad_form(expr)
+        P, q = extractor.quad_form(expr)
+        # Handle 0.5 factor.
+        P = 2*P
```
akshayka: If I understand correctly, `P` and `q` are tensor-to-problem-data tensors, not the actual problem data. If that's right, could you modify the comment above to indicate that? That'd make it easier for others to maintain/develop this code. We might even rename the variables `P` and `q` to `params_to_P` and `params_to_q`, as you did in another method.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -151,32 +159,46 @@ def quad_form(self, expr):
...
+            P_size = P.shape[0]*P.shape[1]
+            P_list.append(P.reshape((P_size, 1), order='F'))
+            q_list.append(q)
+
+        # Stitch together Ps and qs and constant.
+        P = sp.hstack(P_list)
+        # Stack q with constant offset as last row.
+        q = np.stack(q_list, axis=1)
+        q = sp.vstack([q, constant])
+        return P, q
```
akshayka: Are `P` and `q` here tensors that map parameters to the actual `P` and `q`? If so, can you add a sentence of documentation?

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
SteveDiamond: That's right, it's incomplete. Same for other functions copied over. I've made them all notimplemented.

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
SteveDiamond: True.

cvxpy/utilities/coeff_extractor.py
SteveDiamond: Basically sparse matrices don't support the operation needed, at least with simple multiplication. It would make more sense to keep things sparse and figure out a good way of doing the operation on sparse matrices. I've added some comments.

SteveDiamond: Actually I fixed this, removed the conversion.

cvxpy/utilities/coeff_extractor.py
SteveDiamond: Yeah it's kind of a mess. Added comments.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
SteveDiamond: Correct.

SteveDiamond:
@akshayka I addressed all your comments.

akshayka:
Thanks Steven! The documentation helps a lot. LGTM.

SteveDiamond:
I'm pretty sure it was fine, since ``cache.param_cone_prog`` saved whatever was in `` data[s.PARAM_PROB]``. But to make things clearer I renamed the field to ``cache.param_prog``.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -14,21 +14,126 @@
...
+                                        lower_inequality,
+                                        group_constraints)
 from cvxpy.utilities.coeff_extractor import CoeffExtractor
-
+import cvxpy.settings as s
 import numpy as np
+import scipy.sparse as sp
+
+
+class ParamQuadProg(object):
```
rileyjmurray, line 34: We now have two ``Param_xyz_Prog`` classes: ``ParamQuadProg`` and ``ParamConeProg``. Both of these classes inherit directly from ``object``.

I think it would be good to add an abstract class "ParamProg" which specifies the functions that a ``Param_xyz_Prog`` class must implement. The ``ParamConeProg`` and ``ParamQuadProg`` classes would then inherit from ``ParamProg``. Doing this will make it easier for us to add DPP support for quasi-convex programming and geometric programming, down the line.

The ``ParamProg`` class could be inherit directly from ``object``, and be defined in as a single file in ``cvxpy/problems/``. The function signatures should come with docstrings specifying purpose and return-types, even though they're ultimately left un-implemented.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -80,30 +186,26 @@ def apply(self, problem):
                 con = lower_inequality(con)
             cons.append(con)
 
+        # Reorder constraints to Zero, NonPos.
```
rileyjmurray, line 189: The following comment is outside this PR's scope, but I want to bring it up anyway, before I forget:

When it comes to conic constraints, we might want to consider creating (and using) ``NonNeg`` rather than ``NonPos``. Right now there is a discrepancy between thinking of linear inequalities as constraining an affine expression to lay in the *nonpositive* orthant, and thinking of linear matrix inequalities as constraining an affine expression to lay in the cone of *postive*-semidefinite matrices. This discrepency complicates dual variable interpretations when we think about Lagrangians. There might even be places in cvxpy code where comments or object datatype say "NonPos", but the literal problem data stored is actually with respect to a "NonNeg" constraint.

cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
```diff
@@ -444,10 +444,9 @@ def add_model_soc_constr(self, model, variables,
             # Ignore empty constraints.
             if ind:
                 lin_expr_list.append((ind, val))
-                lin_rhs.append(vec[i])
             else:
                 lin_expr_list.append(None)
-                lin_rhs.append(0.0)
+            lin_rhs.append(vec[i])
```
rileyjmurray, line 449: Please make sure CPLEX tests pass locally after this change. Right now CPLEX tests fail due to a dual variable issue. If you update CPLEX tests with the lines:

https://github.com/rileyjmurray/cvxpy/commit/2698c33f2641a53e75c8cc274d1c416a777ae665#diff-d45ad349a495678d5a89e0e562e076b8R743-R750

then the local tests should all pass.

cvxpy/reductions/solvers/solver.py
```diff
@@ -15,12 +15,72 @@
...
+    soc : list of int
+        A list of the second-order cone dimensions.
+    psd : list of int
+        A list of the positive semidefinite cone dimensions, where the
+        dimension of the PSD cone of k by k matrices is k.
+    """
+    def __init__(self, constr_map):
+        self.zero = int(sum(c.size for c in constr_map[Zero]))
+        self.nonpos = int(sum(c.size for c in constr_map[NonPos]))
+        self.exp = int(sum(c.num_cones() for c in constr_map[ExpCone]))
```
rileyjmurray, line 46: This definition isn't consistent with the docstring. It gives the number of exponential cones, rather than the total dimension of exponential cones (the latter is 3x the former).

The docstring could just be "the number of 3-dimensional exponential cones".

A longer-term comment: higher-dimensional exponential cones are possible. Such cones only really save a few epigraph variables, but in the future there might be solvers which support for this feature.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
SteveDiamond: That makes a lot of sense. Maybe the starting point would be creating NonNeg constraints, as you mentioned.

SteveDiamond:
@rileyjmurray thanks for reviewing! I addressed all your comments in my diff.

SteveDiamond:
I merged in. All the points have been addressed.

SteveDiamond:
Thanks for reviewing!

PR #969, opened 2020-03-17
Fixed link to CVXPY paper

adishavit:
Original link seems wrong.


adishavit:
Alternatively, if you want to keep the arxiv link, I think this is the correct one: https://arxiv.org/pdf/1603.00943.pdf

SteveDiamond:
Thanks!

PR #976, opened 2020-03-26
Update index.rst

kylematoba:
Fix warm start example.


rileyjmurray:
Thanks for catching this!

The TravisCI build failure is of course unrelated (and has since been fixed).

PR #981, opened 2020-03-30
Change use of * as matmul to UserWarning

juesato:
Addresses https://github.com/cvxgrp/cvxpy/issues/980#


SteveDiamond:
Thanks for addressing this! @rileyjmurray I thought we added this warning already?

rileyjmurray:
@SteveDiamond  We added a deprecation warning. Turns out deprecation warnings are suppressed by default, and so many users would not see the warning. This PR switches to a UserWarning, which ensures the warning is seen unless explicitly suppressed.

rileyjmurray:
The CI builds show this works as expected. Thanks @juesato!

Issue #982, opened 2020-04-01
Power control example has problems


keithbriggs:
Running the code at https://www.cvxpy.org/examples/dgp/power_control.html, gives the objective value 0.5000000001090454 (with all elements of p sitting at the lower bound of 0.1), not 1.9868421072460272 as on the webpage.

Worse, changing the objective from sum to:

    objective=cp.Minimize(cp.sum_squares(p))

gives the optimal objective value as 1.0000000065572575e-10, still with p=[0.1 0.1 0.1 0.1 0.1].   The correct answer for that p should be 5e-2.

akshayka:
Thanks for reporting! I’ll look into this.

akshayka:
I've updated the example on the website (modifying the problem data slightly so the constraints on `p` are not tight), and initiated a pull request to fix the bug you saw when using the `sum_squares` atom.

@keithbriggs , thank you for opening this issue — and, belatedly, for opening many other issues in this repository. Getting feedback and bug reports from users is extremely useful for us, and we're grateful that you've taken the time to document the issues you've seen.

keithbriggs:
And thank you for the quick fix! This issue was actually found by a student of mine. I'm doing my bit to promote cvxpy amongst colleagues and students.

akshayka:
You're most welcome.

We've pushed a new version of CVXPY (1.0.29) to PyPi that contains a fix for this issue. Thanks again!

PR #983, opened 2020-04-01
[DGP] Fix quad_over_lin canon

akshayka:
The quad_over_lin canon was previously missing a transformation. Because
the sum_squares atom is implemented using quad_over_lin, this lead to
errors when the sum_squares atom was used in a DGP program.

Issue #982 surfaced this problem.


SteveDiamond:
Looks good!

PR #987, opened 2020-04-04
[DGP] Always eval params

akshayka:
Evaluate parameters when solving log-log convex programs. DGP requires a
specialized DPP ruleset to ensure the corresponding DCP program is
DCP. We haven't worked that ruleset out yet, so for now always evaluate the
parameters.


SteveDiamond:
Do we need to include dqcp in the check?

akshayka:
No, not here. DQCP is handled elsewhere (it doesn’t go through this code path). I’ll look into DQCP and handle in a subsequent PR, if it needs modification.

SteveDiamond:
LGTM.

Issue #826, opened 2019-09-25
cvxcore limits nnz of constraint matrix to 2^32


aselviar:
**Describe the bug**
Hi everyone. First of all, thank you all for the great tool you have built. I am facing a problem trying to run a fairly large optimization problem in my opinion. You can see the code below. The variable b size is 500 x 96. What I am trying to do is to match a sum of timeseries profiles (351236 x 15 min timesteps) with a bigger profile by minimizing their difference. With the same formulation and a much smaller problem (672 timesteps and a b variable of the size 10 x 5) the problem is solved in under 2 seconds without a problem.  But when I am running it for the full scale problem I get the error you see below. 

I am running this on Jupyter Lab and python 3.7.4. The python installation is done with conda.

**To Reproduce**

X_opt = cp.Constant(np.asarray(X.iloc[:,:500])) # the array size is (35136,500)
K_opt = cp.Constant(np.asarray(K.YearlyDemand)) # the vector size is 96
b = cp.Variable((500,96),boolean = True, value = np.zeros((500,96)))
Y_opt = cp.Constant(np.asarray(y)) # the vector size is 35136

constraints = []

constraints.append( cp.sum(b, axis = 0) == 1 ) # the sum of the elements of every column of b must be equal to 1
constraints.append( cp.sum(b, axis = 1) <= 1 ) # the sum of the elements of every row of b must be smaller or equal to 1

objective = cp.Minimize(cp.sum(cp.abs(Y_opt-cp.sum((cp.diag(K_opt)*((X_opt@b).T)).T, axis = 1))))

prob = cp.Problem(objective, constraints)

prob.solve(solver = cp.GLPK_MI, verbose = True)

**Expected behavior**
I would expect the problem to solve as with the much smaller problem. But when I run this one, RAM usage explodes up to 100 GB (about 99% of the available RAM on the server). After a while the RAM usage goes down and then a periodical swinging begins (RAM goes up and down from 50% to 100% every few minutes). From the error and after a lot of googling my suspicion is that the problem is too big for the memory and that at some point data is getting broken down to smaller pieces. I do not think it reaches to the point, where the solver does its work. I tried to optimize the code by vectorizing everything (current version) and trying not to have loops etc. in the formulation. But this did not change anything. Do you guys have any clue if this is a bug or a limitation? Or do you maybe have an idea on how to solve this?

**Output**
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<timed exec> in <module>

[...]

PR #990, opened 2020-04-08
a test for cvxcore overflows (and other minor changes)

rileyjmurray:
@SteveDiamond following our slack thread, here's a PR containing a test to cause a cvxcore overflow error. I created a new file within ``cvxpy/tests`` which won't be discovered by automatic testing suites. The intention is that we execute that file with command line arguments only when necessary. The current file contains a test which takes about 100GB RAM; the overflow happens in Eigen. To see what's going on in Eigen you're going to need to execute the python test script within GDB, and make sure cvxcore was built without GCC compiler optimizations (change line 50 of setup.py to use flag ``"-O0"``).

[One of these commits](https://github.com/cvxgrp/cvxpy/commit/2698c33f2641a53e75c8cc274d1c416a777ae665) modified ``cvxpy/cvxcore/src/ProblemData.hpp``. Our normal test suite passed after those changes, and those changes did seem to address the ``int`` vs ``long`` issue in ``getI``, ``getJ``, ``getV`` from #826. However, the changes didn't actually help with the Eigen error mentioned above. Since I'd rather not mess with cvxcore without good reason, I rolled those changes back.

Other changes:
 *  ``test_curvature.py`` to resolve deprecation warnings.
 * CPLEX tests, in keeping with #920.
 * an incorrect docstring regarding the exponential cone.

### DIFF
--- a/cvxpy/cvxcore/src/ProblemData.hpp
+++ b/cvxpy/cvxcore/src/ProblemData.hpp
@@ -77,9 +77,9 @@ public:
   /**
    * Returns the data vector V as a contiguous 1D numpy array.
    */
-  void getV(double *values, int num_values) {
+  void getV(double *values, long num_values) {
     std::vector<double> V = TensorV[param_id][vec_idx];
-    for (int i = 0; i < num_values; i++) {
+    for (long i = 0; i < num_values; i++) {
       values[i] = V[i];
     }
   }
@@ -87,9 +87,9 @@ public:
   /**
    * Returns the row index vector I as a contiguous 1D numpy array.
    */
-  void getI(double *values, int num_values) {
+  void getI(double *values, long num_values) {
     std::vector<int> I = TensorI[param_id][vec_idx];
-    for (int i = 0; i < num_values; i++) {
+    for (long i = 0; i < num_values; i++) {
       values[i] = I[i];
     }
   }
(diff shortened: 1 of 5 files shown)


akshayka:
Riley, just to make sure I understand: do you mean that commit 2698c33f2641a53e75c8cc274d1c416a777ae665 fixed the overflow in `getI`, etc., but exposed another bug, in Eigen? If that's the case, I think that commit should get merged in. There's no harm in accepting longs instead of ints.

Additionally, if you have the error message  / stack trace / debug output describing the overflow in Eigen, can you paste it here? (Or maybe it's in a related issue, which I haven't seen?)

rileyjmurray:
@akshayka that commit *might have fixed* (perhaps "probably fixed") the overflow in ``getI``, etc...

I use the qualifications there, because the test case provided in this commit never triggered overflow in the same way as #826. The Eigen error I mention occurs both with and without the change in commit 2698c33f2641a53e75c8cc274d1c416a777ae665. 

I guess I was thinking that if I don't know that a change to cvxcore demonstrably fixes something specific, then I shouldn't make that change. But at the same time- local tests did pass with the change in 2698c33f2641a53e75c8cc274d1c416a777ae665. If you think the change is simple and natural enough, then I can add it back to this PR. Just give the word if so.

rileyjmurray:
I'll also post the stacktrace here later today.

SteveDiamond:
@rileyjmurray the commit is a good start! Sadly the functions that interface between cvxcore and cvxpy are pretty hard to change. For example, if you change a function like getI you need to regenerate the swig interface, and there may be more to it as well, like changing cvxcore.i. This is something we can work on together.

SteveDiamond:
@rileyjmurray why don't you merge this? The small changes are nice.

PR #995, opened 2020-04-11
[DGP] Fix canonicalization of eye_minus_inv atom

akshayka:
The canonicalization of eye_minus_inv was previously incorrect. This was
discovered during internal testing. A test is added to ensure that
checks for correctness of the canonicalization.


SteveDiamond:
LGTM. Why was it working before out of curiosity?

akshayka:
The reduction to a dcp program was wrong, so it emitted a non equivalent problem before. That problem was solved, but the answer was not the answer to the DGP program. The test wasn’t specific enough to catch the bug. I noticed that the output was wrong while manually auditing the atoms.

Issue #999, opened 2020-04-18
Matrix power (sort of)


SteveDiamond:
I'd like to add an atom that implements the following function. Let ``x in R^n`` and ``A in R^{m x n}``. Then ``f(A,x)`` outputs an m-vector, with entries given by
```
f(A,x)_i = x_1^A_{i1}*x_2^A_{i2}*...x_n^A_{in}.
```
So it's sort of like matrix multiplication, but where you take powers instead of multiplying by coefficients and instead of summing you multiply. 

The idea comes from [this paper](https://arxiv.org/pdf/2001.03649.pdf) I wrote on log-linear dynamical systems. I haven't needed an atom yet because I do the log-transform by hand, but @moehle suggested a posynomial formulation (a sum of such functions), and in general it would be cool to see what we can compose this with.

Anyway this function is not hard to add, but I'm not sure what name to call it and am looking for suggestions.

SteveDiamond:
After some thought, my proposal is the “geometric matrix multiplication”, ‘’gmatmul’’ in cvxpy. Mathematically, denoted A♢x 😜

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rileyjmurray:
I like this idea! And the notation is nice. Since python doesn't allow the creation of new infix operators, we'd probably have to stick with a named function like ``cp.gmatmul`` as you said.

akshayka:
If x is elementwise positive, each entry in the expression you’ve written would be a monomial, ie log-log affine. So you could compose lots of things with it.

I agree, it’s a cool atom!

PR #1004, opened 2020-04-18
added gmatmul

SteveDiamond:
I added the gmatmul atom discussed in #999 . I need some help though, because when I make the docs no API entry is created for gmatmul.


SteveDiamond:
Also @akshayka do I need to add some DPP logic?

cvxpy/atoms/gmatmul.py
```diff
@@ -0,0 +1,120 @@
+"""
+Copyright 2013 Steven Diamond
```
akshayka, line 2: Should this be "Copyright, the CVXPY authors", as in https://github.com/cvxgrp/cvxpy/blob/master/copyright.txt?

doc/source/tutorial/dgp/index.rst
```diff
@@ -591,6 +592,14 @@ and returns a vector or matrix.
      - |convex| log-log convex
      - |incr| incr.
 
+   * - :ref:`gmatmul(A, x) <gmatmul>`
+
+       :math:`A \in \mathbf{R}^{m \times n}`
+     - :math:`\left[\begin{matrix}\prod_{j=1}^n x_j^{A_{1j}} \\\vdots \\\prod_{j=1}^n x_j^{A_{mj}}\end{matrix}\right]`
+     - :math:`x \in \mathbf{R}^n_{++}`
+     - |affine| log-log affine
+     - None
```
akshayka, line 601: This atom is increasing in x, right? (Since its domain is x positive?)

akshayka:
> Also @akshayka do I need to add some DPP logic?

Yes, some DPP logic will be necessary.

I haven't merged the DPP for DGP branch (dgp-dpp) into master yet. I think the branch is ready for review (or will very soon be ready for review), and all the tests are passing.

If this change gets merged before dgp-dpp is merged, I can rebase and add the DPP logic.

akshayka:
> I need some help though, because when I make the docs no API entry is created for gmatmul.

You'll need to modify the appropriate file in

https://github.com/cvxgrp/cvxpy/tree/master/doc/source/api_reference

to tell Sphinx to generate an API entry in the website docs.

doc/source/tutorial/dgp/index.rst
SteveDiamond: No, because you could for example have $A = -1$.

SteveDiamond: The actual logic of the atom is increasing if A is all non-negative, decreasing if A is all non-positive. But that seemed to involved for the tutorial description.

SteveDiamond:
> > I need some help though, because when I make the docs no API entry is created for gmatmul.
> 
> You'll need to modify the appropriate file in
> 
> https://github.com/cvxgrp/cvxpy/tree/master/doc/source/api_reference
> 
> to tell Sphinx to generate an API entry in the website docs.

That worked. How do I make the DGP tutorial entry link to it though?

cvxpy/atoms/gmatmul.py
SteveDiamond: Thanks! This is fixed.

SteveDiamond:
I got the linking from the DGP atoms page to work (just needed to delete the old build files before building html again). Anyway, from my point of view the PR is done.

doc/source/tutorial/dgp/index.rst
akshayka: Oh, right. I think we could include that fact in the documentation. For example, the documentation on power() for DGP says incr when p > 0, decr when p < 0. If there's a discrepancy between what's documented and what's implemented, that might lead to confusion.

SteveDiamond: Ok I added this logic to the docs.

SteveDiamond:
Thanks!

PR #1010, opened 2020-04-24
DPP for DGP

akshayka:
This change defines and implements DPP for DGP. It also adds differentiation for DGP problems.

The DPP ruleset for DGP is the log-log analogue of the ruleset for DCP. Whereas multiplication was special cased in DCP, powers are special-cased in DGP. The expression `x**a` is allowed so long as both `x` and `a` are not parametrized. If `a` is parameterized, it must be a Parameter, not a compound expression containing parameters. The sign of an exponent can be real, but the sign of all other Parameters must be positive.

This change adds keyword arguments `dpp=True/False` to `is_dcp` and `is_dgp`. When True, these methods apply the DPP ruleset. This change also adds a keyword argument `enforce_dgp=True/False` to `problem.solve`: when True, a DPPError is raised on DPP violations, instead of a warning.

This change also includes critical bug fixes and performance improvements for DPP-DCP and ASA form.

For DCP, this change fixes a bug in which DPP wasn't applied correctly to constraints; on master, it's possible to write non-DPP constraints like `x == log(param)`, which leads to canonicalization errors and incorrect results. This change fixes that.

For ASA form, this change pre-computes and caches a reduced representation of the "problem data tensor," by eliminating  rows that are identically zero to produce a reduced tensor; additionally, the indptr and indices for the solver's CSC data matrix are pre-computed. As a result, the multiplication of the (reduced) problem data tensor and the parameter vector (which now emits a dense data vector for the CSC matrix), and the subsequent construction of the solver's problem data matrix, is often orders of magnitude faster. The problem data tensor is reduced by identifying the nonzero rows, and then (carefully) building a CSR matrix that represents the reduced tensor. (Naively converting to a CSR matrix is extremely inefficient, and often leads to OOM errors.)

Additionally, a fast path for `apply_param_jac` is added, for small problems.

Finally, this change also includes web documentation, to describe the new features.


cvxpy/atoms/elementwise/power.py
```diff
@@ -104,106 +110,153 @@ class power(Elementwise):
...
+                p = 0
+                w = None
+
+            self.p_rational, self.w = p, w
+            self.approx_error = float(abs(self.p_rational - p))
         super(power, self).__init__(x)
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        # Throw error if negative and power doesn't handle that.
```
SteveDiamond, line 151: Could you comment on removing this logic? It seems fine to remove, just wondering if there's any concerns.

cvxpy/atoms/elementwise/power.py
```diff
@@ -104,106 +110,153 @@ class power(Elementwise):
...
+        # a member of the atom. However, this doesn't work because power
+        # is a special case: while in general parameters in a DGP program
+        # must be positive, they can have any sign when appearing as an
+        # exponent, since in this case we don't need to take the log
+        # (eg, in a monomial x_1^a_1x_2a^2, a_1 and a_2 don't need to be
+        # positive). If the parameter p were an arg and was negative,
+        # then x^p would get falsely flagged as unknown curvature under DGP.
+        #
+        # So, as a workaround, we overload the parameters method.
+        return self.args[0].parameters() + self.p.parameters()
```
SteveDiamond, line 215: Interesting workaround. I ran into a similar issue with gmatmul and maybe should do the same.

cvxpy/atoms/elementwise/power.py
```diff
@@ -298,15 +380,9 @@ def copy(self, args=None, id_objects={}):
         """
         if args is None:
             args = self.args
-        # Avoid calling __init__() directly as we do not have p and max_denom.
```
SteveDiamond, line 301: Any idea what was going on here originally? The power code was all written by someone else so I don't have a handle on the details.

cvxpy/atoms/elementwise/power.py
SteveDiamond: Nvm, I see you did that ;)

cvxpy/atoms/gmatmul.py
```diff
@@ -95,15 +100,36 @@ def is_atom_concave(self):
         """
         return False
 
+    def parameters(self):
+        # The exponent matrix, which is not an argument, may be parametrized.
+        return self.args[0].parameters() + self.A.parameters()
```
SteveDiamond, line 105: Random comment, it would perhaps be useful in some cases to allow A to be a variable (when X is constant). Same with exponents in power. I couldn't find a clean way to handle this situation within DGP, but maybe you can think of one.

cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -58,6 +58,95 @@ def get_parameter_vector(param_size,
...
+
+    The problem data tensor A is a matrix of shape (m, p), where p is the
+    length of the parameter vector. The product A@param_vec gives the
+    entries of the problem data matrix for a solver;
+    the solver's problem data matrix has dimensions(n_constr, n_var + 1),
+    and n_constr *( n_var + 1) = m. In other words, each row in A corresponds
+    to an entry in the solver's data matrix.
+
+    This function removes the rows in A that are identically zero, since these
+    rows correspond to zeros in the problem data .It also returns the indices
```
SteveDiamond, line 72: typo.

cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -58,6 +58,95 @@ def get_parameter_vector(param_size,
...
+    shape = (n_constr, var_length + 1)
+    indices = nonzero_rows % (n_constr)
+
+    # cols holds the column corresponding to each row in nonzero_rows
+    cols = nonzero_rows // n_constr
+    n_cols = shape[1]
+
+    # construction of the indptr: scan through cols, and find
+    # the structure of the column index pointer
+    indptr = np.zeros(n_cols + 1, dtype=np.int32)
```
SteveDiamond, line 130: Wow this is all a lot more complicated than I expected! Nice job!

cvxpy/problems/problem.py
```diff
@@ -431,6 +505,17 @@ def get_problem_data(self, solver, gp=False):
 
         if self._cache.param_prog is not None:
             # fast path, bypasses application of reductions
+            if gp:
+                dgp2dcp = self._cache.solving_chain.get(Dgp2Dcp)
+                # Parameters in the param cone prog are the logs
+                # of parameters in the original problem (with one exception:
+                # parameters appearing as exponents (in power atoms) are
+                # unchanged).
+                old_params_to_new_params = dgp2dcp.canon_methods._parameters
```
SteveDiamond, line 514: Also the powers in gmatmul are not logs.

cvxpy/problems/problem.py
```diff
@@ -738,32 +836,62 @@ def backward(self):
...
-            parameter.gradient = dparams[parameter.id]
+
+        if not gp:
+            for param in self.parameters():
+                param.gradient = dparams[param.id]
+        else:
+            dgp2dcp = self._cache.solving_chain.get(Dgp2Dcp)
+            old_params_to_new_params = dgp2dcp.canon_methods._parameters
+            for param in self.parameters():
+                # Note: if param is an exponent of a power atom, then
```
SteveDiamond, line 866: How are you testing whether a param is from a power atom?

cvxpy/reductions/dgp2dcp/atom_canonicalizers/__init__.py
```diff
@@ -105,3 +107,14 @@ def variable_canon(self, variable, args):
...
             return log_variable, []
+
+    def parameter_canon(self, parameter, args):
+        del args
+        # Swaps out positive parameters for unconstrained variables.
+        if parameter in self._parameters:
+            return self._parameters[parameter], []
+        else:
+            log_parameter = Parameter(parameter.shape, name=parameter.name(),
+                                      value=np.log(parameter.value))
```
SteveDiamond, line 118: How can we access parameter value already?

doc/source/tutorial/advanced/index.rst
```diff
@@ -986,18 +1017,96 @@ DPP. You can represent a parametric quadratic form like so:
...
+    monomial = (x**a)**a
+    print("Is DPP? ", monomial.is_dgp(dpp=True))
+    print("Is DGP? ", monomial.is_dgp(dpp=False))
+
+This code snippet prints
+
+::
+
+    Is DPP? False
+    Is DCP? True
```
SteveDiamond, line 1086: Should be "Is DGP? True"

cvxpy/atoms/elementwise/power.py
akshayka: I don't think there are any concerns. I wanted the atom's value to accurately reflect the function. The logic that was removed has to do with the specific graph implementation used for Dcp2Cone --- so while the solver won't be able to, in these cases, assign negative values of `x`, we can still evaluate a power at a negative value.

cvxpy/atoms/elementwise/power.py
akshayka: In the original code, there were arguments to the constructor that were thrown away. I just decided to not throw those arguments away.

cvxpy/atoms/gmatmul.py
akshayka: Hmm, I'll think about that, and maybe address it in a future PR if I get inspired.

cvxpy/cvxcore/python/canonInterface.py
akshayka: Thanks! I tried two or three different implementations, some of which were simpler but which failed when `A` had many rows (such as in the inpainting example). This implementation, while fairly low-level, was the fastest (and used the least memory).

cvxpy/problems/problem.py
akshayka: done

cvxpy/problems/problem.py
akshayka: I added a comment. param.id is in dparams iff param was an exponent (in a power atom or gmatmul).

It's important that the exponents in gmatmul & power are attributes of the expression, and not args; this prevents them from being replaced by new parameters in Dgp2Dcp. I've added comments to this effect in gmatmul and power.

cvxpy/reductions/dgp2dcp/atom_canonicalizers/__init__.py
akshayka: `Dgp2Dcp` assumes that parameter values are not `None`. This is just for convenience, because it means that the first time `problem.solve` is called (in which case all parameters must have values), no extra logic is required to hook up the parameter values in the DGP problem to the values of the parameters in the DCP problem.

Since `Dgp2Dcp` is only called from `problem.solve` anyway, I thought this was fine.
In particular, using the values here doesn't affect curvature analysis for the DCP problem, since the curvature of a parameter only depends on its attributes. 

I've modified `Dgp2Dcp.accepts` to explicitly check that parameters have been set.

doc/source/tutorial/advanced/index.rst
akshayka: Thanks, good catch!

cvxpy/atoms/gmatmul.py
akshayka: @SteveDiamond , I thought about this. The way to do it would be to define an exponential atom in which you can specify the base (e.g., Euler's number e, a Constant, a Parameter, ...).  This atom would be log-log convex.

akshayka: I don't intend to add such an atom in this PR, but if you want to add something like it in the future, you're welcome to! It would definitely be useful when fitting monomials to data.

SteveDiamond: Makes sense, thanks!

cvxpy/reductions/dgp2dcp/atom_canonicalizers/__init__.py
SteveDiamond: Good point, when we call solve there need to be parameter values. It makes me wonder why we even have the eval parameters reduction.

SteveDiamond:
Thanks for the quick replies! LGTM.

PR #1018, opened 2020-05-05
Idea: DGP wrap

SteveDiamond:
This is not a complete PR, just getting an idea out there. @akshayka pointed out that if you have a function ``f(x)`` then the function ``exp(f(log x))`` will become ``f(x)`` after the dgp2dcp mapping. This mapping can be used to generate many more DGP compatible functions. This PR is one approach to exploiting that mapping, by wrapping up atomic functions. I think all that is further needed is a dgp2dcp transformation for the ``dgp_wrap`` expression (besides obvious things like tests and documentation).

Anyway, I'd love to hear thoughts on this PR, and whether this feature would be useful/desirable.


SteveDiamond:
To expand the motivation, I don't see this as a widely used feature. It's more a convenience when exploring DGP formulations, so that we don't need to create new atoms. But that doesn't mean we wouldn't add new DGP atoms, if they turn out to be helpful.

akshayka:
> Anyway, I'd love to hear thoughts on this PR, and whether this feature would be useful/desirable.

It's a cool feature! It makes sense to me to move forward with it.

It's kind of interesting, conceptually, and it shows that LLCPs are (from a modeling perspective) much more expressive than geometric programs.

SteveDiamond:
Thanks! To me, it's cool to emphasize that just as every LLCP is a convex program, every convex program is an LLCP :) 

How should I handle the DPP aspect? My inclination is to say dgp_wrap is not DPP, but not sure exactly how to express that or if that's too crude.

akshayka:
> How should I handle the DPP aspect? My inclination is to say dgp_wrap is not DPP, but not sure exactly how to express that or if that's too crude.

I'd also just say that dgp_wrap is not DPP, at least for now.

I think you can express that by checking if a DPP scope is active in `is_log_log_convex`, something like

```
if u.scopes.dpp_scope_active():
    return False
```

Below is example usage of `dpp_scope_active()`

https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/atoms/elementwise/power.py#L223

SteveDiamond:
I looked into this idea more, and it turns out there are a lot of implementation issues. I'm going to close this PR. I like the idea conceptually though.

PR #1020, opened 2020-05-09
Create NonNeg class, use NonNeg for ConicSolver, major web doc update

rileyjmurray:
Well this took forever and a day, but here we go. The motivation for this PR was just to update web documentation. In the process it became clear that ``ConicSolver.format_constraints`` had been returning problem data for the nonnegative orthant, when it was labeled as the nonpositive orthant. The problem data was always correctly passed to solvers, so this isn't a correctness issue per-say, but it did mean that writing a new solver interface would involve a major headaches. 

In the process of correcting the above issue I introduced a ``NonNeg`` class within the file ``cvxpy/constraints/nonpos.py`` (maybe we should change the file name). I then modified all conic solver interfaces to use NonNeg constraint objects rather than NonPos constraint objects -- again, this didn't actually mean changing logic in the solver interfaces. Since the data was already been passed as though it were NonNeg, changes to most solver interfaces were a matter of find-and-replace.

I moved one of the ConeDims classes to ``cone_matrix_stuffing.py``, since the data returned by ``ParamConeProg``  (defined in that file) can only be understood when using that ConeDims class. I moved another ConeDims class to ``qp_matrix_stuffing.py`` (the file was previously defined in ``solver.py``, but only ever imported in the QP code path). I modified the ConeDims field to have NonNeg instead of NonPos for conic programs, but left the ConeDims class for QPs unchanged.

Somehow in a previous PR I failed to drop SuperSCS as we had intended. This PR also fixes that.


cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -24,14 +25,70 @@
...
+                                        lower_ineq_to_nonpos,
                                         group_constraints)
-from cvxpy.utilities.coeff_extractor import CoeffExtractor
 import cvxpy.settings as s
+from cvxpy.utilities.coeff_extractor import CoeffExtractor
 import numpy as np
 import scipy.sparse as sp
 
 
+class ConeDims(object):
```
SteveDiamond, line 36: I pulled ConeDims into Solver since it's repeated in qp_matrix_stuffing and cone_matrix_stuffing. Should the two instances differ at all?

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -185,7 +242,7 @@ def apply(self, problem):
             if isinstance(con, Equality):
                 con = lower_equality(con)
             elif isinstance(con, Inequality):
-                con = lower_inequality(con)
+                con = lower_ineq_to_nonpos(con)
```
SteveDiamond, line 245: Can you comment on why it's nonpos for QPs?

cvxpy/reductions/solvers/conic_solvers/conic_solver.py
```diff
@@ -213,11 +160,10 @@ def format_constraints(self, problem, exp_cone_order):
         restruct_mat = []  # Form a block diagonal matrix.
         for constr in problem.constraints:
             total_height = sum([arg.size for arg in constr.args])
-            if type(constr) in [Zero, NonPos]:
-                # Both of these constraints have a single argument.
-                # c.T * x + b (<)= 0 if and only if c.T * x (<)= -b.
-                # Need to negate to switch from NonPos to NonNeg.
+            if type(constr) == Zero:
                 restruct_mat.append(-sp.eye(constr.size, format='csr'))
```
SteveDiamond, line 164: The negative here is right?

doc/source/contributing/index.rst
```diff
@@ -4,35 +4,472 @@ Contributing
...
+ - invert.
+
+The first three functions are very easy (often trivial) to write.
+The remaining functions are called in order: ``apply`` stages data for ``solve_via_data``,
+``solve_via_data`` calls the *Awesome* solver by way of the existing third-party
+``AwesomePy`` package, and ``invert`` transforms the output from ``AwesomePy`` into
+the format that CVXPY expects.
+
+Key goals in this process are that the output of ``apply`` should be as close as possible
+to the *Awesome*'s standard form, and that ``solve_via_data`` should kept short.
```
SteveDiamond, line 245: should be kept short.

SteveDiamond: pointing out a typo, not a comment ;)

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
rileyjmurray: I'm not sure if the two instances should differ. It's possible that when QP matrix stuffing says "NonPos" it actually means NonPos. Since I'm not familiar with the QP interfaces I didn't test that. I think it's fair to have different ConeDims classes, depending on different target solvers. However the current ConeDims for QPs contains a lot of excess information (e.g. exponential cone, PSD cone).

Separately -- I'm pretty sure the ConeDims that was in ``solver.py`` was never imported in the ConicSolver codepath. I tracked references and the only usage appeared to be for QP's, so that's why I moved it there.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
rileyjmurray: The function ``lower_ineq_to_nonpos`` is just a renaming of ``lower_inequality``. I changed the name because for ConicSolvers I needed to "lower" inequalities in the opposite direction than was assumed by ``lower_inequality``.

So conic solvers use ``lower_ineq_to_nonneg``, because I changed ConicSolver to use the NonNeg class. QP solvers use the original function (now with a new name) because I didn't introduce the NonNeg class into the code path.

cvxpy/reductions/solvers/conic_solvers/conic_solver.py
rileyjmurray: To be honest I don't know why the negative is there, but I'm pretty sure it results in dual variables being passed correctly. It certainly means equality-constraint dual variables are being passed the same way they were before the PR.

I would have had to keep a negative there for ``NonNeg`` constraints, but I avoided that by creating ``lower_ineq_to_nonneg`` (which means the sign is flipped elsewhere).

rileyjmurray:
@SteveDiamond the ECOS_BB PR is here: https://github.com/cvxgrp/cvxpy/pull/973. That PR mentions dropping SuperSCS as part of an add-on for the ECOS_BB issue, which I somehow neglected to follow-through on. We discussed the user-displayed error message and documentation changes as part of that PR.

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
SteveDiamond: Ok, that's fine.

SteveDiamond:
Thanks for pointing out the PR to drop ECOS_BB. Not sure why it slipped my mind. Anyway, the PR looks fine if you add tests for NonNeg.

doc/source/updates/index.rst
```diff
@@ -1,22 +1,104 @@
...
+Highlights
+~~~~~~~~~~
+
+:ref:`Disciplined parametrized programming <dpp>` or "DPP" provides a way to state your problem
+in more symbolic terms. Taking advantage of DPP can shorten the amount of time it takes
+for CVXPY to pass your problem to a numerical solver. DPP also provides the basis for differentiating
+the map from parameters to the solution of an optimization problem.
+
+Certain solvers (SCS and OSQP) now support differentiating the map from the parameters of an
+optimization problem to the optimal solution of that problem. This feature allows for more general
```
akshayka, line 21: Differentiation is only implemented for SCS, not OSQP (unless I missed a recent change).

doc/source/updates/index.rst
```diff
@@ -1,22 +1,104 @@
...
+
+CVXPY 1.1
+---------
+
+Highlights
+~~~~~~~~~~
+
+:ref:`Disciplined parametrized programming <dpp>` or "DPP" provides a way to state your problem
+in more symbolic terms. Taking advantage of DPP can shorten the amount of time it takes
+for CVXPY to pass your problem to a numerical solver. DPP also provides the basis for differentiating
```
akshayka, line 17: I might change this to:

:ref:`Disciplined parametrized programming <dpp>` or "DPP" is a ruleset for constructing parametrized problems in CVXPY. Taking advantage of DPP can decrease the time it takes CVXPY to repeatedly canonicalize a parametrized problem. DPP also provides ...

doc/source/updates/index.rst
akshayka: You could also link to the derivatives section if you want (the name of the link is just 'derivatives')

doc/source/updates/index.rst
```diff
@@ -1,22 +1,104 @@
...
+
+This release resolves bugs in detecting when a problem falls into the category of
+"disciplined quasi-convex programming" (DQCP).
+
+Known issues
+~~~~~~~~~~~~
+
+DPP problems with many CVXPY Parameters can take a long time to compile.
+
+DPP doesn't compose with Disciplined Quasi-Convex Programming (DQCP).
```
akshayka, line 65: nit: no dash in quasiconvex, here and elsewhere. Maybe change to "Disciplined quasiconvex programming (DQCP) doesn't support DPP"?

doc/source/contributing/index.rst
rileyjmurray: Gotcha!

doc/source/updates/index.rst
rileyjmurray: Change made

doc/source/updates/index.rst
rileyjmurray: There's currently a merge conflict related to derivative documentation. I'm working on it now.

PR #1022, opened 2020-05-09
Additional example on commodity flow

okpoti2:
@SteveDiamond Based on our earlier discussions, I am submitting this pull request so the example on Multi-commodity Network Flow Problem can be merged.


SteveDiamond:
Sorry for the slow response! It looks good to me, merging it in.

PR #1027, opened 2020-05-18
QP Solvers Fixes

bstellato:
I have made a few fixes for the QP solvers.

These are quite standard:
- The `apply` function in the `QpSolver` class was rewriting the `data` dict and, therefore, deleting the `ParamQuadProg` element inside. This means that CVXPY was reparsing QPs all the times, also with parametric programs.

- I have deleted the `apply` function in `osqp_qpif.py` so that it uses the one defined in `QpSolver`

I would like to ask **your feedback** on this:

- Gurobi can return the solution if it hits the maximum time `TimeLimit`. I have assigned this case to the status `s.USER_LIMIT`. This situation happens when solving long mixed-integer programs: the solver has a feasible solution but it hasn't proven its optimality yet. At the moment, CVXPY just discards that solution and throws an error. With this PR Gurobi can return a feasible solution, if `solCount > 0`. 

- I did the same when OSQP hits the maximum time so that it returns the last ADMM iterate (this is the same as what is happening with SCS).

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -34,7 +34,7 @@ class OSQP(QpSolver):
         # Map OSQP statuses back to CVXPY statuses
         status = self.STATUS_MAP.get(solution.info.status_val, s.SOLVER_ERROR)
 
-        if status in s.SOLUTION_PRESENT + [s.USER_LIMIT]:
+        if status in s.SOLUTION_PRESENT:
             opt_val = solution.info.obj_val + inverse_data[s.OFFSET]
             primal_vars = {
                 OSQP.VAR_ID:
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -17,6 +17,7 @@ class OSQP(QpSolver):
                   3: s.INFEASIBLE_INACCURATE,
                   -4: s.UNBOUNDED,
                   4: s.UNBOUNDED_INACCURATE,
+                  -6: s.USER_LIMIT,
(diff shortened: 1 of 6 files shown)


cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -27,37 +28,13 @@ def import_solver(self):
...
-        return data, inv_data
-
     def invert(self, solution, inverse_data):
         attr = {s.SOLVE_TIME: solution.info.run_time}
 
         # Map OSQP statuses back to CVXPY statuses
         status = self.STATUS_MAP.get(solution.info.status_val, s.SOLVER_ERROR)
 
-        if status in s.SOLUTION_PRESENT:
+        if status in s.SOLUTION_PRESENT + [s.USER_LIMIT]:
```
SteveDiamond, line 37: Should we update solution present to include user limit?

SteveDiamond:
RE questions, USER_LIMIT status is definitely better than throwing away the solution.

bstellato:
Thanks for the comments! I think adding `USER_LIMIT` to `SOLUTION_PRESENT` would work for continuous solvers. However, if the problem is mixed-integer you might get an error in trying to access the solution when no incumbent has been found so far.

Maybe we can add `USER_LIMIT` to `SOLUTION_PRESENT` and then add a check in the mixed-integer solver interfaces to see if an incumbent was found, e.g., `model.solCount > 0` for Gurobi. That's what I did in the last commit. 

I have also added a small test to solve the problem from scratch vs with cached `ParamQuadProg` with every installed QP solver. Does it work for you?

SteveDiamond:
Thanks! Looks great.

SteveDiamond:
We don't need to update any other solvers, because USER_LIMIT only appears in the gurobi_qpif and osqp_qpif interfaces.

PR #1031, opened 2020-05-22
Default solvers with both Conic and QP interfaces to the QP one.

bstellato:
Solvers with both Conic and QP interfaces default to their QP interface when solving LPs/QPs.

This is a much more efficient way to call GUROBI and CPLEX.


SteveDiamond:
LGTM.

PR #1032, opened 2020-05-22
Accelerate parametric QPs

### LINKED ISSUE #1030: Parametric QPs slower on 1.1.0a4 compared to 1.0.31
**Describe the bug**
CVXPY 1.1.0a4 takes longer to solve a sequence of QPs  compared to 1.0.

**To Reproduce**

```python
import cvxpy as cp
import numpy as np
from time import time
from tqdm import tqdm

print("CVXPY version: ", cp.__version__)

# Problem data.
m = 300
n = 200
np.random.seed(1)
A = np.random.randn(m, n)
b = np.random.randn(m)
gamma = cp.Parameter(nonneg=True, name="gammix")
x = cp.Variable(n)
N = 100
solver = cp.OSQP  
# solver = cp.GUROBI


def solve_sequence(solver):

    print("Solver", solver)

    objective = cp.Minimize(cp.sum_squares(A @ x - b) + gamma * cp.norm(x, 1))
    constraints = [0 <= x, x <= 1]

    t_solve_full = 0
    t_start = time()
    for n in tqdm(range(N)):
        gamma.value = 1
        prob.solve(solver=solver)
        t_solve_full += prob.solver_stats.solve_time/N
    t_full = time() - t_start


    t_solve_param = 0
    t_start = time()
    for n in tqdm(range(N)):
        gamma.value = 1
        prob.solve(solver=solver)
        t_solve_param += prob.solver_stats.solve_time/N
    t_param = time() - t_start

    print("Time full: %.2e" % t_full)
    print("Avg solve time full: %2e" % t_solve_full)
    print("Time param: %.2e" % t_param)
    print("Avg solve time param: %2e" % t_solve_param)

solve_sequence(solver)
```

**Expected behavior**
1.1.0a4 should be faster than 1.0.31 given the new parametric infrastructure.

**Output**

This is the output with *OSQP* (new version is slower):
```
CVXPY version:  1.1.0a4
Solver OSQP
100%|███████████████████████████████████████████████████████████████| 100/100 [00:16<00:00,  6.05it/s]
100%|███████████████████████████████████████████████████████████████| 100/100 [00:13<00:00,  7.19it/s]
Time full: 1.65e+01
Avg solve time full: 1.168879e-01
Time param: 1.39e+01
Avg solve time param: 1.157574e-01
```

```
CVXPY version:  1.0.31
Solver OSQP
100%|█████████████████████████████████████████| 100/100 [00:14<00:00,  6.68it/s]
100%|█████████████████████████████████████████| 100/100 [00:03<00:00, 28.02it/s]
Time full: 1.50e+01
Avg solve time full: 1.214789e-01
Time param: 3.57e+00
Avg solve time param: 7.173415e-03
```

This is the output with *GUROBI* (no noticeable improvement):
```
CVXPY version:  1.1.0a4
Solver GUROBI
100%|███████████████████████████████████████████████████████████████| 100/100 [00:23<00:00,  4.34it/s]
100%|███████████████████████████████████████████████████████████████| 100/100 [00:20<00:00,  4.81it/s]
Time full: 2.30e+01
Avg solve time full: 1.301685e-01
Time param: 2.08e+01
Avg solve time param: 1.350439e-01
```

```
CVXPY version:  1.0.31
Solver GUROBI
100%|█████████████████████████████████████████| 100/100 [00:20<00:00,  4.93it/s]
100%|█████████████████████████████████████████| 100/100 [00:21<00:00,  4.72it/s]
Time full: 2.03e+01
Avg solve time full: 1.305531e-01
Time param: 2.12e+01
Avg solve time param: 1.440245e-01
```

**Version**
 - OS: Mac OSX Catalina
 - CVXPY Version: 1.0.31 vs 1.1.0a4

**Observations**
- OSQP is way faster to solve the parametric program in 1.0.31. It looks like it is not warm-started nor cached.

- Gurobi takes similar time in both cases. However, the time improvement seems really minor given the fact that in the `full` case we are always recreating the cvxpy problem.

Do you guys have any idea why this is happening?

bstellato:
Fixes #1030. It includes the commits in PR #1031 .

I have merged the warm-start fix 254e404c623a6e036ab054adcfe1977ccecfe2c0 @SteveDiamond made and included the @akshayka changes from branch [qp_reduce_asa_map](https://github.com/cvxgrp/cvxpy/tree/qp_reduce_asa_map).

These are the benchmark scripts with profiling and timing outputs
[test_param_prog.zip](https://github.com/cvxgrp/cvxpy/files/4670508/test_param_prog.zip)

Output. There is a good improvement now for OSQP both in terms of warm-start and cvxpy time.  Canonicalization is still ~3x slower for QPs compared to Conic programs for this example.
```
CVXPY version:  1.1.0a4
Solver:  OSQP
Full solve
 - Avg total time: 3.7903 sec
 - Avg cvxpy time: 0.7871 sec
 - Avg solve time: 3.0031 sec
Param solve
 - Avg total time: 0.1721 sec
 - Avg cvxpy time: 0.1064 sec
 - Avg solve time: 0.0658 sec
Solver:  SCS
Full solve
 - Avg total time: 2.1966 sec
 - Avg cvxpy time: 1.2895 sec
 - Avg solve time: 0.9071 sec
Param solve
 - Avg total time: 0.3405 sec
 - Avg cvxpy time: 0.0352 sec
 - Avg solve time: 0.3053 sec

CVXPY version:  1.0.31
Solver:  OSQP
Full solve
 - Avg total time: 3.1999 sec
 - Avg cvxpy time: 0.1766 sec
 - Avg solve time: 3.0233 sec
Param solve
 - Avg total time: 0.2719 sec
 - Avg cvxpy time: 0.2034 sec
 - Avg solve time: 0.0684 sec
Solver:  SCS
Full solve
 - Avg total time: 1.1476 sec
 - Avg cvxpy time: 0.2132 sec
 - Avg solve time: 0.9344 sec
Param solve
 - Avg total time: 0.5300 sec
 - Avg cvxpy time: 0.2215 sec
 - Avg solve time: 0.3085 sec
```


bstellato:
From a quick snakeviz inspection, it looks like cvxpy 1.1.0a4 is a very thin layer around the solver call in Conic problems. Not so much for QPs.

Conic
Screen Shot 2020-05-22 at 5 46 33 PM

QPs
Screen Shot 2020-05-22 at 5 49 53 PM

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -35,9 +35,11 @@ def _is_lp(self):
 
 
 def _solve_as_qp(problem, candidates):
-    if _is_lp(problem) and candidates['conic_solvers']:
-        # OSQP can take many iterations for LPs; use a conic solver
-        # instead
+    if _is_lp(problem) and \
+            [s for s in candidates['conic_solvers'] if s not in candidates['qp_solvers']]:
```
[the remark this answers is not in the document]
SteveDiamond: Oops sorry I merged the other PR about this :(

SteveDiamond:
Probably to speed up QPs completely, you would also need to apply Akshay's method to the P matrix.

akshayka:
@bstellato , I think Steven's right by the way --- the following snippet

```python
            if sp.issparse(self.P):
                tensor_application = self.P @ sp.csc_matrix(param_vec[:, None])
```

is going to be very slow. The matrix `self.P` is usually very sparse, and `param_vec` is always dense. 

You should be replicate what I did to speed this up (use `canonInterface.reduce_problem_data_tensor` to get a dense reduced version of P and the indices & indptr for constructing the data matrix). You can look at the commit which introduced that function to get a sense of how to use it.

You don't necessarily have to do ^ in this PR, though. Once you address the comment I left, we can merge this in.

cvxpy/reductions/solvers/solving_chain.py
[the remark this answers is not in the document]
bstellato: I thought that that's the intended behavior, i.e., we use one of the default open-source solvers if they are available. If we want to use Gurobi, we can specify it and it becomes the only available candidate solver. If we do not specify it, we use ECOS for LPs and Gurobi does not matter. Maybe I am missing something.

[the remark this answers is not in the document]
akshayka: You’re right, I had assumed that Gurobi had a higher priority than ECOS, but I was wrong. (The priorities are here https://github.com/cvxgrp/cvxpy/blob/7abf8509aad7478d20c8988500db6cb4cbd12bc3/cvxpy/reductions/solvers/defines.py#L63)

bstellato:
Thank you @akshayka for the tip on tuna. I have adapted your functions `reduce_problem_data_tensor` and `get_matrix_and_offset_from_tensor` (now called `get_matrix_from_tensor` with an option to return the offset. With `P` it does not return any offset. I edited the docstrings accordingly).

I have also added a test to make sure matrix `P` is correct when parameters vary.

I also took the liberty to remove lines https://github.com/cvxgrp/cvxpy/blob/f41ef797c5ff4c54af8e6b8c433fef8275a5adc7/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L73-L78 from the OSQP QP interface. They were used to check if the sparsity pattern does not change when we have OSQP present in the cache. I believe they might be unnecessary because, if the cache is there, the problem should be just changing the values of the parameters (objective and constraints are not editable). Happy to bring them back if I got this wrong.

Now I get these timings:
```
CVXPY version:  1.1.0a4
Solver:  OSQP
---
Param solve
 - Avg total time: 0.1476 sec
 - Avg cvxpy time: 0.0768 sec
 - Avg solve time: 0.0708 sec
Solver:  SCS
---
Param solve
 - Avg total time: 0.3768 sec
 - Avg cvxpy time: 0.0405 sec
 - Avg solve time: 0.3363 sec
```

`ParamQuadProg` is still slightly slower than `ParamConeProg`, but less than 2x. Here are the updated OSQP performance profiles:
[benchmark.zip](https://github.com/cvxgrp/cvxpy/files/4693374/benchmark.zip)

Here are two pictures from tuna visualizer:

Before:
before

After
after2

Let me know what you think!

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -153,25 +178,37 @@ def param_value(idx):
...
-            self.q, param_vec, self.x.size)
+        # Old
+        #  if param_vec is None:
+        #      tensor_application = self.P
+        #  else:
+        #      if sp.issparse(self.P):
+        #          tensor_application = self.P @ sp.csc_matrix(param_vec[:, None])
+        #      else:
+        #          tensor_application = self.P @ param_vec
+        #  P = tensor_application.reshape((self.x.size, self.x.size), order='F').tocsc()
```
akshayka, line 189: Delete please

cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py
```diff
@@ -153,25 +178,37 @@ def param_value(idx):
...
+
+        # New
```
akshayka, line 191: & delete this as well

bstellato: Sorry I left the comments. It should be fixed now.

cvxpy/tests/test_param_quad_prog.py
```diff
@@ -33,6 +34,45 @@ def assertItemsAlmostEqual(self, a, b, places=2):
...
+            prob.solve(solver=solver)
+
+            # Get data with new parameter
+            gamma.value = gamma_val_new
+            data_param_new, _, _ = prob.get_problem_data(solver)
+            prob.solve(solver=solver)
+            x_gamma_new = np.copy(x.value)
+
+            # Check if data match
+            self.assertAlmostEqual(spa.linalg.norm(data_param_new['P'] - data_scratch['P']), 0)
```
akshayka, line 71: use `np.testing.assert_allclose(a.todense(), b.todense())` -- we'll get useful debugging output when/if it fails.

akshayka:
Looks good modulo some minor comments I've left. Thanks for doing this!

bstellato:
Done! I have used numpy testing to compare arrays. 

One comment: if we really want to save time/memory, we can form only the upper triangular part of `P` since it is symmetric. It might help speeding it up but I believe for now it is good.

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -70,12 +70,7 @@ def solve_via_data(self, data, warm_start, verbose, solver_opts,
         if solver_cache is not None and self.name() in solver_cache:
```
SteveDiamond, line 70: getting rid of same_pattern check is fine. Can you get rid of the same_pattern boolean altogether though?

SteveDiamond:
One comment about same_pattern, otherwise looks good!

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
bstellato: Sounds good! I removed it.

SteveDiamond:
@bstellato looks like you need to loosen the tolerance on the new param qp test. That's the cause of the Appveyor failure.

SteveDiamond:
Nicely done! We're ready to release 1.1 IMO.

PR #1034, opened 2020-05-23
Anaconda --> conda in install docs

akshayka:
and fix broken link.

conda is a package/environment management system, anaconda is a distribution


SteveDiamond:
Good catch!

Issue #797, opened 2019-08-26
SCIP solver support 


wfrece:
**Is your feature request related to a problem? Please describe.**
I would like to have SCIP as a solver for CVXPY, so that I do not have to use native PySCIPOpt to optimize MILP problem sets. 

**Describe the solution you'd like**
Self-explanatory - please add SCIP lib to the current list of supported solvers. 

**Describe alternatives you've considered**
Using PySCIPOpt directly is an option, but its difficult to do, mainly because of the extra libs that I'm wanting to use in CVXPY that don't seem usable if I'm not also using CVXPY to do the solving. 

**Additional context**
None.

wfrece:
Hey Steven,

Thanks for getting back to me. We need it, but I'm not sure i'm a
sophisticated enough of a developer to do this. I don't understand most of
the code in the PR that you sent, and I barely understand enough of CVXPY
to even use it. We got something working with CVXPY and GUROBI, but we
can't use GUROBI because it costs money. We really need to use SCIP, which
is the leading open source (free) MILP solver, but we really need help with
having it be supported in CVXPY. Anyone who can maybe assist us with this
task?

rileyjmurray:
@wfrece my advice would be to ask one of SCIP's developers to take the lead on this. Right now CVXPY does not have a good open-source MIP solver, and SCIP might be a great place to look (since it is under very active development, and has so many people involved in the project).

SCIP developers are listed here: https://scip.zib.de/index.php#developers. If you can get the attention of someone on their python side of things (e.g. Matthias Miltenberger, Felipe Serrano), then we can work with them to get an interface in CVXPY.

rimaddo:
Hi, did you ever manage to find someone to do this? I am a developer who studied maths/operations research at university. This looks really interesting, I had a look at [#776](https://github.com/cvxgrp/cvxpy/pull/776) and I think I could replicate for SCIP, if you are still looking for someone to help out?

adishavit:
That would be fantastic Rachel.
I was still hoping this would be possible - so any help you may
provide would be greatly appreciated!!

rileyjmurray:
@rimaddo we'd love a SCIP interface! I'm actually in the process of writing up documentation to help people write new solver interfaces. You can find the working draft of that documentation at [this link](https://github.com/rileyjmurray/cvxpy/blob/prep1.1docs/doc/source/contributing/index.rst).

[...]

PR #1035, opened 2020-05-24
SCIP interface

rimaddo:
# Description

Adds a [SCIP](https://scip.zib.de/) solver interface as requested in [#797](https://github.com/cvxgrp/cvxpy/issues/797).

This is a bit of an in-progress PR since there is a test (`test_socp_1`) which is not yet passing (fails on checking the complementary) and also a non-linear constrain logging warning, which I will continue to look at.

However apart from this I think that the bulk of the functionality is present.

It would be great to get some feedback about if this looks like it is progressing in the right direction and if there are any changes needed.

Many thanks


cvxpy/__init__.py
```diff
@@ -15,18 +15,40 @@
 """
 
 __version__ = "1.1.0a4"
+
+import cvxpy.interface.scipy_wrapper
 from cvxpy.atoms import *
-from cvxpy.constraints import NonPos, Zero, SOC, PSD
+from cvxpy.constraints import NonPos, PSD, SOC, Zero
+from cvxpy.error import DCPError, DGPError, DPPError, SolverError, disable_warnings, enable_warnings, warnings_enabled
+from cvxpy.expressions.constants import CallbackParam, Constant, Parameter
```
rimaddo, line 23: One possible issue; I have a setting which will auto-alphabetise imports, which has created some noise in a couple of files that I have added to like this. If you would like me to revert this please let me know and I will do so.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,434 @@
...
+    "inforunbd": s.UNBOUNDED_INACCURATE,
+    # ERROR
+    "userinterrupt": s.SOLVER_ERROR,
+    "memlimit": s.SOLVER_ERROR,
+    "sollimit": s.SOLVER_ERROR,
+    "nodelimit": s.SOLVER_ERROR,
+    "totalnodelimit": s.SOLVER_ERROR,
+    "stallnodelimit": s.SOLVER_ERROR,
+    "restartlimit": s.SOLVER_ERROR,
+    "unknown": s.SOLVER_ERROR,
```
rimaddo, line 63: I wasn't completely sure about this mapping, I originally had a lot of the errors set as `USER_LIMIT`, for example `memlimit`  however since this was updated to `SOLUTION_PRESENT` and I didn't think this could be guaranteed then I have reverted a lot to `SOLVER_ERROR`.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,434 @@
...
+        model.setHeuristics(SCIP_PARAMSETTING.OFF)
+        model.disablePropagation()
+
+        # Set model verbosity
+        hide_output = not verbose
+        model.hideOutput(hide_output)
+
+        # User-input parameters
+        if solver_opts:
+            model.setParams(**solver_opts)
```
rimaddo, line 278: For this, if I am passing params to the SCIP model should they not be in `solver_opts` directly but instead `solver_opts["scip_params"]` more like the cplex implementation? i.e. are `solver_opts` actually for `cvxpy`? Would like to add some tests around using params to understand better perhaps.

cvxpy/__init__.py
SteveDiamond: No it's fine.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
SteveDiamond: ``USER_LIMIT`` is new, so the semantics are still loose. To me, it seems useful to capture all the cases where some solution is returned. Another approach would be to be more literal and cover every case where some kind of user imposed limit led to the solver not terminating, but that doesn't seem as useful.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
SteveDiamond: Either way is ok. Are SCIP parameters really extensive and complicated? Then I would start with passing in a ``scip_params`` dict. @rileyjmurray what's your opinion?

cvxpy/__init__.py
SteveDiamond: Actually I'd prefer you revert these changes and turn off the setting while you're working on this. I hadn't seen the full scope of the changes yet when I commented before.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,434 @@
...
+            model=model,
+            variables=variables,
+            constraints=constraints,
+            data=data,
+            dims=dims,
+            solver_opts=solver_opts,
+        )
+        return solution
+
+    def _define_data(self, data: Dict[str, Any]) -> Tuple:
```
SteveDiamond, line 180: For better or worse, we don't use type annotation in cvxpy.

SteveDiamond:
Hi! It's looking good so far. I made a few comments. There are some new docs up on contributing solver interfaces: https://www.cvxpy.org/contributing/index.html#solver-interfaces

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
SteveDiamond: So to be clear, my suggestion is to use ``USER_LIMIT`` only when a solution can be returned.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
SteveDiamond: I consulted with other devs. Type annotations are fine. It would be nice to have more of them in cvxpy!

cvxpy/__init__.py
rimaddo: No worries, reverted.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: Thanks for clarifying. Is there any status where if a solution is present is undetermined until further inspection? Most "limits" I would say could fall into this category, for example if you reach a time limit that doesn't guarantee a solution has been found, you could hit your time limit and have no solution or a sub-optimal solution.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: Great, would be happy to help add some more after this if it would be helpful, let me know.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
SteveDiamond: You can do the inspection in the invert or solve_via_data method, right? I would do it there and call it USER_LIMIT if the solution is present, otherwise SOLVER_ERROR.

rimaddo: Right yes, will do!

cvxpy/tests/test_conic_solvers.py
```diff
@@ -15,16 +15,17 @@
...
 import cvxpy as cp
-import unittest
+from cvxpy import SCIP
 from cvxpy.error import SolverError
 from cvxpy.tests.base_test import BaseTest
-from cvxpy.tests.solver_test_helpers import StandardTestECPs, StandardTestSDPs
-from cvxpy.tests.solver_test_helpers import StandardTestSOCPs, StandardTestLPs
-from cvxpy.tests.solver_test_helpers import StandardTestMixedCPs
-from cvxpy.reductions.solvers.defines import INSTALLED_SOLVERS, INSTALLED_MI_SOLVERS
+from cvxpy.tests.solver_test_helpers import StandardTestSOCPs, StandardTestLPs, StandardTestMixedCPs
```
SteveDiamond, line 27: It's good to merge the imports, but you need to add ''StandardTestECPs, StandardTestSDPs''. That's why the build is failing.

rimaddo: Sorry I missed that the last couple of days, will get more time to spend on this over the weekend and try to address the last bits!

SteveDiamond:
Is this done? We're going to release 1.1 soon (maybe today?), but we can merge this in after that release.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -975,13 +982,57 @@ def test_nag_socp_1(self):
...
+
+    def test_scip_lp_4(self):
+        StandardTestLPs.test_lp_4(solver=SCIP)
+
+    def test_scip_socp_0(self):
+        StandardTestSOCPs.test_socp_0(solver=SCIP)
+
+    # Fails when checking complementary, TODO: fix.
+    # def test_scip_socp_1(self):
+    #     StandardTestSOCPs.test_socp_1(solver=SCIP, places=3)
```
rimaddo, line 1015: All tests pass apart from this one.

rimaddo:
> Is this done? We're going to release 1.1 soon (maybe today?), but we can merge this in after that release.

It's done apart from one test [here](https://github.com/cvxgrp/cvxpy/pull/1035/files#r433502857) which fails, I haven't been able to work out how to correct this yet. It has incorrect complementary and dual values.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -975,13 +982,56 @@ def test_nag_socp_1(self):
...
+        StandardTestLPs.test_lp_3(solver=SCIP)
+
+    def test_scip_lp_4(self):
+        StandardTestLPs.test_lp_4(solver=SCIP)
+
+    def test_scip_socp_0(self):
+        StandardTestSOCPs.test_socp_0(solver=SCIP)
+
+    def test_scip_socp_1(self):
+        StandardTestSOCPs.test_socp_1(solver=SCIP, places=3, duals=False)
```
rimaddo, line 1014: This now passes this test, for which it was previously failing on the dual/complementary. However I am just skipping these tests via `duals=False`. The function in SCIP to get the dual is `getDualsolLinear` and there does not seem to be a non-linear equivalent. Additionally there are some issues which mention that duals can only be provided for linear constraints in SCIP, for example [here](https://github.com/SCIP-Interfaces/PySCIPOpt/issues/228). Therefore, based on this I think that this is just not possible to compute with SCIP. Is this an acceptable solution for you?

SteveDiamond: I see. In that case, it's fine to only return dual values for linear constraints.

SteveDiamond:
When this is finished, add @akshayka or @rileyjmurray as reviewers.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,430 @@
...
+
+
+class VariableTypes:
+    """Variable type constants."""
+    BINARY = "BINARY"
+    INTEGER = "INTEGER"
+    CONTINUOUS = "CONTINUOUS"
+
+
+class SCIP(SCS):
```
rileyjmurray, line 78: I'm a bit of a stickler for inheriting from a solver that doesn't have a clear connection at the solver at hand. For an example where I support doing this, there's GLPK inheriting from CVXOPT, because we use the CVXOPT python package as our interface with the GLPK solver.

That said-- you've already done a substantial amount of work on this PR, and this type of inheritance structure is not unique to your contribution. So, my request is that this docstring explain which aspects of the SCS class the SCIP class accesses or relies upon (e.g. dual variable recovery).

If for some reason you're up for removing inheritance from SCS, I'd of course support that as well.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rileyjmurray: It's valid to use *both* direct keyword arguments and a dict keyword argument. For example, CVXPY will allow a command like
```
val1 = True
val2 = 1.0
val3 = 'on'
param_dict = {'option3': val3}
prob.solve(solver='SCIP',scip_param_1=val1, scip_param_2=val2 ,scip_params=param_dict)
```
and the pairs ``{'scip_param1': val1, 'scip_param2': val2, 'scip_params': param_dict}`` will be visible during a call to ``solve_via_data``.

If your solver interface only plans on exposing simple settings (like a maximum number of branch and bound nodes), then I think a user should be able to specify such a parameter without providing an entire ``scip_params'' dict. But it's fair to assume that fancy solver settings will be passed as a dict. The MOSEK interface is an example where both types of arguments are allowed.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,430 @@
...
+                )
+
+                eq_dual.update(leq_dual)
+                dual_vars = eq_dual
+
+            return Solution(status, opt_val, primal_vars, dual_vars, {})
+        else:
+            return failure_solution(status)
+
+    def solve_via_data(
```
rileyjmurray, line 133: The formatting of the body of this function makes it unnecessarily long. I think the following snippet is far more readable for people using standard landscape-oriented displays.
```
    def solve_via_data(
            self,
            data: Dict[str, Any],
            warm_start: bool,
            verbose: bool,
            solver_opts: Dict[str, Any],
            solver_cache: Dict = None,
    ) -> Solution:
        """Returns the result of the call to the solver."""
        from pyscipopt.scip import Model
        model = Model()
        A, b, c, dims = self._define_data(data)
        variables = self._create_variables(model, data, c)
        constraints = self._add_constraints(model, variables, A, b, dims)
        self._set_params(model, verbose, solver_opts)
        solution = self._solve(model, variables, constraints, data, dims)
        return solution
```

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,430 @@
...
+                    if ctype == ConstraintTypes.EQUAL
+                    else (expression <= b[i])
+                )
+                constraints.append(constraint)
+            else:
+                constraints.append(None)
+
+        return constraints
+
+    def add_model_soc_constr(
```
rileyjmurray, line 357: Since the dual variables for SOC don't work yet, there is a high chance that someone else will modify this function some day. Right now it's formatted in a way which I think will be hard for people to parse.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -0,0 +1,430 @@
...
+        ] + [
+            model.addVar(
+                obj=0,
+                name="soc_x_%d" % i,
+                vtype=VariableTypes.CONTINUOUS,
+                lb=None,
+                ub=None,
+            )
+            for i in rows[1:]
+        ]
```
rileyjmurray, line 399: This code block in particular is very hard to read. I've never seen a list comprehension span 8 lines before. Can you please rewrite this code block so that it uses a standard for-loop?

cvxpy/tests/test_conic_solvers.py
```diff
@@ -975,13 +982,56 @@ def test_nag_socp_1(self):
     def test_nag_socp_2(self):
         StandardTestSOCPs.test_socp_2(solver='NAG')
 
-    def test_ecos_socp_3(self):
+    def test_nag_socp_3(self):
```
rileyjmurray, line 985: Thanks for catching this!

cvxpy/tests/test_conic_solvers.py
```diff
@@ -15,16 +15,23 @@
...
 
 import math
+import unittest
+
 import numpy as np
 import scipy.linalg as la
+
 import cvxpy as cp
-import unittest
+from cvxpy import SCIP
```
rileyjmurray, line 24: Don't import SCIP. It should be possible to use a solver without a user calling an import like this.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -975,13 +982,56 @@ def test_nag_socp_1(self):
...
 
-    def test_ecos_socp_3(self):
+    def test_nag_socp_3(self):
         # axis 0
         StandardTestSOCPs.test_socp_3ax0(solver='NAG')
         # axis 1
         StandardTestSOCPs.test_socp_3ax1(solver='NAG')
 
 
+@unittest.skipUnless(SCIP in INSTALLED_SOLVERS, 'SCIP is not installed.')
```
rileyjmurray, line 992: The condition should be ``'SCIP' in INSTALLED_SOLVERS``.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -975,13 +982,56 @@ def test_nag_socp_1(self):
...
+class TestSCIP(unittest.TestCase):
```
rileyjmurray, line 993: During calls to tests, use ``solver='SCIP'``.

cvxpy/tests/test_conic_solvers.py
rileyjmurray: @rimaddo the GitHub issue you linked doesn't seem to be about nonlinear models in the usual sense of the word, rather, it's about a linear relaxation for a mixed integer linear program. I've looked through the SCIP documentation and found [this mention](http://scip-interfaces.github.io/PySCIPOpt/docs/html/classpyscipopt_1_1scip_1_1NLRow.html#ad970d559414db82732bf4ac5af1a0b9b) of dual variables to nonlinear constraints. Could that function be relevant for SOC dual variable recovery here?

rimaddo: Thanks for the link, apologies for the slow reply on this, thanks for pointing out. I will have a look today and see if I can use it.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: Will do

cvxpy/tests/test_conic_solvers.py
rimaddo: Sure, this and the next couple make sense, I will change all to a direct string rather than the string variable!

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: No worries, that makes sense. I have removed it and am now inheriting from `ConicSolver` instead. I think I had it in not so much from need but more because I did a lot of learning from looking at the Gurobi implementation because this is the solver I am most familiar with, and it's inherits from `SCS`. One suggestion if you don't mind me making it, it could be really nice to have an abstract base class which is the standard thing to inherit off? Since there is a very set pattern and set of methods which every implementation class needs.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: Sure, changed as suggested.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
rimaddo: I've added a `scip_params` option since many of the more advanced params have `/`'s in them which can't be passed as kwargs and added some test around it see [here](https://github.com/rimaddo/cvxpy/blob/feature/SCIP-interface/cvxpy/tests/test_conic_solvers.py#L1047) and the 4 following.

cvxpy/tests/test_conic_solvers.py
rileyjmurray: This file should not import solver-related constants. Whenever the SCIP constant is needed, pass it as an explicit string (that's the most common way users specify a solver).

rileyjmurray: Sorry, ignore this comment. (And a few others I left in error.)

rileyjmurray:
@rimaddo, if you're having trouble with the dual variables, I wouldn't consider that a hard requirement for us to merge this PR. You've done a tremendous amount of work already. Even documenting where the interface stands on dual variables would be a fair resolution as far as I'm concerned.

rimaddo:
> @rimaddo, if you're having trouble with the dual variables, I wouldn't consider that a hard requirement for us to merge this PR. You've done a tremendous amount of work already. Even documenting where the interface stands on dual variables would be a fair resolution as far as I'm concerned.

Hi @rileyjmurray yeah I haven't been able to figure out why but I can never get that one case to work, not sure if I'm missing something obvious or what. Thanks for clarifying that it's not a hard requirement, I will add some documentation on where dual variables are at over the weekend as suggested. Thanks!

doc/source/install/index.rst
```diff
@@ -169,6 +169,18 @@ CVXPY supports the NAG solver.
...
 See the `NAG <https://www.nag.co.uk/nag-library-python>`_ website for installation instructions.
 
+Install with SCIP support
+---------------------------
+
+CVXPY supports the SCIP solver.
+Simply install SCIP such that you can ``from pyscipopt.scip import Model`` in Python.
+See the `PySCIPOpt <https://github.com/SCIP-Interfaces/PySCIPOpt#installation>`_ github for installation instructions.
+
+It is important to note that the currently the SCIP solver does not consistently calculate duals correctly, therefore if you are looking to use the dual values specifically this may not be the best solver choice for you.
```
rimaddo, line 179: I added a duals note here @rileyjmurray, wasn't sure where to put it hope this is okay, let me know if you want any changes.

rileyjmurray: I would rephrase it as follows.

"CVXPY's SCIP interface does not recover dual variables for nonlinear constraints. If you require dual variables for a continuous nonlinear problem, you will need to use another solver. We welcome additional contributions to the SCIP interface, to recover dual variables for nonlinear constraints in continuous problems."

That comment assumes the interface recovers dual variables to linear equations and linear inequalities correctly. That is the case, right?

rimaddo: Sorry for the slow reply, `test_socp_1` is the test that I have being trying to fix but `test_lp_2` also skips checking the dual. I think I looked into this one less because it was already being skipped for `CBC`

rimaddo: I've updated without the `nonlinear`?

rileyjmurray: @rimaddo I wasn't aware that CBC duals have been turned off. I'm okay with leaving things as-is, then. After merging in the PR we can create a GitHub issue for SCIP dual variables. (Probably also CBC dual variables.)

rimaddo:
> This is an awesome contribution --- I suspect a lot of people will find it useful. Thanks so much for your work!

No worries, it was interesting to do! Thanks

PR #1040, opened 2020-05-28
Change raw latex to math so equations display correctly

tommyod:
Hello! First and foremost: thank you for the great library.

Some equations seem to not display correctly in the examples, see [this page](https://www.cvxpy.org/examples/applications/robust_kalman.html) for an example

## Before

Here's what it looks like in Firefox:

image

## After

Changing the lines to `.. math::` seems to do the trick:

image

I also have a question. I see that the files below are identical:
```
./examples/notebooks/WWW/robust_kalman.ipynb
./doc/source/examples/applications/robust_kalman.rst
```

**Question.** Is the `.rst` the result of a manual export from the notebook, or is it automated? If it's not automated, I think it would be better to have the code in one place and automate this.

### DIFF
--- a/doc/source/examples/applications/ant_array_min_beamwidth.rst
+++ b/doc/source/examples/applications/ant_array_min_beamwidth.rst
@@ -37,7 +37,7 @@ performed by bisection, where the interval which contains the optimal
 value is bisected according to the result of the following feasibility
 problem:
 
-.. raw:: latex
+.. math::
 
    \begin{array}{ll}
    \mbox{minimize}   &  0 \\
@@ -56,7 +56,7 @@ the antenna array configuration and specs.
 Once the optimal beamwidth is found, the solution :math:`w` is refined
 with the following optimization:
 
-.. raw:: latex
+.. math::
 
    \begin{array}{ll}
    \mbox{minimize}   &  \|w\| \\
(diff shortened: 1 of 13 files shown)


SteveDiamond:
It is automated. Thanks for fixing this!

tommyod:
@SteveDiamond If it's automated, won't this PR be without effect once the notebooks overwrite the `.rst` files? Where is the automation step? The only file that I found that seems to be exporting notebooks is `examples/machine_learning/generate_rst.sh`.

SteveDiamond:
We convert the .ipynb files to .rst files using a jupyter function. So it's not generated each time, only when we call the function.

tommyod:
Alright! Where is the jupyter function located? 

If I want to read through the examples and do some proofreading as I go, it is meaningless to do it in the `.rst` files since they will be overwritten? In other words: If I make changes, should I do it in the `.rst` files or `.ipynb` files?

SteveDiamond:
Changing the ``.rst`` files is correct. They would be overwritten if we changed the ``.ipynb`` file and needed to generate the  ``.rst``, but this is only done manually. It's not automated when we build cvxpy or something. We make the  ``.rst`` files using ``ipython nbconvert --to rst notebook.ipynb``.

PR #1041, opened 2020-05-29
Updated DCP Tutorial Doc

kevkevinpal:
**Few Documentation changes in DCP Tutorial doc**

1. Was looking at the tutorial and noticed some of the code was outdated and updated it to python 3.7.7

2. I noticed that we return UNKNOWN for one of the examples but I guessing after updates we now return QUASICONVEX so I updated the doc to show that

3. lastly the photos for the explanation on how the atoms build up the expression did not include quasi-convex and it would be make more sense if it matched the change mentioned above. I am not an expert at photo editing so it probably look better than what I could do.


SteveDiamond:
I really appreciate your work, but I think it would be confusing to discuss quasiconvexity in the DCP tutorial.

kevkevinpal:
> I really appreciate your work, but I think it would be confusing to discuss quasiconvexity in the DCP tutorial.

I agree, do you think it would be better to replace the current example to one which is a convex function that returns UNKOWN but returns CONVEX when instead of sqrt we apply l2 norm?

 I can see if I can find such expressions so we don't need to explain/show quasiconvexity in the tutorial.

PR #1062, opened 2020-06-19
Cache auxiliary output for some solvers

akshayka:
We currently cache the OSQP/SCS results dicts in problem._solver_cache. This
is useful when debugging a call to solve. This
change also caches the ECOS results dict and the entire Gurobi model;
this makes it possible to inspect the solver's performance
programmatically, without looking at the verbose output.

We should implement similar changes for all solvers, in a future change. It might be
a good idea to expose cached solver information in a public API in a
future PR.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
@@ -56,7 +56,8 @@ class GUROBI(QpSolver):
 
         # Start populating attribute dictionary
         attr = {s.SOLVE_TIME: model.Runtime,
-                s.NUM_ITERS: model.BarIterCount}
+                s.NUM_ITERS: model.BarIterCount,
+                s.EXTRA_STATS: model}
 
         # Map GUROBI statuses back to CVXPY statuses
         status = self.STATUS_MAP.get(model.Status, s.SOLVER_ERROR)
@@ -202,8 +203,6 @@ class GUROBI(QpSolver):
         except Exception:  # Error in the solution
             results_dict["status"] = s.SOLVER_ERROR
 
-        if solver_cache is not None:
-            solver_cache[self.name()] = model
         results_dict["model"] = model
 
         return results_dict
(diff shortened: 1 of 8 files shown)


cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -202,6 +202,8 @@ def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=No
         except Exception:  # Error in the solution
             results_dict["status"] = s.SOLVER_ERROR
 
+        if solver_cache is not None:
+            solver_cache[self.name()] = model
```
SteveDiamond, line 206: Actually won't this clash with gurobi_conif? The name is GUROBI in both cases.

akshayka: Oh, yes, that's true. For the purpose of giving the user access to the solver's internal state, that's fine; but I guess what's not fine is the solver might mistakenly use cached state that doesn't belong to it.

I suppose there are two different uses for storing the solver's state: (1) reporting extra information about the solve back to the user, and (2) giving other parts of CVXPY (solver interfaces, the backward / forward methods) access to cached state.

So instead of piggybacking onto the solver_cache, we should instead come up with a separate API for (1). Here's one proposal: we add a new attribute to `SolverStats`, called `extra_stats` (or `solver_specific_stats`, or `misc`) that stores auxiliary information returned by the solver. For ECOS, SCS, and OSQP, it'd store a results dict. For Gurobi, it'd store the Gurobi model. etc. How does that sound?

SteveDiamond: That sounds good!

akshayka: Okay, done! @SteveDiamond , see what you think.

SteveDiamond: Looks good!

PR #1064, opened 2020-06-27
Add support for custom kkt solvers in cvxopt interface

### LINKED ISSUE #1042: Improve support for alternative kktsolver functions in cvxopt
CVXOPT lets users provide a function handle that implements an algorithm to solve linearized KKT systems. Right now CVXPY partially exposes that functionality, by allowing users to select one of two pre-made KKT solvers ("chol" is the default, "robust" uses an LDL factorization designed to handle linearly dependent constraints). It would be nice if this functionally was fully-exposed.

From what I can tell, it looks like CVXPY behaves correctly if a user provides a CVXOPT-compliant function handle for the KKT solver. (See [here](https://github.com/cvxgrp/cvxpy/blob/387c48e08b230664e4bb092bf4e32a5d3573347e/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py#L323) for how we parse this setting.) The only clear limitation is at [line 161](https://github.com/cvxgrp/cvxpy/blob/387c48e08b230664e4bb092bf4e32a5d3573347e/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py#L161), where we assume the KKT solver is one of the two pre-defined options. The computational expense ``remove_redundant_rows`` is potentially huge, so it would be good to avoid if a user provided their own KKT solver.

My current plan is to personally make the necessary changes here; I'm creating this GitHub issue largely so I don't lose track of this task.

rileyjmurray:
CVXOPT lets users pass a function handle which implements an algorithm to solve linearized KKT systems. This PR exposes that functionality to CVXPY users, and resolves #1042.

Under this PR, a user can now select a KKT solver by specifying one of six strings, or a custom function handle.
 * The strings correspond to five cvxopt built-in solvers, and the "robust" solver which is part of the cvxpy source code.
 * Users are directed to a specific part of cvxpy source code to determine how to implement a KKT solver of their own.

This functionality is only really relevant when a user is confident that a given cvxpy formulation leads to a suitably structured cone program. So from a practical perspective, I wouldn't recommend using this feature with nonlinear cvxpy atoms. If someone wants to take advantage of this feature for nonlinear convex optimization problems, they should formulate their problem directly as a cone program (e.g. with SOC constraints, Schur complements, etc...).

Remarks related to the main purpose of this PR
 1. I haven't added dedicated tests for this, however, I've slightly changed how cvxpy formulates its custom robust KKT solver. By doing this, the existing tests for cvxopt act as tests for the new proposed API. I'm happy to add more tests if anyone thinks that's a good idea.
 2. It might make sense to remove redundant constraints when using any one of the five CVXOPT solvers. Right now, the cvxpy interface to cvxopt only applies a presolve procedure when using cvxopt's ``chol`` KKT solver.

Changes incidental to this PR
 * Eigenvalue computation in ``Constant`` expressions was being performed with an incorrect call to [scipy.sparse.linalg.eigsh](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigsh.html#scipy.sparse.linalg.eigsh). Without the changes made here, the attribute ``A._bottom_eig`` of a Constant ``A`` was actually being set to the *maximum* eigenvalue when ``A`` is a symmetric PSD matrix which is exactly singular. At the time of first making this PR I'm not sure if my changes here are the best way to proceed. I'll investigate this further, and indicate when this part of the PR is ready for code review.
 * Formatting of an ``.rst`` file in web documentation.

### DIFF
--- a/cvxpy/expressions/constants/constant.py
+++ b/cvxpy/expressions/constants/constant.py
@@ -203,7 +203,7 @@ class Constant(Leaf):
 
         # Compute bottom eigenvalue if absent.
         if self._bottom_eig is None:
-            ev = smallest_eig_near_ref(self.value, EIGVAL_TOL)
+            ev = extremal_eig_near_ref(self.value, EIGVAL_TOL, low=True)
             self._bottom_eig = ev
 
         return self._bottom_eig >= -EIGVAL_TOL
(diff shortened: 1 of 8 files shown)


SteveDiamond:
It looks fine on the whole. A lot of changes, and I don't follow them all, but if all the tests still pass it's fine.

SteveDiamond:
Maybe you can add a custom kkt solver test where you just pass in the robust kkt solver function?

rileyjmurray:
Mutliple tests have been added for the CVXOPT kktsolver option. One of these tests involved creating a ``StandardTestLP`` with redundant equality constraints, which I've added to the default tests for most solvers.

I also I've determined that the changes I made to eigenvalue computation were appropriate. I've added some additional assertions in a unittest to support this.

PR #1065, opened 2020-06-29
Remove Gurobi conic interface bottleneck

rluce:
The conic interface to Gurobi processes the linear constraints through a DOK sparse matrix, which can easily induce a big overhead.   Consider the following example:
```
import numpy as np                                                              
import scipy as sp                                                              
import cvxpy as cp                                                              
import time                                                                     
                                                                                
m = 500                                                                         
n = 5000                                                                        
b = np.random.rand(m)                                                           
for d in np.linspace(0.1, 0.8, 3):                                              
    A = sp.sparse.random(m, n, density=d)                                       
    start_time = time.time()                                                    
    n = A.shape[1]                                                              
    x = cp.Variable(n)                                                          
    # Force routing through the conic interface, and make problem               
    # infeasible so that solve time is practically zero.                        
    prob = cp.Problem(cp.Minimize(0),                                           
            [A @ x >= b, x >= 0, cp.sum_squares(x) <= 3.14, x <= -1])           
    prob.solve(solver='GUROBI', verbose=False)                                  
    print("m={:5d} n={:4d} d={:4.2f}: time={:4.1f}".format(                     
        m, n, d, time.time() - start_time))
```

On my local machine I find for current upstream master:
```
m=  500 n=5000 d=0.10: time= 6.2
m=  500 n=5000 d=0.45: time=26.2
m=  500 n=5000 d=0.80: time=45.7
```

In this PR the data is processed in CSR format (similar to the QP interface), and I get:
```
m=  500 n=5000 d=0.10: time= 0.8
m=  500 n=5000 d=0.45: time= 2.9
m=  500 n=5000 d=0.80: time= 5.1
```

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
@@ -103,7 +103,8 @@ class GUROBI(SCS):
         """Returns the solution to the original problem given the inverse_data.
         """
         status = solution['status']
-        attr = {s.EXTRA_STATS: solution['model']}
+        attr = {s.SOLVE_TIME: solution[s.SOLVE_TIME],
+                s.EXTRA_STATS: solution['model']}
 
         primal_vars = None
         dual_vars = None
--- a/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
@@ -212,9 +212,6 @@ class GUROBI(SCS):
             variables += new_vars
             soc_start += constr_len
 
-        gur_constrs = eq_constrs + ineq_constrs + new_leq_constrs + soc_constrs
-        model.update()
-
         # Set parameters
         # TODO user option to not compute duals.
         model.setParam("QCPDual", True)
@@ -234,16 +231,11 @@ class GUROBI(SCS):
             # Only add duals if not a MIP.
             # Not sure why we need to negate the following,
             # but need to in order to be consistent with other solvers.
+            vals = []
             if not (data[s.BOOL_IDX] or data[s.INT_IDX]):
-                vals = []
-                for lc in gur_constrs:
-                    if lc is not None:
-                        if isinstance(lc, gurobipy.QConstr):
-                            vals.append(lc.QCPi)
-                        else:
-                            vals.append(lc.Pi)
-                    else:
-                        vals.append(0)
+                lin_constrs = eq_constrs + ineq_constrs + new_leq_constrs
+                vals += model.getAttr('Pi', lin_constrs)
+                vals += model.getAttr('QCPi', soc_constrs)
                 solution["y"] = -np.array(vals)
                 solution[s.EQ_DUAL] = solution["y"][0:dims[s.EQ_DIM]]
                 solution[s.INEQ_DUAL] = solution["y"][dims[s.EQ_DIM]:]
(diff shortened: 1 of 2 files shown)


SteveDiamond:
This seems fine. Do you know why test_gurobi_lp_3 is failing? It's failing on master too, but this would be a good place to fix it.

rluce:
> Do you know why test_gurobi_lp_3 is failing? It's failing on master too, but this would be a good place to fix it.

I'll take a look.

rileyjmurray:
@rluce I found this error in my other PR, but didn't bring it up. It seems that GUROBI is returning an "infeasible, inaccurate" status code for a problem which is in fact unbounded. The test problem currently has five variables, but the test will still fail if you modify it to have only a single variable. There is a chance that this is a problem with the GUROBI solver, rather than just a problem with the CVXPY interface to GUROBI.

rluce:
I looked at why `test_gurobi_lp_3` is failing.  The (Gurobi-) solution status for this model `INF_OR_UNBD` which means that the problem is, well, infeasible or unbounded.  This status however is mapped to the cvxpy status `s.INFEASIBLE_INACCURATE` (I not quite sure what inaccurate refers to though).  But then later in `gurobi_qpif.py:invert` this information is not enough to make the correct choice between +/-inf for the objective value (obviously, because it is unknown whether it's infeasible or unbounded).

The reason why Gurobi doesn't conclude simply "Unbounded" is that our presolve applies some dual reduction after which the certificates for infeasibility and unboundedness become indistinguishable. If a strong distinction between infeasible and unbouded is needed, the customary procedure is to disable such presolve reductions.  Generally this is a bad idea because this make presolve less effective and leads to longer solution times, but in the case of a unit test that needs to have this distinction I would suggest to simply set this parameter, i.e.,
```
index 92c55dda..43c757c9 100644
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -869,7 +869,7 @@ class TestGUROBI(BaseTest):
         StandardTestLPs.test_lp_2(solver='GUROBI')
 
     def test_gurobi_lp_3(self):
-        StandardTestLPs.test_lp_3(solver='GUROBI')
+        StandardTestLPs.test_lp_3(solver='GUROBI', DualReductions=0)
 
     def test_gurobi_lp_4(self):
         StandardTestLPs.test_lp_4(solver='GUROBI')
```
With that change the test no longer fails -- should I put it in the PR?

Of course this fixes only the unit test; improving how the `INF_OR_UNBD` status is handled better in general could be another work in the future.  Maybe spit out a warning to promt the user to resolve with `DualReductions=0` to determine the "true" status?

rileyjmurray:
@rluce thanks for looking into things. I think it is a good idea to print a warning of the kind you suggest, when GUROBI returns an "inf or unbounded" status. From a testing perspective I'd like two changes. First, run a version of ``test_gurobi_lp3`` which disables presolve by keyword argument. Then, add another test which allows presolve as usual, but then checks that a ``UserWarning`` was raised.

rluce:
@rileyjmurray For now I tweaked the existing test only.

On second thought the warning may not be the best solution :(. My guess for real-world problems is that Gurobi will almost never conclude "infeasible" or "unbounbed", but say "infeasible or unbounded".  The net result would be that warnings would be spit out just too often, since this would represent the "usual" code path for infeasible or unbounded models. Also, it would be good to have consistent behaviour across solvers; CPLEX has an equivalent stauts "CPX_STAT_INForUNBD".

Maybe we can split the handling of this status off this PR since the focus here is quite different?

rileyjmurray:
@rluce it's okay to split the status handling to another PR. However, if we go that route, my preference would be to leave the test as-is (i.e. still failing) in this PR.

rluce:
OK, so be it, I reverted the change to the test.

Issue #1011, opened 2020-04-27
MOSEK interface: inefficient encoding for SDPs


juesato:
Hi,
I'm unable to solve mid-size SDPs with CVXPY+Mosek (~4000x4000 SDP variable). I can solve the equivalent SDP with YALMIP+Mosek in Matlab. I'm wondering if there is a different way to encode the SDP that I'm missing, or some difference in how CVXPY and YALMIP model the problem?

On Cvxpy, I get a `mosek.Error: rescode.err_space(1051): Out of space.` error. In particular, for Cvxpy, I see 32020003 Constraints (from the Mosek verbose debug output), whereas in Matlab, there are 4001, so it seems the SDP constraint is handled very differently in the two.

Here is the script in Cvxpy:
```
import cvxpy as cp
import numpy as np

def cvx_toy():
  size = 4000
  P = cp.Variable(shape=(size + 1, size + 1))
  x = P[1:, 0]
  X = P[1:, 1:]
  constraints = [
      P == P.T,
      P[0][0] == 1.0,
      P >> 0,
  ]
  diag = cp.atoms.affine.diag.diag
  # b1, b2 = np.random.normal(size=size), np.random.normal(size=size)
  b1 = np.array(range(size)) + 1
  b2 = np.array(range(size)) + 3
  lb, ub = np.minimum(b1, b2), np.maximum(b1, b2)
  constraints += [diag(X) <= (cp.multiply(lb + ub, x) -
                              cp.multiply(lb, ub))]
  obj_cp = cp.sum(x)
  problem = cp.Problem(cp.Minimize(obj_cp), constraints)
  print('Start')
  # problem.solve(solver=cp.SCS)
  problem.solve(solver=cp.MOSEK, verbose=True)
  print(problem.status) 
  print('Opt:', sum(lb))
  print('SDP:', obj_cp.value)
cvx_toy()
```

And the equivalent script in YALMIP (Matlab):
```
function [] = cvxtoy()

addpath(genpath('/home/juesato/code/matlab_packages/yalmip/YALMIP-master'))
addpath(genpath('/home/juesato/install/mosek/9.2/toolbox/r2015a'))

sz = 4000
P = sdpvar(sz+1, sz+1)
x = P(2:sz+1, 1)
X = P(2:sz+1, 2:sz+1)
constraints = [P(1, 1) == 1, P >= 0]
b1 = (1:sz)'
b2 = b1 + 2
lb = min(b1, b2)
[...]

PR #1103, opened 2020-08-11
complete rewrite of MOSEK interface; introduce "Dualize" and "Slacks" reductions

rileyjmurray:
This PR began out of an effort to resolve inefficient compilation of SDPs into the standard-form required by MOSEK (#1011). For continuous problems (note, MOSEK only supports SDP when no integer constraints are present), the best way to do this is to dualize the standard form returned by a CVXPY ParamConeProg object.

Applying dualization actually makes the MOSEK interface far more efficient for all problem classes beyond LP's. Therefore rather than dualizing SDP's only, the dualization is applied to all continuous problems. For discrete problems, there is no option to dualize, so slack variables are required when dealing with mixed-integer problems featuring SOC or ExpCone constraints. This PR introduces ``cvxpy/reductions/cone2cone/affine2direct.py``, which contains two static-method classes ``Dualize`` and ``Slacks``. The MOSEK interface has been completely rewritten to use these reductions, in place of the old implementation.

I expect that once this PR is merged users should see substantial speedups for certain problem classes. I've personally seen order-of-magnitude speedups on large exponential cone programs using this type of transformation. I've also seen a couple problems where MOSEK succeeded on a dualized problem but failed when introducing slack variables, so there is a chance some users will see more accurate solutions following this PR.

This PR includes extensive documentation for the ``Dualize`` and ``Slacks`` reductions. However the only tests are currently confined to the MOSEK interface. I welcome suggestions for dedicated tests that do not rely on the MOSEK interface. 

There are some TODO's listed in the MOSEK interface about making SDP formulation more efficient. Right now symmetric matrix linear operators are defined using dense problem data, rather than sparse problem data. I'm inclined to leave those for others to fix.


rileyjmurray:
@aszekMosek I'd love it if you could review the new MOSEK interface. If you [click this link](https://github.com/cvxgrp/cvxpy/pull/1103/files) you should be able to start a code review.

angeris:
What's fascinating is that a similar problem and solution seems to be reported here: https://github.com/JuliaOpt/MosekTools.jl/issues/40 where dualizing the problem also yields an incredible performance speed up for Mosek.

A weird idea (that somewhat related to this PR but isn't directly so): if we have an SDP with chordal structure, we should be able to detect that and dualize the sparse version of the SDP and pass *this* dualized problem to the solvers. My guess is that this would be quite a bit faster than the current approach for a lot of SDPs (which are generally rather sparse). Otoh, we'd need some way of recovering the primal variables, which are not necessarily obvious.

cc: @SteveDiamond @akshayka

aszekMosek:
@angeris Mosek has always been shouting out loud that one should think well which of the primal/dual pair of problems to implement, esp. for sparse LMIs. https://docs.mosek.com/modeling-cookbook/duality.html#semidefinite-duality-and-lmis

@rileyjmurray Thanks a lot for this! I will definitely review.

angeris:
@aszekMosek Hmm, yeah, perhaps, but I guess my question is why is it not an automatic conversion? Certainly very few people who write `matrix_frac(X, P)` in cvxpy actually know anything about dualization or chordal sparsity. (Either way, we can discuss this outside of the PR.)

SteveDiamond:
This is fantastically documented! I can't comment much on the mosek interface, but the rest looks good. It would be really nice to have a non-mosek test though, that would run on Travis. This is a bit goofy, but you could construct a reduction chain ending in dualize, take the output, make a cvxpy problem, solve it, and run the chain backwards.

rileyjmurray:
@SteveDiamond @akshayka I've written a full set of tests. Assuming the CI tests pass, this should be ready to merge so long as you two and @aszekMosek approve.

rileyjmurray:
@aszekMosek let us know if you would prefer that we merge these changes before waiting for a review.

aszekMosek:
@rileyjmurray Apologies for the delay. I was in fact I was only able to look at it run some tests since yesterday. The mosek_coinf looks great to me, I have no comments there. Thanks! 

Just on a more general note, it works great for LMIs that are natural to dualize, but if the model is already natural to write in "primal" form then the conversion does not pay off, for example
```
P = cp.Variable((200,200), PSD=True)
problem = cp.Problem(cp.Minimize(0), [P[0][0]==3])
problem.solve(solver=cp.MOSEK, verbose=True, save_file="prob.ptf")
```
or #1011 . Of course in this situation the users are not any worse off in that the resulting MOSEK model has "the same order of inefficiency" as before, or as with the Slack() model. Until there is no SDP dualizer or more SDP presolve in MOSEK it is hard to find a perfect way out of this. 

Another remark is that suddenly dualizing all continuous problems can come as a surprise to some users. We don't have such a good coverage of the cvxpy usage to make statistical claims here.

rileyjmurray:
@aszekMosek thanks for the quick reply!

You're right that these changes have little benefit when a CVXPY user formulates their problem in what MOSEK calls the "primal" form. The issue is that CVXPY automatically compiles all cone programs into an affine conic form ``A @ x + b \in K``, where ``x`` is unconstrained. We won't be able to efficiently translate user-specified "primal formulations" to MOSEK's API until we change that base assumption. I can say that I'm interested in changing that base assumption, however, it's probably a ways off (maybe suitable for CVXPY 1.2). If we were to make that change then we would need to be far more intelligent about when to dualize versus when to use slacks.

Thanks for pointing out that the automatic dualization might surprise users. Often in PR's we make changes without updating some pre-release notes. @SteveDiamond how would you like me to log this change on release notes in the web documentation? I don't want to change the live website until a new version is released.

akshayka:
Awesome work @rileyjmurray! LGTM.

For the website, you could log the changes in a separate branch, to be merged into gh-pages when this change is released.

Do you anticipate these reduction being useful for other solvers in addition to MOSEK?

SteveDiamond:
Akshay's suggestion sounds good. Just follow the steps in the gh-pages part of ``Makefile``, but using a different branch.

aszekMosek:
There will be some models where Dualize() will lead to a less efficient Mosek model unless the user explicitly specifies ``mosek.iparam.intpnt_solve_form : mosek.solveform.dual`` in the ``mosek_params`` to internally dualize (thus undo the cvxpy  dualization). I saw this for example when I tested on the example from https://nbviewer.jupyter.org/github/MOSEK/Tutorials/blob/master/wasserstein/wasserstein-bary-reg.ipynb So that might be worth pointing out for the users to be aware of.

rileyjmurray:
@akshayka the reductions will be useful for any solver which requires a primal standard form ``A @ x == b, x in K``. I think this format is actually pretty standard for SDP solvers (e.g. SeDuMi and SDPNAL+).

@aszekMosek I'll be sure to emphasize the option of telling MOSEK to perform low-level dualization in the documentation. In follow-up PRs it might make sense to expose that as a simple keyword argument (like ``save_file``, or ``verbose``).

PR #1108, opened 2020-08-13
remove gurobi model update

sjiekak:
gurobi model update sends modification to the gurobi engine. 
When the engine is on a different machine, it causes additional traffic and longer optimization. 

I contacted gurobi support, and they said : 

> It is needed only in some specific cases such as, e.g., if you are working on a compute server or a cloud and want to retrieve specific information from different scenarios. For this you have to change the ScenarioNumber parameter and then call model.update() to make the server actually retrieve the scenario data. For "normal" MIP solving, you don't need to call model.update() when altering your mode

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
@@ -328,7 +328,6 @@ class GUROBI(SCS):
                     lb=-gp.GRB.INFINITY,
                     ub=gp.GRB.INFINITY)
             ]
-        
 
         new_lin_constrs = []
         for i, row in enumerate(rows):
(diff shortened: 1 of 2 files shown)


sjiekak:
@rluce @rileyjmurray

rluce:
> When the engine is on a different machine, it causes additional traffic and longer optimization.

Each call to `Model.update()` causes one network message to the cloud/remote server, so if you are solving very tiny models the overhead from these messages could become noticeable in the total time of communication and solving.  The optimization time itself is not affected by superfluous update calls.  (Probably needless to say, but these superfluous update calls are mostly irrelevant if you run cvxpy/Gurobi on you local machine.)

Beyond the example you gave above (scenario data management), such update calls are also needed to query attributes of *any* newly added entities (variables, constraints, etc.).  Looking at the code of the two interfaces in the current state, it looks like it is safe to remove the superfluous update calls.  The reason why I left them in the code when I touched it was maintainability:  Right now there is no intermediate querying of added entities, but maybe in the future someone unaware of this technical niche aspect needs to do that and will run into weird errors.

I don't have any strong feelings about the proposed PR; it's the call of the cvxpy maintainers.

SteveDiamond:
+1 in favor of maintainability. I'd rather keep the update calls and avoid bugs in the future. @sjiekak if it's a major issue for your use case feel free to fork cvxpy and change it there.

sjiekak:
Thanks for the feedback both of you.

PR #1121, opened 2020-09-01
Change cone2cone unittest to reflect ECOS_BB change

rileyjmurray:
The unit tests previously had ``@skipunless`` decorators, which checked if an MI-SOCP solver is installed. Now that ECOS_BB is back, that list had to be explicitly updated to exclude ECOS_BB from a test where it gave an incorrect (suboptimal) solution. 

Excluding ECOS_BB does not indicate the cone2cone reduction is broken, since the test passes when other MI-SOCP solvers are installed.

My plan is to merge this PR once CI tests pass.

### DIFF
--- a/cvxpy/tests/solver_test_helpers.py
+++ b/cvxpy/tests/solver_test_helpers.py
@@ -528,7 +528,6 @@ def mi_socp_1():
              norm(x,2) <= y[1]
              x[0] + x[1] + 3*x[2] >= 0.1
              y <= 5, y integer.
-    and solve with MOSEK.
     """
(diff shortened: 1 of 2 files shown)


rileyjmurray:
A final note: the failing test was ``test_mi_socp_2``. The failing case was when ``affine = ['s',0]``, which indicated that slack variables were used to turn linear inequalities into linear equations, however, second-order cone constraints were handled directly (without slack variables). I mention this because if someone gets the low-level problem data we pass to ECOS_BB in that case, then that problem data could be useful in an ECOS_BB bug report.

PR #1143, opened 2020-10-09
replace np.eye with sp.eye in qp2quadform

akshayka:
Replace a dense identity with a sparse one in `power_canon` for QPs.


SteveDiamond:
Looks good! Merging.

PR #1151, opened 2020-10-20
Lazily reduce problem data tensor

akshayka:
Reduce problem data tensor on first call to apply parameters.

An intermediate `ParamConeProg` is constructed when canonicalizing cone
programs; this `ParamConeProg` is never actually evaluated, so reducing
its tensor is unnecessary.

@sbarratt

### DIFF
--- a/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
+++ b/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
@@ -114,19 +114,8 @@ class ParamConeProg(ParamProb):
 
         # The variable
         self.x = x
-
-        # Form a reduced representation of A, for faster application of
-        # parameters.
-        if np.prod(A.shape) != 0:
-            reduced_A, indices, indptr, shape = (
-                canonInterface.reduce_problem_data_tensor(A, self.x.size)
-            )
-            self.reduced_A = reduced_A
-            self.problem_data_index = (indices, indptr, shape)
-        else:
-            self.reduced_A = A
-            self.problem_data_index = None
-
+        self.reduced_A = None
+        self.problem_data_tensor = None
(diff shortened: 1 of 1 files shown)


SteveDiamond (approved):
LGTM

akshayka:
@SteveDiamond @sbarratt The derivative tests are failing with


  File "C:\conda\envs\testenv\lib\site-packages\ecos.py", line 62, in solve
1898    return _ecos.csolve((m,n1,p), c, data, indices, colptr, h, dims, A.data, A.indices, A.indptr, b, **kwargs)
1899TypeError: 'eps' is an invalid keyword argument for this function


This appears to be related to the recent release of diffcp, which included ECOS. Could one of you look into it?

sbarratt:
Ah, whoops, looks like eps is no longer a valid argument if the solver is
"ECOS".

Shane

SteveDiamond:
Not sure what to do here. We have a naming clash with the ``solver`` keyword. It's used in cvxpy and in diffcp.

akshayka:
> Not sure what to do here. We have a naming clash with the `solver` keyword. It's used in cvxpy and in diffcp.

That's unfortunate.

I guess we can't really change CVXPY's interface. Can we change the keyword argument name in `diffcp` to `method`, instead of `solver`?

akshayka:
> > Not sure what to do here. We have a naming clash with the `solver` keyword. It's used in cvxpy and in diffcp.
> 
> That's unfortunate.
> 
> I guess we can't really change CVXPY's interface. Can we change the keyword argument name in `diffcp` to `method`, instead of `solver`?

er or I guess `solve_method`, since CVXPY also uses `method`

akshayka:
Rebased on top of b8a54937311063db42ac9a49e8ddce5520920b42 to retrigger the build ...

PR #1152, opened 2020-10-20
force SCS in diffcp

akshayka:
Forces `diffcp` to use `SCS`, since the `diffcp` conic interface already assumes this.


SteveDiamond (approved):
LGTM

akshayka:
The push/branch builds failed because they pulled `diffcp` 1.0.15, instead of the latest version. The PR builds succeeded (they got the latest version), so I'm merging this change.

PR #1162, opened 2020-10-28
Check shape before checking equality in osqp interface

akshayka:
numpy arrays need to have conforming shapes before checking for
elementwise equality


SteveDiamond (approved):
LGTM

PR #1163, opened 2020-10-29
error when parameter value is sparse

sbarratt:
Temporarily addresses https://github.com/cvxgrp/cvxpy/issues/1159


akshayka (approved):
LGTM, thanks!

PR #1165, opened 2020-10-29
Add sparse parameters to CVXPY

sbarratt:
Here's my first pass (it doesn't work.)

Here's the API I'm proposing:
```python
import cvxpy as cp
import numpy as np
import scipy.sparse

m, n = 50, 10

b = np.random.randn(m)
Avalue = scipy.sparse.random(m, n, density=.5, format='coo')

x = cp.Variable(n)
A = cp.Parameter((m, n), sparse=True, nonzero=Avalue.nonzero())
problem2 = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b)))
A.value = Avalue.data
problem2.solve()
```

### DIFF
--- a/cvxpy/expressions/constants/parameter.py
+++ b/cvxpy/expressions/constants/parameter.py
@@ -93,7 +93,7 @@ class Parameter(Leaf):
         """NumPy.ndarray or None: The numeric value of the parameter.
         """
         if self.sparse:
-            return self.Indexer @ self._parameter.value
+            return (self.Indexer @ self._parameter.value).reshape(self.shape, order='C')
         else:
             return self._value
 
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Attributes of variables and parameters should go in leaf.py. (Parameters and variables are supposed to stay aligned in their API.) Maybe instead of sparse=True, the input could be a binary sparse matrix with the sparsity pattern? And then ``param.value = `` sets the nonzeros?

sbarratt:
But the parameter is no longer an object right, it's actually an expression, with a parameter under the hood for the nonzero entries? How would I implement that cleanly?

SteveDiamond:
The parameter would need to be rewritten in ``cvx_attr2constr.py``. Right now only variables get rewritten there, but parameters could be included.

PR #1166, opened 2020-10-31
Replace nose with pytest

### LINKED ISSUE #1157: Use pytest instead of nose?
I see that `nose` is used to run the tests, but [the docs state](https://nose.readthedocs.io/en/latest/) "Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership." It has not been updated for several years on [github.com/nose-devs/nose](https://github.com/nose-devs/nose).

Want me to create a PR to swap out nose as the test runner in favor of [pytest](https://github.com/pytest-dev/pytest)?

tommyod:
This PR closes #1157 . It replaces nose with pytest, more specifically:

- CI/CD scripts and documentation have been changed.
- I changed the order of `flake8` and the tests. The rationale is that flake8 should go first since it's faster, causing the build to fail fast if there is an error with syntax.
- There are some differences in `cvxpy/tests/test_constant_atoms.py`, but the git diff is mostly syntax. The only real change is that I use the `pytest.mark.parametrize` decorator on the `test_constant_atoms` function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then CXX=g++; fi
 install: source continuous_integration/Travis/install_dependencies.sh
 script:
-  - bash continuous_integration/Travis/test_script.sh
   - flake8 .
+  - bash continuous_integration/Travis/test_script.sh
 after_success:
   - if [[ "$COVERAGE" == "true" ]]; then coveralls; fi
 deploy:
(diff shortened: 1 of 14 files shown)


rileyjmurray:
This is great work! You are very right that we should have been running flake8 before the full suite of unit tests. I will take a detailed look at these changes soon.

rileyjmurray (commented):
@akshayka @SteveDiamond do you want ``cvxpy/performance_tests`` to run on CI?

rileyjmurray:
It looks like ``test_benchmarks`` doesn't take much time to run, so I suppose we might as well add that to CI builds. I've resolved a merge conflict. Once the CI builds pass following that change, I'll merge this in.

PR #1168, opened 2020-11-01
Cplex conflict refiner

tomas789:
TL;DR: This is a very basic implementation but already very helpful. But I guess with conflict refiner it is the same as with sex. Better bad than none.

I'm using similar functionality already and I think others might find it useful. It is certainly not perfect. Read the comments for details.

I exported that directly from the module just because it was easy. I'm more than happy to move it to better place if there is one.

### DIFF
--- a/cvxpy/__init__.py
+++ b/cvxpy/__init__.py
@@ -31,3 +31,4 @@ from cvxpy.settings import (CVXOPT, GLPK, GLPK_MI, CBC, CPLEX, OSQP, NAG,
 from cvxpy.transforms import linearize, partial_optimize, suppfunc
 from cvxpy.reductions import *
 from cvxpy.reductions.solvers.defines import installed_solvers
+from cvxpy.utilities import cplex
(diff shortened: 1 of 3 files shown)


SteveDiamond:
@tomas789 this is interesting, I wasn't aware of conflict refiners. I'd like to see this developed a bit more before it goes in cvxpy. Why don't you add it to https://github.com/cvxgrp/cvxpyanalyzer? This is a repo for utilities to run on top of cvxpy.

tomas789:
That has a very simple reason. I did not know about that ;) If you think that is a better place for this, just close this PR and I'll move it over there.

Is there a link from documentation to the other repo?

SteveDiamond:
Give me some time to develop the docs, then we can move this over. I haven't advertised the cvxpy analyzer yet, but I'll add something to the cvxpy docs.

SteveDiamond:
Copied over to cvxpyanalyzer.

PR #1169, opened 2020-11-01
Cplex variable names

tomas789:
It exports Variable's `.name` property as variable name to Cplex. It is useful when debugging an infeasibility as Cplex will actually print the variable name. It gets even more useful when combined with the conflict refiner (PR #1168). 

This has two main limitations at the moment:

1. Cvxpy actually creates lots of variables under the hood, sometimes just copying ones created by the user. This often does not preserve the user-specified names. I'll try to address parts of this issue with future PRs.
2. It only works for variables and not for constraints. They are more complex to tackle. That is a story for another time I guess.


SteveDiamond:
LGTM.

PR #1170, opened 2020-11-01
Cplex variable names: Bugfix

tomas789:
Relates to my previous PR #1169. There was a problem with multidimensional variables. The `param_prob.id_to_var` names corresponded to number of variables but the `q` variable is if course for each item in that vector/matrix/tensor. Test suite did not catch it as it likely does not have Cplex test cases.

### DIFF
--- a/cvxpy/reductions/inverse_data.py
+++ b/cvxpy/reductions/inverse_data.py
@@ -22,7 +22,7 @@ class InverseData(object):
 
     def __init__(self, problem):
         varis = problem.variables()
-        self.id_map, self.var_offsets, self.x_length, self.var_shapes = \
+        self.id_map, self.var_offsets, self.var_names, self.x_length, self.var_shapes = \
             InverseData.get_var_offsets(varis)
 
         self.param_shapes = {}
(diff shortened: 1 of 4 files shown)


tomas789:
I've just noticed that on my machine one test is failing but it the CI all looks good. `cvxpy.tests.test_quad_form.TestNonOptimal.test_singular_quad_form`

It just says
```
tomaskrejci@Tomass-MBP cvxpy % python -m nose2 cvxpy.tests.test_quad_form.TestNonOptimal.test_singular_quad_form
.E
======================================================================
ERROR: cvxpy.tests.test_quad_form.TestNonOptimal.test_singular_quad_form (nose2.loader.LoadTestsFailure)
----------------------------------------------------------------------
  File "cvxpy/cvxpy/tests/test_quad_form.py", line 70, in test_singular_quad_form
    xopt = x.value.A.flatten()
AttributeError: 'numpy.ndarray' object has no attribute 'A'
```

The error message makes sense to me and the test seems broken. Question is why it does not show up in the CI.

SteveDiamond:
This PR is perhaps too large for the issue it's addressing. Is it really necessary to pass the name info down like this?

SteveDiamond:
Thanks for pointing out the failing test. It wasn't being run by nosetests for some reason. I've fixed it on master.

SteveDiamond:
Merged into ``qp_var_names`` branch.

PR #1172, opened 2020-11-07
Allow for custom solvers

nrontsis:
## Purpose
Allows the use of custom solvers in `cvxpy`. This can be helpful when prototyping new solvers, writing official solvers for `cvxpy`, or developing custom solvers tailored to a specific application. 

To use a custom solver, one has to implement a solver class that is a child of `QpSolver` or `ConicSolver`. Then an instance of this new solver class is passed to `problem.solve(solver=.)`. Note that conflicts with existing solvers are avoided because custom solvers are specified by passing objects that are instances of the `Solver` class, while standard solvers are specified via strings (e.g. "OSQP"). Furthermore, when a user specifies a custom solver, then no standard solvers are suggested as "candidates" via `cvxpy`. That was done on purpose for simplicity.

## Minimal usage example
```python
import cvxpy as cp
from cvxpy.reductions.solvers.qp_solvers.osqp_qpif import OSQP


class CUSTOM_OSQP(OSQP):
    MIP_CAPABLE = False

    def name(self):
        return "CUSTOM_OSQP"

    def solve_via_data(self, *args, **kwargs):
        print("Solving with a custom QP solver!")
        super().solve_via_data(*args, **kwargs)


x = cp.Variable()
quadratic = cp.square(x)
problem = cp.Problem(cp.Minimize(quadratic))
problem.solve(solver=CUSTOM_OSQP())
```

See also the documentation and the tests added for more examples/details.

## ToDo List:
- [x] Tests
- [x] Documentation

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -545,17 +545,18 @@ class Problem(u.Canonical):
                                 solver=None,
                                 gp=False):
         """
-        Find candiate solvers for the current problem. If solver
+        Find candidate solvers for the current problem. If solver
         is not None, it checks if the specified solver is compatible
         with the problem passed.
 
         Arguments
         ---------
-        solver : string
-            The name of the solver with which to solve the problem. If no
-            solver is supplied (i.e., if solver is None), then the targeted
-            solver may be any of those that are installed. If the problem
-            is variable-free, then this parameter is ignored.
+        solver : Union[string, Solver, None]
+            The name of the solver with which to solve the problem or an
+            instance of a custom solver. If no solver is supplied
+            (i.e., if solver is None), then the targeted solver may be any
+            of those that are installed. If the problem is variable-free,
+            then this parameter is ignored.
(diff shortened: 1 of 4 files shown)


rileyjmurray:
This is extremely cool! I'll add Steven as a reviewer.

SteveDiamond:
This is interesting! The custom solver would in practice probably need to define the ``invert`` method as well, right?

nrontsis:
Right, I have updated the docs to reflect that.

SteveDiamond (approved):
It all looks good to me, in particular how small the changes are to the canonicalization procedure.

nrontsis:
Hey, are there any other changes that you might want to see for this to get merged? I would happy to work on them :rocket:

SteveDiamond:
@akshayka what do you think?

akshayka (approved):

akshayka:
Looks good to me. Great work on the elegant implementation!

PR #1174, opened 2020-11-12
Allow zero length variables

nrontsis:
Allows zero-length `cvxpy.Variable`s.

This is to increase consistency with other standard libraries that deal with matrix concepts (e.g. `numpy`, `scipy` and `pandas`).

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -94,7 +94,7 @@ class Leaf(expression.Expression):
             raise ValueError("Expressions of dimension greater than 2 "
                              "are not supported.")
         for d in shape:
-            if not isinstance(d, numbers.Integral) or d <= 0:
+            if not isinstance(d, numbers.Integral) or d < 0:
                 raise ValueError("Invalid dimensions %s." % (shape,))
         self._shape = tuple(np.int32(d) for d in shape)
 
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -88,8 +88,8 @@ class TestExpressions(BaseTest):
                          "Cannot set more than one special attribute in Variable.")
 
         with self.assertRaises(Exception) as cm:
-            Variable((2, 0))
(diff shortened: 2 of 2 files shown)


akshayka:
Does this have any practical use, besides consistency?

nrontsis:
My use case is to increase generality. Sometimes I have optimisation problems that I solve with a function:
```python
def solve_with_cvxpy(number_of_nodes):
   ...
```
that when e.g. `number_of_nodes=1` a zero-length variable is created somewhere in the process. Currently, I resort to writing two different methods, i.e.
```python
def solve_with_cvxpy_for_one_node():
   ...

def solve_with_cvxpy_for_many_nodes(number_of_nodes):
   ...
```
and using if statements to switch between the two.

See also https://github.com/pytorch/pytorch/issues/5014 for a related discussion.

akshayka:
Interesting, thanks for sharing context.

@SteveDiamond , what do you think? Would zero-length variables break any assumptions in, e.g., cvxcore, or matrix stuffing?

rileyjmurray:
Chiming in. I don't think this should be merged in without some tests showing how zero-dimensional variables behave "correctly" in a given model. Also, we need to keep in mind that allowing Variables of dimension zero means we're allowing Expressions of dimension zero, so it's not enough to only test against the Variable class.

akshayka:
Yes, I wouldn't be surprised if our codebase implicitly assumes nonzero
size in many places.

nrontsis:
@rileyjmurray I think that zero length expressions are already supported:
```python
>>> cvxpy.Variable((2, 1))[:, []]
Expression(CONSTANT, UNKNOWN, (2, 0))
```

SteveDiamond:
Zero length expressions are not officially supported. Getting support for them would be quite nice actually, and increase consistency with numpy, but I suspect we would need changes to cvxcore. @nrontsis can you try solving problems involving zero length variables and expressions? This is a change that will require a *lot* of testing and investigation.

PR #1180, opened 2020-11-19
Add hint to docs about changes to ECOS_BB

lumbric:
This might also help somebody who is using 1.1.0 <= CVXPY < 1.1.6 and is
wondering why ECOS_BB is not installed.


SteveDiamond:
LGTM

PR #1181, opened 2020-11-22
Return correct domain

RSchwan:
I was trying to get the domain of an expression and was surprised that it always returned an empty array. After some investigation I found out that the domain property function in `leaf.py` always returns an empty array instead of the calculated domain (see changed line). To me this looks like a clear mistake/bug. I'm just a bit worried that no one noticed this since the part of the codebase is already over 3 years old.

Also, it seems like this isn't the only mistake in regard to returning the correct domain. For example, this is now correct after the change:
``` python
import cvxpy as cp

X = cp.Variable((3, 3), PSD=True)

constraints = []

objective = cp.Minimize(cp.norm(X, 2))

all_domains = objective.args[0].domain  # domain of the objective function
for arg in constraints:
  for l in range(len(arg.args)):
    for dom in arg.args[l].domain:
      all_domains.append(dom)  # domain on each side of constraints
print(all_domains) # returns '[PSD(Expression(AFFINE, UNKNOWN, (3, 3)))]'
```
but the slightly modified but equivalent problem has still an empty domain:
``` python
import cvxpy as cp

X = cp.Variable((3, 3))

constraints = [X >> 0]

objective = cp.Minimize(cp.norm(X, 2))

all_domains = objective.args[0].domain  # domain of the objective function
for arg in constraints:
  for l in range(len(arg.args)):
    for dom in arg.args[l].domain:
      all_domains.append(dom)  # domain on each side of constraints
print(all_domains) # returns '[]'
```

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -269,7 +269,7 @@ class Leaf(expression.Expression):
             domain.append(self >> 0)
         elif self.attributes['NSD']:
             domain.append(self << 0)
-        return []
+        return domain
 
     def project(self, val):
         """Project value onto the attribute set of the leaf.


rileyjmurray:
Hi @RSchwan. I think there may be a misunderstanding about the meaning of the word "domain" in this context. From our doc-strings: https://www.cvxpy.org/api_reference/cvxpy.atoms.html?highlight=domain#cvxpy.atoms.atom.Atom.domain and https://www.cvxpy.org/api_reference/cvxpy.expressions.html?highlight=domain#cvxpy.expressions.expression.Expression.domain, the domain is intended to impose constraints on the region where the expression is finite. In your case, the 2-norm of a matrix is finite everywhere, so there is no restriction on the domain. (@SteveDiamond pleast correct me if I'm wrong on that).

rileyjmurray:
@RSchwan I've taken another look, and I can see that the code obviously contains some kind of mistake. Either the current implementation populates a list that it never should have used, or the current implementation erroneously returns an empty list. I think we're in the first of these two situations, but I need @SteveDiamond and @akshayka to confirm.

SteveDiamond:
There is a mistake in the code and the PR is correct. I guess people don't use domains very much, so this never got noticed.

rileyjmurray:
Gotcha. @RSchwan regarding
```
import cvxpy as cp

X = cp.Variable((3, 3))
constraints = [X >> 0]
objective = cp.Minimize(cp.norm(X, 2))
```
The way CVXPY is designed it won't be able to infer properties of an Expression based on other constraints. So if you have
```
x = cp.Variable()
f = cp.Minimize(x)
cons = [x >= 1]
prob = cp.Problem(f, cons)
```
The objective ``f`` won't know that ``x`` is positive. Extracting that kind of information requires parsing the entire model for symbolic properties. For example, if we had the infeasible constraints ``cons = [x >= 1, cp.norm(x + 1) <= 0.5]`` then realistically we can't infer the domain of ``f`` until we've certified the problem as infeasible.

RSchwan:
@rileyjmurray regarding your last comment. Sure, the objective doesn't know about the constraint `x >= 1` but in my opinion the constant `x >= 1` itself would infer the domain `x >= 1`. The same with simple constraints like `X >> 0`. The constraint is so simple (in a sense the same as `X = cp.Variable((3, 3), PSD=True)`) that we could infer the domain constraint `X >> 0`. Hence, the constraint `cp.norm(x + 1) <= 0.5` should also infer the domain constraint `cp.norm(x + 1) <= 0.5`.

To clarify my point. If I have the following optimization problem (with changes of pull request)
```python
import cvxpy as cp

x = cp.Variable(nonneg=True)

constraints = []

objective = cp.Minimize(cp.sqrt(x))

all_domains = objective.args[0].domain  # domain of the objective function
for arg in constraints:
  for l in range(len(arg.args)):
    for dom in arg.args[l].domain:
      all_domains.append(dom)  # domain on each side of constraints
print(all_domains) # returns '[Inequality(Constant(CONSTANT, ZERO, ())), Inequality(Constant(CONSTANT, ZERO, ()))]'
```
we see that the domain are two constraints. One is the constraint `x >= 0` from the definition of `x`, and one is `x + 1 >= 0` which is the domain of `cp.sqrt()` (see [power.py#L364](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/atoms/elementwise/power.py#L364) where the constraint comes from). Hence, there a simple cases there the domain constraints can be determined beforehand. In the case of atoms (`cp.sqrt()` in our example) the domain is implemented. So wouldn't it make sense that we have a similar behavior for simple constraints like `x >= 0` or `X >> 0`, or does this only make sense for atoms?

I'm bringing all these questions up because I'm currently trying to understand the implementation of the [DCCP](https://github.com/cvxgrp/dccp) package. In the [paper](https://stanford.edu/~boyd/papers/pdf/dccp.pdf) Section 3 the algorithm has to be initialized in the interior of the domain (see [here](https://github.com/cvxgrp/dccp/blob/master/dccp/problem.py#L127) for the implementation). Hence, I am just a bit worried that cvxpy doesn't calculate the domain correctly. This is also the origin of this pull request.

SteveDiamond:
In DCCP, constraints are converted into soft-constraints (via the slack variables), so the domain in question is the intersection of the domains of all of the functions present. The constraints themselves do not restrict the domain.

RSchwan:
@SteveDiamond thanks for the clarification. I guess this makes sense then. But I guess my pull request is still valid, since it's a clear bug in the code.

SteveDiamond:
For sure, the PR is good. Merging now.

PR #1199, opened 2020-12-21
1191 fix cvxopt status

### LINKED ISSUE #1191: Missing implementation of some status return values of cvxopt.glpk.ilp()
**Describe the bug**

The function `cvxopt.glpk.ilp()` may return following values as `status`: `'optimal'`, `'feasible'`, `'undefined'`, `'invalid formulation'`, `'infeasible problem'`, `'LP relaxation is primal infeasible'`, `'LP relaxation is dual infeasible'`, `'unknown'`. AFAICS the following values are not handled correctly by CVXPY and cause a KeyError:

* `'invalid formulation'`
* `'infeasible problem'`

Help text of `ilp()` mentions possible return values::

```
>>> from cvxopt.glpk import ilp
>>> help(ilp)
ilp(...)
    Solves a mixed integer linear program using GLPK.
    
    (status, x) = ilp(c, G, h, A, b, I, B)
...
    status       if status is 'optimal', 'feasible', or 'undefined',
                 a value of x is returned and the status string 
                 gives the status of x.  Other possible values of              status are:  'invalid formulation', 
                 'infeasible problem', 'LP relaxation is primal 
                 infeasible', 'LP relaxation is dual infeasible', 
                 'unknown'.
    
    x            a (sub-)optimal solution if status is 'optimal', 
                 'feasible', or 'undefined'.  None otherwise

```

These are the values which can are handled atm AFAICS:

https://github.com/cvxgrp/cvxpy/blob/b5b0f8b15b1f10106301d2b921ccfc6cbbbaf044/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py#L53-L63

There seem to be some unnecessary values too, they might be used by other subclasses. I haven't checked other versions of CVXOPT, this might have changed over time of course.

**To Reproduce**
I have a huge a example which can reproduce the problem. Unfortunately I haven't figured out yet how to reproduce it with a minimal problem. I'll add it to the ticket as soon as I get this done.

If one simply uses a problem with contradicting constraints, cvxopt will return `'LP relaxation is primal infeasible'`. I'm not sure what `'infeasible problem'` means exactly and how to construct one.

**Expected behavior**

The problem should be actually infeasible, so I expected `cp.Problem.solve()` to return without raising a `KeyError` exception and with the status field set to `infeasible` or something similar.

**Output**

That's what PyCharm says when I run my unit tests on my project:

```
...
/opt/miniconda3/envs/myenv/lib/python3.7/site-packages/cvxpy/problems/problem.py:395: in solve
/opt/miniconda3/envs/myenv/lib/python3.7/site-packages/cvxpy/problems/problem.py:746: in _solve
    self, data, warm_start, verbose, kwargs)
/opt/miniconda3/envs/myenv/lib/python3.7/site-packages/cvxpy/reductions/solvers/solving_chain.py:326: in solve_via_data
    solver_opts, problem._solver_cache)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cvxpy.reductions.solvers.conic_solvers.glpk_mi_conif.GLPK_MI object at 0x7fc0ec038510>
data = {'A': None, 'G': <16x12 sparse matrix, tc='d', nnz=44>, 'b': None, 'bool_vars_idx': [0, 4, 1, 5, 2, 6, ...], ...}
warm_start = True, verbose = False, solver_opts = {}, solver_cache = {}

    def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=None):
        import cvxopt
        import cvxopt.solvers
        # Save original cvxopt solver options.
        old_options = cvxopt.glpk.options.copy()
        # Silence cvxopt if verbose is False.
        if verbose:
            cvxopt.glpk.options["msg_lev"] = "GLP_MSG_ON"
        else:
            cvxopt.glpk.options["msg_lev"] = "GLP_MSG_OFF"
    
        # Apply any user-specific options.
        # Rename max_iters to maxiters.
        if "max_iters" in solver_opts:
            solver_opts["maxiters"] = solver_opts["max_iters"]
        for key, value in solver_opts.items():
            cvxopt.glpk.options[key] = value
    
        try:
            results_tup = cvxopt.glpk.ilp(data[s.C],
                                          data[s.G],
                                          data[s.H],
                                          data[s.A],
                                          data[s.B],
                                          set(int(i) for i in data[s.INT_IDX]),
                                          set(int(i) for i in data[s.BOOL_IDX]))
            results_dict = {}
            results_dict["status"] = results_tup[0]
            results_dict["x"] = results_tup[1]
        # Catch exceptions in CVXOPT and convert them to solver errors.
        #except ValueError as e:
        except RuntimeError as e:
            results_dict = {"status": "unknown"}
    
        # Restore original cvxopt solver options.
        self._restore_solver_options(old_options)
    
        # Convert CVXOPT results to solution format.
        solution = {}
>       status = self.STATUS_MAP[results_dict['status']]
E       KeyError: 'infeasible problem'

/opt/miniconda3/envs/myenv/lib/python3.7/site-packages/cvxpy/reductions/solvers/conic_solvers/glpk_mi_conif.py:90: KeyError
```

**Version**
 - OS: Ubuntu
 - CVXPY Version: 

``` >>> cp.__version__                                                                                                                  
'1.1.3'
>>> cvxopt.__version__                                                                                                              
'1.2.5'
```

**Additional context**

This seems to be very similar to  #812, but since I use CVXPY 1.1.3 the fix from #910 should be included already. So I guess this is still open.

b1a0:
This fixes #1191, which occurs because there is no status code 'infeasible' in CVXOPT/GLPK.  
The test case added is an MIP that is infeasible but integer relaxable and triggers the status 'infeasible problem' in GLPK. 

I particularly want to stress, that I did only add this test for solver='GLPK_MI', since I don't have other solvers installed and the test might be an edge case where some solvers are failing. Still, it would be interesting to see what other solvers have to say about this test, which is interesting in itself.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py
@@ -52,7 +52,7 @@ class CVXOPT(ECOS):
     # Map of CVXOPT status to CVXPY status.
     STATUS_MAP = {"optimal": s.OPTIMAL,
                   "feasible": s.OPTIMAL_INACCURATE,
-                  "infeasible": s.INFEASIBLE,
+                  "infeasible problem": s.INFEASIBLE,
                   "primal infeasible": s.INFEASIBLE,
                   "LP relaxation is primal infeasible": s.INFEASIBLE,
                   "LP relaxation is dual infeasible": s.UNBOUNDED,
--- a/cvxpy/tests/solver_test_helpers.py
+++ b/cvxpy/tests/solver_test_helpers.py
@@ -516,8 +516,24 @@ def mi_lp_2():
     return sth
 
 
-# TODO: Add a test-case for an infeasible integer-linear program
-#  (infeasible only due integrality constraints).
+def mi_lp_3():
+    # infeasible (but relaxable) test case
+    x = cp.Variable(4, boolean=True)
+    from cvxpy.expressions.constants import Constant
+    objective = cp.Maximize(Constant(1))
+    constraints = [x[0] + x[1] + x[2] + x[3] <= 2,
+                   x[0] + x[1] + x[2] + x[3] >= 2,
+                   x[0] + x[1] <= 1,
+                   x[0] + x[2] <= 1,
+                   x[0] + x[3] <= 1,
+                   x[2] + x[3] <= 1,
+                   x[1] + x[3] <= 1,
+                   x[1] + x[2] <= 1]
+    obj_pair = (objective, -np.inf)
+    con_pairs = [(c, None) for c in constraints]
+    var_pairs = [(x, None)]
+    sth = SolverTestHelper(obj_pair, var_pairs, con_pairs)
+    return sth
 
 
 def mi_socp_1():
@@ -635,6 +651,13 @@ class StandardTestLPs(object):
     @staticmethod
     def test_mi_lp_2(solver, places=4, **kwargs):
         sth = mi_lp_2()
+        sth.solve(solver, **kwargs)
+        sth.verify_objective(places)
+        sth.verify_primal_values(places)
+
+    @staticmethod
+    def test_mi_lp_3(solver, places=4, **kwargs):
+        sth = mi_lp_3()
         sth.solve(solver, **kwargs)
         sth.verify_objective(places)
         sth.verify_primal_values(places)
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -420,6 +420,9 @@ class TestMosek(unittest.TestCase):
     def test_mosek_mi_lp_2(self):
         StandardTestLPs.test_mi_lp_2(solver='MOSEK')
 
+    def test_mosek_mi_lp_3(self):
+        StandardTestLPs.test_mi_lp_3(solver='MOSEK')
+
     def test_mosek_mi_socp_1(self):
         StandardTestSOCPs.test_mi_socp_1(solver='MOSEK')
 
@@ -615,6 +618,9 @@ class TestCBC(BaseTest):
     def test_cbc_mi_lp_2(self):
         StandardTestLPs.test_mi_lp_2(solver='CBC')
 
+    def test_cbc_mi_lp_3(self):
+        StandardTestLPs.test_mi_lp_3(solver='CBC')
+
 
 @unittest.skipUnless('GLPK' in INSTALLED_SOLVERS, 'GLPK is not installed.')
 class TestGLPK(unittest.TestCase):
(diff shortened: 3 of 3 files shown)


SteveDiamond (approved):
Looks good to me!

SteveDiamond:
Thanks for fixing this!

b1a0:
**TL;DR:** 
As opposed to the claim in the commit message, I mistakenly added the test to the solver SCIP in addition to correctly testing GLPK_MI. The easy fix might be to just remove this addition from the commit, but while considering this I realized that this example might run into further problems in the future when people try to add this to test other solvers.

**Explanation:** 
I forgot to remove the test case from the SCIP solver in my commit. This is untested on my system (tests were most likely skipped due to no installation of SCIP). I want to make sure that other people who have SCIP installed do not run into a failed test because SCIP can't solve this problem. Does the CI have an installed SCIP solver and was this tested?

For example trying to solve this problem with ECOS_BB results in 

- `prob.status` 'infeasible_inaccurate', 
- `prob.value` returns no value to compare with the objective `-inf` and 
-  a UserWarning, which would make this test fail if it was added to TestECOS_BB: `UserWarning: Solution may be inaccurate. Try another solver, adjusting the solver settings, or solve with verbose=True for more information.`

Is the objective value `-inf` for infeasible problems solver dependent? Would it be better to check infeasibility differently?

For reproducibility, the minimal standalone example:

```
import numpy as np
import cvxpy as cp
from cvxpy.expressions.constants import Constant

solver = 'ECOS_BB'

x = cp.Variable(4, boolean=True)
objective = cp.Maximize(Constant(1))
constraints = [x[0] + x[1] + x[2] + x[3] <= 2,
               x[0] + x[1] + x[2] + x[3] >= 2,
               x[0] + x[1] <= 1,
               x[0] + x[2] <= 1,
               x[0] + x[3] <= 1,
               x[2] + x[3] <= 1,
               x[1] + x[3] <= 1,
               x[1] + x[2] <= 1]
obj_pair = (objective, -np.inf)
con_pairs = [(c, None) for c in constraints]
var_pairs = [(x, None)]

prob = cp.Problem(obj_pair[0], [c for c, _ in con_pairs])
prob.solve(solver=solver, verbose=True)
```

SteveDiamond:
The objective value should always be ``-inf`` for infeasible problems.

SteveDiamond:
There was a bug where "infeasible inaccurate" and "unbounded inaccurate" were not returning the correct ``problem.value``. I fixed it on master.

b1a0:
Thank you for the clarification, I added the test to all solvers.

SteveDiamond:
Ok, thanks!

PR #1202, opened 2020-12-30
Continuous integration improvements

rileyjmurray:
This PR makes the following general changes
 * Remove attempts at deploying conda packages to cvxgrp (since @h-vetinari has set up a good conda-forge build process).
 * Change target python versions from 3.5 -- 3.7 to 3.6 -- 3.9.
 * Remove a file ``appveyor_install.ps1`` that was accidentally re-created when 1.1 was merged onto master.

Notable changes specific to TravisCI (linux and macOS builds):
 * NumPy and SciPy versions are now chosen based on python version. I left comments in  ``/continuous_integration/Travis/install_dependencies.sh`` explaining which NumPy/SciPy versions seem to be required for a given version of python,
 * Only build against python 3.7, 3.8, 3.9. So, only AppVeyor builds test against python 3.6.
 * Consolidated code in a bash script we used to set up the test environment.

I didn't make direct changes to PyPI deployments, but we should keep an eye out when we bump cvxpy's version next time to make sure PyPI deployments do go as expected.


continuous_integration/AppVeyor/configure_conda_and_install_cvxpy.ps1
```diff
@@ -3,8 +3,7 @@ conda config --add channels oxfordcontrol
 conda create -n testenv --yes python=$env:PYTHON_VERSION mkl=2018.0.3 pip pytest numpy scipy
 conda activate testenv
 "python=$env:PYTHON_VERSION" | Out-File C:\conda\envs\testenv\conda-meta\pinned -encoding ascii
-conda install --yes lapack ecos multiprocess
-conda install -c conda-forge --yes scs
+conda install --yes lapack ecos scs
```
SteveDiamond, line 6: the SCS in main conda works now?

SteveDiamond (approved):
It looks good to me! Thanks for figuring out the right numpy and scipy versions.

continuous_integration/AppVeyor/configure_conda_and_install_cvxpy.ps1
rileyjmurray: This line change didn't actually affect where we get SCS. The script had declared conda-forge as a default channel at line 1, so the old version was redundant in specifying conda-forge at line 7.

PR #1210, opened 2021-01-14
Xpress speed improvements

merraksh:
The main change is improved performance for problems with many equality AND many inequality constraints. Previously Xpress added both classes of constraints as one, causing performance loss due to a Python loop. Now it only uses API functions. Other additions: statistics on CPU time and objective value; an rST entry.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
@@ -91,6 +91,7 @@ class XPRESS(QpSolver):
 
         return Solution(status, opt_val, primal_vars, dual_vars, attr)
 
+
     def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=None):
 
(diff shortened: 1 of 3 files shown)


cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
```diff
@@ -221,7 +210,7 @@ def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=No
         # Set options if compatible with Xpress problem control names
 
         self.prob_.setControl({i: solver_opts[i] for i in solver_opts
-                               if hasattr(xp.controls.__dict__, i)})
```
SteveDiamond, line 224: Are you sure we don't need the ``hasattr``? Why was it there before?

SteveDiamond (approved):
It all looks good to me!

SteveDiamond:
Thanks for the PR!

merraksh:
Thank you!

PR #1211, opened 2021-01-15
Power cones

rileyjmurray:
Well @SteveDiamond and @akshayka, I've got a big one for ya!

## What are power cones?

The 3D power cone with scale parameter ``\alpha`` in ``(0, 1)`` is
```
{ (x,y,z) : x^{\alpha}  y^{1-\alpha} >= |z|, x >= 0 , y >= 0 }.
```
This cone is supported by SCS and MOSEK (V9 and higher). This cone can be generalized to have a "long left-hand side", which gives us the ``n``D power cone
```
{ (w,z) \in \R^{n-1} \times \R:  \prod_{i=1}^{n-1} w_i^{\alpha_i}  >= |z|, w >= 0 }.
```
For the ``n``D power cone, we need ``\alpha`` is a componentwise positive ``n-1``-vector that sums to one.

Although the ND power cone is more sophisticated than the 3D power cone, any ``n``D power cone can be expressed with ``n-2`` 3D power cones and ``n-2`` auxiliary variables.  The 3D power cone with ``\alpha = 1/2`` is the same as the [rotated quadratic cone](https://docs.mosek.com/modeling-cookbook/cqo.html#rotated-quadratic-cones) (or, from CVXPY's standpoint, the epigraph of ``quad_over_lin``).

Power cones are exponential-cone representable. It's easy to adapt Proposition 8.1 from the appendix of https://arxiv.org/abs/2006.06811 to obtain a formal proof of this fact. Despite the fact that power cones are exponential-cone representable, power cones generally exhibit better numerics. For example, it's possible to find analytic expressions for central rays of power cones, while this is not the case for the exponential cone. There's also the fact that the reformulation from the standard 3D power cone to exponential cones requires *two* 3D exponential cones. So, it's obviously good to avoid that.

## What are power cones good for?

Several CVXPY atoms are currently being approximated by second-order cones, when they could be represented exactly by power cones. These atoms include ``geo_mean``, ``pnorm``, and ``power``, which have downstream uses in ``harmonic_mean``, ``inv_prod``, and ``mixed_norm``. The MOSEK Modeling Cookbook has [a section on power cones](https://docs.mosek.com/modeling-cookbook/powo.html).

My interest in power cones comes from their usefulness in convex relaxations to nonconvex signomial and polynomial optimization problems. Specifically, I'm going to have some undergrads explore the idea mentioned in the penultimate paragraph of https://arxiv.org/abs/2006.06811.

## How does this PR incorporate power cone support?

I created two constraint classes within ``cvxpy/constraints/power.py``: ``PowCone3D`` and ``PowConeND``. The ``PowCone3D`` class affects the cvxpy codebase in a way that's very similar to ``ExpCone``. For example, I modified the following code to account for ``PowCone3D``:
 * the ``ConeDims`` class for conic solvers,
 * the ``ConeMatrixStuffing`` and ``ConicSolver`` classes,
 * the solver interfaces for SCS and MOSEK,
 * the ``Dualize`` reduction (NOTE: *before merging this PR, I need to update the Slacks reduction*),
 * ``solver_test_helpers.py``, and
 * ``test_conic_solvers.py``.

The ``PowConeND`` class is unusual. It's the only cone we define that isn't acceptable by any of our current solvers. In order to make this work, I created a new reduction within ``cvxpy/reductions/cone2cone``: the ``Exotic2Common`` reduction. For now, this reduction only handles transforming a ``PowConeND`` constraint into an equivalent (vectorized) ``PowCone3D`` constraint. This reduction is inserted in the ``construct_solving_chain`` function of ``cvxpy/reductions/solvers/solving_chain.py``. I also streamlined some existing code in that function so we only loop over constraints once (rather than 5 times). The tests for ``PowConeND`` can be found in ``test_cone2cone.py``.

## What's missing in this PR?

Both ``PowCone3D`` and ``PowConeND`` have some not-implemented functions. For ``PowCone3D`` there's no implementation for computing residual. For ``PowConeND`` there's no implementation of ``save_dual_variables``. The dual variable recovery for ``PowConeND`` will probably be hard to implement. Remember: ``PowConeND`` is the only constraint class we have that never touches the ``ParamConeProgram`` or ``ConeMatrixStuffing`` classes. There's a chance that dual variable recovery for ND power cones will require modifications to ``Exotic2Common``.

This PR makes no effort to implement power cone representations for user-facing atoms! We should have a careful discussion about how this might be done.

## Remarks

The scope of ``Exotic2Common`` could expand in the future. For example, we might support the epigraph of the relative entropy function defined on ``\R^n \times \R^n \times \R``. The "exotic cones" line thinking should help us prepare for eventually supporting extensible solvers like Hypatia (Julia) and Alfonso (MATLAB).

The atoms like ``geo_mean`` and ``pnorm`` currently use SOCP approximations. As it happens, those approximations are valid for the power cone as well. It might make sense to automatically write these atoms in terms of power cones (with no regard to the target solver) and then to convert them to SOCP approximations if necessary (say, when working with CVXOPT). Interestingly, ECOS would have the option of using the SOCP approximation, or using an exact reformulation in terms of exponential cones.


cvxpy/constraints/power.py
```diff
@@ -0,0 +1,183 @@
...
+    def num_cones(self):
+        return self.x.size
+
+    def cone_sizes(self):
+        return [3]*self.num_cones()
+
+    def is_dcp(self, dpp=False):
+        if dpp:
+            with scopes.dpp_scope():
+                return all(arg.is_affine() for arg in self.args)
```
rileyjmurray, line 68: I also need to check that ``self.alpha`` isn't a Parameter.

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,183 @@
...
+        w >= 0
+        """
+        Expression = cvxtypes.expression()
+        W = Expression.cast_to_const(W)
+        if not (W.is_real() and W.is_affine()):
+            raise ValueError("Invalid first argument.")
+        z = Expression.cast_to_const(z)
+        if z.ndim > 1 or not (z.is_real() and z.is_affine()):
+            raise ValueError("Invalid second argument.")
+        # Check t has one entry per cone.
```
rileyjmurray, line 110: Check that ``z`` has one entry per cone. (``t`` refers an to argument in ``SOC``'s init function)

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,183 @@
...
+            msg = "PowerConeND requires left-hand-side to have at least two terms."
+            raise ValueError(msg)
+        alpha = Expression.cast_to_const(alpha)
+        if alpha.shape != W.shape:
+            raise ValueError("Argument dimensions %s and %s are not equal."
+                             % (W.shape, alpha.shape))
+        if np.any(alpha.value <= 0):
+            raise ValueError("Argument alpha must be entry-wise positive.")
+        if np.any(np.abs(1 - np.sum(alpha.value, axis=axis)) > PowConeND._TOL_):
+            raise ValueError("Argument alpha must sum to 1 along axis %s." % axis)
```
rileyjmurray, line 127: Add tests that this behaves as expected.

cvxpy/tests/solver_test_helpers.py
```diff
@@ -453,6 +461,76 @@ def expcone_socp_1():
...
+    expect_epis = expect_x ** 2
+    expect_x = np.round(expect_x, decimals=5)
+    expect_epis = np.round(expect_epis, decimals=5)
+    expect_y_square = 25
+    var_pairs = [(x, expect_x),
+                 (epis, expect_epis),
+                 (y_square, expect_y_square)]
+    expect_ineq1 = 0.7793969212001993
+    expect_ineq2 = 2.865602615049077 / 10
+    con_pairs = [(constraints[0], None),  # hard to find until later ...
```
rileyjmurray, line 491: Set this based on a common value determined by SCS and MOSEK (offline)

cvxpy/tests/solver_test_helpers.py
```diff
@@ -453,6 +461,76 @@ def expcone_socp_1():
...
+    x = cp.Variable(shape=(3,))
+    expect_x = np.array([0.06393515, 0.78320961, 2.30571048])
+    hypos = cp.Variable(shape=(2,))
+    expect_hypos = None
+    objective = cp.Maximize(cp.sum(hypos) - x[0])
+    arg1 = cp.hstack([x[0], x[2]])
+    arg2 = cp.hstack(([x[1], 1.0]))
+    con_pairs = [
+        (x[0] + x[1] + 0.5 * x[2] == 2, None),
+        (cp.constraints.PowCone3D(arg1, arg2, hypos, [0.2, 0.4]), None)
```
rileyjmurray, line 523: Set the expected values based on offline calculations with SCS and MOSEK

cvxpy/reductions/cone2cone/exotic2common.py
```diff
@@ -0,0 +1,88 @@
...
+    if W.ndim == 1:
+        W = reshape(W, (W.size, 1))
+        alpha = np.reshape(alpha, (W.size, 1))
+    n, k = W.shape
+    if n == 2:
+        can_con = PowCone3D(W[0, :], W[1, :], z, alpha[0, :])
+    else:
+        T = Variable(shape=(n-2, k))
+        arg1, arg2, arg3, arg4 = [], [], [], []
+        for j in range(k):
```
SteveDiamond, line 58: Is there any way this can be vectorized? Not a huge deal.

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,183 @@
...
+        self.W = W
+        self.z = z
+        self.alpha = alpha
+        self.axis = axis
```
SteveDiamond, line 131: Should you check that axis is not None?

cvxpy/reductions/cone2cone/exotic2common.py
rileyjmurray: A vectorized version of this would require constructing a somewhat complicated sparse matrix by hand, and I didn't have the time to do that this time around. I'll add a TODO.

cvxpy/constraints/power.py
rileyjmurray: Since the default value is numeric, a user would have to supply ``None`` for that to happen. We don't have such a check for SOC constraints, so I think we can leave it as-is.

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,196 @@
...
+            msg = "Argument alpha must have entries in the open interval (0, 1)."
+            raise ValueError(msg)
+        arg_sizes = [self.x.size, self.y.size, self.z.size, self.alpha.size]
+        if min(arg_sizes) != max(arg_sizes):
+            msg = ("All arguments must have the same size. Provided arguments are"
+                   "of size %s" % str(arg_sizes))
+            raise ValueError(msg)
+        super(PowCone3D, self).__init__([self.x, self.y, self.z],
+                                        constr_id)
+        pass
```
akshayka, line 48: Delete the `pass` statement.

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,196 @@
...
+from cvxpy.utilities import scopes
+import numpy as np
+
+
+class PowCone3D(Constraint):
+
+    def __init__(self, x, y, z, alpha, constr_id=None):
+        """
+        x[i]**alpha * y[i]**(1-alpha) >= |z[i]|  for all i
+        x >= 0, y >= 0
```
akshayka, line 28: Can you include a small docstring documenting the types of x, y, z, and alpha (i.e., `Expression`s and a numeric constant)?

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,196 @@
...
+            raise ValueError("Invalid second argument.")
+        # Check z has one entry per cone.
+        if (W.ndim <= 1 and z.size > 1) or \
+           (W.ndim == 2 and z.size != W.shape[1-axis]) or \
+           (W.ndim == 1 and axis == 1):
+            raise ValueError(
+                "Argument dimensions %s and %s, with axis=%i, are incompatible."
+                % (W.shape, z.shape, axis))
+        if W.ndim == 2 and W.shape[axis] <= 1:
+            msg = "PowerConeND requires left-hand-side to have at least two terms."
```
akshayka, line 131: `PowerConeND` --> `PowConeND`

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, W, z, alpha, axis=0, constr_id=None):
+        """
+        \\prod_i w_i^{\\alpha_i} >= |z|
+        w >= 0
+        """
+        Expression = cvxtypes.expression()
+        W = Expression.cast_to_const(W)
+        if not (W.is_real() and W.is_affine()):
+            raise ValueError("Invalid first argument.")
```
akshayka, line 119: Should this error message and the next one be more descriptive, like they are in `PowCone3D`?

cvxpy/constraints/power.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class PowConeND(Constraint):
+
+    _TOL_ = 1e-6
+
+    def __init__(self, W, z, alpha, axis=0, constr_id=None):
+        """
+        \\prod_i w_i^{\\alpha_i} >= |z|
+        w >= 0
```
akshayka, line 114: Include a small docstring, with basic facts like PowConeND generalizes PowCone3D?

cvxpy/problems/problem.py
```diff
@@ -826,7 +826,7 @@ def backward(self):
         obtaining the gradient of a solution with respect to the Parameters.
         In other words, it calculates the sensitivities of the Parameters
         with respect to perturbations in the optimal Variable values. This
-        can be useful for integrating CVXPY into automatic differentation
+        can be useful for integrating CVXPY into automatic differentiation
```
akshayka, line 829: oops thanks!

cvxpy/reductions/cone2cone/exotic2common.py
```diff
@@ -0,0 +1,92 @@
...
+supported by ParamConeProg. If ParamConeProg is updated
+to support more cones, then it may be necessary to change
+this file.
+"""
+
+
+def pow_nd_canon(con, args):
+    """
+    con : PowerConeND
+        I can extract metadata from this.
```
akshayka, line 38: Was this a note to yourself? At least change this to the royal "we" :)

cvxpy/reductions/cone2cone/exotic2common.py
```diff
@@ -0,0 +1,92 @@
...
+        alpha = alpha.T
+    if W.ndim == 1:
+        W = reshape(W, (W.size, 1))
+        alpha = np.reshape(alpha, (W.size, 1))
+    n, k = W.shape
+    if n == 2:
+        can_con = PowCone3D(W[0, :], W[1, :], z, alpha[0, :])
+    else:
+        T = Variable(shape=(n-2, k))
+        arg1, arg2, arg3, arg4 = [], [], [], []
```
akshayka, line 57: Should these names be more descriptive? Even `x`, `y`, `z`, and `alpha` would be better.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -52,38 +52,54 @@ class ConeDims(object):
         A list of the positive semidefinite cone dimensions, where the
         dimension of the PSD cone of k by k matrices is k.
     """
+
+    EQ_DIM = s.EQ_DIM
+    LEQ_DIM = s.LEQ_DIM
+    EXP_DIM = s.EXP_DIM
+    SOC_DIM = s.SOC_DIM
+    PSD_DIM = s.PSD_DIM
+    P3D_DIM = 'p3'
```
akshayka, line 61: Should `P3D_DIM` be defined in `settings`, for symmetry with the other cones?

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -214,8 +230,11 @@ def construct_solving_chain(problem, candidates, gp=False, enforce_dpp=False):
         solver_instance = slv_def.SOLVER_MAP_CONIC[solver]
         if (all(c in solver_instance.SUPPORTED_CONSTRAINTS for c in cones)
                 and (has_constr or not solver_instance.REQUIRES_CONSTR)):
-            reductions += [ConeMatrixStuffing(),
-                           solver_instance]
+            if ex_cos:
+                reductions += [Exotic2Common(),
+                               ConeMatrixStuffing(), solver_instance]
+            else:
+                reductions += [ConeMatrixStuffing(), solver_instance]
```
akshayka, line 237: Sorry, a nit, but I would change this to:

```
if ex_cos:
  reductions += [Exotic2Common()]
reductions += [ConeMatrixStuffing(), solver_instance]
```

akshayka (commented):
This looks great Riley! I left just minor comments.

akshayka (approved):
(Approved, modulo the minor comments I left.)

cvxpy/constraints/power.py
rileyjmurray: Nice catch! This function was actually written in a way that required ``alpha`` to be a vector. I've added automatic promotion of scalar ``alpha`` to vector ``alpha`` and written a proper docstring.

cvxpy/reductions/cone2cone/exotic2common.py
rileyjmurray: It was! Lol. Nice catch.

cvxpy/reductions/cone2cone/exotic2common.py
rileyjmurray: Good point!

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
rileyjmurray: There are comments in ``settings`` that say these constants should be moved into the files related to matrix stuffing. I was really thinking of hard-coding the literal values of ``s.EXP_DIM``, ``s.SOC_DIM``, etc... into the class attributes here (like with ``P3D_DIM``), but I thought that would be a bit much without checking with you guys. I should probably commit to one of these paths.

Let me know if you'd like me to switch to string literals for these attributes of the ``ConeDims`` class. I can make any necessary changes in the solver interface files. I could also just leave a "TODO" indicating that as a future task.

rileyjmurray:
Leaving some TODO's for myself before merging:
 * Update the ``Slacks`` reduction in ``cone2cone.py``. Update that code path in the MOSEK interface.
 * Add a unit-test for mixed-integer power cone programs.

rileyjmurray:
@SteveDiamond @akshayka I've also updated the web documentation and removed some case-checking in tests that were put in place for Python 2. Let me know if things still look good to you.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
rileyjmurray: @akshayka let me know what you think

rileyjmurray:
Merging in, per Steven's go-ahead on Slack.

PR #1224, opened 2021-01-25
Add log_normcdf atom

ali-tny:
This PR adds the `log_normcdf` atom from the Matlab library, using a quadratic approximation written as a composition of cvxpy functions.

Closes https://github.com/cvxgrp/cvxpy/issues/1219.

The first commit uses [the same approximation as the Matlab implementation](https://github.com/cvxr/CVX/blob/master/functions/%40cvx/log_normcdf.m). I update the approximation in the second commit to bound the absolute error, at the cost of a small increase in the error between 1 and 2:
image
(i clipped the y limit on the relative to make it a bit clearer - the relative error jumps to 1 after x=3, where the approximation is just 0)
and the error of the new approximation is lower than the old for all x < -4 as well (and equal for x > 4)


ali-tny:
Not sure what the problem with the windows builds is - at the start of the conda environment creation it gets:
```
CondaHTTPError: HTTP 500 INTERNAL SERVER ERROR for url <https://conda.anaconda.org/oxfordcontrol/win-64/repodata.json>

...

Could not find conda environment: testenv
You can list all discoverable environments with `conda info --envs`.
```
but then does a load of installing then only later fails with

 ```
conda activate testenv
Could not find conda environment: testenv
You can list all discoverable environments with `conda info --envs`.
Command exited with code 1
```

seems a strange error considering the changes in this PR - i was hoping it was intermittent but force-pushed to run it again and got the same problem

File: cvxpy/atoms/elementwise/log_normcdf.py [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+                0.09003495,
+                0.08200854,
+                0.01371543,
+                0.04641081,
+            ]
+        )
+    )
+    b = np.array([[3.0, 2.0, 1.0, 0.0, -1.0, -2.5, -3.5]]).reshape(-1, 1)
+
+    is_scalar = not hasattr(x, "shape")
```
SteveDiamond, line 44: This is interesting. Normally we use ``Expression.cast_to_const`` to make sure inputs are cvxpy objects.

SteveDiamond:
> Not sure what the problem with the windows builds is - at the start of the conda environment creation it gets:
> 
> ```
> CondaHTTPError: HTTP 500 INTERNAL SERVER ERROR for url <https://conda.anaconda.org/oxfordcontrol/win-64/repodata.json>
> 
> ...
> 
> Could not find conda environment: testenv
> You can list all discoverable environments with `conda info --envs`.
> ```
> 
> but then does a load of installing then only later fails with
> 
> ```
> conda activate testenv
> Could not find conda environment: testenv
> You can list all discoverable environments with `conda info --envs`.
> Command exited with code 1
> ```
> 
> seems a strange error considering the changes in this PR - i was hoping it was intermittent but force-pushed to run it again and got the same problem

This can't be from your PR obviously. @rileyjmurray any idea what's going on?

rileyjmurray:
It looks like something's up with conda: https://github.com/conda/conda/issues/10475.

cvxpy/atoms/elementwise/log_normcdf.py
ali-tny: ah ok. I mainly did this so that if the input was a float rather than a numpy array, the output value would be a float rather than a 1-element numpy array - maybe that doesn't matter though. I had a look for a util but missed the `cast_to_const`!

I can simplify it to just be:
```
x = Expression.cast_to_const(x)
x_size = functools.reduce(lambda i, j: i * j, x.shape, 1)
```
and avoid the whole `if / else` block

ali-tny: have updated to use `cast_to_const` in a new commit - can rebase and squash once reviewed

ali-tny: also I've just realised I can just use `x.size` instead of the reduce so i'll use that... (lol)

SteveDiamond (approved):
LGTM

SteveDiamond:
This all looks good! Thanks for the PR!

PR #1230, opened 2021-02-01
use oldest supported numpy in pyproject.toml

### LINKED ISSUE #1229: cvxpy 1.1.7 no longer compatible with numpy<1.20?
I use python 3.7.9 with cvxpy 1.1.7 and numpy 1.18.1.
However as of Sunday we get an error message that the module is compiled with a newer numpy version. Before the weekend everything was still working.

What happened on sunday 31 january that this throws an error now?
Issue is resolved by upgrading numpy to 1.20, however isnt cvxpy supposed to support numpy>1.15? I want to stay on 1.18.1 still.

import cvxpy as cp
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/__init__.py:18: in <module>
    from cvxpy.atoms import *
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/atoms/__init__.py:20: in <module>
    from cvxpy.atoms.geo_mean import geo_mean
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/atoms/geo_mean.py:20: in <module>
    from cvxpy.utilities.power_tools import (fracify, decompose, approx_error, lower_bound,
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/utilities/power_tools.py:18: in <module>
    from cvxpy.atoms.affine.reshape import reshape
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/atoms/affine/reshape.py:18: in <module>
    from cvxpy.atoms.affine.hstack import hstack
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/atoms/affine/hstack.py:18: in <module>
    from cvxpy.atoms.affine.affine_atom import AffAtom
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/atoms/affine/affine_atom.py:22: in <module>
    from cvxpy.cvxcore.python import canonInterface
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/cvxpy/cvxcore/python/__init__.py:3: in <module>
    import _cvxcore
E   ImportError: numpy.core.multiarray failed to import
------------------------------- Captured stderr --------------------------------
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd

akshayka:
In theory this should fix #1229.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["numpy >= 1.15", "scipy >= 1.1.0", "setuptools>=40.8.0", "wheel"]
+requires = ["oldest-supported-numpy", "scipy >= 1.1.0", "setuptools>=40.8.0", "wheel"]


SteveDiamond (approved):

BastiaanvG:
@akshayka 
With this change I still get a conflict. 

> ERROR: Some build dependencies for git+https://github.com/BastiaanvG/cvxpy (from -r requirements.txt (line 6)) conflict with the backend dependencies: numpy==1.14.5 is incompatible with numpy>=1.15.

(1.14.5 is the minimum supported numpy version)

The fix should include removing the minimum numpy version under setup_requires in setup.py.

akshayka:
@SteveDiamond @rileyjmurray can we relax the minimum NumPy version in setup.py? Or is 1.15 special?

akshayka:
@bastiaanvG which build system is giving you that error? Is that from pip, or some other tool?

rileyjmurray:
@akshayka I do not believe numpy can be relaxed from 1.15 to 1.14. There was a whole compatibility debacle about that a while back.

BastiaanvG:
It works. It just throws an error as setup.py has a conflicting build specification. Isnt the purpose of using a pyproject.toml to get the build specification out of setup.py?

pyproject.toml
```diff
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["numpy >= 1.15", "scipy >= 1.1.0", "setuptools>=40.8.0", "wheel"]
\ No newline at end of file
+requires = ["oldest-supported-numpy", "scipy >= 1.1.0", "setuptools>=40.8.0", "wheel"]
```
rileyjmurray, line 2: I'm happy to merge this in, but we should keep in mind that our solver dependencies might end up making ``oldest_supported_numpy`` (https://pypi.org/project/oldest-supported-numpy/) inadequate.

akshayka:
@rileyjmurray Another option is to pin the version in the pyproject.toml to the minimum version accepted by CVXPY. I guess that's more natural.

pyproject.toml
akshayka: Yeah. Not sure whether we should merge this or attempt to find a better solution.

rileyjmurray:
@akshayka yes we could pin CVXPY's minimum supported version, but then we'd have to do that for different platforms (at the very least, different python versions), and that's precisely what ``oldest-supported-numpy`` tries to avoid. I'm unfortunately not able to look into these things in detail right now. I think the best thing would be to try your proposed fix, but understanding that this fix may not be final.

BastiaanvG:
Isnt the better option to just only exlcude the latest numpy version inside the pyproject.toml? Instead of forcing the oldest?

rileyjmurray:
@BastiaanvG if we did that then we would build against numpy 1.19 in all situations, and might break things for numpy 1.18 users. Forcing oldest maximizes our chances of compatibility without making assumptions on user's configurations.

rileyjmurray:
@akshayka @SteveDiamond given the number of people affected, I think we should go ahead with the candidate fix we have so far. We probably need to increment cvxpy's version to 1.1.8 in order to trigger deployments, however.

akshayka:
Sounds good to me. I'll merge this in.

akshayka:
@SteveDiamond , can you cut a new release?

SteveDiamond:
Done.

roberthuisman:
Thanks for putting the effort into resolving the problem.

However as of now cvxpy 1.1.8 cannot be installed on python 3.7 since the oldest-supported-numpy is 1.14.5.

ERROR: Some build dependencies for cvxpy==1.1.8 from https://files.pythonhosted.org/packages/e6/3b/6a6650ed48503c4e4b2f497ce40517fe427f3cfacfc8fb43ff52f89bc4c8/cvxpy-1.1.8.tar.gz#sha256=444068d4eda9ffcd43578895174489d4cef36b28ba7ae8a96ab9ef9571d2b4ff conflict with the backend dependencies: numpy==1.14.5 is incompatible with numpy>=1.15.

BastiaanvG:
@roberthuisman 
Remove this line or remove the versioning requirement. And it will work
https://github.com/cvxgrp/cvxpy/blob/1007d3a2d90e45aaf92314c774b1cbfdc55f0283/setup.py#L77

This line, install requirements can stay the same (see: https://pypi.org/project/oldest-supported-numpy/):
https://github.com/cvxgrp/cvxpy/blob/1007d3a2d90e45aaf92314c774b1cbfdc55f0283/setup.py#L75

qci-vanstone:
> Remove this line or remove the versioning requirement. And it will work
> https://github.com/cvxgrp/cvxpy/blob/1007d3a2d90e45aaf92314c774b1cbfdc55f0283/setup.py#L77

If that is the case, can the `setup_requires` version be relaxed or removed from cvxpy? Or do we have to build this from source on python 3.7?

Issue #1232, opened 2021-02-03
Allow single MI solver != ECOS_BB


phschiele:
First of all, thank you for maintaining this awesome package!

**Describe the bug**
An [error is raised](https://github.com/cvxgrp/cvxpy/blob/71fa8221d5903f78f8273eaa7a222e571c493117/cvxpy/problems/problem.py#L611) if exactly one MI solver is installed, even if it is not `ECOS_BB`, which needs to be called explicitly.
I propose a slightly narrower condition in [this PR](https://github.com/cvxgrp/cvxpy/pull/1233).

```python
# ECOS_BB must be called explicitly.
if len(slv_def.INSTALLED_MI_SOLVERS) == 1 and solver != s.ECOS_BB:
    msg = ...
    raise error.SolverError(msg)
```

**To Reproduce**
With only the default solvers installed, run:
```
pip uninstall ecos
pip install cplex
```
```python
import cvxpy as cp

x_bool = cp.Variable(boolean=True)
obj = cp.Minimize(cp.abs(x_bool - 0.2))
p = cp.Problem(obj, [])
p.solve()
```

**Expected behavior**
Do not raise the exception if a single MI solver except `ECOS_BB` is installed.

**Version**
 - OS: Ubuntu 18.04
 - CVXPY Version: 1.1.7

**Additional context**
I acknowledge that this is an edge case. My proposed solution only defines the condition a bit narrower, which matches the intended use case and does not increase complexity.

phschiele:
Closed with #1233

PR #1233, opened 2021-02-03
Allow single MI solver != ECOS_BB [issue 1232]

phschiele:
The corresponding issue for this PR can be found here: #1232 

While the change is only one line, I followed the contributing guidelines:

- [x]  no regression change
- [x]  no benchmark timing change (`37.38s` pre-change vs. `37.28s` after the change)
- [x]  flake8 passing

A UT which checks exactly this scenario is likely not relevant for most users and indeed would be quite tricky to include.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -608,7 +608,7 @@ class Problem(u.Canonical):
 
         if self.is_mixed_integer():
             # ECOS_BB must be called explicitly.
-            if len(slv_def.INSTALLED_MI_SOLVERS) == 1 and solver != s.ECOS_BB:
+            if slv_def.INSTALLED_MI_SOLVERS == [s.ECOS_BB] and solver != s.ECOS_BB:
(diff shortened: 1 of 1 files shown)


SteveDiamond:
@phschiele can you clarify the bug? It only happens when the user doesn't have ecos installed, right?

phschiele:
> @phschiele can you clarify the bug? It only happens when the user doesn't have ecos installed, right?

@SteveDiamond Precisely. It happens if a single (valid MI) solver is installed, which is not `ecos`.
I have added a more detailed description in [this issue](https://github.com/cvxgrp/cvxpy/issues/1232).

rileyjmurray:
@SteveDiamond this PR fixes an issue affecting people who use the ``cvxpy-base`` conda package (which was created to avoid hard dependencies on ECOS and other software licensed under the GPL).

SteveDiamond (approved):
LGTM

PR #1234, opened 2021-02-03
python-version-specific numpy versions in pyproject.toml

akshayka:
This might help with #1229 (per the suggestion from @BastiaanvG )

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,9 @@
 [build-system]
 requires = [
-    "numpy==1.15; python_version='3.6'",
-    "numpy==1.15; python_version='3.7'",
-    "numpy==1.16; python_version='3.8'",
-    "numpy==1.19; python_version='3.9'",
+    "numpy==1.15; python_version=='3.6'",
+    "numpy==1.15; python_version=='3.7'",
+    "numpy==1.16; python_version=='3.8'",
+    "numpy==1.19; python_version=='3.9'",
     "scipy >= 1.1.0",
     "setuptools>=40.8.0",
     "wheel"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,10 @@
 [build-system]
-requires = ["oldest-supported-numpy", "scipy >= 1.1.0", "setuptools>=40.8.0", "wheel"]
+requires = [
+    "numpy==1.15; python_version='3.6'",
+    "numpy==1.15; python_version='3.7'",
+    "numpy==1.16; python_version='3.8'",
+    "numpy==1.19; python_version='3.9'",
+    "scipy >= 1.1.0",
+    "setuptools>=40.8.0",
+    "wheel"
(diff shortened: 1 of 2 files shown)


rileyjmurray:
We should see how the CI builds go. Although, I don't think the CI will adequately test this change. I suggest the following:

I merge these changes into a branch on my fork. This will trigger CI builds that deploy to a test version of PyPI. Once those deployments succeed we should test installation from pip as though we were end-users. We only really need to do this with linux, since linux pip users have to install from a source distribution. @akshayka, could you also run such test installs on a macOS machine? I'd want to make sure installation from source also works there. (For Windows we can rely the fact that we have wheels, which I think are unaffected by this issue.)

akshayka:
That sounds like a good plan to me.

@rileyjmurray : I no longer have a Mac, went all in on Linux some time ago.

@SteveDiamond : Do you have access to macOS to test on?

Otherwise hopefully someone from the community will be able to help out.

rileyjmurray:
@akshayka I have an old mac (relatively recent macOS) that I can use. If you can take the linux tests then we'd be good.

akshayka:
@rileyjmurray Let me know once the CI running on your fork has uploaded to the test repository on PyPI.

rileyjmurray:
@akshayka there's a cvxpy 1.1.8 on test pypi. I'm running into installation from source on mac. However, I think that trouble is because when I specify the index url to use test pypi, the installation process can't find dependencies like ``oldest-supported-numpy``. I'm tinkering with it now and let you know if I think these errors are okay.

akshayka:
Riley, thanks for trying that all out!

For your first point, about finding other dependencies, you can pass the following flag to pip

```
--extra-index-url https://pypi.org/simple
```

So something like

```
https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple cvxpy
```

should hopefully work?

For your second point (first edit), maybe we should just hardcode the NumPy versions in the `pyproject.toml` (as shown in the oldest-supported-numpy docs). Something like:

```
[build-system]
requires = [
    "wheel",
    "setuptools",
    "numpy==1.15; python_version=='3.6'",
    "numpy==1.16; python_version=='3.7'",
    # more numpy requirements...
]
```

Or whatever the correct numpy requirements are. At least then we should hopefully avoid the error encountered when installing NumPy < 1.15? There are only a few versions of Python we're supported, so this seems reasonable to me.

So my proposal would be to revert the `oldest-supported-numpy` change and instead hardcode python-version-specific NumPy versions.

akshayka:
@rileyjmurray @SteveDiamond want to take a look? I used the Python versions from `continuous_integration/Travis/install_dependencies.sh`

SteveDiamond (approved):
LGTM, and I strongly endorse the manual pinning to Python versions.

rileyjmurray (approved):

Issue #1160, opened 2020-10-28
Reducing problem data tensor can be a bottleneck


akshayka:
**Describe the bug**
The `reduce_problem_data_tensor` function in `canonInterface.py` can be very slow.

**To Reproduce**
```python3
import cvxpy as cp
import numpy as np

m, n = 100000, 1000
X = np.random.randn(m, n)                                                              
y = np.random.randn(n)                                      
beta = cp.Variable(m)                                         
prob = cp.Problem(cp.Minimize(cp.sum_squares(y - X.T @ beta) +  cp.norm(beta, 1)))
prob.solve()
```

**Output**
![image](https://user-images.githubusercontent.com/1994308/97473793-9e3a8480-1908-11eb-8d71-220f8d9edced.png)

**Version**
 - OS: Ubuntu 20.10
 - CVXPY Version: 1.1.7

**Additional context**
`reduce_problem_data_tensor` was introduced in https://github.com/cvxgrp/cvxpy/pull/1010, where it was found that removing the rows in the tensor that are identically zero could in some cases yield orders of magnitude speed ups when multiplying the tensor by the parameter vector. This function is implemented in Python and performs a for loop over the nnz in A, which is possibly the bottleneck when nnz is large.

One option is to move this function into C. A better option would be to construct a more efficient implementation of this function in Python

akshayka:
@sbarratt and I have separately looked into this, and neither of us was able to find a Python implementation that was much faster (and didn't OOM our machines on large problems). So it seems like a C++ port of the current implementation would be the way to go (with say a pybind11 binding).

Mojusko:
Same problem here? Is there a way to disable it?

rileyjmurray:
I don't know anything about disabling this function and falling back on an old implementation (it seems unlikely that we'd leave something like that in). However, as a more general note, I think it might be possible to get some noticeable performance improvements to the current python implementation by making very small changes. I've had to do an almost identical operation before, just with zero columns of a matrix

https://github.com/rileyjmurray/sageopt/blob/001fc80b24fe5ee7df24890af6d4d935ffd2983a/sageopt/coniclifts/utilities.py#L59-L66

The difference is that I iterate over the lists/arrays directly, rather than iterating over an index that I use to access arrays. So for example, I'd change
```
    reduced_row = np.zeros(old_row.shape)
    for i in range(reduced_row.size):
        reduced_row[i] = old_row_to_new_row[old_row[i]]
```
to
```
[...]

PR #1235, opened 2021-02-04
Performance improvement for reducing problem data tensor [issue 1160]

phschiele:
In issue #1160, @akshayka and @rileyjmurray were disucssing ways to speed up `reduce_problem_data_tensor()`.

Using [line_profiler](https://github.com/pyutils/line_profiler) I have profiled the optimization problem stated in the issue, narrowing down the lines consuming most of the CPU time:
```
Total time: 124.153 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   111         2         17.0      8.5      0.0      old_row_to_new_row = {
   112         2    5266879.0 2633439.5      4.2          unique_old_row[i]: i for i in range(unique_old_row.size)
   113                                               }                                        
   ...
   119  10043002   21723051.0      2.2     17.5      for i in range(reduced_row.size):
   120  10043000   26991628.0      2.7     21.7          reduced_row[i] = old_row_to_new_row[old_row[i]]                              
   ...
   148  10043000   23828321.0      2.4     19.2          while col == prev_col and i < nnz - 1:
   149  10021000   21721225.0      2.2     17.5              i += 1
   150  10021000   23244796.0      2.3     18.7              col = cols[i]
```

Using the additional arguments `return_inverse` and `return_counts` of `np.unique()`, the time spent in the function was reduced from `124s` to `1.15s`. 

I haven't worked with CSR matrices extensively, so please let me know if I am overlooking something.

### Contributing guideline checks:

- [x]  no regression change
- [x] slight improvement in benchmark timings (34.76s pre-change vs. 31.02s after the change)
- [x] flake8 passing
I noticed many regression test sensitivities when trying out different approaches, so the function seems to be well tested. If you can think of any additional test case which I should add, please let me know.

### DIFF
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -135,7 +135,6 @@ def reduce_problem_data_tensor(A, var_length, quad_form=False):
     positions, counts = np.unique(cols, return_counts=True)
     indptr[positions+1] = counts
     indptr = np.cumsum(indptr)
-    # assert np.all(indptr == indptr_bak)
 
     return reduced_A, indices, indptr, shape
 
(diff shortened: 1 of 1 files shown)


akshayka (approved):
This is great! Thank you!

Issue #1238, opened 2021-02-12
Stop resetting warning filters


colinc719:
**Describe the bug**
Multiplying an expression has a path that will reset existing warning filters.

**To Reproduce**
```python
import warnings

import cvxpy


class MyWarning(Warning):
    pass


with warnings.catch_warnings():
    warnings.simplefilter("error", MyWarning)
    [1, 2, 3] * cvxpy.Variable(3)
    warnings.warn(MyWarning("my msg"))
```

**Expected behavior**
I would expect
```
MyWarning: my msg
```
to be raised
(which happens if you remove the `[1, 2, 3] * cvxpy.Variable(3)` line)

**Version**
 - CVXPY Version: 1.1.10

rileyjmurray:
I'm not sure if I follow your example. We reset the warning filters to ensure that users are alerted if there are unknown places in their codebase where matmul happens by ``*``. The idea is that running non-compliant code will indicate how many times the deprecated syntax is used (not just "whether or not the deprecated syntax is used somewhere").

However, it seems that by resetting filters with this purpose in mind, we're also resetting any filters a user might have set for other purposes. Is that correct?

colinc719:
Yeah, the biggest thing is that it's resetting filters that are completely unrelated to cvxpy.

But, I also don't think you should be resetting filters at all.  We actually came across this bug because we were trying to turn this warning into an error after fixing all cases in our codebase.  I think you should just respect user's warning filters.

akshayka:
Python3 docs say

>  Repetitions of a particular warning for the same source location are typically suppressed.

(https://docs.python.org/3/library/warnings.html).

So maybe the warning behavior will work as we intend it, even if CVXPY doesn't touch filters.

akshayka:
Sorry, accidentally hit the close button.

Riley, I misunderstood your comment. You'd like to raise a warning every time. In that case, can we just use an "always" filter, instead of resetting warnings.

colinc719:
I think those decisions about whether to warn every time or just once should be up to the user.  If they want to warn once, they can:
```python
# library code
def do_something_bad():
    warnings.warn(UserWarning("I did a bad thing"))



# user code
with warnings.catch_warnings():
    warnings.simplefilter("once", UserWarning)
    for _ in range(3):
        do_something_bad()
```
output:
```
<ipython-input-48-207c6614fee3>:2: UserWarning: I did a bad thing
  warnings.warn(UserWarning("I did a bad thing"))
[...]

PR #1240, opened 2021-02-14
Matmul warning

rileyjmurray:
This addresses #1238 and is intended to supersede #1239. It removes side-effects of our existing warnings for accessing matmul through ``*`` and introduces a new ``MatmulWarning`` class for that purpose.

This PR also updates the web doc source to mention the importance of cvxpy 1.1.10 for the numpy installation issue.

### DIFF
--- a/cvxpy/error.py
+++ b/cvxpy/error.py
@@ -66,9 +66,3 @@ class ParameterError(Exception):
     """Error thrown for accessing the value of an unspecified parameter.
     """
     pass
-
-
-class MatmulWarning(Warning):
-    """Warning raised for accessing matmul by the * operator.
-    """
-    pass
--- a/cvxpy/error.py
+++ b/cvxpy/error.py
@@ -66,3 +66,9 @@ class ParameterError(Exception):
     """Error thrown for accessing the value of an unspecified parameter.
     """
     pass
+
+
+class MatmulWarning(Warning):
+    """Warning raised for accessing matmul by the * operator.
+    """
+    pass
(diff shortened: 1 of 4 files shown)


akshayka:
I think raising a MatmulWarning instead of a DeprecationWarning is bit odd. Production could might have policies that promote DeprecationWarning's to errors. A custom MatmulWarning will go undetected, unless a human sees it.

Why not raise a DeprecationWarning, which is a very standard type of warning? Humans will understand it, and so will codebases. Its semantics are very clear.

akshayka:
That said I don't feel strongly, so I won't block this! I'll leave it to @SteveDiamond

rileyjmurray:
We changed from a DeprecationWarning to a UserWarning a while back, because DeprecationWarnings are not as visible to users: https://github.com/cvxgrp/cvxpy/issues/980 and https://github.com/cvxgrp/cvxpy/pull/981.

I see your point about the semantics of DeprecationWarning though. One option would be to raise a DeprecationWarning in addition to something a user is more likely to see. I'd be okay with removing the custom MatmulWarning if we raised both a DeprecationWarning and UserWarning. I think both are actually justified, as the DeprecationWarning says "you should change this code and may have to change code elsewhere to be compliant in the future", while a UserWarning says "this might not have done what you thought it was going to do." 

@akshayka what do you think about the two-warning option?

akshayka:
@rileyjmurray , that sounds good to me!

SteveDiamond:
I'm also fine with the two warnings.

rileyjmurray:
Great! I'll change it to two warnings and merge it in later today.

Issue #1213, opened 2021-01-17
DCP rules and diagonal submatrices


tschm:
Hello,
We all know how to maximize

`
mu' x - r * x' Cov x.
`

Here mu is a vector of expected returns, r is a parameter for risk-aversion (nonneg=True) and Cov is PSD covariance matrix. There is also what's called a concentration which is x' cvx.diag(cvx.diag(Cov)) x. So I am using only the diagonal entries of Cov. Note that this matrix will be PSD if Cov is PSD. Nevertheless, cvxpy fails with DCP problems. How would I approach this problem? manz thanks...

phschiele:
Hi @tschm!
Perhaps it helps to add a reproducible example to talk about, so lets consider this optimization from the [portfolio optimization notebook](https://colab.research.google.com/github/cvxgrp/cvx_short_course/blob/master/applications/portfolio_optimization.ipynb):

```python
# Generate data for long only portfolio optimization.
import numpy as np
import cvxpy as cp

np.random.seed(1)
n = 10
mu = np.abs(np.random.randn(n, 1))
Sigma = np.random.randn(n, n)
Sigma = Sigma.T.dot(Sigma)

w = cp.Variable(n)
gamma = cp.Parameter(nonneg=True)
gamma.value = 1
ret = mu.T@w 
risk = cp.quad_form(w, Sigma)
concentration = cp.quad_form(w, cp.diag(cp.diag(Sigma)))
prob = cp.Problem(cp.Maximize(ret - gamma*risk - concentration), 
               [cp.sum(w) == 1, 
                w >= 0])
prob.solve()
```
Indeed, this fails due to the DCP error you mentioned.
However, changing `cp.diag(cp.diag(Sigma))` to `np.diag(np.diag(Sigma))` already solves the issue.

What would also work is passing the `value` of the first `diag` function to the second one:
`cp.diag(cp.diag(Sigma).value)`

```python
diag1 = cp.diag(Sigma)
print(diag1.is_psd())
diag2 = cp.diag(diag1.value)
print(diag2.is_psd())
>> True
>> True
#####
diag1 = cp.diag(Sigma)
print(diag1.is_psd())
diag2 = cp.diag(diag1)
print(diag2.is_psd())
>> True
>> False
```

It seems to build down to the implementation of `is_nonneg` in the `diag_mat` class.
[...]

PR #1242, opened 2021-02-17
Extend is_nonneg() of diag_mat to check if arg is psd [issue 1213]

phschiele:
This PR adds @SteveDiamond 's proposed improvement for the `is_nonneg()` check in the `diag_mat` class, as per discussion in #1213. Also, the tests are extended to cover this case.

### Contributing guideline checks:

- [x] no regression change
- [x] no relevant change in benchmark timings (25.33s pre-change vs. 25.05s after the change)
- [x] flake8 passing


SteveDiamond:
LGTM! Thanks for fixing this!

PR #1246, opened 2021-02-23
Use try/except to get Gurobi iterations 

### LINKED ISSUE #1194: How to deal with the error when using Gurobi with cvxpy ：Unable to retrieve attribute 'BarIterCount'
How to deal with the error when using Gurobi with cvxpy ：AttributeError: Unable to retrieve attribute 'BarIterCount'.

I have an Integer programming problem, using cvxpy and set gurobi as solver. When the number of variables is small, the result is ok. After the number of variables reach a level of like 43x13x6, then this error occured. I suppose it may be caused by the scale of the problem, in which the gurobi solver can not estimate the BarIterCount, which is the max Iterations needed.

Thus, I wonder, is there any way to manually set the BarItercount attribute of gurobi through the interface of the cvx? Or whether there exists other way to solve this problem?

Thanks for any suggestions you may provide for me.

The error is as follows:

  s.NUM_ITERS: model.BarIterCount,
  File "src\gurobipy\model.pxi", line 343, in gurobipy.gurobipy.Model.__getattr__
  File "src\gurobipy\model.pxi", line 1842, in gurobipy.gurobipy.Model.getAttr
  File "src\gurobipy\attrutil.pxi", line 100, in gurobipy.gurobipy.__getattr
AttributeError: Unable to retrieve attribute 'BarIterCount'

bstellato:
Quick fix #1194 . 

It looks like Gurobi does not create the attributes BarIterCount and IterCount when the problem is large.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
@@ -54,9 +54,24 @@ class GUROBI(QpSolver):
         constraints_grb = model.getConstrs()
         m = len(constraints_grb)
 
+
+        # Note: Gurobi does not always fill BarIterCount
+        # and IterCount so better using try/except
+        try:
+            bar_iter_count = model.BarIterCount
+        except AttributeError:
+            bar_iter_count = 0
+        try:
+            simplex_iter_count = model.IterCount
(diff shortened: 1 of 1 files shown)


rileyjmurray:
Thanks, @bstellato! Merging in.

Issue #1147, opened 2020-10-14
A problem that could be solved using CVXOPT in v1.0.31 fails in v1.1.6


jeffrey-hokanson:
**Describe the bug**
I suspect there has been a regression in the reduction for CVXOPT.  The example below runs without error in CVXPY v1.0.31, but fails in CVXPY v1.1.6.

**To Reproduce**
```
import numpy as np
import cvxpy as cp

M = 2
m = 3
X = np.random.randn(M, m)

alpha = cp.Variable(M, nonneg = True)
beta = cp.Variable()
obj = cp.Maximize(beta)
x = cp.Parameter(m)
p = cp.Parameter(m)

constraints = [ X.T @ alpha == beta * p + x, cp.sum(alpha) == 1,]
prob = cp.Problem(obj, constraints)

x.value = np.mean(X, axis = 0)
p.value = X[0] - x.value
prob.solve(solver = 'CVXOPT', verbose =True, warm_start = True)
print("alpha", alpha.value)
print("beta", beta.value)

assert np.all(np.isclose(alpha.value,np.eye(M)[:,0]))
assert np.isclose(beta.value, 1)
```

Note, this will succeed if switching to SCS or if `M`>=  `m` in both CVXPY v1.0.31 or v1.1.6.  

**Expected behavior**
This code should run and pass the assertion statements. 

**Output**

In CVXPY v1.0.31
```
     pcost       dcost       gap    pres   dres   k/t
 0:  3.8781e-16  5.5511e-16  2e+00  0e+00  3e+00  1e+00
 1: -8.2427e-01 -8.2427e-01  2e-01  3e-16  3e-01  9e-02
 2: -9.9824e-01 -9.9824e-01  2e-03  1e-16  3e-03  1e-03
 3: -9.9998e-01 -9.9998e-01  2e-05  2e-16  3e-05  1e-05
 4: -1.0000e+00 -1.0000e+00  2e-07  3e-16  3e-07  1e-07
 5: -1.0000e+00 -1.0000e+00  2e-09  7e-17  3e-09  1e-09
Optimal solution found.
alpha [9.99999999e-01 8.78634363e-10]
beta 0.9999999982427316

```

In CVXPY v1.1.6

```
Traceback (most recent call last):
  File "break_cvxpy.py", line 24, in <module>
    prob.solve(solver = 'CVXOPT', verbose =True, warm_start = True)
[...]

PR #1250, opened 2021-02-27
Fix CVXOPT's handling of degenerate problems

rileyjmurray:
For #1147. The case-check at line 121 was meant to identify when a problem had no conic constraints. However, it actually identified when a problem had more linear equations than variables (in which case the constraints are either redundant or the problem is trivially infeasible). This change should get us the behavior that was intended from the start.

Despite the simplicity of this change, this PR should not be merged until (1) all CI tests pass and (2) some reviewers approve (I'll request reviewers once tests pass).

Edit: I realize this PR was made on a branch of the main cvxpy repository, rather than my fork. That wasn't intended and I'll avoid that in the future. That said, I see no harm in continuing with this as-is.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py
@@ -118,7 +118,7 @@ class CVXOPT(ECOS):
         data[s.B] = b[:len_eq].flatten()
         if data[s.B].shape[0] == 0:
             data[s.B] = None
-        if len_eq > A.shape[1]:
+        if len_eq >= A.shape[0]:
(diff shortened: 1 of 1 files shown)


SteveDiamond:
The change makes sense, but seems to have caused failures on travis.

rileyjmurray:
I see. Apparently, my PyCharm configuration wasn't running ``test_constant_atoms.py``, since that file doesn't contain any ``unittest.TestCase`` class. The issue is that the optimization problems generated there often have only equality constraints when compiled into standard form (no conic constraints). I'm able to see these errors locally now, and I'll fix them soon.

rileyjmurray:
Will be resolved as part of #1427 (beyond that PR's initial scope).

PR #1261, opened 2021-03-03
Fix issue 1260

SteveDiamond:
This resolves 1260 by calling ``np.conj(P).T`` instead of ``P.H``, which only works for numpy matrices. I added a test based on issue 1260 as well.


rileyjmurray (approved):
Looks good to me! I'd merge in once tests pass.

PR #1264, opened 2021-03-04
Fix bug in C-style reshape

akshayka:
This change fixes a bug in the canonicalization of C-style reshape.

Previously a transpose was omitted when reshaping matrices.

This change also removes a stray `print` statement.

See the change for an example of a `reshape` that previously did not work.


SteveDiamond (approved):
LGTM

Issue #1177, opened 2020-11-16
gurobi time limit might cause cvxpy key error exception


sjiekak:
**Describe the bug**

cvxpy fails with an unexpected `KeyError: 'primal'` exception when using gurobi conic solver finishes without finding a solution

**To Reproduce**

Unfortunately I can't provide the code right now but here are the steps: 
- Submit a problem which takes at least T seconds to gurobi to find a solution
- Set a time limit lower than T. ` prob.solve(solver="gurobi", timeLimit=T-1, verbose=true)`
- gurobi returns with status [TIME_LIMIT](https://www.gurobi.com/documentation/9.0/refman/optimization_status_codes.html#sec:StatusCodes)
- gurobi doesn't set the solution variable X in the model
- cvxpy try to access the gurobi solution variable : `solution["primal"] = np.array([v.X for v in variables])`  at https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py#L229
- the error `AttributeError : ("Unable to retrieve attribute 'X'",)` is raised, and caught silently at https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py#L242
- now the key "primal" is not set in the solution
- later in invert_data, cvxpy try to access  "primal" https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py#L113
- Fails with KeyError: 'primal'

**Expected behavior**

cvxpy should return with either a solver error or a user limit, whatever you deem more accurate

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

```
  File "/tmp/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 396, in solve
    return solve_func(self, *args, **kwargs)
  File "/tmp/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 754, in _solve
    self.unpack_results(solution, solving_chain, inverse_data)
  File "/tmp/lib/python3.8/site-packages/cvxpy/problems/problem.py", line 1058, in unpack_results
    solution = chain.invert(solution, inverse_data)
  File "/tmp/lib/python3.8/site-packages/cvxpy/reductions/chain.py", line 79, in invert
    solution = r.invert(solution, inv)
[...]

PR #1270, opened 2021-03-11
Gurobi: fix for when time limit is hit with no solution

ashamraeva:
A fix for #1177, following our discussion.

Here is a brief summary.

### Context:

Sometimes Gurobi hits the time limit before finding a solution or proving that the problem is infeasible or unbounded. Currently cvxpy raises an unexpected exception in this case, as it tries to unpack a solution.

### What we want

1. that cvxpy stops raising an unexpected exception,
2. to be able to identify this case so that the user could retry with a higher time limit.

### The simplest way to achieve the above given current code

1. map this case to `INFEASIBLE_INACCURATE` status (and not to `USER_LIMIT`, as it is in `SOLUTION_PRESENT` list, and then tries to unpack a solution at each given opportunity resulting in another unexpected exception),
2. pass `attr` with details from the solver to `failure_solution`.

### Result

1. cvxpy returns successfully with status `INFEASIBLE_INACCURATE`,
2. the solver status is accessible via `solver_stats` -> `extra_stats`, so that one can check if Gurobi raised the `TIME_LIMIT` status and act upon it.

### Tests

I have added a test to capture the expected behavior described above:

1. build a model which triggers our case,
2. test if no exception is raised,
3. test if Gurobi status is accessible via `solver_stats` -> `extra_stats`,
4. if Gurobi changes its algorithms so that a solution is found or the termination reason is different, skip the test as it's not relevant anymore.

The test might be a bit clumsy, but I figured it's better than nothing... 

Looking forward to your thoughts!

### DIFF
--- a/cvxpy/reductions/solution.py
+++ b/cvxpy/reductions/solution.py
@@ -18,7 +18,7 @@ import numpy as np
 import cvxpy.settings as s
 
 
-def failure_solution(status):
+def failure_solution(status, attr=None):
     """Factory function for infeasible or unbounded solutions.
 
     Parameters
@@ -37,7 +37,9 @@ def failure_solution(status):
         opt_val = -np.inf
     else:
         opt_val = None
-    return Solution(status, opt_val, {}, {}, {})
+    if attr is None:
+        attr = {}
+    return Solution(status, opt_val, {}, {}, attr)
 
 
 class Solution(object):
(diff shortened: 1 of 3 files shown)


SteveDiamond (approved):
I think this is good. You use a lot of test features I wasn't aware of :)

Issue #1176, opened 2020-11-16
Unclear exception


adishavit:
I'm getting this error with Gurobi: **`AttributeError: Unable to retrieve attribute 'X'`**  

The full stack:
```
Traceback (most recent call last):
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\IPython\core\interactiveshell.py", line 3418, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-948b20075a95>", line 1, in <module>
    prob.solve(solver=solver, verbose=0, TimeLimit=time_limit)
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\problems\problem.py", line 396, in solve
    return solve_func(self, *args, **kwargs)
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\problems\problem.py", line 754, in _solve
    self.unpack_results(solution, solving_chain, inverse_data)
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\problems\problem.py", line 1058, in unpack_results
    solution = chain.invert(solution, inverse_data)
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\reductions\chain.py", line 79, in invert
    solution = r.invert(solution, inv)
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\reductions\solvers\qp_solvers\gurobi_qpif.py", line 67, in invert
    x = np.array([x_grb[i].X for i in range(n)])
  File "C:\Users\Public\Miniconda3\envs\...\lib\site-packages\cvxpy\reductions\solvers\qp_solvers\gurobi_qpif.py", line 67, in <listcomp>
    x = np.array([x_grb[i].X for i in range(n)])
  File "var.pxi", line 91, in gurobipy.Var.__getattr__
  File "var.pxi", line 119, in gurobipy.Var.getAttr
  File "attrutil.pxi", line 100, in gurobipy.__getattr
AttributeError: Unable to retrieve attribute 'X'
```
The problem solves fine or returns infeasible with slightly different constraints/constants.  
What does this mean?

Versions:
```
cvxopt                    1.2.5                    pypi_0    pypi
cvxpy                     1.1.7            py37h1834ac0_0    conda-forge
cvxpy-base                1.1.7            py37h414f9d2_0    conda-forge
[...]

PR #1272, opened 2021-03-12
Gurobi: fix for when time limit is hit with no solution (QP case)

ashamraeva:
Similar to my recent fix in #1270 for [gurobi_conif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py), we have encountered the same issue in [gurobi_qupif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py) this time. It is also documented in issue #1176. This PR should fix it.

The changes are **identical** to the one in #1270 **up to the following differences**:

1. the scope of the changes is [gurobi_qupif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py) instead of [gurobi_conif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py),
2. there was no need to explicitly pass `attr` with details from the solver, as it is already passed,
3. for the test, I had to slightly change the model to trigger the [gurobi_qupif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py) case instead of [gurobi_conif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py),
4. as to the location of the test, I have appended it to the [test_qp_solvers.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/tests/test_qp_solvers.py) suite.

Thanks for your thoughts!

### DIFF
--- a/cvxpy/tests/test_qp_solvers.py
+++ b/cvxpy/tests/test_qp_solvers.py
@@ -499,3 +499,43 @@ class TestQp(BaseTest):
         prob = Problem(obj)
         prob.solve()
         self.assertAlmostEqual(obj.value, 1.0)
+
+    def test_gurobi_time_limit_no_solution(self):
+        """Make sure that if Gurobi terminates due to a time limit before finding a solution:
+            1) no error is raised,
+            2) solver stats are returned.
+            The test is skipped if something changes on Gurobi's side so that:
+            - a solution is found despite a time limit of zero,
+            - a different termination criteria is hit first.
+        """
+        from cvxpy import GUROBI
+        if GUROBI in INSTALLED_SOLVERS:
+            import gurobipy
+            objective = Minimize(self.x[0])
+            constraints = [self.x[0] >= 1]
+            prob = Problem(objective, constraints)
+            try:
+                prob.solve(solver=GUROBI, TimeLimit=0.0)
+            except Exception as e:
+                self.fail("An exception %s is raised instead of returning a result." % e)
+
+            extra_stats = None
+            solver_stats = getattr(prob, "solver_stats", None)
+            if solver_stats:
+                extra_stats = getattr(solver_stats, "extra_stats", None)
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_qp_solvers.py
```diff
@@ -499,3 +499,43 @@ def test_square_param(self):
...
+            - a different termination criteria is hit first.
+        """
+        from cvxpy import GUROBI
+        if GUROBI in INSTALLED_SOLVERS:
+            import gurobipy
+            objective = Minimize(self.x[0])
+            constraints = [self.x[0] >= 1]
+            prob = Problem(objective, constraints)
+            try:
+                prob.solve(solver=GUROBI, TimeLimit=0.0)
```
SteveDiamond, line 518: Isn't this the same problem as in the previous test you wrote? So was that previous test using the gurobi_qpif.py? In that case you should add a SOC constraint or something to make it not a QP.

ashamraeva: The difference between the tests is at the `constraints` level:

- in the previous test:  `constraints = [cp.square(self.x[0]) <= 1]`,
- in this test: `constraints = [self.x[0] >= 1]`.

I checked by debugging the tests, that the 1st case indeed passes by gurobi_conif.py and the 2nd passes by gurobi_qupif.py :blush: And that they indeed fail without their respective changes and pass with them.

SteveDiamond: Thanks!

SteveDiamond (approved):
LGTM

PR #1276, opened 2021-03-12
Faster GurobiQP

rostyboost:
This PR:
1) makes use of `addMVar` to vectorize the creation of variables
2) implements a fast seeding mechanism + caching for DPP problems
**Currently**: recreate a `model` object from scratch and set the `Start` attribute on each variable sequentially from the last model solution (warm start, no way to specify starting point manually).
**This PR**: re-use the `model` object without recreating it and seed it in a vectorized fashion (for variables as well as constraints) from the `value` attributes of the leafs. This allows for specifying an arbitrary seed from the top-level `cp.Variable` objects prior to any model run, but also makes the `model` itself cached in the case where the problem is DPP. However there is no "automatic" warm-start behavior anymore.

Since this PR drops support for Gurobi < 9 and changes the warm start behavior, I'd like to gauge the appetite of the maintainers before investing more time into testing+backward compatibility considerations.

Internal benchmark on small/medium-size MIPs show significant performance improvements.


akshayka:
@gglockner , would you or someone else at Gurobi like to provide input on this PR?

@bstellato , I've added you as a reviewer for this PR, since you recently merged in a PR modifying the warm-start behavior of the Gurobi interface.

A high-level comment: I'm not sure how many of our users rely on Gurobi < 9. If some do, one option is to check the version of the gurobipy package in the interface, and fall-back to the old implementation if the version is < 9.

akshayka:
Pinging @gglockner once more. A change like this would likely make Gurobi a lot more useful to many people

gglockner:
I'm interested in looking at this but it's going to be a couple of weeks until I get time. Perhaps @rluce might be able to look at it sooner.

rluce:
Give me a few days, I'll go over the changes.

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -112,114 +136,35 @@ def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=No
...
         else:
-            obj = grb.QuadExpr()
-            if P.count_nonzero():  # If there are any nonzero elms in P
-                P = P.tocoo()
-                obj.addTerms(0.5*P.data, vars=list(x[P.row]),
-                             vars2=list(x[P.col]))
-            obj.add(grb.LinExpr(q, x))  # Add linear part
-            model.setObjective(obj)  # Set objective
-        model.update()
+            model = solver_cache['model']
```
rluce, line 156: So if the model still sits in the cache, this pathh assumes that the all data changes are confied to b, g, P and q.  Is it guaranteeed that F and A haven't changed?

rluce:
A few thoughts on the performance impact mentioned above:
- Can you post a (maybe contrieved?) example that shows the warm start performance improvement?
- What is contributing more to the improved performance:  The (now) skipped rebuilding of the model, or the different pathway to set Start values for the variables?
- In most cases it is not necessary to use MVar objects for good model building performance (in contrast to using `addMConstr` and friends.)  The fact that you set the `Start` attribute in a vectorized syntax is only _a little_ more than syntactic sugar.  That said, you should get almost the same performance by using patterns like
```
x = model.addVars(3)
[...]
s = np.array([1,2,3])
model.setAttr('Start', x, s)
```
In my experience there will be pleny of users on Gurobi <9, so if you could check the performance without using MVars that would be great.

> However there is no "automatic" warm-start behavior anymore.
I'm not familiar enough with the current warm starting behaviour to see all the implications.   Could you explain what the direct implication for existing code would be?

SteveDiamond:
@rostyboost it would be great if you could look at this again when you get a chance. I'm pushing for a new release and trying to get a lot of these PRs in there.

rostyboost:
> * Can you post a (maybe contrieved?) example that shows the warm start performance improvement?

Yes. I will work on publishing a benchmark here.

> * What is contributing more to the improved performance:  The (now) skipped rebuilding of the model, or the different pathway to set Start values for the variables?

Avoiding the rebuild of the model is by far the major contributing factor to the improved performance.

> In my experience there will be pleny of users on Gurobi <9, so if you could check the performance without using MVars that would be great.

Yes, your intuition is correct: the use of `MVars` only brings minor performance gains. However it arguably makes the code substantially simpler to read. But if there are still plenty of Gurobi cvxpy users on < 9 , it makes sense to keep compatibility with older versions.

> I'm not familiar enough with the current warm starting behaviour to see all the implications. Could you explain what the direct implication for existing code would be?

I am not sure how widely used current warm-start mechanism is. The direct implication of current PR proposal means that without manually seeding the solver with a solution from a previous run (or known by other means), there won't be a warm start happening. There might be a way to not break the current behavior by detecting when no user-specified seed have been provided and to re-use the last solution as a seed in such cases. I'll think about it.

rileyjmurray:
@rostyboost would you like to keep this PR open? Just comment here if so and we can get the ball rolling. I think that overall I'd like to retain our current warm-start behavior and support Gurobi v8, but your idea of caching the Gurobi model is valuable.

rileyjmurray:
Closing due to inactivity. Happy to re-open if someone would like to pick up where things left off!

Issue #1263, opened 2021-03-04
Problem with solution in CVXPY in v1.0.31 returns unbounded in v1.1.0


kmonson:
1.0.31 produces a solution and 1.1.0 reports the problem as unbounded. (It produces the same result with any of the 1.1.* versions.)

It looks like it's generating very different problems for the solver when I change versions.

A and G are different dimensions, dims is different, h is different, and the inverse data is very different (only start shown). c is identical.

Version 1.0.31:
```
[{'A': <8761x35040 sparse matrix of type '<class 'numpy.float64'>'
	with 35040 stored elements in Compressed Sparse Column format>,
  'G': <123005x35040 sparse matrix of type '<class 'numpy.float64'>'
	with 175198 stored elements in Compressed Sparse Column format>,
  'b': array([-0., -0., -0., ..., -0., -0., -0.]),
  'c': array([-36.73, -36.47, -36.57, ...,   0.  ,   0.  ,   0.  ]),
  'dims': (zero: 8761, nonpos: 123005, exp: 0, soc: [], psd: []),
  'h': array([ -0.,  -0.,  -0., ..., 800., 800., 800.]),
  'offset': array([0.])},
 Chain(reductions=[<cvxpy.reductions.dcp2cone.dcp2cone.Dcp2Cone object at 0x12c3953d0>, <cvxpy.reductions.cvx_attr2constr.CvxAttr2Constr object at 0x12c395580>, <cvxpy.reductions.eval_params.EvalParams object at 0x12c395c70>, <cvxpy.reductions.dcp2cone.cone_matrix_stuffing.ConeMatrixStuffing object at 0x12c3917f0>, <cvxpy.reductions.solvers.conic_solvers.ecos_conif.ECOS object at 0x12b0c8910>]),
 [<cvxpy.reductions.inverse_data.InverseData object at 0x12c391190>,
  [{0: ('var7186', Variable((8760,)), None),
    1: ('ene', Variable((8760,)), None),
    2: ('dis', Variable((8760,)), None),
    3: ('ch', Variable((8760,)), None)},
   {0: ('pv_out', Variable((8760,), nonneg=True), None),
    1: ('ene', Variable((8760,)), None),
    2: ('dis', Variable((8760,)), None),
    3: ('ch', Variable((8760,)), None)},
   {30: 30,
    34: 34,
    38: 38,
```

Version 1.1.0:
```
[...]

PR #1277, opened 2021-03-13
Fix NonPos

SteveDiamond:
NonPos is not working correctly in the current master. This change fixes the problem.


SteveDiamond:
Addresses #1263

akshayka (approved):
LGTM

PR #1278, opened 2021-03-14
Update problem.py

tschm:
Adding variable and parameter dict

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -356,6 +356,14 @@ class Problem(u.Canonical):
         """
         return self._solver_stats
 
+    @property
+    def parameters(self):
+        return {parameter.name(): parameter for parameter in self.problem.parameters()}
+
+    @property
+    def variables(self):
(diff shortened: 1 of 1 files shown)


cvxpy/problems/problem.py
```diff
@@ -356,6 +356,14 @@ def solver_stats(self):
         """
         return self._solver_stats
 
+    @property
+    def parameters(self):
+        return {parameter.name(): parameter for parameter in self.problem.parameters()}
```
SteveDiamond, line 361: Could you change these methods to "param_dict" and "var_dict"? We already have parameters and variables methods, and we can't break that interface.

SteveDiamond:
Thanks for doing a PR! We can walk you through the process, don't worry :) You will need to add tests though. Just make a new commit to the fork you're basing the PR from.

tschm:
Sorry, I can't open appveyor because of an outdated security configuration on your side?

SteveDiamond:
I'm closing this since you moved things to a different PR.

PR #1284, opened 2021-03-15
Minor edit to verbose compilation output

akshayka:
Only print that compilations will take less time when parameters are involved


SteveDiamond (approved):
LGTM

PR #1295, opened 2021-03-18
update python requires and readme

mar-heaven:
cvxpy/problems/problems.py lines 57 new syntax used requires python >=3.6


SteveDiamond:
@akshayka this relates to the new compilation message. Maybe we should change the format string?

SteveDiamond:
Resolved by removing the f strings.

PR #1300, opened 2021-03-20
Set prob.value to prob.objective.value

SteveDiamond:
This PR sets prob.value to prob.objective.value when a solution is present. I had to fix a number of .value functions for individual atoms. The solutions I chose were not always the most elegant, so I'm open to improvements.

### DIFF
--- a/cvxpy/atoms/elementwise/ceil.py
+++ b/cvxpy/atoms/elementwise/ceil.py
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import cvxpy.settings as s
 from cvxpy.atoms.elementwise.elementwise import Elementwise
 import numpy as np
 import scipy.sparse as sp
@@ -26,7 +27,8 @@ class ceil(Elementwise):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        return np.ceil(np.around(values[0], decimals=4))
+        decimals = np.int(np.abs(np.log10(s.ATOM_EVAL_TOL)))
+        return np.ceil(np.around(values[0], decimals=decimals))
 
     def sign_from_args(self):
         """Returns sign (is positive, is negative) of the expression.
--- a/cvxpy/atoms/elementwise/ceil.py
+++ b/cvxpy/atoms/elementwise/ceil.py
@@ -26,7 +26,7 @@ class ceil(Elementwise):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        return np.ceil(values[0])
+        return np.ceil(np.around(values[0], decimals=4))
 
     def sign_from_args(self):
         """Returns sign (is positive, is negative) of the expression.
--- a/cvxpy/atoms/lambda_sum_largest.py
+++ b/cvxpy/atoms/lambda_sum_largest.py
@@ -44,7 +44,7 @@ class lambda_sum_largest(lambda_max):
         Requires that A be symmetric.
         """
         eigs = LA.eigvals(values[0])
-        return sum_largest(eigs, self.k).value
+        return sum_largest(np.real(eigs), self.k).value
 
     def get_data(self):
         """Returns the parameter k.
--- a/cvxpy/atoms/lambda_sum_largest.py
+++ b/cvxpy/atoms/lambda_sum_largest.py
@@ -43,8 +43,8 @@ class lambda_sum_largest(lambda_max):
 
         Requires that A be symmetric.
         """
-        eigs = LA.eigvals(values[0])
-        return sum_largest(np.real(eigs), self.k).value
+        eigs = LA.eigvalsh(values[0])
+        return sum_largest(eigs, self.k).value
 
     def get_data(self):
         """Returns the parameter k.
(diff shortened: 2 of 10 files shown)


cvxpy/atoms/elementwise/ceil.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, x):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        return np.ceil(values[0])
+        return np.ceil(np.around(values[0], decimals=4))
```
rileyjmurray, line 29: Is this change to make the ceil atom reflect numerical precision issues in solutions? Like, since we consider ``x.value == 3.00000001`` to more or less be feasible for ``x <= 3``, we want ``cp.ceil(expr).value`` to evaluate to 3 when ``expr.value == 3.0000001``?

If so, I think it's a pragmatic approach. The only improvement I'd suggest is to set the decimals as a named constant somewhere. That way if a user has some sensitive application they can do something like ``decimals=8``. We store other such constants in https://github.com/cvxgrp/cvxpy/blob/66a437daed434a486ccd0faa1786b97d400bd57c/cvxpy/settings.py#L166.

cvxpy/atoms/lambda_sum_largest.py
```diff
@@ -44,7 +44,7 @@ def numeric(self, values):
         Requires that A be symmetric.
         """
         eigs = LA.eigvals(values[0])
-        return sum_largest(eigs, self.k).value
+        return sum_largest(np.real(eigs), self.k).value
```
rileyjmurray, line 47: Rather than taking the real part of the eigenvalues, we should use ``LA.eigvalsh`` ("h" for "Hermitian"). For symmetric matrices eigvalsh is faster and more reliable. We use it here: https://github.com/cvxgrp/cvxpy/blob/c88f09215252c932c101c842dd684a60c16ea4cf/cvxpy/atoms/lambda_max.py.

It would also be good to update this atom: https://github.com/cvxgrp/cvxpy/blob/c88f09215252c932c101c842dd684a60c16ea4cf/cvxpy/atoms/log_det.py to use eigvalsh.

cvxpy/atoms/log_det.py
```diff
@@ -35,7 +35,7 @@ def numeric(self, values):
         and is equivalent to the nuclear norm of the matrix logarithm of A.
         """
         sign, logdet = LA.slogdet(values[0])
-        if sign == 1:
+        if np.abs(sign) == 1:
```
rileyjmurray, line 38: If the domain of this atom is the set of positive definite matrices, then it seems to me that the current value function is correct as-is.

From the numpy documentation "In all cases, the determinant is equal to sign * np.exp(logdet)." So sign == -1 would imply A is not positive definite.

cvxpy/atoms/log_det.py
```diff
@@ -104,7 +104,10 @@ def _domain(self):
 
     @property
     def value(self):
-        if not np.allclose(self.args[0].value, self.args[0].value.T.conj()):
+        if not np.allclose(self.args[0].value,
```
rileyjmurray, line 107: I think we should use tolerances defined explicitly in ``cvxpy/settings.py``. (This might require creating a new tolerance.)

cvxpy/atoms/elementwise/ceil.py
SteveDiamond: That's right. There were cases in the tests where the optimal value obtained by the "solver" would be say 3, but the objective would be ``ceil(x)`` where ``x.value = 3.00001``.

SteveDiamond: I agree with the idea of having a global tolerance setting. I would like it to be just one value, so we would need to convert into decimals.

cvxpy/atoms/lambda_sum_largest.py
SteveDiamond: Fixed.

cvxpy/atoms/log_det.py
SteveDiamond: The problem is values[0] might be constrained to be positive definite, but numerically it might not be positive definite.

cvxpy/atoms/log_det.py
SteveDiamond: Fixed.

cvxpy/atoms/elementwise/ceil.py
SteveDiamond: Fixed.

cvxpy/atoms/log_det.py
rileyjmurray: In that case, I think the thing to do would be if ``sign != 1``, then re-compute slogdet with ``values[0] + PSD_TOL**2 * np.eye(n)``, where ``PSD_TOL`` is whatever tolerance we use to decide when a matrix is positive semidefinite. If we get ``sign != 1`` at that point, then returning ``+inf`` would be appropriate. I realize this is some funny business but I think funny business is preferable to crazy behavior like allowing large negative eigenvalues.

SteveDiamond: Sorry, what I actually meant is that values[0] might not be numerically symmetric. I've pushed a new fix, which applies log_det to the symmetric part. It goes back to checking if ``sign == 1``.

rileyjmurray (approved):
Looks good to me!

PR #1309, opened 2021-03-24
Add USER_LIMIT status to init so that it can be accessed with cvxpy.USER_LIMIT

### LINKED ISSUE #1308: USER_LIMIT status cannot be accessed like the others
**Describe the bug**
When using cvxpy, one can access its statuses via `cvxpy.OPTIMAL_INACCURATE` for example. However, when it comes to the `USER_LIMIT` status, calling `cvxpy.USER_LIMIT` gives the following message:

`Cannot find reference 'USER_LIMIT' in '__init__.py' `

**To Reproduce**
Just call `cvxpy.USER_LIMIT` similarly to `cvxpy.OPTIMAL_INACCURATE`.

**Expected behavior**
The behavior should be the same as for the other statuses (`cvxpy.USER_LIMIT` should be accessible).

**Version**
 - OS: ubuntu 20.04
 - CVXPY Version: 1.1.11

**Additional context**
I believe the only change needed is adding `USER_LIMIT` to [\_\_init\_\_.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/__init__.py) at lines 27-30:
```
from cvxpy.settings import (CVXOPT, GLPK, GLPK_MI, CBC, CPLEX, OSQP, NAG,
                            ECOS, SCS, DIFFCP, GUROBI, MOSEK, XPRESS, SCIP, ECOS_BB,
                            OPTIMAL, UNBOUNDED, INFEASIBLE, SOLVER_ERROR, ROBUST_KKTSOLVER,
                            OPTIMAL_INACCURATE, UNBOUNDED_INACCURATE, INFEASIBLE_INACCURATE)
```

Please let me know if my understanding is correct and I could submit a PR.

ashamraeva:
Fixes #1308.

I did not add any tests, as I could not find any similar ones (but I did check locally that it indeed works :blush: ). Let me know if it's okay.

### DIFF
--- a/cvxpy/__init__.py
+++ b/cvxpy/__init__.py
@@ -27,7 +27,7 @@ from cvxpy.error import DCPError, DPPError, DGPError, SolverError, disable_warni
 from cvxpy.settings import (CVXOPT, GLPK, GLPK_MI, CBC, CPLEX, OSQP, NAG,
                             ECOS, SCS, DIFFCP, GUROBI, MOSEK, XPRESS, SCIP, ECOS_BB,
                             OPTIMAL, UNBOUNDED, INFEASIBLE, SOLVER_ERROR, ROBUST_KKTSOLVER,
-                            OPTIMAL_INACCURATE, UNBOUNDED_INACCURATE, INFEASIBLE_INACCURATE)
+                            OPTIMAL_INACCURATE, UNBOUNDED_INACCURATE, INFEASIBLE_INACCURATE, USER_LIMIT)
 from cvxpy.settings import get_num_threads, set_num_threads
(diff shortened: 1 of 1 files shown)


ashamraeva:
Hm, not sure why tests are failing (all pass locally). Would you be so kind to have a look? Thank you!

SteveDiamond:
The test failed because it couldn't download one of the solvers for some reason. Anyway, the PR looks good, I'm going to merge it. Thanks!

SteveDiamond:
The CI always breaks for some dumb reason. It takes a lot of time to maintain the system.

ashamraeva:
Good to know, thanks!

PR #1310, opened 2021-03-28
raise NotImplementedError instead of return NotImplemented

r-barnes:


r-barnes:
Test failures appear to be flakes?

cvxpy/interface/scipy_wrapper.py
```diff
@@ -26,7 +26,7 @@ def wrap_bin_op(method):
     """
     def new_method(self, other):
         if isinstance(other, exp.Expression):
-            return NotImplemented
```
SteveDiamond, line 29: This one you can't change. It helps with operator overloading.

SteveDiamond:
Yeah sorry about the confusing tests. We're in the process of revamping our CI. The AppVeyor test is real. See my comment.

cvxpy/interface/scipy_wrapper.py
r-barnes: Thanks; reverted.

SteveDiamond:
LGTM. Thanks for doing this!

PR #1311, opened 2021-03-29
Switch to using safe default arguments

r-barnes:
When `[]` or `{}` or any other mutable type is used as a default argument, this can have unexpected results as discussed [here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) and [here](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) for reasons explained [here](https://web.archive.org/web/20200221224620/http://effbot.org/zone/default-values.htm).

Rather than relying on programmer discipline not to abuse these default arguments, it is generally considered best practice to default value the arguments to `None` and check this within the function body.

This PR fixes all the instances I found, excluding the `doc/` subdir, which I assume consists largely of upstream code.

### DIFF
--- a/cvxpy/atoms/affine/add_expr.py
+++ b/cvxpy/atoms/affine/add_expr.py
@@ -81,7 +81,7 @@ class AddExpression(AffAtom):
 
     # As __init__ takes in the arg_groups instead of args, we need a special
     # copy() function.
-    def copy(self, args=None, id_objects={}):
+    def copy(self, args=None, id_objects=None):
         """Returns a shallow copy of the AddExpression atom.
 
         Parameters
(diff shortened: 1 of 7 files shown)


rileyjmurray:
@r-barnes thanks for catching this! I've run into this issue on other projects and also use the default-None approach. I'll merge this in once CI tests pass.

rileyjmurray:
@r-barnes can you resolve the merge conflicts on your end?

r-barnes:
@rileyjmurray : Done.

SteveDiamond (approved):
LGTM

SteveDiamond:
Thanks for doing this!

PR #1312, opened 2021-03-29
Use lazy interpolation for logging

r-barnes:
For performance reasons it's usually best to avoid performing string interpolation for loggers. If the logger is not emitting messages, then the interpolation is not performed and time is saved. If the logger is emitting, the logger functions handle the interpolation.


SteveDiamond:
LGTM.

PR #1313, opened 2021-03-29
Ensure variables are initialized appropriately

r-barnes:
Trying to reduce situations where attributes are defined outside of `__init__` resulting in unstable class interfaces


SteveDiamond (approved):
LGTM

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -178,10 +178,10 @@ def apply_parameters(self, id_to_param_value=None, zero_offset=False,
                     canonInterface.reduce_problem_data_tensor(
                         self.A, self.x.size))
                 self.reduced_A = reduced_A
-                self.problem_data_index = (indices, indptr, shape)
```
SteveDiamond, line 181: This change seems to break the tests.

SteveDiamond:
I merged this in manually.

PR #1316, opened 2021-03-29
Classes do not need to inherit from object in Python3

r-barnes:
There are a number of instances where the code says:
```
class foo(object):
```
this PR switches them all to
```
class foo:
```
As explained [here](https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object), the first form was necessary for Python 2, but is not needed in Python 3. Since CVXPY requires Python >=3.5, this pattern is no longer necessary.


SteveDiamond:
Thanks! LGTM.

PR #1319, opened 2021-03-29
Add some more type annotations

r-barnes:
Adds additional type annotations

### DIFF
--- a/cvxpy/interface/matrix_utilities.py
+++ b/cvxpy/interface/matrix_utilities.py
@@ -165,7 +165,7 @@ def from_1D_to_2D(constant):
         return constant
 
 
-def convert(constant, sparse=False, convert_scalars=False):
+def convert(constant, sparse=False, convert_scalars: bool = False):
     """Convert to appropriate type.
     """
     if isinstance(constant, (list, np.matrix)):
@@ -222,7 +222,7 @@ def sign(constant):
     return (min_val >= 0, max_val <= 0)
 
 
-def is_complex(constant, tol=1e-5) -> bool:
+def is_complex(constant, tol: float = 1e-5) -> bool:
     """Return (is real, is imaginary).
 
     Parameters
@@ -287,7 +287,7 @@ def is_hermitian(constant) -> bool:
     return is_symm, is_herm
 
 
-def is_sparse_symmetric(m, complex=False) -> bool:
+def is_sparse_symmetric(m, complex: bool = False) -> bool:
     """Check if a sparse matrix is symmetric
 
     Parameters
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -472,7 +472,7 @@ class Problem(u.Canonical):
         """
         cls.REGISTERED_SOLVE_METHODS[name] = func
 
-    def get_problem_data(self, solver, gp=False, enforce_dpp=False, verbose=False):
+    def get_problem_data(self, solver, gp: bool = False, enforce_dpp: bool = False, verbose: bool = False):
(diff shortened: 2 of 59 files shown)


r-barnes:
This PR makes the value of type annotations slightly more apparent. For instance, consider:
```
    def __div__(self, other: Union[int, float]) -> float:
        if not isinstance(other, (int, float)):
            raise NotImplementedError()
```
The code currently defensively checks the input type, but can only happen at runtime. The type annotation now provides a way for users to find the problem preemptively via static analysis.

r-barnes:
Fixed merge conflicts.

SteveDiamond:
This seems to break the build for some reason, related to imports.

SteveDiamond:
Also we're having some flake8 issues on our Travis CI (for some reason Travis is not showing up in PR checks). The main issue is "E252 missing whitespace around parameter equals", which seems to be from the new type signatures.

r-barnes:
@SteveDiamond : Do you use any sort of auto-formatter? e.g. Black or AutoPep8?

SteveDiamond:
I use black in all my other projects. (I didn't know about it when I started cvxpy.) I'll discuss with the other devs, but I could do a PR to start using black, and that would resolve this issue (and only use flake8 in CI for syntax errors).

r-barnes:
I can fix this by hand if you prefer not convert, but I thought I'd ask before going to that work.

SteveDiamond:
> I can fix this by hand if you prefer not convert, but I thought I'd ask before going to that work.

I don't want to put you through that trouble if it can be avoided. We'll figure out if we can switch to black in a day or two.

SteveDiamond:
Hi @r-barnes , after internal discussion we decided that we need more time to consider a change to black. For the moment I fixed the flake8 issues in the current build manually (#1327 ).

r-barnes:
Thanks @SteveDiamond . I have flake8 and testing working locally now, so hopefully I can avoid a repeat of that, and hopefully the CI will get back online.

I've backed out the previous version of this PR in favour of one that concentrates on adding boolean typing. Circular dependencies were causing issues with the previous version. I'll come back to it later.

r-barnes:
Added integer and floating point annotations too.

r-barnes:
Just checking in on this.

cvxpy/interface/matrix_utilities.py
```diff
@@ -165,7 +165,7 @@ def from_1D_to_2D(constant):
         return constant
 
 
-def convert(constant, sparse=False, convert_scalars=False):
+def convert(constant, sparse=False, convert_scalars: bool = False):
```
SteveDiamond, line 168: Add annotation for "sparse"?

cvxpy/problems/problem.py
```diff
@@ -822,8 +822,8 @@ def _invalidate_cache(self) -> None:
     def _solve(self,
                solver=None,
                warm_start=True,
-               verbose=False,
-               gp=False, qcp=False, requires_grad=False, enforce_dpp=False, **kwargs):
+               verbose: bool = False,
+               gp: bool = False, qcp: bool = False, requires_grad: bool = False, enforce_dpp: bool = False, **kwargs):
```
SteveDiamond, line 826: This line looks a bit long.

SteveDiamond:
@r-barnes sorry for the slow response! I have a couple tiny comments, otherwise LGTM.

SteveDiamond:
@r-barnes I went ahead and merged this in. Thanks again for all your contributions!

r-barnes:
Oh, sorry, this slipped through the cracks on my end.

PR #1323, opened 2021-03-30
Populate solver_stats for ECOS_BB, ECOS & SCS

### LINKED ISSUE #1322: solver_stats are empty when using ECOS_BB
**Describe the bug**
When using ECOS_BB with cvxpy, the `solver_stats` attribute of `problem` is empty, while for other solvers it usually contains some useful solver-specific optimization details.

**To Reproduce**
Just solve any problem with ECOS_BB and look at `problem.solver_stats`.

**Expected behavior**
To see some solver-specific info in `problem.solver_stats`, similar to other solvers.

**Version**
 - OS: ubuntu 20.04
 - CVXPY Version: 1.1.11

**Additional context**
I believe the only change needed here is to pass `attr` explicitly at lines 104-106 in [ecos_bb_conif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/ecos_bb_conif.py) instead of an empty dict (when a solution is present) or instead of None (when there is no solution):

```
        attr = {}
        attr[s.SOLVE_TIME] = solution["info"]["timing"]["tsolve"]
        attr[s.SETUP_TIME] = solution["info"]["timing"]["tsetup"]
        attr[s.NUM_ITERS] = solution["info"]["iter"]
        attr[s.EXTRA_STATS] = solution

        if status in s.SOLUTION_PRESENT:
            ...
            return Solution(status, opt_val, primal_vars, dual_vars, {})
        else:
            return failure_solution(status)
```

Let me know if my suggestion is fine and I will submit a PR.

ashamraeva:
Fixes #1322.

I have taken this opportunity to pass solver stats for ECOS and SCS as well when there is no solution (similar to [gurobi_conif.py](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py)).

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/ecos_bb_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/ecos_bb_conif.py
@@ -101,9 +101,9 @@ class ECOS_BB(ECOS):
                     inverse_data[self.EQ_CONSTR]
                 )
                 dual_vars.update(eq_duals)
-            return Solution(status, opt_val, primal_vars, dual_vars, {})
+            return Solution(status, opt_val, primal_vars, dual_vars, attr)
         else:
-            return failure_solution(status)
+            return failure_solution(status, attr)
 
     def solve_via_data(self, data, warm_start, verbose, solver_opts, solver_cache=None):
         import ecos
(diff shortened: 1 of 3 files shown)


SteveDiamond:
LGTM.

PR #1326, opened 2021-03-31
Suppress warning about unused variable

r-barnes:
Fixes
```
cvxpy/cvxcore/src/LinOpOperations.cpp: In function ‘Tensor get_rmul_mat(const LinOp&, int)’:
cvxpy/cvxcore/src/LinOpOperations.cpp:729:7: warning: variable ‘arg_cols’ set but not used [-Wunused-but-set-variable]
  729 |   int arg_cols;
      |       ^~~~~~~~
```
But you may want to just drop the `arg_cols` variable here.

### DIFF
--- a/cvxpy/cvxcore/src/LinOpOperations.cpp
+++ b/cvxpy/cvxcore/src/LinOpOperations.cpp
@@ -726,16 +726,12 @@ Tensor get_rmul_mat(const LinOp &lin, int arg_idx) {
                       : 1;
 
   // Interpret as row or column vector as needed.
-  int arg_cols;
   int result_rows;
   if (lin.get_args()[0]->get_shape().size() == 0) {
-    arg_cols = 1;
     result_rows = 1;
   } else if (lin.get_args()[0]->get_shape().size() == 1) {
-    arg_cols = lin.get_args()[0]->get_shape()[0];
     result_rows = 1;
   } else {
-    arg_cols = lin.get_args()[0]->get_shape()[1];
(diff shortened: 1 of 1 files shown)


cvxpy/cvxcore/src/LinOpOperations.cpp
```diff
@@ -738,6 +738,7 @@ Tensor get_rmul_mat(const LinOp &lin, int arg_idx) {
     arg_cols = lin.get_args()[0]->get_shape()[1];
     result_rows = lin.get_args()[0]->get_shape()[0];
   }
+  (void)arg_cols; //Suppress warning about variable that is set but not used
```
akshayka, line 741: If I understand correctly, `arg_cols` is actually not used? If so we should just delete `arg_cols`, instead of keeping it and suppressing the warning. (ie delete lines 729, 732, 735, 741).

akshayka:
> But you may want to just drop the `arg_cols` variable here.

Yes, let's just drop `arg_cols`

cvxpy/cvxcore/src/LinOpOperations.cpp
r-barnes: Okay, I've updated the PR to remove `arg_cols`.

r-barnes:
I've updated the PR to drop `arg_cols`.

r-barnes:
Bump

akshayka (approved):
LGTM

PR #1337, opened 2021-04-13
Disallow sparse constants in DGP

akshayka:
Sparse constants cannot be elementwise positive, since they generally have many entries which are zero. So in particular they shouldn't be used in DGP.

Without this change, the `is_pos` method fails with an error (a raised Python exception) on sparse constants.

In the future it might be possible to have an overloaded notion of sparsity for DGP, where the zero entries actually represent the value 1.


SteveDiamond (approved):
LGTM

PR #1359, opened 2021-05-03
Logistic atom - improved gradient calculation

adamw00000:
Connected to this question on Stack Overflow: https://stackoverflow.com/questions/67062485/overflow-while-using-logistic-function-in-cvxpy.

Current implementation of logistic atom overflows when one of the elements in the input vector exceeds ~710 due to np.exp invocation. Atom evaluation doesn't have this problem - it uses np.logaddexp to handle big values.

I've rewritten gradient implementation to utilize np.logaddexp call, which prevents overflow from happening. First we calculate the solution in the logarithmic scale, and only after we take an exponent of the resulting value.

### DIFF
--- a/cvxpy/atoms/elementwise/logistic.py
+++ b/cvxpy/atoms/elementwise/logistic.py
@@ -75,6 +75,5 @@ class logistic(Elementwise):
         """
         rows = self.args[0].size
         cols = self.size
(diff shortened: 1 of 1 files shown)


SteveDiamond:
LGTM! The test failure is unrelated.

PR #1364, opened 2021-05-09
Annotations for warm_start, verbose, and shape

r-barnes:


SteveDiamond:
FYI a couple builds are failing on travis globally because of ECOS BB issues.

rileyjmurray (approved):
Looks good to me! Although, it's worth noting that our second-order conic solvers don't support warm-start. So although the argument appears in ``apply(...)`` or ``solve_via_data(...)``, it can't really be used.

SteveDiamond (approved):
LGTM

PR #1368, opened 2021-05-13
Move Travis CI to GitHub Actions

phschiele:
Hi @SteveDiamond and @rileyjmurray,

as discussed, here is an initial PR to move over the Travis CI to GitHub Actions.
You mentioned that the scope of the CI should match Travis 100%, so I stayed close to the existing implementation and reused a large portion of the existing bash scripts. I'll leave some further comments below.

**Potential follow-ups** (would be great to get your opinion on):
- Adding a cache for the dependency installation
- Further simplify e.g. the deploy step using existing marketplace actions
- Move over windows CI as well
- Make the CI more robust using an advanced dependency manager (e.g. poetry or pipenv)

**Open questions**:
- ~There is still another Travis file for cvxcore. Is that still relevant?~ Removed

**Remaining ToDos:**
- [x] Activate tests again, I disabled them to move on with development here, as they are also failing in master (related to ECOS).
- [x] Only run deploy step for pushes to master
- [x] Add relevant secrets to the GitHub repo


.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
...
+    env:
+      RUNNER_OS: ${{ matrix.os }}
+      PYTHON_VERSION: ${{ matrix.python-version }}
+      COVERAGE: "False"
+      DEPLOY_PYPI_SOURCE: "${{ matrix.deploy_pypi_source == 'True' && 'True' || 'False' }}"
+      USE_OPENMP: "${{ matrix.openmp == 'True' && 'True' || 'False' }}"
+      PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
+      PYPI_USER: ${{ secrets.PYPI_USER }}
+      PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      PYPI_API_ENDPOINT: ${{ secrets.PYPI_API_ENDPOINT }}
```
phschiele, line 47: These secrets need to be defined in the repo settings.

.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          if [ "$RUNNER_OS" = "macOS" ]; then CC=gcc ; fi
+          if [ "$RUNNER_OS" = "macOS" ]; then CXX=g++; fi
+          source continuous_integration/GitHub/install_dependencies.sh
+      - name: test
+        run: |
+          flake8 .
+          bash continuous_integration/GitHub/test_script.sh
+      - name: coverage
+        run: if [[ "$COVERAGE" == "True" ]]; then coveralls; fi
```
phschiele, line 66: It seems that the coverage run is always `False`, but I added it to keep the possibility to enable it again.

continuous_integration/GitHub/install_dependencies.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# This script is meant to be called by the "install" step defined in
```
phschiele, line 2: This script was simplified quite a bit by using the existing marketplace action for setting up conda.

cvxpy/problems/problem.py
```diff
@@ -472,7 +472,8 @@ def register_solve(cls, name, func) -> None:
         """
         cls.REGISTERED_SOLVE_METHODS[name] = func
 
-    def get_problem_data(self, solver, gp: bool = False, enforce_dpp: bool = False, verbose: bool = False):
+    def get_problem_data(self, solver, gp: bool = False, enforce_dpp: bool = False,
+                         verbose: bool = False):
```
phschiele, line 476: Just changed this line to make flake8 pass.

.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
+name: build
+
+on:
+    pull_request:
+    push:
+        branches:
+            - master
+
+jobs:
+  cleanup-runs:
```
phschiele, line 10: Cancels previous GH actions workflows in your branch (if you commit again before the last tests finished).
Not executed for tag or master pushes.

.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+    - uses: rokroskar/workflow-run-cleanup-action@master
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
+  build:
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash -l {0}
```
phschiele, line 21: Required to have the conda environment activated in all bash scripts, see [use a default shell](https://github.com/conda-incubator/setup-miniconda#use-a-default-shell).

.github/workflows/build.yml
```diff
@@ -0,0 +1,68 @@
...
+          if [ "$RUNNER_OS" = "macOS" ]; then CXX=g++; fi
+          source continuous_integration/GitHub/install_dependencies.sh
+      - name: test
+        run: |
+          flake8 .
+          bash continuous_integration/GitHub/test_script.sh
+      - name: coverage
+        run: if [[ "$COVERAGE" == "True" ]]; then coveralls; fi
+      - name: deploy
+        run: bash continuous_integration/GitHub/deploy_script.sh
```
phschiele, line 68: If the deployment to takes place is currently based on `cvxpy.__version__` in the `__init__.py` file. So I assume one should only run this stage on master/tag pushes.

SteveDiamond: That's correct.

.github/workflows/build.yml
SteveDiamond: Thanks, yeah we probably want coverage again at some point.

SteveDiamond:
Thanks so much for doing this @phschiele! You can ignore the travis.yml for cvxcore (actually delete it). To enable the tests, just mark the failing tests as skipped. They're for ECOS BB, which has known bugs.

SteveDiamond:
Another nice thing to add would be CI going back to Python 3.5 (since in theory we support 3.5+).

.github/workflows/build.yml
phschiele: Done in 8208cb5

phschiele:
@SteveDiamond Thanks for your swift feedback, all mentioned comments are included now.

SteveDiamond:
Ok this generally looks great, I added a few secrets but need time/help to get all of them.

SteveDiamond:
@phschiele I believe I added all the secrets. Let me know if I missed one.

phschiele:
> @phschiele I believe I added all the secrets. Let me know if I missed one.

@SteveDiamond Great, we can test it by reverting the last commit (and reverting again if that test works). Should we give it a try?

rileyjmurray (approved):
Looks good to me! Thanks for all this work, @phschiele!

SteveDiamond:
I checked with Riley that all the secrets are there. I'm going to merge this in. Thanks again @phschiele !

PR #1375, opened 2021-05-18
More types

r-barnes:
Includes some alphabetizing of import lines.

### DIFF
--- a/cvxpy/atoms/affine/reshape.py
+++ b/cvxpy/atoms/affine/reshape.py
@@ -40,7 +40,7 @@ class reshape(AffAtom):
     order : F(ortran) or C
     """
 
-    def __init__(self, expr, shape: Tuple[int], order: str = 'F') -> None:
+    def __init__(self, expr, shape: Tuple[int, ...], order: str = 'F') -> None:
         if isinstance(shape, numbers.Integral):
             shape = (int(shape),)
         if len(shape) > 2:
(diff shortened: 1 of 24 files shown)


SteveDiamond:
Thanks! Some of the changes don't seem compatible with python 3.5.

rileyjmurray:
@SteveDiamond the Python Foundation dropped support for 3.5 a while ago. We had a conversation about dropping support for 3.5 the last time I updated Travis/AppVeyor. The remaining mentions of 3.5 support in the docs should probably be updated to reflect no support for 3.5, and our GH Actions CI should probably not build for 3.5. Pinging @akshayka to make sure I'm not mis-remembering.

SteveDiamond:
Hmm ok, I guess that's fine to drop 3.5. We had to abandon f-strings at one point because someone made a case for keeping 3.5.

cvxpy/settings.py
```diff
@@ -164,7 +164,7 @@
 UNKNOWN = "UNKNOWN"
 
 # Numerical tolerances
-EIGVAL_TOL = 1e-10
+EIGVAL_TOL: float = 1e-10
```
SteveDiamond, line 167: add types for all the values here?

SteveDiamond: actually why would we do types for literals?

SteveDiamond (approved):
LGTM

SteveDiamond:
Thanks @r-barnes ! It looks good to me. I was a little confused about types for constants.

cvxpy/settings.py
r-barnes: I've removed that now. I think Pyre at least prefers types for globals that are defined by function calls or similar, but I don't think it has a preference for literals. This was here as part of an attempt to fix something else.

PR #1376, opened 2021-05-18
use flexible numpy version in pyproject.toml

naphatkrit:
the only version number that matters for ABI interface purposes is the first minor version number. Therefore, use flexible versions here to allow pip to select a newer version of numpy if it so wishes. This speeds up the install time on linux for python 3.9 because a pre-built numpy wheel for 1.19.0 does not exist at that python version, but 1.19.5 does exist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,9 @@
 [build-system]
 requires = [
-    "numpy==1.15; python_version=='3.6'",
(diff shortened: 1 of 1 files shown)


naphatkrit:
(just signed the CLA)

SteveDiamond:
Thanks!

Issue #1367, opened 2021-05-12
Install order with numpy can break cvxpy


naphatkrit:
**Describe the bug**
Installing a pinned version of numpy together with cvxpy in the same pip command will cause cvxpy to be built against the wrong version of numpy. This is presumably due to pip building all packages first before installing, meaning at build time, cvxpy does not know which version of numpy it should build for.

**To Reproduce**
On a clean python environment (I used the `python:3.7` docker image):

```
$ pip install numpy==1.19.2 cvxpy==1.1.1
...
$ python
Python 3.7.10 (default, Apr 10 2021, 16:02:25)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cvxpy
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
```

If we instead run a command to install numpy first, it works fine:
```
$ pip install numpy==1.19.2 && pip install cvxpy==1.1.12
...
$ python
Python 3.7.10 (default, Apr 10 2021, 16:02:25)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cvxpy
>>>
```

If we instead pin numpy to its latest version, that also works fine:
```
$ pip install numpy==1.20.3 cvxpy==1.1.1
...
$ python
Python 3.7.10 (default, Apr 10 2021, 16:02:25)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cvxpy
>>>
```
Presumably, this will break on the next major numpy upgrade.

**Expected behavior**
cvxpy should install correctly in all cases, to allow it to be installed as part of a pinned requirements.txt file.

**Version**
 - OS: MacOS 11.12.3, on docker
 - CVXPY Version: 1.1.12

h-vetinari:
You're using two different versions of cvxpy, is that intentional?
```
$ pip install numpy==1.19.2 cvxpy==1.1.1
$ pip install numpy==1.19.2 && pip install cvxpy==1.1.12
$ pip install numpy==1.20.3 cvxpy==1.1.1
```

naphatkrit:
That’s a typo, meant to say 1.1.12

h-vetinari:
Based on the error
```
>>> import cvxpy
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
```
[...]

PR #1378, opened 2021-05-19
Build Manylinux Wheels

phschiele:
Hi @akshayka,

as discussed in #1367, this PR uses `cibuildwheel` to build manylinux wheels.
I have taken some inspiration from your and @bstellato's implementations and everything seems to be working as expected.

The only issue I found was that we need to run the manylinux build on the `manylinux_2_24` image. This means that the wheels only work with distros based on Debian 9 or higher (e.g. Ubuntu 18.04+). When running with other flags here, I got the following error:

```
ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain
```
I used [auditwheel-symbols](https://github.com/messense/auditwheel-symbols) to inspect which symbols are causing the issues. Below are the outputs I got for manylinux 2014 and 2010, respectively.

```
_cvxcore.cpython-38-x86_64-linux-gnu.so is not manylinux_2_17(aka manylinux2014) compliant because it links the following forbidden libraries:
libstdc++.so.6  offending symbols: _ZNSt12out_of_rangeC1EPKc@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_appendEPKcm@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEED1Ev@@GLIBCXX_3.4.21, _ZNSt11logic_errorC2ERKS_@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_createERmm@@GLIBCXX_3.4.21, _ZNSt16invalid_argumentC1EPKc@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE6appendEPKc@@GLIBCXX_3.4.21
libstdc++.so.6  offending symbols: __cxa_throw_bad_array_new_length@@CXXABI_1.3.8
```

```
_cvxcore.cpython-38-x86_64-linux-gnu.so is not manylinux_2_12(aka manylinux2010) compliant because it links the following forbidden libraries:
libstdc++.so.6  offending symbols: __cxa_throw_bad_array_new_length@@CXXABI_1.3.8
libstdc++.so.6  offending symbols: _ZNSt12out_of_rangeC1EPKc@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_appendEPKcm@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEED1Ev@@GLIBCXX_3.4.21, _ZNSt11logic_errorC2ERKS_@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE9_M_createERmm@@GLIBCXX_3.4.21, _ZNSt16invalid_argumentC1EPKc@@GLIBCXX_3.4.21, _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE6appendEPKc@@GLIBCXX_3.4.21
libc.so.6       offending symbols: memcpy@@GLIBC_2.14
```


continuous_integration/GitHub/deploy_script.sh
```diff
@@ -1,29 +0,0 @@
-#!/bin/bash
-
-source activate testenv
-conda config --add channels conda-forge
-conda install --yes requests twine
-pip install readme_renderer
```
phschiele, line 6: Was `readme_renderer` actually used? If so, we could add it back as a separate build step.

SteveDiamond: I believe readme_renderer is being used as part of the upload build, though I couldn't tell you what for exactly.

rileyjmurray: I tried to remove it before and was unable (some errors appeared, I think with the PyPI upload). The reason why it's installed via pip is that we can't get the right version from conda.

rileyjmurray:
There isn't much here that I can intelligently comment on. @phschiele my one question is: has this been tested for correctness somehow? In the past when I've set up our CI or deployment tools, I've made deployments to "Test PyPI" and my own conda channel.

phschiele:
> There isn't much here that I can intelligently comment on. @phschiele my one question is: has this been tested for correctness somehow? In the past when I've set up our CI or deployment tools, I've made deployments to "Test PyPI" and my own conda channel.

@rileyjmurray In terms of testing, I downloaded the artifacts from the actions tab (e.g. the latest run [here](https://github.com/cvxpy/cvxpy/actions/runs/860620649)) and verified that the Linux wheel works as expected.
I suppose it would be quite easy to upload every "master merge" to Test PyPI and only the releases to PyPI itself, what do you think?

rileyjmurray:
@phschiele in principle I'd be a big fan of uploading all master merge's to Test PyPI. The one issue there is that PyPI (and I believe Test PyPI) requires new version numbers for every release. Since only a portion of master merge's get incremented version numbers, that seems to prevent uploads for each master merge. Do you see a simple way around that?

phschiele:
@rileyjmurray You're right, seems like even on Test PyPI one cannot overwrite existing versions. 
I guess a workaround could be to simply append the git hash to the filenames for Test PyPI, but not sure if that's a great idea.

rileyjmurray:
@phschiele in that case I think it's not worth fussing with Test PyPI. However per your edit, @SteveDiamond @akshayka we might want to look at using GH Actions to increment version numbers (personally I think that's something to consider separately from this PR.)

akshayka:
This looks great! I like that this lets us easily build with OpenMP (which we should do, at some point in the future).

Eventually, we should test and build for Windows using Github Actions as well (@rileyjmurray, do you agree?). I believe all that will require is adding a new OS to the build matrix. But we can handle that in a future change.

akshayka (approved):

SteveDiamond (approved):

PR #1388, opened 2021-05-21
add py.typed file so that users of cvxpy get type info from the package

naphatkrit:
see https://www.python.org/dev/peps/pep-0561/, search for "py.typed" on the page.

tested by installing locally, importing, and running mypy.


SteveDiamond:
@r-barnes are you familiar with py.typed? It seems fine to me, just wanted a second opinion.

r-barnes:
@SteveDiamond : I'm afraid I'm not familiar with the `py.typed` system.

@naphatkrit: In the link you provided I see
> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. 
At this point cvxpy is able to provide type annotations for a number of public signatures, but by no means all. My parsing of the above sentence is that `py.typed` would perform checking within the package itself ("wish to support type checking"), but I think the purpose of this is to indicate to code importing cvxpy that cvxpy as type annotation support.

Could you clarify?

Is there any downside to adding `py.typed` before complete type annotation coverage is present?

naphatkrit:
> At this point cvxpy is able to provide type annotations for a number of public signatures, but by no means all. My parsing of the above sentence is that py.typed would perform checking within the package itself ("wish to support type checking"), but I think the purpose of this is to indicate to code importing cvxpy that cvxpy as type annotation support.
py.typed just indicates that your package provides type information, for users of the package.

> Is there any downside to adding py.typed before complete type annotation coverage is present?
i think the main downside will be that some types may be looser than expected and so users might get false confidence.

naphatkrit:
just to demonstrate,
```
import cvxpy

reveal_type(cvxpy.sum)
prob = cvxpy.Problem()
reveal_type(problem.solve)
```

running mypy on this file will yield:
```
test.py:3: note: Revealed type is 'def (expr: Any, axis: Any =, keepdims: builtins.bool =) -> Any'
test.py:4: error: Missing positional argument "objective" in call to "Problem"
test.py:5: error: Name 'problem' is not defined
test.py:5: note: Revealed type is 'Any'
Found 2 errors in 1 file (checked 1 source file)
```
See how `cvxpy.sum` has types info, but `Problem.solve` doesn't. so users of `cvxpy.sum` will get extra checks, users of `Problem.solve` will continue to get no type info (no regression from today)

naphatkrit:
That said, I'll leave it up to this group to determine when is an appropriate time to ship a PR like this one.

SteveDiamond:
I feel comfortable merging this after this discussion.

PR #1391, opened 2021-05-22
Move windows CI to GitHub Actions

phschiele:
As mentioned by @akshayka, it could be convenient to also include Windows builds in the GitHub Action CI.
We can directly use the shell scripts, as windows runners on GitHub come with bash (through the Windows install of git).
While the install seems fine, trying out many configurations I always got different kinds of errors when running the test script. 
A common one was this
```
File "C:\Miniconda\envs\test\lib\site-packages\scipy\sparse\linalg\isolve\iterative.py", line 10, in <module>
    from . import _iterative
ImportError: DLL load failed: The specified module could not be found.
```
suggesting that some library (or path?) issue with scipy (Full log [here](https://github.com/cvxpy/cvxpy/pull/1391/checks?check_run_id=2647229096)).

Does that seem familiar, or do you have an idea what this could be related to?
I'll play around with it some more in the meantime. 

cc: @SteveDiamond @rileyjmurray


SteveDiamond:
@phschiele have you seen the Appveyor scripts? https://github.com/cvxpy/cvxpy/tree/master/continuous_integration/AppVeyor

phschiele:
@SteveDiamond Yes, I saw those. I suppose one could simply call the `configure_conda_and_install_cvxpy.ps1` script from GitHub as well.
However, I thought it would be ideal to have only a single script that works on Linux, macOS, and Windows.
The only real differences between the AppVeyor and GitHub scripts is that the former fixes `mkl=2018.0.3` and adds `
conda config --add channels oxfordcontrol`. I have experimented with those settings, but it didn't fix the issues.

SteveDiamond:
Got it, just checking that you had seen it. I totally agree it would be ideal if the same script more or less worked for all platforms.

phschiele:
There are also some warnings during the compilation of cvxcore on windows, but I doubt it is related:
```
cl : Command line warning D9002 : ignoring unknown option '-O3'
cl : Command line warning D9002 : ignoring unknown option '-std=c++11'
```

rileyjmurray:
Personally, I wouldn't place too much importance on the same setup script being used for all operating systems. That's a convenience that comes as a far second to each operating system having a reliable CI setup. I think it's a major improvement to have everything on GH Actions even if different operating systems need separate setup scripts.

All that said -- it's strange that the differences between the install scripts aren't affecting whether this DLL error shows up. I agree that the warnings about cvxcore are unrelated. This looks like a more generic scipy issue. @phschiele any luck finding others with similar issues online?

SteveDiamond:
@phschiele @rileyjmurray I released a new patch and the github actions didn't upload any wheels. Any idea what's going on here? Is something wrong with get_version?

phschiele:
@SteveDiamond Looks like the deploy step was skipped in the [runs](https://github.com/cvxpy/cvxpy/actions/runs/872642749).
The condition is 
`if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')`, which I have copied from the [qdldl build](https://github.com/oxfordcontrol/qdldl-python/blob/master/.github/workflows/build.yml). 
I'll take a closer look at why it didn't trigger and let you know here.

SteveDiamond:
Ok sounds good. We used to test if the version had increased relative to the version on pypi.

phschiele:
@SteveDiamond I think I found the issue. Our build was configured to only run on pushes to master. Comparing to the setup on qdldl:
`on: [push, pull_request]`
You can see that this creates two runs, one for the merge to master and one for the tag:
image
_______________
We only got the master merge, which did not trigger the deploy step:
image

SteveDiamond:
That used to happen, not sure what went wrong this time. I did push the tag.

phschiele:
@SteveDiamond In the `.github/workflows/build.yml` file, we need to replace:
```yml
on:
    pull_request:
    push:
        branches:
            - master

```
with 
```yml
on: [push, pull_request]
```

SteveDiamond:
I tried that but it said it was a syntax error.

phschiele:
@SteveDiamond 
```yml
        branches:
            - master
 ```
Also needs to be removed, such that the `on` block is only the single line `on: [push, pull_request]`

I am also looking around if it is possible to retrigger for the `1.1.13` tag once that is included.

**Comment by phschiele** [comment]
@SteveDiamond From what I can see there are three options:
1. We could either download the artifacts from the action and run the `twine upload` command locally
2. Delete the tag, push, and create it again, something like [this](https://stackoverflow.com/questions/7947078/git-tag-delete-and-re-add)
3. Create a new tag

**Comment by SteveDiamond** [comment]
I deleted the tag and pushed but that didn't seem to do anything.

**Comment by phschiele** [comment]
@SteveDiamond right, I still see the same tag from an hour ago.
So you did the following?:
```bash
git push --delete origin v1.1.13
git tag --delete v1.1.13  <- perhaps the local tag must also be deleted
git tag v1.1.13
git push origin v1.1.13
```

SteveDiamond:
I hadn't deleted the local tag.

phschiele:
@SteveDiamond Two things went wrong:
1. It needs to be `--repository-url` not `-repository-url` in the `twine upload` command in the `build.yml` file
2. The upload for the `OPENMP` build failed because there we don't build the wheel (at the moment). We can ignore that for now and I'll create a follow-up for that one.

PR #1396, opened 2021-05-24
Skip Deploy Step Openmp Build

phschiele:
We do not build the wheels using openmp at the moment, thus the deployment fails for that config. 
This PR skips this step accordingly.
@SteveDiamond


SteveDiamond (approved):
LGTM

PR #1398, opened 2021-05-25
Move Windows CI to GitHub Action v2

phschiele:
Hi @SteveDiamond, @rileyjmurray, @akshayka 

closed the first PR, as it got too cluttered from our deployment investigation.

To get the CI working on windows I had to:
- bump `scipy=1.1 numpy=1.15` to `scipy=1.3 numpy=1.16` on Python 3.6 and 3.7, this fixed the "DLL not found" error
- skip two benchmark tests -> They caused the "Error 3" failures during pytest. I found some comments in the codebase there is a known bug affecting some windows/`scipy` combinations. I suspect it is a similar case here
- remove the printing of `osqp_version`, could be added back with if condition if necessary

Overall the complexity got reduced quite a bit.

Note: seems like one job failed due to the connection error again, so it needs to be retriggered by an admin. 
I'll check if other users also have that issue when using `setup-miniconda@v2`


SteveDiamond:
Once the tests pass this generally looks good to merge to me.

.github/workflows/build.yml
```diff
@@ -22,19 +22,19 @@ jobs:
...
           - os: ubuntu-18.04
             python-version: 3.8
             openmp: "True"
           - os: ubuntu-18.04
             python-version: 3.8
-            openmp: "False"
           - os: macos-10.15
             python-version: 3.8
-            openmp: "False"
             deploy_pypi_source: "True"
```
SteveDiamond, line 35: Should we get rid of this?

SteveDiamond:
We could skip building wheels and source when we're not uploading to pypi.

.github/workflows/build.yml
phschiele: Wouldn't it still be helpful e.g. for early adopters of new Python versions, before a wheel is provided?

SteveDiamond: fair enough

SteveDiamond:
LGTM!

Issue #228, opened 2015-09-24
Missing atom: gammaln


Bonnevie:
I heard Boyd joking at a recent talk that you never get requests for new atoms, so here is one:
I am a bit hamstringed by the lack of gammaln (logarithmic gamma function) in your list of atoms. It's definitely convex (e.g. see https://proofwiki.org/wiki/Log_of_Gamma_Function_is_Convex_on_Positive_Reals or just plot the second derivative), so I think it would be a fruitful addition to cvxpy. It occurs with some frequency when working with probabilistic models as many exponential family distributions have a gamma function embedded in their partition function. 
By extension, the multivariate gamma (multigammaln in scipy) would also be useful.

SteveDiamond:
I'd suggest approximating the log_gamma function as a piecewise linear function. If you find a really good approximation we can add it to cvxpy. This is how log_norm_cdf works in CVX.

Bonnevie:
I am told the Lanczos approximation is the way to go with gammaln. Do you need an implementation in C? It's available in the cmath package as "lgamma".

SteveDiamond:
The point of approximating the function isn't to evaluate it. It's to create a representation that CVXPY can work with and incorporate into the final form of the problem. Specifically, lgamma needs to be represented as partial optimization over a cone program, e.g.,

lgamma(x) = minimize_{over t} c^Tt
                     subject to dx + At + b in K

where t in R^n is a new variable, d in R^m, A in R^{m x n} and b in R^m are known constants, and K is a convex cone of a supported type (nonnegative and equal to 0 are the easiest to think about).

SteveDiamond:
I'm closing this. Feel free to reopen it if there's more interest/progress that you want to contribute.

cdipaolo:
Hi @Bonnevie @SteveDiamond, I also ran into this issue (in the same problem space regarding exponential family ML estimates) and wanted to share the approximations I came up with to get around it. I'm sure there's something much more sophisticated that would work better, but this appears to work well enough and might help someone who runs into this issue later. This doesn't feel robust enough to add directly in to the library, but perhaps a future iteration of it would work for that purpose.

The short of it is that you can use the following expression as a drop-in log-gamma replacement in CVXPY:
```python
x = cp.Variable()
approx_log_gamma = lambda x: cp.maximum(
    2.18382 - 3.62887*x,
    1.79241 - 2.4902*x,
    1.21628 - 1.37035*x,
    0.261474 - 0.28904*x,
[...]

PR #1400, opened 2021-05-26
loggamma approximation

SteveDiamond:
This adds the approximation of the log-gamma function from #228. I followed the naming convention from scipy.

### DIFF
--- a/cvxpy/atoms/__init__.py
+++ b/cvxpy/atoms/__init__.py
@@ -82,6 +82,7 @@ from cvxpy.atoms.elementwise.huber import huber
 from cvxpy.atoms.elementwise.inv_pos import inv_pos
 from cvxpy.atoms.elementwise.kl_div import kl_div
 from cvxpy.atoms.elementwise.log import log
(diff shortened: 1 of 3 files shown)


rileyjmurray:
I think it would be good to add tolerance parameters here. Like ``loggamma(x, rtol, atol)`` with relative and absolute tolerances. To keep things simple we can just use ``rtol=0.27`` and ``atol=0.04`` as defaults (since these hold for the given implementation) and raise NotImplementedError's when these values are changed.  Alternatively, the function name could be changed to ``loggamma_approx`` and we can dispense with tolerances. @akshayka what do you think?

SteveDiamond:
For context we already have a ``log_normcdf`` which is an approximation.

rileyjmurray:
Ah! In that case, I think this is fine. Do you want to add this atom and the other approximation to the web documentation? Or save that for a release of 1.2?

SteveDiamond:
Maybe I'll add the docs later. I'd like to merge this in now so I don't forget.

PR #1411, opened 2021-06-03
remove model update within soc constraint for loop

KarenOP:
I'm working on a model with > 10k SOC constraints. The model updates are adding 60s and there's a warning from Gurobi that says
```
Warning: excessive time spent in model updates.
```
In the for loop while adding SOC constraints, there's a Gurobi `model.update()` that's causing this.
Since you're not using the model to get anything between adding these constraints and then optimizing, there's no real reason to have it there at all.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
@@ -329,7 +329,6 @@ class GUROBI(SCS):
                     lb=-gp.GRB.INFINITY,
                     ub=gp.GRB.INFINITY)
             ]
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Thanks @KarenOP for figuring this out! I want to setup better gurobi testing before we merge this in. @phschiele it should be very easy, just install gurobipy as part of the build. I can take a look this weekend.

phschiele:
@SteveDiamond I can add a PR to include gurobipy (and perhaps some of the other solvers on the roadmap) tomorrow.

SteveDiamond:
Thanks @KarenOP ! LGTM.

PR #1413, opened 2021-06-04
Gurobi CI

phschiele:
@SteveDiamond Adding Gurobi CI first to unblock #1411 
The remaining solvers will be added in a follow-up.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -982,7 +982,7 @@ class TestGUROBI(BaseTest):
         StandardTestLPs.test_lp_2(solver='GUROBI')
 
     def test_gurobi_lp_3(self) -> None:
-        StandardTestLPs.test_lp_3(solver='GUROBI')
+        StandardTestLPs.test_lp_3(solver='GUROBI', InfUnbdInfo=1)
 
     def test_gurobi_lp_4(self) -> None:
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_conic_solvers.py
```diff
@@ -982,7 +982,7 @@ def test_gurobi_lp_2(self) -> None:
         StandardTestLPs.test_lp_2(solver='GUROBI')
 
     def test_gurobi_lp_3(self) -> None:
-        StandardTestLPs.test_lp_3(solver='GUROBI')
+        StandardTestLPs.test_lp_3(solver='GUROBI', InfUnbdInfo=1)
```
phschiele, line 985: The problem is unbounded below, but Gurobi would only return `inf`. Adding this parameter also returns the direction in which the problem is unbounded.

cvxpy/tests/test_qp_solvers.py
```diff
@@ -51,13 +51,13 @@ def setUp(self) -> None:
         self.offset = Variable(1, name='offset')
         self.quadratic_coeff = Variable(1, name='quadratic_coeff')
 
-        T = 100
```
phschiele, line 54: I had to reduce the size of `control` and `smooth_ridge` due to size limitations in the free/trial version.

rileyjmurray (approved):

Issue #1414, opened 2021-06-04
Add an interface to SciPy-accessible LP solvers


Michael-git96:
Scipy version 1.6.1 and later versions have access to HiGHS solvers (https://www.maths.ed.ac.uk/hall/HiGHS/#guide) for LPs via the linprog function  which are comparable with GLPK and other solvers in terms of performance especially for large sparse LP problems which uses SciPy.sparse matrices.

The aim would be to have a new interface with SciPy so that users can access the HiGHS solvers through CVXPY. 

I have forked and cloned the repo and have already started to work on such an interface.

Michael-git96:
I have been working on a new Solver interface to incorporate the HiGHS solvers (https://www.maths.ed.ac.uk/hall/HiGHS/#guide) via SciPy and by following the steps in the contributing section of CVXPY (https://www.cvxpy.org/contributing/index.html#solver-interfaces) I managed to get a working solution however I had two questions:

1) The SciPy linprog() function has a parameter called "method" (https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html), however that parameter is already being used in Problem.solve() (line 385 in https://github.com/cvxpy/cvxpy/blob/master/cvxpy/problems/problem.py) . Should I ask users to specify options within a separate dictionary e.g. prob.solve(verbose=True, solver="SCIPY", scipy_options={method='highs-ipm', 'options'={'tol':1e-6}}) ? Or should I change the name of the parameter to something like "scipy_method" and it can then be included in the documentation?

2) linprog() does not return any dual variables, however it does return the slack variables. In the invert function within the scipy_conif.py file I created, I return the dual_vars as None. Are there any consequences of doing this, or is there a way to calculate the dual variables which isn't time consuming?

rileyjmurray:
For (1), I think having an explicit keyword argument ``scipy_method`` is the way to go. That wouldn't go in the docs for the Problem class, it would just be mentioned in [this part of the web docs](https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options). The ``SCIPY`` class should also have the ability to process arguments like 
```...
d={'method':'highs-ipm', 'tol':1e-6}
prob.solve(solver='SCIPY', scipy_options=d)
```
then your interface will effectively do something like
```
method = d.pop('method', 'interior_point')  # 'interior_point' is the SciPy default
# ^ extract any keyword arguments that SciPy doesn't put in "options"
[...]

Issue #1395, opened 2021-05-24
CVXPY not working with scipy.sparse matrices and verbose=True


Michael-git96:
**Describe the bug**
I recently upgraded cvxpy to version 1.1.12 and now when I run an optimisation with a sparse constraint matrix and verbose=True, I get an error.

**To Reproduce**
import cvxpy as cp
import numpy as np
from scipy import sparse

x = cp.Variable((10, 1))
c = -np.ones((10,))

objective = cp.Minimize(c.T@x)

G = sparse.eye(10)
h = np.ones((10, 1))*10
prob = cp.Problem(objective, [G@x <= h])

res = prob.solve(verbose=False)

**Expected behavior**
It used to work fine with the previous version, and it still works fine if verbose=False. Setting verbose=False in the code above will lead to the correct result.

**Version**
 - OS: Windows
 - CVXPY Version: 1.1.12

SteveDiamond:
This works on master. We will put out a patch soon, for the moment try installing from source.

Michael-git96:
Ok, thank you, I will install from the source and wait for the patch. If this has already been fixed I will just go ahead and close this issue.

PR #1416, opened 2021-06-06
Add an interface to SciPy-accessible LP solvers - Issue: #1414

Michael-git96:
This new interface with SciPy,  provides access to HiGHS solvers (https://www.maths.ed.ac.uk/hall/HiGHS/#guide) for LPs via the linprog function (https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html) found in scipy.optimize. This pull request is linked to issue #1414. 

This does require a SciPy version > 1.6.1 as earlier versions did not have access to the HiGHS solvers.


setup.py
```diff
@@ -84,6 +84,6 @@ def is_platform_mac() -> bool:
                       "ecos >= 2",
                       "scs >= 1.1.6",
                       "numpy >= 1.15",
-                      "scipy >= 1.1.0"],
+                      "scipy >= 1.6.1"],
```
phschiele, line 87: Scipy 1.6.1 is quite recent (Feb 21), so this might break things for some users. Do you think it would be better to keep the lower scipy requirement for the `install_requires` section of `setup.py` and raise an informative error if a user tries to use the solvers added in this PR with a version <1.6.1?

rileyjmurray: I agree with @phschiele.

I think the best way forward is to not set the ``method`` parameter in ``linprog``, unless that parameter was specified by the user. We can trust SciPy to select the best default method for a given version.

If the user specifies a method string, we can check if it's compatible with the given version of SciPy. If the method is not compatible then we could tell them what version of SciPy they'd need to install to get that method. On top of that, it might be good to raise a warning if someone (a) does not set the ``method`` parameter and (b) has SciPy version < 1.6.1.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -0,0 +1,142 @@
+"""
+Copyright 2021 Michael Sommerauer
```
rileyjmurray, line 2: The copyright notice we use now is "The CVXPY Developers".

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -0,0 +1,142 @@
...
+        # Extract solver options which are not part of the options dictionary
+        if solver_opts:
+            if "scipy_options" not in solver_opts:
+                raise ValueError("All parameters for the SCIPY solver should "
+                                 "be incased within a dictionary called "
+                                 "scipy_options e.g. \n"
+                                 "prob.solve(solver='SCIPY', verbose=True,"
+                                 " scipy_options={'method':'highs-ds', 'maxiter':10000})")
+            meth = solver_opts['scipy_options'].pop("method", "highs-ds")
+            bnds = solver_opts['scipy_options'].pop("bounds", (None, None))
```
rileyjmurray, line 109: "Bounds" is interpreted with respect to the single vectorized decision variable, which won't be transparently related to the CVXPY model specified by the user. For the time being we should raise an error if a user specifies "bounds".

Michael-git96: Two of the lp tests (lp_3 and lp_4 in cvxpy/tests/solver_test_helpers.py) were passed and have bounds within the constraints similar to:

`# Infeasible
x = cp.Variable((10, 1))
objective = cp.Minimize(cp.sum(x))
prob = cp.Problem(objective, [x >= 1, x <= 0])
prob.solve(verbose=True, solver="SCIPY")`

As far as I can tell, as long as the bounds for CVXPY are contained within the constraints of the Problem instance, they will be passed to the solver through the constraints. With the "bounds" parameter within the scipy_options dictionary the user has the ability to specify exactly the same bounds:

`# Infeasible
x = cp.Variable((10, 1))
objective = cp.Minimize(cp.sum(x))
prob = cp.Problem(objective)
prob.solve(verbose=True, solver="SCIPY", scipy_options={'bounds':(1, 0)})`

From what I can see, the only downside of keeping the bounds parameter, is that if the user were to try another solver and only specified the bounds through that parameter, then the other solver would not see those bounds. But if the user has specified the bounds using the first example above, then everything should work as normal. 

Should I still go ahead and raise an error if the user specifies "bounds" within scipy_options to force the user to include them within the constraints of the CVXPY Problem instance?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
Michael-git96: I have changed this in the latest commit, thanks

setup.py
Michael-git96: Unfortunately SciPy does not do a version check at the moment and adjust the method appropriately. The default method chosen is always "interior-point" most likely due to legacy reasons, hence I have used the same default and added in a warning which explains to the user the benefits of specifying the method and using the HiGHs solvers which require a SciPy version >= 1.6.1.

I have added the warnings in my latest commit however I think they are not triggered when verbose is set to False.

rileyjmurray: Hmm. @Michael-git96, perhaps we could manually check the SciPy version and set the default that way? For example:
```
import scipy as sp
from packaging import version

# set our default solve method
cur_version = version.Version(sp.__version__)
if cur_version >= version.Version('1.6.1'):
    method = 'highs'
else:
    method = 'interior-point'

# let the user override our default
if 'method' in scipy_options:
    method = scipy_options['method']

File: cvxpy/reductions/solvers/conic_solvers/scipy_conif.py (hunk 3 above)
Comment (rileyjmurray) [reply cvxpy/reductions/solvers/conic_solvers/scipy_conif.py:119]: The equivalence between specifying bounds in a CVXPY Constraint (or a Variable attribute) and in ``scipy_options`` is only really valid for a problem with a single vectorized decision variable, and for problems that use only linear algebraic constraints. Some of our "atoms" are LP representable but require introducing additional helper variables that the user isn't told about. For example, we introduce helper variables for the 1-norm, infinity-norm, and the sum-k-largest atoms. In these cases, if someone set ``bounds=(1, 10)``, then they would unknowingly affect the helper variables we introduced while canonicalizing the problem. I think the best thing to do right now is to raise an error. @SteveDiamond, @akshayka, what do you two think?

Comment (akshayka) [reply cvxpy/reductions/solvers/conic_solvers/scipy_conif.py:119]: I agree that it's best to raise an error. It would be very confusing to have to separate but slightly different notions of box constraints (CVXPY constraints and SciPy bounds).

File: setup.py (hunk 1 above)
Comment (Michael-git96) [reply setup.py]: @rileyjmurray - thank you for the suggestion, I have gone ahead and implemented the default method based on the version of SciPy installed.

File: cvxpy/reductions/solvers/conic_solvers/scipy_conif.py (hunk 3 above)
Comment (Michael-git96) [reply cvxpy/reductions/solvers/conic_solvers/scipy_conif.py]: Thanks @akshayka and @rileyjmurray, I have gone ahead and ensured that CVXPY raises an error if the user specifies the 'bounds' parameter.

**Review by rileyjmurray — CHANGES_REQUESTED** [review]
This mostly looks good! I left one tiny comment about unittests. Other than that, the only thing that remains is updating documentation in the following places:
 * Our table of solver capabilities: https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#choosing-a-solver. You can list the solver as "SciPy / HiGHS".
 * Setting solver options: https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advahttps://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rstnced/index.rst#setting-solver-options. You can just provide a one-line statement much like the SCIP description.
 * Installation: https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst. Mention that SciPy's "interior-point" and "revised-simplex" implementations are written in python and are always available, while HiGHS is a C++ solver that comes bundled with SciPy version 1.1.6 and higher.

**Comment by Michael-git96** [comment]
@rileyjmurray - I made the commits for the documentation but I am a bit worried that it may not be that clear, because at the moment the solver is called using solver="SCIPY" instead of solver="SciPy/HiGHS". Should I be changing something in the documentation or the code?

File: cvxpy/tests/test_conic_solvers.py [hunk 4]
```diff
@@ -1429,3 +1429,25 @@ def test_ecos_bb_mi_lp_3(self) -> None:
     @pytest.mark.skip(reason="Known bug in ECOS BB")
     def test_ecos_bb_mi_socp_1(self) -> None:
         StandardTestSOCPs.test_mi_socp_1(solver='ECOS_BB')
+
+
+@unittest.skipUnless('SCIPY' in INSTALLED_SOLVERS, 'SCIPY is not installed.')
```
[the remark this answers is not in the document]
Michael-git96: I have removed the redundant check if SciPy is installed.

rileyjmurray:
> @rileyjmurray - I made the commits for the documentation but I am a bit worried that it may not be that clear, because at the moment the solver is called using solver="SCIPY" instead of solver="SciPy/HiGHS". Should I be changing something in the documentation or the code?

Ah! Yes, that's an excellent point. In that case, I think using only "SCIPY" in the table is the way to go.

Michael-git96:
I have changed all the documents so that they include the SCIPY solver. I also added a few more details to the Setting Solver Options section in the tutorial/advanced/index.rst file to clarify that the main benefit of using this solver is its access to the HiGHS solvers.

rileyjmurray (approved):
Looks good to me!

SteveDiamond:
You need to fix the formatting so that ``flake8 .`` passes.

Michael-git96:
To correct the flake8 error, I introduced a new variable which is used in the if statement. I don't think it is particularly elegant or efficient, but it seems to solve the flake8 error.

SteveDiamond:
The test failed because the scipy solver being used raised some kind of error when solving the problem in cvxpy/tests/test_problem.py::TestProblem::test_verbose. That's odd it works for you but doesn't work on all these different platforms (linux, mac) and python versions...

Michael-git96:
@SteveDiamond - Could it be the CVXPY version used for testing? I raised an issue #1395 a few days ago where when verbose was set to True, and sparse matrices were used in the constraints, then it would return an error. I do not think it is linked to the SciPy version but I can double check.

SteveDiamond:
> @SteveDiamond - Could it be the CVXPY version used for testing? I raised an issue #1395 a few days ago where when verbose was set to True, and sparse matrices were used in the constraints, then it would return an error. I do not think it is linked to the SciPy version but I can double check.

#1395 is fixed on master, so that's very unlikely to be the issue.

Michael-git96:
> > @SteveDiamond - Could it be the CVXPY version used for testing? I raised an issue #1395 a few days ago where when verbose was set to True, and sparse matrices were used in the constraints, then it would return an error. I do not think it is linked to the SciPy version but I can double check.
> 
> #1395 is fixed on master, so that's very unlikely to be the issue.

I took a look at the details and noticed that the SCIPY tests ran fine and when it got to TestProblem::test_verbose then the warning that scipy raises about sparse matrices comes up and I believe this is where the error is occuring?
image

(Side note: SciPy currently raises two warnings when using HiGHS solvers: 1. Setting sparse to True, 2. Unknown solver options sparse. This is just a bug in linprog and I will raise an issue on that.)

If this is the case, I am 99% sure that it is the issue (#1395) I raised about having verbose=True and SciPy.sparse constraints. I am struggling to find another explanation and I checked the previous scipy versions (I checked 1.5.2) and there were no issues. Could you please help me figure out why it is failing?  Apologies, I am not that experienced in testing.

Michael-git96:
@SteveDiamond - Taking a closer look at the output it does seem to be an issue only with SCIPY because the other solvers passed the TestProblem::test_verbose fine. So it will not be linked to issue #1395 . 

I checked to see if there was ExpCone, PSD in the SUPPORTED_CONSTRAINTS or if MIP_CAPABLE was set to True, but that is not in scipy_conif.py. It is very odd because I am unable to replicate the issue, but also the SCIPY unittests passed. Could it be an issue with how the problem is formulated within test_verbose?
`p = Problem(cp.Minimize(self.a + self.x[0]),
                            [self.a >= 2, self.x >= 2])`

Michael-git96:
I did some more testing and indeed it seems to be an issue with the "interior-point" and "revised simplex" methods inherent in scipy.optimize.linprog . I believe that, for some reason, those solvers are unable to solve the problem in TestProblem::test_verbose .

Is there a way of replicating that problem?

phschiele:
@Michael-git96 The tests running on the CI use `scipy=1.3` or `scipy=1.5`, depending on the Python version. I assume you used a more recent version for developing the feature, which might explain the discrepancy.
Running locally with `scipy=1.3` on Python 3.8/Ubuntu, I am able to reproduce the error.

Michael-git96:
@phschiele - thank you, I managed to replicate the error with those scipy versions. It seems to be an issue with the scipy.optimize.linprog method "interior-point" or "revised simplex". Whenever those methods are chosen (interior-point is the default if scipy version is less than 1.6.1) the solver fails to solve the test case passed by TestProblem::test_verbose.

Since it seems not not be a problem with the interface I created and more a problem of the solver method chosen I am not sure how best to proceed. I think there are two options:
1) Require a scipy version of 1.6.1 or larger and have the default method as the HiGHS solver so the test passes.
2) Ignore the test when the scipy version is less than 1.6.1 since the standard methods 'interior-point' and 'revised simplex' have issues solving the test case and return status : Numerical difficulties encountered. 

I think 2) may be best because those standard methods still worked well for the unittests in test_conic_solvers.py they just don't seem to work for the test_verbose test cases, however I am open to any suggestions.

SteveDiamond:
> @phschiele - thank you, I managed to replicate the error with those scipy versions. It seems to be an issue with the scipy.optimize.linprog method "interior-point" or "revised simplex". Whenever those methods are chosen (interior-point is the default if scipy version is less than 1.6.1) the solver fails to solve the test case passed by TestProblem::test_verbose.
> 
> Since it seems not not be a problem with the interface I created and more a problem of the solver method chosen I am not sure how best to proceed. I think there are two options:
> 
> 1. Require a scipy version of 1.6.1 or larger and have the default method as the HiGHS solver so the test passes.
> 2. Ignore the test when the scipy version is less than 1.6.1 since the standard methods 'interior-point' and 'revised simplex' have issues solving the test case and return status : Numerical difficulties encountered.
> 
> I think 2) may be best because those standard methods still worked well for the unittests in test_conic_solvers.py they just don't seem to work for the test_verbose test cases, however I am open to any suggestions.

It's fine to skip scipy in the test. Just add SCIPY to the list of skipped solvers on line 273 of test_problem.py

Michael-git96:
@SteveDiamond - I have just added SCIPY to the list on line 273 and I ran pytest using both an earlier version of scipy and the most recent and they seemed to pass.

SteveDiamond (approved):

SteveDiamond:
Great job @Michael-git96 !

Michael-git96:
@SteveDiamond - thank you! I will keep an eye out for any other issues/features that I can help with, I am always happy to contribute.

Michael-git96:
@SteveDiamond - thanks, I joined the discord server but I have not been assigned any roles so I can't see or send messages in the developers chat.

PR #1422, opened 2021-06-18
Add Mosek CI

phschiele:
@SteveDiamond The tests still seem to fail, although it passes in the [PR in my fork](https://github.com/phschiele/cvxpy/pull/13). I have set up the secret there as well (removed it again after the successful test). 

Looks like it's either an access issue again, or something went wrong with the setup of the secret.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,6 +47,7 @@ jobs:
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
       PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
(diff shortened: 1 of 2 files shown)


SteveDiamond:
I set the secret again and relaunched the build, but it still failed. Not sure what's going on.

SteveDiamond:
Also when mosek is installed it will be used by default. I'm not sure we want the tests using mosek everywhere, need to think about this.

phschiele:
@SteveDiamond Feel free to cherry-pick the two commits/copy the changes into a new PR to see if that resolves the issue.

RE default solver:
Perhaps using [parametrized tests](https://docs.pytest.org/en/6.2.x/parametrize.html) could be an option. Given that depending on the users' local setup, either MOSEK or ECOS/SCS will be used, it would be nice to test all cases, especially if it does not add a lot of code overhead.

SteveDiamond:
Merged in via #1431

Issue #1425, opened 2021-06-21
dual variables for SciPy LP solver interfaces 


rileyjmurray:
From the SciPy user digest, as of SciPy 1.7.0:

> When ``linprog`` is used with ``method`` ``'highs'``, ``'highs-ipm'``, or
> ``'highs-ds'``, the result object now reports the marginals (AKA shadow
> prices, dual values) and residuals associated with each constraint.

Adding support for this would follow up on  #1416. Pinging @Michael-git96.

Michael-git96:
@rileyjmurray - thanks for letting me know, glad to see the dual variables are returned for the HiGHS solvers in the latest SciPy version. I will create a pull request linked to this issue.

PR #1426, opened 2021-06-23
Adding dual variables for SciPy LP solver interfaces #1425

Michael-git96:
Adding dual variables when the SciPy solver is used with methods "highs", "highs-ds" or "highs-ipm" and the Scipy version is >= 1.7.0 . Linked to Issue #1425 .

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
@@ -19,7 +19,9 @@ import cvxpy.settings as s
 from cvxpy.constraints import NonNeg, Zero
 from cvxpy.reductions.solution import failure_solution, Solution
 from cvxpy.reductions.solvers.conic_solvers.conic_solver import ConicSolver
+from cvxpy.reductions.solvers import utilities
 from distutils.version import StrictVersion
+import scipy  # For version checks
 import warnings
 
 
@@ -92,7 +94,6 @@ class SCIPY(ConicSolver):
 
     def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
         from scipy import optimize as opt
-        import scipy
 
         # Set default method which can be overriden by user inputs
         if (StrictVersion(scipy.__version__) < StrictVersion('1.6.1')):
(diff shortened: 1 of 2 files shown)


Michael-git96:
Two things I would like to note:

1) I left the parameter duals as False in the tests in cvxpy/tests/test_conic_solvers.py, since the dual variables are only returned when the user has a SciPy version >= 1.7.0 and the method is one of the HiGHS methods. 

2) I believe the HiGHS methods return the dual variables with the sign switched, so I had to switch it back in line 186 and 190 within cvxpy/reductions/solvers/conic_solvers/scipy_conif.py . I ran a simple LP and calculated the Lagrangian dual manually and it seemed like the sign was switched in the result, but I would appreciate it if someone could double check that.

Michael-git96:
Update on the two points above:

1) I ran the SCIPY tests with dual set to True (and a SciPy version of 1.7.0) and they passed, so the duals work when they are returned.

2) I found a note in the documentation of linprog for SciPy 1.7.0 (https://docs.scipy.org/doc/scipy-1.7.0/reference/optimize.linprog-highs.html#optimize-linprog-highs) which supports the suspicion I had about the dual variables being returned with the sign switched:
> The result fields ineqlin, eqlin, lower, and upper all contain marginals, or partial derivatives of the objective function with respect to the right-hand side of each constraint. These partial derivatives are also referred to as “Lagrange multipliers”, “dual values”, and “shadow prices”. The sign convention of marginals is opposite that of Lagrange multipliers produced by many nonlinear solvers.

rileyjmurray:
Thanks for doing this, @Michael-git96!

> Update on the two points above:
> 
> 1. I ran the SCIPY tests with dual set to True (and a SciPy version of 1.7.0) and they passed, so the duals work when they are returned.

Could you change the tests so that the check for duals is performed when the appropriate SciPy version is installed? You could look at adding a class attribute like ``SCIPY.DUALS = StrictVersion(scipy.__version__) >= StrictVersion('1.7.0')`` and then in the tests change lines like ``StandardTestLPs.test_lp_0(solver='SCIPY', duals=False)`` to ``StandardTestLPs.test_lp_0(solver='SCIPY', duals=SCIPY.DUALS)`` (you'd need to import the SCIPY solver object right after https://github.com/cvxpy/cvxpy/blob/693a58a7d784bca7c0b6460f462ec22326555f6d/cvxpy/tests/test_conic_solvers.py#L1434).

Michael-git96:
Thanks for the suggestion @rileyjmurray, I have added in the class attribute so now the duals will be tested if a SciPy version higher than or equal to 1.7.0 is installed.

rileyjmurray (approved):

rileyjmurray:
Looks good to me. Thanks again, @Michael-git96!

Issue #1406, opened 2021-05-28
Add a relative entropy atom (consistent with SciPy's rel_entr).


rileyjmurray:
CVXPY has a ``kl_div`` atom that's consistent with SciPy's ``kl_div``. SciPy also has a function called ``rel_entr`` that only differs from ``kl_div`` by a linear expression. The ``rel_entr`` function is consistent with Wikipedia's definition of relative entropy and KL divergence (that is, Wikipedia considers these as two names for the same function).

We should add an elementwise ``rel_entr`` atom to CVXPY to be consistent with SciPy, and expose the more "canonical" relative entropy function in a natural way. This should be very easy to implement. If someone wants to contribute this atom we'd need (1) the atom itself, (2) some test cases for the atom, (3) web documentation for the atom.

See also #733.

Michael-git96:
@rileyjmurray - I would be up to give it a try if no one else has already started working on it?

phschiele:
@Michael-git96 I have also just started with this issue. Let's align through discord.

rileyjmurray:
Closed by #1427 and #1439.

Issue #733, opened 2019-05-23
Update kl_div documentation


robdmc:
**Describe the bug**
There is a discrepancy in the mathematical definition cvxpy uses for kl_div

https://www.cvxpy.org/api_reference/cvxpy.atoms.elementwise.html#kl-div

and the standard mathematical definition, which wikipedia embraces.
https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence

There might be some numerical trickery you guys are using that is unfamiliar to me, but it's really not obvious to me how the two expressions are equivalent. 

I posted a similar question to your forum, but given it's possible this is a bug, I thought I'd raise it here as well.

**To Reproduce**
In numpy syntax where x and y are 1-d arrays, the standard expression for KL Divergence
is

D = sum(x * log(x / y))

In your implementation this expression would be

D = sum(kl_div(x, y)) = sum(x * log(x / y) - x + y)

Unless I'm missing something, those don't appear to be the same thing.

rileyjmurray:
There is a subtle distinction between "[KL divergence](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.kl_div.html)" and "[relative entropy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.rel_entr.html#scipy.special.rel_entr)" in scientific python code. CVXPY follows the convention of SciPy. Most references in mathematics do not mention this difference, because these two functions are the same if you sum over their outputs, and the input vectors ``x``, ``y`` are probability distributions.


You might say that ``kl_div`` is a little awkward compared to ``rel_entr``, and honestly I'd agree with you there. However from an optimization perspective, ``kl_div`` makes more sense, because it is the perspective of the [convex conjugate](https://en.wikipedia.org/wiki/Convex_conjugate) of the exponential function. For this reason (and the simple fact that ``cvxpy.kl_div`` does exactly what the documentation says it does), I'm making this issue as resolved.

robdmc:
Thanks so much for the explanation.  That makes sense.

RylanSchaeffer:
I'm confused. Is the definition of KL divergence such as in the scipy documentation (https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.kl_div.html) consistent with Wikipedia's definition?

RylanSchaeffer:
[...]

PR #1427, opened 2021-06-25
[New Atom] Relative Entropy (#1406 / #733)

phschiele:
Adding the atom as discussed in #1406 / #733 

- [x] atom
- [x] tests
- [x] documentation

See below for one issue I've encountered while adding the tests.


cvxpy/tests/test_constant_atoms.py
```diff
@@ -328,7 +331,7 @@ def get_indices(size):
     """Get indices for dimension.
     """
     if len(size) == 0:
-        return tuple()
```
phschiele, line 331: Returning `tuple()` here did lead to all tests with `size = tuple()` being silently skipped.
Fixing this has caused some tests unrelated to the PR to fail. 

Some tests fail with errors which seemed to be discussed in older issues already:
```
>           raise TypeError("'h' must be a 'd' matrix of size (%d,1)" %cdim)
E           TypeError: 'h' must be a 'd' matrix of size (9,1)
```
Other tests failed for SCS, but passed for ECOS.
I was also surprised that the expected values for the `cp.norm()` needed to be updated. Were there any updates to the function which would make this expected, such as the default axis?

cvxpy/tests/test_conic_solvers.py
```diff
@@ -211,31 +211,6 @@ def test_complex_matrices(self) -> None:
         self.assertEqual(constraints[2].dual_value.shape, (2, 2))
         self.assertAlmostEqual(sol_scs, n1)
 
-    def test_kl_div(self) -> None:
```
phschiele, line 214: This test seemed to be duplicated from `test_nonlinear_atoms.py`.

I wanted to check if there is a reason to keep this duplication. If not, I can remove some more duplicates in a follow-up.

SteveDiamond: Yeah the only difference I see is the test_nonlinear_atoms tests SCS and ECOS. So if you merge the two tests that's fine.

cvxpy/tests/test_constant_atoms.py
```diff
@@ -135,10 +138,10 @@ def log_sum_exp_axis_1(x): return cp.log_sum_exp(x, axis=1)  # noqa E371
      Constant([5.47722557])),
     (lambda x: cp.norm(x, 1), tuple(), [v_np], Constant([5])),
     (lambda x: cp.norm(x, 1), tuple(), [[[-1, 2], [3, -4]]],
-     Constant([10])),
+     Constant([7])),
```
SteveDiamond, line 141: The change in value here is because cvxpy used to have an idiosyncratic definition of the L1 and LInf norm of matrices, but it was changed to match wikipedia etc.

cvxpy/tests/test_constant_atoms.py
SteveDiamond: That's great you uncovered this issue! That's a lot of errors :( These constant_atoms tests can be surprisingly challenging for solvers. But the cvxopt failures are clearly an input formatting issue.

cvxpy/tests/test_constant_atoms.py
phschiele: Thanks for the clarification. I noticed that it does make a difference it a list of lists or an np.array of those lists is passed:


```py
cp.norm([[-1, 2], [3, -4]], 1).value
>>>> array(7.)
cp.norm(np.array([[-1, 2], [3, -4]]), 1).value
>>>> array(6.)

# Numpy
np.linalg.norm([[-1, 2], [3, -4]],1)
>>>> 6.0
np.linalg.norm(np.array([[-1, 2], [3, -4]]),1)
>>>> 6.0

# Scipy
from scipy.linalg import norm
norm([[-1, 2], [3, -4]],1)
>>>> 6.0
norm(np.array([[-1, 2], [3, -4]]), 1)
>>>> 6.0
```

This is due to the transpose in the following line in the `ndarray_interface`:
https://github.com/cvxpy/cvxpy/blob/693a58a7d784bca7c0b6460f462ec22326555f6d/cvxpy/interface/numpy_interface/ndarray_interface.py#L45-L46

I know there are different conventions around how to parse such lists, so it is most likely expected.

SteveDiamond: Yeah the cvxpy parsing of lists is not aligned with numpy for historical reasons. We should change it, but it would require changing a lot of tests. Taking in lists is not a documented feature of cvxpy so we wouldn't be breaking the API though.

cvxpy/atoms/elementwise/rel_entr.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def is_decr(self, idx) -> bool:
+        """Is the composition non-increasing in argument idx?
+        """
+        if idx == 0:
+            return False
+        else:
+            return True
+
+    def _grad(self, values) -> List[Any]:
```
r-barnes, line 68: You could try `List[Optional[whatever type rel_entr.elemwise_grad_to_diag returns]]`

phschiele: Done in 0d5044a

cvxpy/tests/test_constant_atoms.py
phschiele: I have fixed most of the tests, the three cases below are still failing under CVXOPT only. Strangely, the tests pass for other solvers, and CVXOPT passes for other test cases of these atoms. Perhaps you or @rileyjmurray can give me a pointer for further investigation, as there seemed to be a discussion around this error in [issue 1147](https://github.com/cvxpy/cvxpy/issues/1147).
```py
    (cp.lambda_max, tuple(), [[[2, 0, 0], [0, 3, 0], [0, 0, 1]]], Constant([3])),
...
    (cp.matrix_frac, tuple(), [[1, 2, 3],
                               [[1, 0, 0],
                                [0, 1, 0],
                                [0, 0, 1]]], Constant([14])),
    (cp.matrix_frac, tuple(), [[1, 2, 3],
                               [[67, 78, 90],
                                [78, 94, 108],
                                [90, 108, 127]]], Constant([0.46557377049180271])),
```

phschiele: Looking a bit further, I found that removing the following condition and keeping the `else` case only, all tests pass:
https://github.com/cvxpy/cvxpy/blob/8c1ca2901645497748dccfe65e3b7981bf9bbe7e/cvxpy/reductions/solvers/conic_solvers/cvxopt_conif.py#L123-L130

There is already an [open PR](https://github.com/cvxpy/cvxpy/pull/1250) on this block.

rileyjmurray: @phschiele I'll look at this tonight.

rileyjmurray: @phschiele can you change lines 123 -- 127 to the following?
```
        if len_eq >= A.shape[0]:
            # Then the given optimization problem has no conic constraints.
            # This is certainly a degenerate case, but we'll handle it downstream.
            data[s.G] = None
            data[s.H] = None
```
I believe that will fix things. Let me know if it doesn't! (I made the changes on a branch that's a little old.)

phschiele: Many thanks, @rileyjmurray, all tests are passing after adding this fix.

cvxpy/reductions/dcp2cone/atom_canonicalizers/rel_entr_canon.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def rel_entr_canon(expr, args):
+    shape = expr.shape
+    x = promote(args[0], shape)
+    y = promote(args[1], shape)
+    t = Variable(shape)
+    constraints = [ExpCone(t, x, y)]
+    obj = y - x - t
+    return obj, constraints
```
rileyjmurray, line 29: This is identical to canonicalization for ``kl_div`` (reproduced below)
```
def kl_div_canon(expr, args):
    shape = expr.shape
    x = promote(args[0], shape)
    y = promote(args[1], shape)
    t = Variable(shape)
    constraints = [ExpCone(t, x, y)]
    obj = y - x - t
    return obj, constraints
```
Shouldn't there be some difference between the two?

rileyjmurray: I believe the correct return value here should be ``obj = -t``.

cvxpy/atoms/elementwise/rel_entr.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def __init__(self, x, y) -> None:
+        super(rel_entr, self).__init__(x, y)
+
+    @Elementwise.numpy_numeric
+    def numeric(self, values):
+        x = values[0]
+        y = values[1]
+        # TODO return inf outside the domain
+        return xlogy(x, x/y)
```
rileyjmurray, line 42: SciPy ``rel_entr`` should be used instead. It will correctly identify the domain.

cvxpy/tests/test_nonlinear_atoms.py
```diff
@@ -71,32 +71,54 @@ def test_entr(self) -> None:
...
+
+        constrs = [cvx.sum(v_prob) == 1]
+        rel_entr_prob = cvx.Problem(cvx.Minimize(obj_rel_entr), constrs)
+        p_refProb.value = npSPriors
+        rel_entr_prob.solve(solver=cvx.SCS)
+        rel_entr_prob.solve(solver=cvx.SCS, verbose=True)
+        self.assertItemsAlmostEqual(v_prob.value, npSPriors, places=3)
+        rel_entr_prob.solve(solver=cvx.ECOS, verbose=True)
+        self.assertItemsAlmostEqual(v_prob.value, npSPriors)
+
```
rileyjmurray, line 121: Earlier I made a comment that I think canonicalization of rel_entr is incorrect. Looking at this test I'm more confident of that comment. It may well be that this test passes, but note that it only uses ``rel_entr`` with two probability distributions. In this case ``sum(rel_entr(x,y))`` returns the same value as ``sum(kl_div(x,y))``. A test needs to be written where these atoms are applied without using a probability distribution as input. Let me know if you need help finding such a test problem.

rileyjmurray (changes requested):
I believe ``rel_entr`` is incorrectly being canonicalized, and that tests are passing only because ``sum(rel_entr(x,y))`` and ``sum(kl_div(x,y))`` are the same when applied to probability distributions.

cvxpy/tests/test_nonlinear_atoms.py
phschiele: > I believe `rel_entr` is incorrectly being canonicalized, and that tests are passing only because `sum(rel_entr(x,y))` and `sum(kl_div(x,y))` are the same when applied to probability distributions.

@rileyjmurray I have addressed your comments in the latest commit. The test I have added is quite simple, but it is sensitive to the change in `rel_entr_canon.py` and now gives the expected results. Please let me know if you had a more practical example in mind, though. 

I have also added the SciPy function you mentioned, and I have done the same for the `kl_div` atom.

cvxpy/tests/test_nonlinear_atoms.py
```diff
@@ -71,32 +71,70 @@ def test_entr(self) -> None:
...
+        kl_div_prob.solve()
+        self.assertItemsAlmostEqual(x.value, y.value)
+        self.assertItemsAlmostEqual(kl_div_prob.value, 0)
+
+        rel_entr_prob = cvx.Problem(cvx.Minimize(cvx.rel_entr(x, y)), constraints=[x + y <= 1])
+        rel_entr_prob.solve()
+        self.assertItemsAlmostEqual(x.value, 0.2178117)
+        self.assertItemsAlmostEqual(y.value, 0.7821882)
+        self.assertItemsAlmostEqual(rel_entr_prob.value, -0.278464)
+
```
rileyjmurray, line 137: Looks good. Can you add a comment like the following?
```
"""
Reference solution computed by passing the following command to Wolfram Alpha:
minimize x*log(x/y) subject to {x + y <= 1, 0 <= x, 0 <= y}
"""
```
I don't know how you personally computed the reference solution, but I did just verify it on Wolfram Alpha.

rileyjmurray (commented):
I left one tiny comment about documenting a test. Once that's addressed I believe this is good to merge.

cvxpy/tests/test_nonlinear_atoms.py
phschiele: Exactly, I have used Wolfram Alpha to get external validation. Added the comment in the latest commit.

rileyjmurray (approved):
@akshayka @SteveDiamond @bstellato. Our [governance doc](https://github.com/cvxpy/org/blob/main/governance.md#non-financial-decisions) requires approval from at least half the project maintainers before changes are made to the public API. I'm approving now, but before we merge we'll need approval from one of you as well.

SteveDiamond (approved):
LGTM. Great job @phschiele !

PR #1430, opened 2021-06-29
Specify solvers in tests

phschiele:
As suggested by @SteveDiamond, this PR defines solvers for all tests which currently don't specify a solver.
The idea is that a test should not depend on the users' setups.

### DIFF
--- a/cvxpy/tests/test_derivative.py
+++ b/cvxpy/tests/test_derivative.py
@@ -12,7 +12,7 @@ def perturbcheck(problem, gp: bool = False, delta: float = 1e-5,
     """Checks the analytical derivative against a numerical computation."""
     np.random.seed(0)
     if not problem.parameters():
-        problem.solve(gp=gp, requires_grad=True, eps=eps, **kwargs)
+        problem.solve(solver=cp.DIFFCP, gp=gp, requires_grad=True, eps=eps, **kwargs)
         problem.derivative()
         for variable in problem.variables():
             np.testing.assert_equal(variable.delta, 0.0)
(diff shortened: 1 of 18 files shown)


cvxpy/tests/test_derivative.py
```diff
@@ -277,30 +277,26 @@ def test_infeasible(self) -> None:
         param = cp.Parameter()
         problem = cp.Problem(cp.Minimize(param), [x >= 1, x <= -1])
         param.value = 1
-        try:
-            problem.solve(requires_grad=True)
-        except cp.SolverError:
```
phschiele, line 282: Solving the infeasible/unbounded problems does not raise a solver error, so the below code was not tested.
The errors are only raised once the `backward()` or `derivative()` functions are called.

SteveDiamond:
This looks pretty reasonable. Did you have any particular criteria for using ECOS vs SCS? Also maybe there should be more OSQP in there?

SteveDiamond:
Also for context of others, the idea here is to make the tests fully deterministic no matter what solvers the user has installed. It will also help integrate the MOSEK CI.

phschiele:
> This looks pretty reasonable. Did you have any particular criteria for using ECOS vs SCS? Also maybe there should be more OSQP in there?

I did not have strict criteria for selecting between ECOS/SCS, although I did try many tests with both solvers.
As you can see in the commit history, some tests only passed using ECOS across all environments.

Regarding OSQP, I thought it is getting good coverage through `test_qp_solvers`, but I have further added a few explicit calls in 5e55ca3.

SteveDiamond (approved):
LGTM

SteveDiamond:
Fantastic! Thanks so much! Merging in.

PR #1434, opened 2021-07-01
Fix old print statements

r-barnes:
Fixes some Python 2 print statements, now that Python2 is gone forever.

But since these issues are still here, maybe these files haven't been used in a _long_ time and should be removed?


r-barnes:
@SteveDiamond - Tests are passing.

akshayka:
Let's delete these files, you're right @r-barnes, we don't need them. Good catch!

r-barnes:
@akshayka : Should I remove `cvxpy/cvxcore/tests/python/364A_scripts/**` or `cvxpy/cvxcore/tests/python/**`?

akshayka:
Let's delete `cvxpy/cvxcore/tests/python/**`. The code is unused, and incorrect. (cvxcore has changed a lot since these tests were written.)

In the future we could and perhaps should write proper tests for cvxcore.

r-barnes:
@akshayka : I've removed those files along with `cvxpy/cvxcore/tests/.keep` because there was nothing left in the directory `tests/` subdir. Let me know if I should add that back.

akshayka (approved):

PR #1435, opened 2021-07-01
[CI] Only install Mosek if license is available

phschiele:
Since Github action secrets are not copied to forks, the CI fails on PRs after adding Mosek.

This PR skips the Mosek install in case the secret is not available. It is not an ideal solution, as there is some chance that a test only fails on Mosek after the PR is merged.


SteveDiamond:
LGTM.

PR #1437, opened 2021-07-08
Update performance_utils.py

tschm:
Some cosmetics raised by our linter


SteveDiamond:
Could you explain the motivation here a bit? It looks quite similar.

akshayka (approved):

akshayka:
> Could you explain the motivation here a bit? It looks quite similar.

They look like minor readability changes. LGTM.

PR #1441, opened 2021-07-15
Update SCS status map to use status vals rather than string matching

bodono:
Matching the string is somewhat brittle. Matching the status values should be more robust and more backwards compatible.


bodono:
It looks like this is failing the linter checks, but for lines that I haven't changed. What do you want me to do in this case?

bodono:
Thanks @phschiele

SteveDiamond (approved):
LGTM

PR #1442, opened 2021-07-15
Use builtin types instead of deprecated numpy aliases

### LINKED ISSUE #1334: Do not use numpy deprecated type
When using numpy v1.20.0, the type `numpy.complex` is marked as deprecated.

This is problematic because cvxpy uses it [here](https://github.com/cvxgrp/cvxpy/blob/ef0f6702f7db865e0beb1550477780fd323fbac8/cvxpy/interface/numpy_interface/ndarray_interface.py#L49) (and probably in other places as well).

Would it be possible to change it so this library can be fully compatible with the latest versions of numpy ?

I was thinking of using `numpy.complex_` instead. If the change is that simple i will be happy to open a PR myself.

hashstat:
See https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

Fixes #1334

### DIFF
--- a/cvxpy/atoms/axis_atom.py
+++ b/cvxpy/atoms/axis_atom.py
@@ -78,7 +78,7 @@ class AxisAtom(Atom):
         else:
             m, n = self.args[0].shape
             if self.axis == 0:  # function apply to each column
-                D = sp.csc_matrix((m*n, n), dtype=np.float)
+                D = sp.csc_matrix((m*n, n), dtype=float)
                 for i in range(n):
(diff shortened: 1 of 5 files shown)


SteveDiamond:
Thanks @hashstat this looks good! You need to sign the CLA, or if you can't get that to work you can post something like "I agree to the CLA".

hashstat:
@SteveDiamond I attempted to sign the CLA twice without success. :disappointed: 

I agree to the CLA.

PR #1445, opened 2021-07-20
Support for Gurobi Environments

sjschneider:
Adds support for Gurobi Environments: https://www.gurobi.com/documentation/9.1/refman/environments.html

Environments store configuration parameters, licensing information, and provide other capabilities.  This is critical for Gurobi ISV licenses, which may not have a license file that is set via an environmental variable.

The environment is specified in the cp.solve() method with an optional 'env' argument that is detected in the solver_opts dict.  StandardTests are employed to detect changes to the parameters in the resulting model.  The StandardTest methods were modified to return a SolverTestHelper rather than None.  Although the tests only use two of these methods, all methods were modified to consistency and to encourage re-use of these StandardTests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
@@ -154,7 +154,17 @@ class GUROBI(SCS):
 
         n = c.shape[0]
 
-        model = gurobipy.Model()
+        # Create a new model
+        if 'env' in solver_opts:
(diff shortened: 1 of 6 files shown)


SteveDiamond:
@rileyjmurray what do you think about the solver tests returning the SolverTestHelper? Otherwise looks promising.

sjschneider:
Any outstanding concerns before this can be merged?

SteveDiamond:
This looks fine to me.

PR #1456, opened 2021-07-31
Fix Constant.is_psd() and Constant.is_nsd()

### LINKED ISSUE #1451: quad_form(x,P).is_dcp() is broken
**Describe the bug**
Gives answer `True` for an obviously `False` case.

**To Reproduce**
```
import cvxpy
import numpy as np
P = np.array([[1, 2], [2, 1]])
x = cvxpy.Variable(2)
q = cvxpy.quad_form(x, P)
q.is_dcp()
``` 

**Expected behavior**
Expecting an answer of `False`

**Output**
```
Python 3.8.8 (default, Apr 13 2021, 12:59:45) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import cvxpy as cp
>>> import numpy as np
>>> P = np.array([[1, 2], [2, 1]])
>>> x = cp.Variable(2)
>>> q = cp.quad_form(x, P)
>>> q.is_dcp()
True
>>> print(cp.__version__)
1.1.14
```

**Version**
 - OS: macOS 11.5.1
 - CVXPY Version:  1.1.14

**Additional context**
Version 1.1.1 seems to be the last version that yields `False`.

rileyjmurray:
This fixes #1451. The only change to non-test code is to apply a keyword argument to SciPy's ``eigsh`` so that it returns eigenvalues sorted by value rather than absolute value. Some earlier tests should have caught the bug in #1451, but they tested Parameter objects rather than Constant objects.

We should revisit this in the near future. The largest or smallest eigenvalues returned by ``eigsh`` when running in shift-invert mode are sorted differently than if we don't use shift-invert mode. Although (updated!) tests are passing, we should take a more careful look to make sure we won't run into issues with this different sorting. 

We should also add tests for semidefiniteness based on the Gershgorin Circle Theorem, as I mentioned in #1451.

Edit: I've marked this as work-in-progress to make sure we have appropriate discussions.

### DIFF
--- a/cvxpy/tests/test_problem.py
+++ b/cvxpy/tests/test_problem.py
@@ -1521,7 +1521,11 @@ class TestProblem(BaseTest):
             constraints = [X == [[2, 0], [0, 2]] - C]
             prob = Problem(obj, constraints)
             result = prob.solve(solver=s.CVXOPT)
-            self.assertItemsAlmostEqual(constraints[0].dual_value, psd_constr_dual)
+            new_constr_dual = (constraints[0].dual_value + constraints[0].dual_value.T)/2
+            # ^ Symmetrizing is valid, because the dual variable is with respect to an
+            #   unstructured equality constraint. Dual optimal solutions are non-unique
+            #   in this formulation, up to symmetrizing the dual variable.
+            self.assertItemsAlmostEqual(new_constr_dual, psd_constr_dual)
 
         # Test the dual values with SCS.
         C = Variable((2, 2), symmetric=True)
--- a/cvxpy/tests/test_quad_form.py
+++ b/cvxpy/tests/test_quad_form.py
@@ -137,7 +137,7 @@ class TestNonOptimal(BaseTest):
     def test_psd_exactly_tolerance(self) -> None:
         """Test that PSD check when eigenvalue is exactly -EIGVAL_TOL
         """
-        P = np.array([[-EIGVAL_TOL, 0], [0, 10]])
+        P = np.array([[-0.999*EIGVAL_TOL, 0], [0, 10]])
         x = cvxpy.Variable(2)
         # Forming quad_form is okay
         with warnings.catch_warnings():
@@ -149,7 +149,7 @@ class TestNonOptimal(BaseTest):
     def test_nsd_exactly_tolerance(self) -> None:
         """Test that NSD check when eigenvalue is exactly EIGVAL_TOL
         """
-        P = np.array([[EIGVAL_TOL, 0], [0, -10]])
+        P = np.array([[0.999*EIGVAL_TOL, 0], [0, -10]])
         x = cvxpy.Variable(2)
         # Forming quad_form is okay
         with warnings.catch_warnings():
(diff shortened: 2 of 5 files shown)


SteveDiamond:
It looks good to me. Could you say more about the different sorting? It's not according to size?

rileyjmurray:
@SteveDiamond I've gone back and adjusted the implementation to be clearer about how it works. 

Basically, the problem with the old implementation (the one in cvxpy 1.1.1) was that it labeled some values as ``_top_eig`` and ``_bottom_eig`` when those values weren't actually the max or min eigenvalue of the matrix we were testing. However, those values *were* valid eigenvalues that we could use to determine if a matrix was PSD or NSD up to some tolerance. The old (v1.1.1) implementation basically worked as follows for testing if a matrix is PSD:
* If the matrix was in fact PSD up to our prescribed tolerance, then the ``_bottom_eig`` was being set to the PSD matrix's *maximum* eigenvalue (which would trivially be >= our tolerance).
* If the matrix was *not* PSD up to our tolerance, then ``_bottom_eig`` was equal to the largest eigenvalue of the matrix *that was less than our tolerance*.

This PR's proposed implementation basically takes the same approach, but without using bad variable names. I've left detailed comments in the code explaining what it does.

cvxpy/tests/test_problem.py
```diff
@@ -1521,7 +1521,11 @@ def test_psd_duals(self) -> None:
             constraints = [X == [[2, 0], [0, 2]] - C]
             prob = Problem(obj, constraints)
             result = prob.solve(solver=s.CVXOPT)
-            self.assertItemsAlmostEqual(constraints[0].dual_value, psd_constr_dual)
+            new_constr_dual = (constraints[0].dual_value + constraints[0].dual_value.T)/2
+            # ^ Symmetrizing is valid, because the dual variable is with respect to an
+            #   unstructured equality constraint. Dual optimal solutions are non-unique
+            #   in this formulation, up to symmetrizing the dual variable.
+            self.assertItemsAlmostEqual(new_constr_dual, psd_constr_dual)
```
rileyjmurray, line 1528: This test started to fail after I made a (necessary) change to the CVXOPT interface that affected presolve when using ``chol`` as the KKT solver. The cause of the failure was a non-unique dual optimal solution, as documented here.

cvxpy/tests/test_quad_form.py
```diff
@@ -137,7 +137,7 @@ def test_non_psd(self) -> None:
     def test_psd_exactly_tolerance(self) -> None:
         """Test that PSD check when eigenvalue is exactly -EIGVAL_TOL
         """
-        P = np.array([[-EIGVAL_TOL, 0], [0, 10]])
+        P = np.array([[-0.999*EIGVAL_TOL, 0], [0, 10]])
```
rileyjmurray, line 140: This factor could be more like 0.999999. Unfortunately, with this proposed implementation I couldn't set things up so that *exactly* hitting an eigenvalue tolerance still let the tests pass.

SteveDiamond (approved):
LGTM

PR #1461, opened 2021-08-19
Missing factor of 1/2 in documentation for quad_form

hardikparwana:
A factor of 1/2 missing from documentation for quad_form


hardikparwana:
Sorry, my bad. It was correct.

PR #1463, opened 2021-08-19
Validate constraints on problem creation

### LINKED ISSUE #1462: Strange error handling of constraints
**Describe the bug**
The behavior when accidentally passing in an expression in the list of constraints (as opposed to an equality/inequality), is very odd. I'm not sure what is happening.

**To Reproduce**
I encountered the bug while solving a portfolio optimization problem like below:
```
import cvxpy as cp
import numpy as np

n = 3
np.random.seed(1)
mu = np.random.randn(n)
Sigma = np.random.randn(n, n)
Sigma = Sigma.T @ Sigma

w = cp.Variable(n)
obj = cp.Maximize(mu.T@w - cp.quad_form(w, Sigma))
constr = [cp.sum(w), w >= 0]
prob = cp.Problem(obj, constr)
prob.solve(verbose=True)
```

**Expected behavior**
Some sort of error should be raised in this case. Definitely the problem shouldn't solve!

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

**Version**
 - OS: Mac
 - CVXPY Version: 1.1.15

**Additional context**

akshayka:
Problem canonicalization fails silently when an expression is included
in a problem's list of constraints.

This change modifies the Problem constructor to check that each
constraint is in fact a Constraint (or a bool)

Previously DQCP included lazily evaluated
constraints in the constraint list. This change modifies the DQCP
reduction to store the lazy constraints elsewhere.

Fixes #1462

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -23,6 +23,7 @@ import cvxpy.utilities as u
 import cvxpy.utilities.performance_utils as perf
 from cvxpy import error
 from cvxpy import settings as s, Constant
+from cvxpy.constraints.constraint import Constraint
 from cvxpy.constraints import Equality, Inequality, NonPos, Zero, NonNeg
 from cvxpy.expressions import cvxtypes
 from cvxpy.interface.matrix_utilities import scalar_value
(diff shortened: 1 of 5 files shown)


SteveDiamond (approved):
LGTM

rileyjmurray (approved):
Looks good to me as well!

PR #1476, opened 2021-09-14
Better log in mosek

SCUTJcfeng:


SteveDiamond:
Sorry I didn't notice this PR. Can you comment on why this is better?

SCUTJcfeng:
Sorry I didn't comment on this PR...

I prefer to use `s.LOGGER` since default mosek logger is stdout, by using `s.LOGGER`, I can redirect output to log file.

Issue #1474, opened 2021-09-11
SCS requirement issue


SteveDiamond:
**Describe the bug**
Currently SCS 1.1.6 is the minimum required version in CVXPY. However, the ``acceleration_lookback`` keyword option is used internally (https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/scs_conif.py#L312). This I believe was added in SCS 2.0. For SCS 1.1.6 the second solve causes an error because the keyword is not supported.

We should either update the minimum required SCS version or gate the second SCS solve on having SCS >= 2.0.

**To Reproduce**
The issue was forwarded to me so I don't have the code to reproduce. Any example that triggers the second SCS solve will cause the bug.

**Expected behavior**
We should not have the error due to the unsupported ``acceleration_lookback`` keyword.

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

**Version**
 - OS: N/A
 - CVXPY Version: 1.1.15

**Additional context**

akshayka:
> We should either update the minimum required SCS version or gate the second SCS solve on having SCS >= 2.0.

Let's gate the second solve on having SCS >= 2.0. It was only introduced due to instability with AA, so it's not relevant to older versions of SCS.

PR #1481, opened 2021-09-26
Fix SCS requirement issue (Issue #1474)

phschiele:
Issue Link: #1474 

#### MWE to trigger retry

```py
import numpy as np
import cvxpy as cp

n_sec = 20
np.random.seed(315)
mu = np.random.random(n_sec)
random_mat = np.random.rand(n_sec, n_sec)
C = np.dot(random_mat, random_mat.transpose())

x = cp.Variable(n_sec)
prob = cp.Problem(cp.Minimize(cp.QuadForm(x, C)), [cp.sum(x) == 1, 0 <= x, x <= 1, x @ mu >= np.max(mu) - 1e-6])
prob.solve(cp.SCS)
```

Tested under SCS version `2.1.4` and `1.2.6`.


SteveDiamond:
Looks great! Thanks @phschiele !

PR #1484, opened 2021-09-27
Implement trivial inverses for some more atoms

akshayka:
Edge case in which a user takes a sum/cumsum of a scalar atom.

### DIFF
--- a/cvxpy/reductions/dqcp2dcp/inverse.py
+++ b/cvxpy/reductions/dqcp2dcp/inverse.py
@@ -17,11 +17,12 @@ from cvxpy import atoms
 from cvxpy.atoms.affine.add_expr import AddExpression
 from cvxpy.atoms.affine.binary_operators import DivExpression
 from cvxpy.atoms.affine.unary_operators import NegExpression
+from cvxpy.atoms.affine.sum import Sum
 import numpy as np
 
 INVERTIBLE = set(
     [atoms.ceil, atoms.floor, NegExpression, atoms.exp, atoms.log, atoms.log1p,
-     atoms.logistic, atoms.power, atoms.abs])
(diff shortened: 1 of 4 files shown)


akshayka:
Some context: The basic issue is that DQCP allows the root of an expression tree to be transformed by monotone increasing (or decreasing) real functions. The reduction just inverts these functions. Looks like we missed a few inverses in the original implementation.

This part of the grammar is silly, though, because all such a transformation does is yield an equivalent problem. We included it for completeness, but if I were to do it again I'd not allow monotone transformations. Allowing them adds little value, but adds significant maintenance burden ... Perhaps something for CVXPY 2.0 ...

cvxpy/reductions/dqcp2dcp/inverse.py
```diff
@@ -17,11 +17,12 @@
 from cvxpy.atoms.affine.add_expr import AddExpression
 from cvxpy.atoms.affine.binary_operators import DivExpression
 from cvxpy.atoms.affine.unary_operators import NegExpression
+from cvxpy.atoms.affine.sum import Sum
 import numpy as np
 
 INVERTIBLE = set(
     [atoms.ceil, atoms.floor, NegExpression, atoms.exp, atoms.log, atoms.log1p,
-     atoms.logistic, atoms.power, atoms.abs])
+     atoms.logistic, atoms.power, atoms.abs, atoms.cumsum])
```
SteveDiamond, line 25: Why no sum here?

SteveDiamond (approved):
LGTM

cvxpy/reductions/dqcp2dcp/inverse.py
akshayka: Added some comments.

SteveDiamond (approved):

PR #1436, opened 2021-07-04
Sort imports using isort

r-barnes:
Imports are currently ordered somewhat haphazardly: they are not always grouped by type (system, third-party, cvxpy), they are not always in alphabetical order, multiple `from` imports from the same library are not grouped, &c.

This PR applies [isort](https://pypi.org/project/isort/) to the entire cvxpy project to produce a consistent import ordering.

Though large, the PR consists entirely of auto-generated changes achieved by running
```bash
isort --up --tc .
```
in the project root.

PR #1488, opened 2021-09-30
Sort imports

SteveDiamond:
This is replicating #1436 to handle merge conflicts. Imports are sorted by applying ``isort --up --tc .`` to the project root.


SteveDiamond:
@rileyjmurray @akshayka would be great to get one of your opinions on this. If we merge, then going forward we should use ``isort --up --tc`` on new files we add.

PR #1490, opened 2021-10-01
Fixes for isort2 branch

phschiele:
I've commented below the root causes for the different failures (except some whitespace fixes for flake8).

### DIFF
--- a/cvxpy/__init__.py
+++ b/cvxpy/__init__.py
@@ -25,6 +25,7 @@ from cvxpy.expressions.expression import Expression
 from cvxpy.expressions.variable import Variable
 from cvxpy.problems.objective import Maximize, Minimize
 from cvxpy.problems.problem import Problem
+from cvxpy.transforms import linearize, partial_optimize, suppfunc
 from cvxpy.reductions import *
 from cvxpy.reductions.solvers.defines import installed_solvers
 from cvxpy.settings import (CBC, CPLEX, CVXOPT, DIFFCP, ECOS, ECOS_BB, GLPK,
@@ -33,4 +34,3 @@ from cvxpy.settings import (CBC, CPLEX, CVXOPT, DIFFCP, ECOS, ECOS_BB, GLPK,
                             ROBUST_KKTSOLVER, SCIP, SCIPY, SCS, SOLVER_ERROR,
                             UNBOUNDED, UNBOUNDED_INACCURATE, USER_LIMIT,
                             XPRESS, get_num_threads, set_num_threads,)
-from cvxpy.transforms import linearize, partial_optimize, suppfunc
--- a/cvxpy/reductions/__init__.py
+++ b/cvxpy/reductions/__init__.py
@@ -14,6 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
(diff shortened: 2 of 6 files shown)


cvxpy/reductions/__init__.py
```diff
@@ -14,6 +14,10 @@
 limitations under the License.
 """
 
+from cvxpy.reductions.inverse_data import InverseData
+from cvxpy.reductions.reduction import Reduction
+from cvxpy.reductions.solution import Solution
```
phschiele, line 19: This prevents a circular import.

cvxpy/reductions/dcp2cone/atom_canonicalizers/__init__.py
```diff
@@ -26,8 +25,8 @@
 from cvxpy.reductions.dcp2cone.atom_canonicalizers.kl_div_canon import *
 from cvxpy.reductions.dcp2cone.atom_canonicalizers.lambda_max_canon import *
 from cvxpy.reductions.dcp2cone.atom_canonicalizers.lambda_sum_largest_canon import *
-from cvxpy.reductions.dcp2cone.atom_canonicalizers.log1p_canon import *
 from cvxpy.reductions.dcp2cone.atom_canonicalizers.log_canon import *
+from cvxpy.reductions.dcp2cone.atom_canonicalizers.log1p_canon import *
```
phschiele, line 29: `log1p_canon` is defined based on `log_canon`, so it cant be imported first.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -139,7 +139,7 @@ def test_scs_retry(self) -> None:
                            x <= 1,
                            x @ mu >= np.max(mu) - 1e-6])
         prob.solve(cp.SCS)
-        assert prob.status == cp.OPTIMAL
+        assert prob.status in {cp.OPTIMAL, cp.OPTIMAL_INACCURATE}
```
phschiele, line 142: On my local machine, the solution was inaccurate, but the test is just intended to show that it does not crash (as noted in the docstring.

cvxpy/__init__.py
```diff
@@ -25,6 +25,7 @@
 from cvxpy.expressions.variable import Variable
 from cvxpy.problems.objective import Maximize, Minimize
 from cvxpy.problems.problem import Problem
+from cvxpy.transforms import linearize, partial_optimize, suppfunc
 from cvxpy.reductions import *
```
phschiele, line 29: This one was quite implicit.
Importing everything from `reductions`, also makes `cvxpy/reductions/utilities` available as `cvxpy/utilities`. 
In `partial_optimize.py`, `import cvxpy.utilities as u` then imports `cvxpy/reductions/utilities.py` instead of `cvxpy/reductions/utilities.py`.

I could follow up on this by removing `*` imports.

SteveDiamond: Yeah probably best to remove * imports.

SteveDiamond (approved):

SteveDiamond:
Thanks for redoing your fixes! All looks good. Probably should replace ``from cvxpy.reductions import *`` with explicit imports.

PR #1526, opened 2021-11-03
Prevent bad Xpress license from breaking cvxpy

ghackebeil:
On October 31st, the free Community license that ships with Xpress 8.11.0 expired. We observed that even if one is not explicitly using Xpress in cvxpy, the exception that is raised when `import xpress` is executed (to populate `INSTALLED_SOLVERS` in `reductions/solvers/defines.py`) ends up making all of cvxpy unusable. For example, a simple import of cvxpy leads to obscure import errors like:
```python
>>> import cvxpy
  File "~/cvxpy/cvxpy/__init__.py", line 27, in <module>
    from cvxpy.problems.problem import Problem
  File "~/cvxpy/cvxpy/problems/problem.py", line 24, in <module>
    import cvxpy.utilities as u
AttributeError: module 'cvxpy' has no attribute 'utilities'
```

 This can be avoided by simply uninstalling the `xpress` package, but that may not always be as easy as it sounds on deployed production systems.

This PR includes Xpress-specific fixes in `xpress_conif.py` and `xpress_qpif.py` to mask against the various exceptions that get raised depending on whether or not it is the first attempt of `import xpress` or a subsequent attempt. For the first import attempt, a stack trace like the following shows up:
```python
>>> import xpress
  File "~/lib/python3.6/site-packages/xpress/__init__.py", line 256, in <module>
    libver = getversion()
xpress.InterfaceError: Xpress licensing error 10: Your license expired on 31-oct-2021. Please contact support@fico.com to renew it.
```
and on subsequent attempts:
```python
>>> import xpress
  File "~/lib/python3.6/site-packages/xpress/__init__.py", line 162, in <module>
    __version__ = __version_file__.__version__
NameError: name '__version_file__' is not defined
```
There really isn't a way to explicitly refer to `xpress.InterfaceError` in a `try: except:` block for this scenario because the `import xpress` line fails and leaves one without access to the xpress namespace. Therefore, we catch a broader exception type and re-raise it as the exception type handled in `reductions.solvers.solver.is_installed()` (shown below):
```python
    def is_installed(self) -> bool:
        """Is the solver installed?                                                                                                                                       
        """
        try:
            self.import_solver()
            return True
        except ModuleNotFoundError:
            return False
```
An alternative to this PR would be to just catch a broader exception type in the method above (e.g., `Exception`). Whatever the case, cvxpy should not be broken by a third-party package when it is not explicitly used by a user.

**Update:**
This PR was modified to handle a broader exception type in `reductions.solvers.solver.is_installed()`, logging a warning when something other than `ModuleNotFoundError` is caught. Thanks for the quick discussions @akshayka and @SteveDiamond!

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
@@ -56,8 +56,11 @@ class XPRESS(ConicSolver):
     def import_solver(self) -> None:
         """Imports the solver.
         """
-        import xpress
-        self.version = xpress.getversion()
+        try:
+            import xpress
+            self.version = xpress.getversion()
+        except Exception as e:
+            raise ModuleNotFoundError(str(e))
 
     def accepts(self, problem) -> bool:
         """Can Xpress solve the problem?
--- a/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
@@ -56,11 +56,8 @@ class XPRESS(ConicSolver):
     def import_solver(self) -> None:
         """Imports the solver.
         """
-        try:
-            import xpress
-            self.version = xpress.getversion()
-        except Exception as e:
-            raise ModuleNotFoundError(str(e))
+        import xpress
+        self.version = xpress.getversion()
 
     def accepts(self, problem) -> bool:
         """Can Xpress solve the problem?
--- a/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
@@ -22,11 +22,8 @@ class XPRESS(QpSolver):
 
     def import_solver(self) -> None:
 
-        try:
-            import xpress
-            xpress  # Prevents flake8 warning
-        except Exception as e:
-            raise ModuleNotFoundError(str(e))
+        import xpress
+        xpress  # Prevents flake8 warning
 
     def apply(self, problem):
         """Returns a new problem and data for inverting the new solution.
--- a/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py
@@ -22,8 +22,11 @@ class XPRESS(QpSolver):
 
     def import_solver(self) -> None:
 
-        import xpress
-        xpress  # Prevents flake8 warning
+        try:
+            import xpress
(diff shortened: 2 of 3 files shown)


SteveDiamond:
Thanks for figuring this out! I would vote for broadening the exception class handled by ``is_installed``.

ghackebeil:
I'm okay with that as well. I can confirm just changing `ModuleNotFoundError` to `Exception` in `reductions.solvers.solver.is_installed()` fixes this issue for me locally. Feel free to close this PR and push that to master, or just let me know if I should update my PR.

akshayka:
> Thanks for figuring this out! I would vote for broadening the exception class handled by `is_installed`.

We should think about this a bit more, because we recently merged in a PR that tightened the error from `ImportError` to `ModuleNotFoundError`. Too broad exceptions can hide problems unrelated to the interface. Let me find the PR ...

akshayka:
How about we broaden the exception type (so CVXPY can still be used when solver installations are faulty), but also log the error messages associated with failed imports that throw something other than `ModuleNotFoundError`, so that the user can be informed that their solver installation is broken?

akshayka (approved):
LGTM. Thanks!

ghackebeil:
Just pushed something to handle this. It produces output like the following in this situation:
```python
>>> import cvxpy
Using the Community license in this session. If you have a full Xpress license, first set the XPAUTH_PATH environment variable to the full path to your license file, xpauth.xpr, and then restart Python. If you want to use the FICO Community license and no longer want to see this message, set the XPAUTH_PATH environment variable to: /Users/ghackebeil/.pyenv/versions/3.6.5/envs/geli/lib/python3.6/site-packages/xpress/license/community-xpauth.xpr
NB: setting XPAUTH_PATH will also affect any other Xpress products installed on your system.
(CVXPY) Nov 03 02:05:32 PM: Encountered unexpected exception importing solver XPRESS:
InterfaceError('Xpress licensing error 10: Your license expired on 31-oct-2021. Please contact support@fico.com to renew it.',)
(CVXPY) Nov 03 02:05:32 PM: Encountered unexpected exception importing solver XPRESS:
NameError("name '__version_file__' is not defined",)
```
You see two warnings because there are two xpress solver interfaces (with the same name).

ghackebeil:
Just while I'm thinking about it, long term it might make more sense to use some lazy system for checking solver availability (i.e., do not attempt any is_installed() calls until some code requests to use a particular solver).

rileyjmurray:
@ghackebeil you raise a good point, but I'm inclined to merge this as-is. We can revisit lazy imports in the future. @akshayka if you concur then you can go ahead and merge.

PR #1531, opened 2021-11-05
Edits to doc homepage

akshayka:
* Add meta description to homepage.
* Greater emphasis on community.
* Highlight Github discussions.

I think it's beneficial for the community to have two forums, one for real-time conversation (Discord), and one for slower, more in-depth conversation (Discussions). Discussions also frees us from SO's moderation, which sometimes closes questions that I think are relevant (such as this one: https://stackoverflow.com/questions/69768609/constraining-a-variable-to-convex-combination-of-two-variables-cvxpy).

We have already gotten some interesting feature requests from Discussions, which have been promoted to Github issues.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -6,15 +6,19 @@
 Welcome to CVXPY 1.1
 ====================
 
+.. meta::
+   :description: An open source Python-embedded modeling language for convex optimization problems.
+                 Express your problem in a natural way that follows the math.
+   :keywords: convex optimization, open source, software,
+
 **Convex optimization, for everyone.**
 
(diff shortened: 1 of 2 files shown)


doc/source/index.rst
```diff
@@ -6,15 +6,19 @@
 Welcome to CVXPY 1.1
 ====================
 
+.. meta::
```
SteveDiamond, line 9: Thanks for figuring this out! I tried chrome lighthouse yesterday and learned we are missing a meta label.

SteveDiamond (approved):
LGTM

PR #1534, opened 2021-11-06
Gershgorin circle theorem PSD/NSD checks

### LINKED ISSUE #1532: ARPARCK no convergence on the identity matrix
@taozuoqiao pointed out in #1424 that ARPACK can fail to converge on the identity matrix. Moreover, this failure occurs randomly (see below). We should perform PSD/NSD checks by using the Gershgorin Circle Theorem to avoid any failures in detecting when a diagonal matrix is PSD/NSD within a specified tolerance.
```python
In [1]: import numpy as np

In [2]: from scipy.sparse.linalg import eigsh

In [3]: x = np.identity(600)
   ...: n_failed = 0
   ...: for i in range(1000):
   ...:     try:
   ...:         eigsh(x, k=1, which='SA', sigma=-1e-10)
   ...:     except:
   ...:         n_failed += 1
   ...: print(n_failed)
40

In [4]: x = np.identity(600)
   ...: n_failed = 0
   ...: for i in range(1000):
   ...:     try:
   ...:         eigsh(x, k=1, which='SA')
   ...:     except:
   ...:         n_failed += 1
   ...: print(n_failed)
8

```
I'll create a PR soon which addresses diagonal (and diagonally-dominant) matrices.

rileyjmurray:
This resolves #1532 by starting checks for PSD/NSD-ness with the method in https://en.wikipedia.org/wiki/Gershgorin_circle_theorem. If that test passes, then we bypass ARPACK.

### DIFF
--- a/cvxpy/utilities/eigvals.py
+++ b/cvxpy/utilities/eigvals.py
@@ -1,0 +1,103 @@
+import numpy as np
+import scipy.sparse as spar
+import scipy.sparse.linalg as sparla
+
+
+def is_psd_within_tol(A, tol):
+    """
+    Return True if we can certify that A is PSD (up to tolerance "tol").
+
+    First we check if A is PSD according to the Gershgorin Circle Theorem.
+
+    If Gershgorin is inconclusive, then we use an iterative method (from ARPACK,
+    as called through SciPy) to estimate extremal eigenvalues of certain shifted
+    versions of A. The shifts are chosen so that the signs of those eigenvalues
+    tell us the signs of the eigenvalues of A.
+
+    If there are numerical issues then it's possible that this function returns
+    False even when A is PSD. If you know that you're in that situation, then
+    you should replace A by
+
+        A = cvxpy.atoms.affine.wraps.psd_wrap(A).
+
+    Parameters
+    ----------
+    A : Union[np.ndarray, spar.spmatrx]
+        Symmetric (or Hermitian) NumPy ndarray or SciPy sparse matrix.
+
+    tol : float
+        Nonnegative. Something very small, like 1e-10.
+    """
+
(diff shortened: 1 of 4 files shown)


cvxpy/utilities/eigvals.py
```diff
@@ -0,0 +1,103 @@
...
+    Otherwise, return False.
+    """
+    if isinstance(A, spar.spmatrix):
+        diag = A.diagonal()
+        if np.any(diag < -tol):
+            return False
+        A_shift = A - spar.diags(diag)
+        A_shift = np.abs(A_shift)
+        radii = np.array(A_shift.sum(axis=0)).ravel()
+        return np.all(diag - radii >= -tol)
```
SteveDiamond, line 93: Is this right? From wikipedia it looks more like the condition ``np.all(diag + radii > -tol)``.

rileyjmurray: The discs are centered at ``diag`` and then range as low as ``diag - radii`` and as high as ``diag + radii``. Since we're checking if the matrix is PSD, we only need to verify a lower bound on ``diag - radii``.

rileyjmurray: As a sanity check: if you run this test with the matrix of all -1's (which is NSD), then ``diag + radii`` would be positive.

SteveDiamond: But isn't your test a sufficient but not necessary condition for PSD?

SteveDiamond: Nvm, I see that you're using it as a sufficient but necessary condition haha. Ok LGTM.

SteveDiamond (approved):

PR #1535, opened 2021-11-06
User-facing "infeasible or unbounded" status

### LINKED ISSUE #1071: Add a user-facing "infeasible or unbounded" problem status; update solver interfaces
In PR #1065 there was some discussion on how ``test_lp_3`` is failing for GUROBI on CVXPY's master branch. The issue is that under default settings, GUROBI employs a presolve feature which can make it impossible to distinguish between certificates of infeasibility and certificates of unboundedness. CPLEX apparently employs a similar presolve procedure, and currently maps "infeasible or unbounded" to ["unbounded_inaccurate"](https://github.com/cvxgrp/cvxpy/blob/c451fea281dc786dd21b1e7d58f02f3df274fa4c/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py#L195-L197). The pending PR for SCIP (#1035) also maps "infeasible or unbounded" to "unbounded_inaccurate".

It's not appropriate for us to parse an "infeasible or unbounded" status into either "infeasible" or "unbounded" (with or without an "inaccurate" suffix). The most direct solutions are (1) raise a solver failure, or (2) create a new CVXPY status for this situation. Another solution is for CVXPY to automatically re-solve the problem with presolve turned off, before ever returning to the user.

It would be useful if developers of affected solvers could provide opinions on possible ways to proceed. @rluce is with GUROBI, @rkersh is with CPLEX, and @rimaddo is the SCIP contributor. Also pinging @SteveDiamond and @akshayka.

rileyjmurray:
This resolves #1071.  I added a user-facing ``INFEASIBLE_OR_UNBOUNDED`` status. I also changed GUROBI and CPLEX interfaces (conic and QP) to automatically re-solve when they encounter such a status. The documentation has been updated so that users can set ``reoptimize=False`` when using these solvers and get the ``INFEASIBLE_OR_UNBOUNDED`` status directly. I'm open to discussion on what the default behavior here should actually be.

### DIFF
--- a/cvxpy/reductions/solution.py
+++ b/cvxpy/reductions/solution.py
@@ -39,7 +39,7 @@ def failure_solution(status, attr=None) -> "Solution":
         opt_val = None
     if attr is None:
         attr = {}
-    if status == s.UNBOUNDED_INACCURATE:
+    if status == s.INFEASIBLE_OR_UNBOUNDED:
         attr['message'] = """
         The problem is either infeasible or unbounded, but the solver
         cannot tell which. Disable any solver-specific presolve methods
--- a/cvxpy/reductions/solution.py
+++ b/cvxpy/reductions/solution.py
@@ -17,6 +17,15 @@ import numpy as np
 
 import cvxpy.settings as s
 
+INF_OR_UNB_MESSAGE = """
+    The problem is either infeasible or unbounded, but the solver
+    cannot tell which. Disable any solver-specific presolve methods
+    and re-solve to determine the precise problem status.
+
+    For GUROBI and CPLEX you can automatically perform this re-solve
+    with the keyword argument prob.solve(reoptimize=True, ...).
+    """
+
 
 def failure_solution(status, attr=None) -> "Solution":
     """Factory function for infeasible or unbounded solutions.
@@ -40,14 +49,7 @@ def failure_solution(status, attr=None) -> "Solution":
     if attr is None:
         attr = {}
     if status == s.INFEASIBLE_OR_UNBOUNDED:
-        attr['message'] = """
-        The problem is either infeasible or unbounded, but the solver
-        cannot tell which. Disable any solver-specific presolve methods
-        and re-solve to determine the precise problem status.
-
-        For GUROBI and CPLEX you can automatically perform this re-solve
-        with the keyword argument prob.solve(reoptimize=True, ...).
-        """
+        attr['message'] = INF_OR_UNB_MESSAGE
     return Solution(status, opt_val, {}, {}, attr)
 
 
--- a/cvxpy/reductions/solution.py
+++ b/cvxpy/reductions/solution.py
@@ -39,6 +39,12 @@ def failure_solution(status, attr=None) -> "Solution":
         opt_val = None
     if attr is None:
         attr = {}
+    if status == s.UNBOUNDED_INACCURATE:
+        attr['message'] = """
+        The problem is either infeasible or unbounded, but the solver
+        cannot tell which. Disable any solver-specific presolve methods
+        and re-solve to determine the precise problem status.
+        """
     return Solution(status, opt_val, {}, {}, attr)
 
 
--- a/cvxpy/reductions/solution.py
+++ b/cvxpy/reductions/solution.py
@@ -44,6 +44,9 @@ def failure_solution(status, attr=None) -> "Solution":
         The problem is either infeasible or unbounded, but the solver
         cannot tell which. Disable any solver-specific presolve methods
         and re-solve to determine the precise problem status.
+
+        For GUROBI and CPLEX you can automatically perform this re-solve
+        with the keyword argument prob.solve(reoptimize=True, ...).
         """
     return Solution(status, opt_val, {}, {}, attr)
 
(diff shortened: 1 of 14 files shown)


cvxpy/reductions/solution.py
```diff
@@ -39,6 +39,12 @@ def failure_solution(status, attr=None) -> "Solution":
         opt_val = None
     if attr is None:
         attr = {}
+    if status == s.UNBOUNDED_INACCURATE:
```
rileyjmurray, line 42: This should be INFEASIBLE_OR_UNBOUNDED. I'll fix.

doc/source/tutorial/intro/index.rst
```diff
@@ -137,13 +137,22 @@ CVXPY provides the following constants as aliases for the different status strin
...
-For example, to test if a problem was solved successfully, you would use
+To test if a problem was solved successfully, you would use
 
 .. code:: python
 
     prob.status == OPTIMAL
 
+The status ``INFEASIBLE_OR_UNBOUNDED`` is rare. It's used when a solver was able to
+determine that the problem was either infeasible or unbounded, but could not tell which.
+You can determine the precise status by re-solving the problem where you where you
```
SteveDiamond, line 150: typo "where you where you"

cvxpy/reductions/solvers/conic_solvers/nag_conif.py
```diff
@@ -140,14 +140,10 @@ def invert(self, solution, inverse_data):
...
         sln = solution['sln']
 
-        opt_val = None
-        primal_vars = None
-        dual_vars = None
         attr = {}
         if status in s.SOLUTION_PRESENT:
             opt_val = sln.rinfo[0] + inverse_data[s.OBJ_OFFSET]
             nr = inverse_data['nr']
-            x = [0.0] * nr
```
SteveDiamond, line 150: Did you test the NAG solver? It's not being test in the CI, right?

rileyjmurray: It's not being tested on CI and I don't have it locally. This line I removed was to an unreferenced variable (the symbol was reassigned on the next line without ever being used). I can revert changes to NAG if you think that's best.

SteveDiamond: No that sounds fine, just curious.

SteveDiamond (approved):

phschiele:
Hi @rileyjmurray,

I was encountering this issue "in the wild" this week. The problem showed up as unbounded, when it was actually infeasible, which took me a second to figure out. This PR will make such issues easier to debug going forward, so many thanks for the PR! :)

I have replayed the encountered problem using this branch. When setting `verbose=True`, I got an error while trying to print 
https://github.com/cvxpy/cvxpy/blob/8b7f662ab7eacdde8675838e3d581950ede43775/cvxpy/problems/problem.py#L982

This seems to be because `failure_solution()` returns `opt_val = None`, when it returned `opt_val = -np.inf` before.

akshayka:
> This resolves #1071. I added a user-facing `INFEASIBLE_OR_UNBOUNDED` status. I also changed GUROBI and CPLEX interfaces (conic and QP) to automatically re-solve when they encounter such a status.

I am mildly concerned that re-optimizing by default will be problematic for people using these solvers for real-time applications. With this change, they might see some solves taking twice as long as before.

I feel like it might be less disruptive to start with `reoptimize=False`, and have the error message inform the user that they can turn on reoptimization. That said, I don't feel very strongly about this ... What do you think?

We do re-optimize for inaccurate SCS solutions. But we did that because we found that AA could be unstable, to the point that very incorrect solutions were returned.

rileyjmurray:
> I am mildly concerned that re-optimizing by default will be problematic for people using these solvers for real-time applications. With this change, they might see some solves taking twice as long as before.
> 
> I feel like it might be less disruptive to start with `reoptimize=False`, and have the error message inform the user that they can turn on reoptimization. That said, I don't feel very strongly about this ... What do you think?

I think those concerns are valid. Some people might even have contingency plans to handle that scenario (e.g., they already know that their problem could only be unbounded). @SteveDiamond @phschiele what do you think?

rileyjmurray:
> Hi @rileyjmurray,
> 
> I was encountering this issue "in the wild" this week. The problem showed up as unbounded, when it was actually infeasible, which took me a second to figure out. This PR will make such issues easier to debug going forward, so many thanks for the PR! :)
> 
> I have replayed the encountered problem using this branch. When setting `verbose=True`, I got an error while trying to print
> 
> https://github.com/cvxpy/cvxpy/blob/8b7f662ab7eacdde8675838e3d581950ede43775/cvxpy/problems/problem.py#L982
> 
> This seems to be because `failure_solution()` returns `opt_val = None`, when it returned `opt_val = -np.inf` before.

Thanks for catching that! I'll fix before we merge.

SteveDiamond:
> > I am mildly concerned that re-optimizing by default will be problematic for people using these solvers for real-time applications. With this change, they might see some solves taking twice as long as before.
> > I feel like it might be less disruptive to start with `reoptimize=False`, and have the error message inform the user that they can turn on reoptimization. That said, I don't feel very strongly about this ... What do you think?
> 
> I think those concerns are valid. Some people might even have contingency plans to handle that scenario (e.g., they already know that their problem could only be unbounded). @SteveDiamond @phschiele what do you think?

I would vote for a clear warning and reoptimize=False. Definitely we don't want reoptimize without some kind of user awareness, either setting a flag manually or a warning that the reoptimize is happening.

rileyjmurray:
@akshayka @SteveDiamond the default behavior is now ``reoptimize=False``. A warning is raised when a user encounters an ``INFEASIBLE_OR_UNBOUNDED`` status. @phschiele's comment is also taken care of. Merge when ready!

PR #1539, opened 2021-11-09
Update Expression.curvature to include log-log curvature

### LINKED ISSUE #1350: Log-log curvature should be printed for non-DCP expressions
The expression

```
import cvxpy as cp
x = cp.Variable(pos=True)

x*x*x
```

is log-log affine, but

```
repr(x*x*x)
```

prints

```
'Expression(UNKNOWN, NONNEGATIVE, ())'
```

even though

```
print((x*x*x).log_log_curvature)
```

prints

```
LOG-LOG AFFINE
```

(#1348)

akshayka:
Fixes #1350

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -142,6 +142,12 @@ class Expression(u.Canonical):
             curvature_str = s.CONVEX
         elif self.is_concave():
             curvature_str = s.CONCAVE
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
This is fine with me. I agree we need something like #1540 to handle cases with multiple types of curvatures.

PR #1542, opened 2021-11-09
Specify tests paths

phschiele:
Adding this config, one can call
```
pytest
```
instead of 
```
pytest cvxpy/tests
```

The former also so far added tests from `cvxpy/examples/extensions/mixed_integer/tests/test_vars.py`, which fail (at least in my case).
Additionally, the tests from the examples folder could be moved to `cvxpy/tests`, if they are still relevant.


SteveDiamond (approved):
Nice!

Issue #1543, opened 2021-11-10
Python 3.10 compatibility


h-vetinari:
Conda-forge is currently doing the migration of packages for CPython 3.10, the PR for cvxpy is [here](https://github.com/conda-forge/cvxpy-feedstock/pull/66).

Things build and run quite alright, with only a single test failure that seems to be due to changed error messages:
```
================================== FAILURES ===================================
______________________ TestAtoms.test_elemwise_arg_count ______________________

self = <tests.test_atoms.TestAtoms testMethod=test_elemwise_arg_count>

    def test_elemwise_arg_count(self) -> None:
        """Test arg count for max and min variants.
        """
        with self.assertRaises(Exception) as cm:
            cp.maximum(1)
>       self.assertTrue(str(cm.exception) in (
            "__init__() takes at least 3 arguments (2 given)",
            "__init__() missing 1 required positional argument: 'arg2'"))
E       AssertionError: False is not true

cvxpy\tests\test_atoms.py:298: AssertionError
```

SteveDiamond:
Fixed on master.

PR #1544, opened 2021-11-10
Fix test failing under Python 3.10

phschiele:
Issue: #1543 

Python 3.10 prepends the class name to the `__init__()` in the error message, so I use a slightly more generic test.

This should unblock @h-vetinari. As a follow-up, Python 3.10 could also be included in the CI.

### DIFF
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -17,6 +17,7 @@ limitations under the License.
 import unittest
 
 import numpy as np
+import pytest
 import scipy
 import scipy.sparse as sp
 import scipy.stats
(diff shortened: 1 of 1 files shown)


h-vetinari:
Thanks a lot! :)

I tested this in https://github.com/conda-forge/cvxpy-feedstock/pull/66, and the 3.10 builds are passing now. 🥳 

Any ETA when this patch might land? I don't mind carrying it until the next release, but it'd be nice to have a sign-off from the core maintainers before we carry long-running patches...

SteveDiamond (approved):

SteveDiamond:
This fix will be in the next patch, which will be within a month for sure.

SteveDiamond:
We could do a patch sooner if it's important.

h-vetinari:
> 
> 
> We could do a patch sooner if it's important.

There's no need to backport, I can carry the patch on the cvxpy-feedstock until then.

Issue #1541, opened 2021-11-09
Misleading guidance on install page.


SteveDiamond:
**Describe the bug**
The install instructions say to run ``pytest cvxpy/tests`` to test the installation, both via pip and conda. But does this actually work? It doesn't work on my machine. This seems like something that would only work if you installed from source.

I would suggest we drop the instructions for testing.

**To Reproduce**
Try ``pytest cvxpy/tests``

**Expected behavior**
The tests should be run.

**Output**
N/A

**Version**
 - OS: Mac 11.6
 - CVXPY Version: 1.1.17

**Additional context**
Another issue is that for some installations ``pytest`` is not recognized as a command and ``python -m pytest`` must be used instead.

h-vetinari:
It should be possible to do something like `pytest --pyargs cvxpy` without relying on the exact repo layout

phschiele:
@h-vetinari Thanks for the tip, that does work and is better than pointing to the exact package location when installed via pip!

PR #1546, opened 2021-11-11
Update install docs

phschiele:
Issue: #1541

Some small remarks:
- Needed to update two tests to make them work without additional MIP solvers
- The performance tests are not run at the moment

I have tested the pip install using a minimal Dockerfile:

```
FROM python:3.8
# Use any version 3.6-3.10

RUN git clone https://github.com/cvxpy/cvxpy
WORKDIR "/cvxpy"

### Required for SCS tests
RUN apt-get update && apt-get install libopenblas-dev -y
### For Python 3.10, OSQP must be built from source, requiring cmake
RUN apt-get install cmake -y
###

RUN pip install -e .
RUN pip install pytest
RUN pytest
```
I have noticed that without installing BLAS / LAPACK, some SCS tests fail, which might not be immediately obvious for users.
From the [SCS 3.0 docs](https://www.cvxgrp.org/scs/install/python.html#python) it seems that this would be resolved once version 3.0 is available on PyPI.


doc/source/install/index.rst
```diff
@@ -94,12 +81,27 @@ as will `Swig`_ . Once you’ve installed these dependencies, perform the follow
...
+
+  ::
+
+      pytest
+
+To run the tests when CVXPY was not installed from source, use
+
+  ::
+
+      pytest --pyargs cvxpy/tests/
```
SteveDiamond, line 97: Are you sure this works? Only ``pytest --pyargs cvxpy`` works for me.

phschiele: Thanks, changed to `pytest --pyargs cvxpy.tests` in 7c09407, which should work across all setups.

SteveDiamond (approved):
LGTM

PR #1559, opened 2021-11-25
Don’t install a top-level ‘examples’ package

andersk:
Currently `pip install cvxpy` also installs a top-level module named `examples`. Fix `setup.py` to only install modules under `cvxpy`.


SteveDiamond:
Integrating into #1593

PR #1564, opened 2021-11-28
[CI] Adding Sonar

phschiele:
Adding back coverage has been on the backlog for quite some time. This PR brings back coverage and also introduces static code analysis by adding [SonarCloud](https://sonarcloud.io/).

### What
SonarCloud is a tool to track code quality, and it is free for open-source projects.
I have included it in my fork, as seeing a live version is the best way to describe it.
The dashboard can be found [here](https://sonarcloud.io/project/overview?id=cvxpy-sonar).

### How
SonarCloud is linked via their provided GitHub action, and thus should be easy to maintain. 
Setup should be as simple as connecting the tool to the GitHub organization and adding the `SONAR_TOKEN` secret. Retriggering the CI on this PR should yield first results.
Smaller adjustments can then be made through the UI (e.g. set the definition of "New Code", which quality standards to apply, ...).

### Initial results
Some of the bugs and code smells are definitely false positives, many are triggered by the exception tests. These can be marked as such and will be ignored going forward. On the other hand, there are also a few findings where it makes sense to take a closer look. After an initial cleanup of the false positives, it could be worthwhile to link to the dashboard in the "Contribution" section, as it provides actionable tasks (coverage or code smells) that usually are well encapsulated. 

### Intrusiveness 
The tool can be set up on different levels of "intrusiveness". From running completely in the background just to track coverage, to commenting on every PR and even blocking PRs if certain criteria are not met.
I have prepared an [example PR](https://github.com/phschiele/cvxpy/pull/26) which includes code smells and missing coverage. The corresponding analysis can be found [here](https://sonarcloud.io/summary/new_code?id=cvxpy-sonar&pullRequest=26) (the same information would be shown in the bot comment).

### Implementation details:
- `DEPLOY_PYPI_SOURCE` was renamed to `SINGLE_ACTION_CONFIG` as this variable is not only used to decide which config uploads to PyPI, but also which config is used for the `cvxpy-base` tests and coverage
- It would theoretically be possible to run coverage on all configs and combine the outputs for the most accurate results. There are small differences as not all solvers are available for all platforms. I have not done this yet to keep this initial proposal as simple as possible.
- I have added a check to see if the `SONAR_TOKEN` variable exists. This will make sure the action does not fail on forks, which do not have the variable set.
- One can create a coverage badge directly from the SonarCloud UI, if desired
- I'll remove the test dashboard I have created if this PR is merged
- The action runs in parallel to the `cvxpy-base` tests, so the overall time of the CI is not increased.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,0 @@
-[run]
-source = cvxpy
-omit =
-    */python2.7/*
-    */lib-python/2.7/*.py
-    */lib_pypy/_*.py
-    */site-packages/ordereddict.py
-    */site-packages/nose/*
-    */unittest2/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+source = cvxpy
+relative_files = True
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,14 +47,14 @@ jobs:
             python-version: 3.8
           - os: macos-10.15
             python-version: 3.8
-            deploy_pypi_source: "True"
+            single_action_config: "True"
           - os: windows-2019
             python-version: 3.8
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
-      DEPLOY_PYPI_SOURCE: "${{ matrix.deploy_pypi_source == 'True' && 'True' || 'False' }}"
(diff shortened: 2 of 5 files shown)


SteveDiamond:
I think this is very valuable! I would vote for starting off unobtrusively (i.e. tracking in the background). Some of the bugs are real, so sonar is definitely providing value.

SteveDiamond:
Merged via #1565

PR #1565, opened 2021-11-29
[CI] Adding Sonar (v2)

SteveDiamond:


sonar-project.properties
```diff
@@ -0,0 +1,13 @@
+sonar.projectKey=cvxpy
```
phschiele, line 1:
```suggestion
sonar.projectKey=cvxpy_cvxpy
```

SteveDiamond:
Looks like it works. Should I merge @phschiele ?

phschiele:
> Looks like it works. Should I merge @phschiele ?

Yes, looks good!

Issue #1377, opened 2021-05-19
[Request] Huber function, concomitant scale estimation


4aHxKzD:
The Matlab version of CVX offers the option of calling the Huber function as `huber(x,M)` or `huber(x,M,t)`. The latter computes `t+t*huber(x/t,M)`, which is useful for concomitant scale estimation. It would be nice to have the same option with cvxpy.

4aHxKzD:
I noticed that cxvpy relies on scipy for the huber function. Would this make it harder to implement this feature?

rileyjmurray:
It seems that the version of the Huber function with three parameters is just the (shifted) perspective transform of the Huber function with two parameters. In principle, it's not hard for us to accommodate perspective functions, but this isn't something that we've gotten around to yet. See #420 and #833.

rileyjmurray:
This is a great opportunity for someone who wants to contribute a new atom.

CSHugoM:
We are going to try to tackle this issue with other students !

CSHugoM:
For the canonicalization of the huber function, we don't really understand what is the DCPtoCone reduction. Would anyone have a written example of what would be this reduction for the 2 arguments-huber function :
`h(M,x) = 1/2*x^2 if abs(x)<= M `
`h(M,x) = M(abs(x) - 1/2M) else`

Thanks !

rileyjmurray:
@CSHugoM you need to implement the perspective of the Huber function. Here's some information on implementing perspective transforms: https://docs.mosek.com/modeling-cookbook/practical.html?highlight=huber#perspective-functions. Here's information on implementing the basic Huber function: Example 7.4: https://docs.mosek.com/modeling-cookbook/practical.html?highlight=huber#convex-univariate-piecewise-defined-functions.

CSHugoM:
@rileyjmurray Thanks for this ! Sorry for not answering earlier, I was busy during the week end. 
I'll try to implement that, just so you know that I enter an exam period so I might be less active than I was last week

jcbedoyam:
Hi, since there has not been activity since last year I started implementing a solution. However, I have been having trouble with the reductions. I think the canonicalization are the same however I don't get what is the specific optimization problem I have to translate in qp2quad_form and dcp2cone.

SteveDiamond:
[...]

PR #1567, opened 2021-12-02
WIP: add a 3rd argument to Huber function

CSHugoM:
I changed the huber function to add a 3rd argument T which represents the possible scaling of the function.

### DIFF
--- a/cvxpy/atoms/elementwise/huber.py
+++ b/cvxpy/atoms/elementwise/huber.py
@@ -43,17 +43,20 @@ class huber(Elementwise):
         The expression to which the huber function will be applied.
     M : Constant
         A scalar constant.
+    T : Constant
+        Scaling constant if non null
     """
 
-    def __init__(self, x, M: int = 1) -> None:
+    def __init__(self, x, M: int = 1, T: float = 0) -> None:
         self.M = self.cast_to_const(M)
+        self.T = self.cast_to_const(T)
         super(huber, self).__init__(x)
 
     @Elementwise.numpy_numeric
     def numeric(self, values) -> float:
         """Returns the huber function applied elementwise to x.
         """
-        return 2*scipy.special.huber(self.M.value, values[0])
+        return 2*scipy.special.huber(self.M.value, values[0]) if self.T.value == 0 else self.T.value*(1 + 2*scipy.special.huber(self.M.value/self.T.value, values[0]))
 
     def sign_from_args(self) -> Tuple[bool, bool]:
         """Returns sign (is positive, is negative) of the expression.


rileyjmurray:
Thanks so much for giving this a shot, @CSHugoM! Could you please add some tests to verify that it works as expected?

As a starting point, I suggest a test where you minimize this function where the input vectors and scale parameter are confined to a simple box in the positive orthant (say, all components in the interval [1, 2]). You should be able to work out an analytic solution for that problem. I also suggest looking at an existing test with the two-parameter Huber function, using a Variable ``scale`` input, but fixing ``scale == 1``. Then you can check that the behavior is unchanged relative to the two-parameter Huber function.

(Also, sorry for not replying to your GitHub issue. I know I directed you here, and we really appreciated your enthusiasm.)

CSHugoM:
Thanks for your comments !

I wrote the first test in the test_problem.py file. However, when I try to run it with the command : 
python -m pytest test_problem.py::TestProblem::test_huber_3_arg_min

I have an Error : Not Found.

Have you any idea why ? 

Thanks,
Hugo

cvxpy/atoms/elementwise/huber.py
```diff
@@ -43,17 +43,20 @@ class huber(Elementwise):
...
         self.M = self.cast_to_const(M)
+        self.T = self.cast_to_const(T)
         super(huber, self).__init__(x)
 
     @Elementwise.numpy_numeric
     def numeric(self, values) -> float:
         """Returns the huber function applied elementwise to x.
         """
-        return 2*scipy.special.huber(self.M.value, values[0])
+        return 2*scipy.special.huber(self.M.value, values[0]) if self.T.value == 0 else self.T.value*(1 + 2*scipy.special.huber(self.M.value/self.T.value, values[0]))
```
SteveDiamond, line 59: This line is way too long. Break it out into a proper if statement.

CSHugoM: Thanks for pointing this out, I'll take break it out then.

rileyjmurray:
@CSHugoM I left a comment on the GitHub issue that provides more guidance. Take a look and let me know (on that thread) if those resources don't make total sense.

CSHugoM:
@rileyjmurray Thanks for that ! Are you talking about Issue #1377 ? I don't see any new comment on this issue, is this about one of your old comments ?

rileyjmurray:
Huh, I swear left a comment, but it's not there. I just added one again.

rileyjmurray:
@CSHugoM do you feel like you have a handle on what needs to be done? If not, maybe we could have a Skype / Zoom call.

rileyjmurray:
Closing due to inactivity.

PR #1570, opened 2021-12-06
Change logging in MOSEK

aszekMosek:
I propose two changes in the MOSEK log output.

1. Remove automatic printing of infeasibility report/certificate. Since cvxpy does not pass names to MOSEK, this was not very useful in the first place (one has to know the mapping from cvxpy variables to the ``x`` variable in MOSEK to understand the report). Now that CVXPY dualizes continuous problems it is simply impossible even for a fairly expert user to immediately relate that report to the original problem. And that report can have lots of rows, obscuring the interesting part of the optimization log.
2. If we go verbose and log is enabled at all I see no reason why presolve log should be specifically disabled. It is just a few lines and sometimes can actually be informative.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -545,8 +545,6 @@ class MOSEK(ConicSolver):
             print('\n')
             env.set_Stream(mosek.streamtype.log, streamprinter)
             task.set_Stream(mosek.streamtype.log, streamprinter)
-            task.putintparam(mosek.iparam.infeas_report_auto, mosek.onoffkey.on)
-            task.putintparam(mosek.iparam.log_presolve, 0)
 
         # Parse all user-specified parameters (override default logging
         # parameters if applicable).


rileyjmurray:
Sounds good to me! Merging.

JianqiangDing:
@rileyjmurray  Hi, I just upgrade my local cvxpy to the latest version v1.1.18, As @aszekMosek said, cvxpy do a mapping internal from cvxpy variables to variables in MOSEK (this do confuse me a lot because I just wanna check which constraint introduce infeasibility to my problem, and it seems hard to understand that report), but now, the LOG just told me the problem is DUAL_INFEASIBLE, I don't get more details

I just want to ask if there is a better way to debug my problem without using MOSEK directly? any suggestion would be helpful

JianqiangDing:
As the documentation mentioned as follows:
https://docs.mosek.com/latest/pythonapi/parameters.html#mosek.iparam.infeas_report_auto
just set this TAG ON, someone can get the infeasible report from MOSEK, this more relates to MOSEK docs but cvxpy.
MOSEK provide some APIs for supporting problem analyzing, someone may check the link following for more detail
https://docs.mosek.com/latest/toolbox/analyzers.html

aszekMosek:
You can get the infeasibility report in the log if you set that parameter you mentioned. Like you noticed, it will not be very easy to relate the output the cvxpy problem.

https://docs.mosek.com/latest/pythonapi/debugging-infeas.html about infeasibility. Section 8.3.4 has more generic suggestions that can be used from cvxpy.

https://docs.mosek.com/latest/faq/faq.html#cvxpy how to set Mosek parameters from cvxpy.

JianqiangDing:
@aszekMosek Hi, Based on your experiences thus far， I am wondering if solving the problem with MOSEK directly is a better way to debug an infeasible problem, for easy to check the infeasible report?

PR #1575, opened 2021-12-12
Add some types to Variable, Expression, and performance_utils

r-barnes:


cvxpy/expressions/expression.py
```diff
@@ -371,7 +371,7 @@ def shape(self) -> Tuple[int, ...]:
     def is_real(self) -> bool:
         """Is the Leaf real valued?
         """
-        return not self.is_complex()
+        return not self.is_complex
```
rileyjmurray, line 374: @r-barnes there are a *lot* of tests failing due to ValueErrors. This change might be causing some of those failures.

r-barnes: @rileyjmurray : I'll try it with the `()` added back in. It's a little confusing because it's declared `@abc.abstractproperty` earlier in the file.

r-barnes:
@rileyjmurray : This looks like it's passing tests. The errors in the CI appear to be preexisting.

phschiele:
@r-barnes I'll take a look at the test failures.

cvxpy/expressions/expression.py
SteveDiamond: Where is it declared as ``@abc.abstractproperty``? That should be changed.

SteveDiamond:
Error described here: https://stackoverflow.com/questions/52889746/cant-import-annotations-from-future/52890129
Fix might be to pin an older version of pytest for Python 3.6 builds?

cvxpy/expressions/expression.py
r-barnes: @SteveDiamond , Line 382 of that file:
```
    @abc.abstractproperty
    def is_complex(self) -> bool:
        """Is the Leaf complex valued?
        """
        raise NotImplementedError()
```

r-barnes:
@SteveDiamond : I suggest you just drop 3.6 altogether. The Python Foundation is ceasing support for it on 2021-12-23. This is listed on the main Python site [here](https://www.python.org/downloads/) with additional details [here](https://www.python.org/dev/peps/pep-0494/#lifespan).

Using older versions of PyTest to support an EOLed version of Python doesn't seem sustainable.

phschiele:
There were some changes in the conda repo of the package that is causing the error: https://github.com/conda-forge/tomli-feedstock.
Potentially the issue was fixed already, as tests are passing in my recently opened [PR](https://github.com/cvxpy/cvxpy/pull/1584).

cvxpy/expressions/expression.py
SteveDiamond: Ok I changed it.

Issue #1578, opened 2021-12-16
Wrong code comments?


Anjiang-Wei:
**Describe the bug**
In this code example (https://colab.research.google.com/github/cvxpy/cvxpy/blob/master/examples/notebooks/WWW/Channel_capacity_BV4.57.ipynb#scrollTo=NVgWp0w0gRfM for `Capacity of a Communication Channel`) or https://www.cvxpy.org/examples/applications/Channel_capacity_BV4.57.html, I have identified two bugs.
Background information:
```
Convex optimization can be used to find the channel capacity $C$ of a discrete memoryless channel.
Consider a communication channel with input $X(t) \in \{1,2,...,n\}$ and output $Y(t) \in \{1,2,...m\}$. 
This means that the random variables $X$ and $Y$ can take $n$ and $m$ different values, respectively.  

In a discrete memoryless channel, the relation between the input and the output is given by the transition probability:

$p_{ij} = \mathbb{P}(Y(t)=i | X(t)=j)$
```
I first show the code comments written in the example code link, and then I will discuss the potential correction.
(1)
```
p_(i,j) = ℙ(Y(t)=i|X(t)=j), i=1,..,m  j=1,...,m
```
I think the it should be `j=1,...,n` instead of `m`.
(2) 
```
Note that the rows of $P$ must sum to 1 and all elements of $P$ must be positive.
```
I think it should be `the columns of $P$ must sum to 1`, instead of `rows`. The reason is because, for each input X, no matter what it finally becomes after transmitting through a channel, its probability will sum up to 1, so `sum(P_(i, j), axis=j) = 1`, which means the sum of `row`.

Anjiang-Wei:
Also, there are another two minor issues:
(3) When P contains 0, this can fail
```
# c = np.sum(P*np.log2(P),axis=0)
c = np.sum(np.array((xlogy(P, P) / math.log(2))), axis=0)
```

(4) `cp.entr` uses e as base of log. I would suggest using 2 as base (just as the log2 you are invoking in the above script)
```
# I = c@x + cp.sum(cp.entr(y)
I = c@x + cp.sum(cp.entr(y) / math.log(2))
```

keithbriggs:
Thanks for these comments.   I'm not sure who owns the colab and cvxpy webpages, so I don't know how to fix them.

SteveDiamond:
[...]

PR #1582, opened 2021-12-18
Fix bugs in the example of Channel Capacity

Anjiang-Wei:
This PR aims to fix the bugs mentioned in https://github.com/cvxpy/cvxpy/issues/1578
Also, it uses `@` instead of the deprecated `*` for matrix-matrix multiplication

Note that `xlogy(P, P)` is actually `e-based`, and `cp.entr(y)` is also `e-based`. I think base-2 makes more sense (considering the fact that the original code has already used `P*np.log2(P)`, which is `2-based`).


SteveDiamond:
Thanks!

PR #1584, opened 2021-12-18
CPLEX parameter setting

phschiele:
Given the current attention to remote code execution, it's probably a good time to remove the `eval()` statement while setting CPLEX parameters.

Admittedly, it would be quite a niche use case to get user input for solver parameters, but better safe than sorry.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 from collections import namedtuple
+from operator import attrgetter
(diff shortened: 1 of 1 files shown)


rileyjmurray:
@phschiele it looks like this is backward-compatible for those who may have been using it correctly. Can you confirm?

phschiele:
@rileyjmurray Yes, the `attrgetter()` function is available in all supported Python versions and there is a test case covering both successful and unsuccessful parameter changes. 
=> should not break any existing use cases.

rileyjmurray:
Great! In that case, this looks good to me. Thanks for the awesome attention to detail.

PR #1587, opened 2021-12-18
Fix bugs in capacity

Anjiang-Wei:
This is a followup for https://github.com/cvxpy/cvxpy/pull/1582
Sorry I didn't fix all the bugs in the prior PR.

(1) I forgot to import `xlogy`
(2) The objective is actually minimizing `-I`, i.e., `obj = cp.Minimize(-I)`. So it should return `-prob.value` instead of `prob.value`. Otherwise, the capacity will be a negative number (`-0.1887`), which does not make sense.
(3) I forgot to update the final result in the output.
The final result should be `capacity = 1-H(p)=1-H(0.25)=1-0.8113=0.1887`. The formula can be found in:
https://en.wikipedia.org/wiki/Binary_symmetric_channel

And this PR can fix all the above issues.


examples/notebooks/WWW/Channel_capacity_BV4.57.ipynb
```diff
@@ -107,7 +108,7 @@
     "    prob = cp.Problem(obj,constraints)\n",
     "    prob.solve()\n",
     "    if prob.status=='optimal':\n",
-    "        return prob.status, prob.value, x.value\n",
+    "        return prob.status, -prob.value, x.value\n",
```
SteveDiamond, line 111: Maybe maximize instead of minimize then negating?

Anjiang-Wei: That's a good idea. I have updated the PR.

SteveDiamond:
Thanks!

PR #1593, opened 2021-12-24
Add deployment/testing with Python 3.10

SteveDiamond:


SteveDiamond:
@phschiele @akshayka I'm getting pretty close to getting 3.10 working. I'm getting a weird bug I can't reproduce locally involving diffcp: https://github.com/cvxpy/cvxpy/runs/4628602909?check_suite_focus=true#step:6:4333
I'll think about it more but if anything comes to mind let me know.

phschiele:
@SteveDiamond I think I found the issue:
Installing `diffcp` under Python 3.10 installs 1.0.16 instead of 1.0.18, which does not have support for SCS 3.0 yet.
The reason is that on [PyPI](https://pypi.org/project/diffcp/#files) there are neither wheels for 3.10, nor the source.
I'll check why the source was not deployed with the last [tag](https://github.com/cvxgrp/diffcp/runs/4053266136?check_suite_focus=true). After this is fixed, the tests should pass here, and you could even undo the changes made to `diffcp_conif.py`, as this is handled within `diffcp`.

continuous_integration/install_dependencies.sh
```diff
@@ -17,14 +17,18 @@ elif [[ "$PYTHON_VERSION" == "3.9" ]]; then
   # The earliest version of numpy that works is 1.19.
   # Given numpy 1.19, the earliest version of scipy we can use is 1.5.
   conda install scipy=1.5 numpy=1.19 mkl pip pytest lapack ecos scs osqp cvxopt
+elif [[ "$PYTHON_VERSION" == "3.10" ]]; then
+    # The earliest version of numpy that works is 1.19.
+    # Given numpy 1.19, the earliest version of scipy we can use is 1.5.
```
phschiele, line 22:
```suggestion
    # The earliest version of numpy that works is 1.21.
    # Given numpy 1.21, the earliest version of scipy we can use is 1.7.
```

phschiele:
@SteveDiamond I found two more occurrences where the Python versions are listed explicitly: `pyproject.toml` and `sonar-project.properties`.

SteveDiamond:
@phschiele thanks so much for figuring it out! I'm also having some problems building wheels for 3.10. Not totally sure what's going on there: https://github.com/cvxgrp/diffcp/runs/4632902674?check_suite_focus=true#step:7:53

SteveDiamond:
Integrating a version of #1559

SteveDiamond:
@phschiele I think this is good to go. Could you review it when you have time?

.github/workflows/build.yml
```diff
@@ -90,13 +87,42 @@ jobs:
...
+
+      - name: Set Additional Envs
+        shell: bash
+        run: |
+          echo "PYTHON_SUBVERSION=$(echo $PYTHON_VERSION | cut -c 3-)" >> $GITHUB_ENV
+          echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
+
       - name: Build wheels
-        if: ${{env.DEPLOY == 'True' && env.USE_OPENMP != 'True'}}
+        if: ${{'True' == 'True' && env.USE_OPENMP != 'True'}}
```
phschiele, line 120: @SteveDiamond Just a reminder to undo these shortcuts after everything is confirmed working.

SteveDiamond: Thanks! I'm going to let it build the wheels once on master and then undo them.

PR #1598, opened 2022-01-09
WIP: semantic versioning infrastructure

rileyjmurray:
## Description

This PR is meant to resolve #1576 and #1569. It provides infrastructure for managing cvxpy's version number in essentially the same way that SciPy developers manage version numbers (see the first comment in #1576 for details). It sets the working version of cvxpy to 1.2.0 with a flag that master is NOT a release branch and that 1.2.0 has NOT been released. 

I made a handful of changes to the web documentation. We'll want to workshop those changes before merging. Some changes will be necessary to the procedure we use to *deploy* web docs. I don't know what the right changes are at the moment. I've marked the relevant file with a TODO.

I removed some unused files (appveyor, conda recipe, an old MANIFEST file) and marked some files as no longer maintained. I suggest we remove the files that I marked as no longer maintained.

EDIT: Apologies for the useless commit message at the start of these changes.

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [X] Other (Documentation, CI, ...)

PR #1624, opened 2022-01-30
[doc] Multi-version documentation deployment

phschiele:
## Description
This PR enables deploying multiple versions of the web documentation as a second follow-up to #1598.

### Details
Whenever the action is triggered:
1. The docs are built
2. The docs are deployed into a `version/MAJOR.MINOR` folder (overwriting if already exists) on the `gh-pages` branch
3. If the version of the current build is greater or equal to the highest previously deployed version, the docs are also deployed to the root of the `gh-pages` branch.
4. A file called `versions.json` in the root of the `gh-branch` is created or updated, which lists all currently deployed version
5. A dynamic version selector will reflect the changes in 4., such that one can always switch between all deployed versions

### Tests on my fork
I have experimented with the setup in my fork, see [here](https://phschiele.github.io/cvxpy/) (I'll remove this as soon as this PR is merged).

Here are the test cases I ran. Each commit has a `This is test X` on the landing page.
1. Deploy version `1.1.0` -> Deploys to root and `version/1.1`
2. Deploy version `1.2.1` -> Deploys to root and `version/1.2`
3. Deploy version `1.1.1` -> No deploy to root, only overwrites `version/1.1`
4. Deploy version `1.3.0` -> Deploys to root and `version/1.3`

Version dropdown shows:
- latest (1.3)
- 1.3
- 1.2
- 1.1

Note:
Tests for 3.6 are failing as during the conda install `ModuleNotFoundError: No module named 'dataclasses'` is raised. This seems unrelated to the changes in this PR. Looks similar to the failures observed in #1575 .

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/continuous_integration/doc_is_latest.py
+++ b/continuous_integration/doc_is_latest.py
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-# This is a helper file to determine if the doc version that is currently deployed is the latest version
+# This is a helper to determine if the doc version that is currently deployed is the latest version
 
 import argparse
 
--- a/continuous_integration/doc_is_latest.py
+++ b/continuous_integration/doc_is_latest.py
@@ -1,0 +1,27 @@
+"""
+Copyright 2022 the CVXPY developers
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+# This is a helper file to determine if the doc version that is currently deployed is the latest version
+
+import argparse
+
+from packaging import version
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--c', type=str, help='current version')
+parser.add_argument('--l', type=str, help='latest deployed version')
+args = parser.parse_args()
+if not args.l:
+    print(True)
+else:
+    print(version.parse(args.c) >= version.parse(args.l))
--- a/continuous_integration/doc_write_versions.py
+++ b/continuous_integration/doc_write_versions.py
@@ -1,0 +1,34 @@
+"""
+Copyright 2022 the CVXPY developers
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import ast
+import json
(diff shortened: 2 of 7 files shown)


continuous_integration/doc_is_latest.py
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+# This is a helper to determine if the doc version that is currently deployed is the latest version
+
+import argparse
+
+from packaging import version
```
phschiele, line 18: The two helper files were mostly created to leverage version comparisons from the packaging module.

doc/source/conf.py
```diff
@@ -137,7 +137,7 @@ def __getattr__(self, name):
 html_theme = 'cvxpy_alabaster'
 html_sidebars = {
    '**': [
-       'about.html', 'navigation.html', 'searchbox.html',
+       'about.html', 'navigation.html', 'searchbox.html', 'version_selector.html',
```
phschiele, line 140: One could also only add the version selector if the script is run on GH actions.

rileyjmurray: @phschiele do you mean that we could have the selector only on GH Actions, and not when we build locally? If so, do you see an advantage to that?

continuous_integration/doc_is_latest.py
rileyjmurray: Gotcha! Looks good.

continuous_integration/doc_write_versions.py
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License.
+"""
+
+import ast
+import json
+import os
+
+from packaging import version
+
+# This script creates a "versions.json" file listing all deployed version of the documentation
```
rileyjmurray, line 20: Is this file under version control? Or will it be, on the gh-pages branch?

rileyjmurray:
Thanks so much for doing this, @phschiele! It looks good. I only have two minor questions before merging.

continuous_integration/doc_write_versions.py
phschiele: Yes, this file is also committed to the gh-pages branch, compare: https://github.com/phschiele/cvxpy/commits/gh-pages/versions.json

doc/source/conf.py
phschiele: @rileyjmurray Yes, that was the idea. If the only local use case for the docs is debugging, then it's probably not needed.
If someone would actually use locally built docs for whatever reason, I thought it might be weird to still point to the deployed versions.

rileyjmurray: I personally build the docs locally whenever I change documentation source files. Ideally, we should be updating the doc source files whenever we have a PR that adds new features. So I think we should assume this happens fairly often. With that in mind, what would you suggest we do here?

I'll note that we don't need to make an "optimal" decision so long as we document our discussion in this thread. (E.g., by adding comments to conf.py or other files with a permalink to this thread.)

continuous_integration/doc_write_versions.py
phschiele: Sure, done in 9d07dee

doc/source/conf.py
phschiele: I suggest keeping it also for local builds for now, as I don't see a big downside of having it there as well.

As you mention, this can easily be changed if use cases emerge where having the version selector in local builds is a problem. I've linked to this discussion in `conf.py` in 9d07dee.

rileyjmurray (approved):
Great work, @phschiele! Merging.

PR #1637, opened 2022-02-04
docfix for Advanced Features / DPP

TimonKnigge:
Docfix for the DPP ruleset. I'm not entirely sure this captures what the paragraph was trying to say, but there was definitely something wrong with the documentation beforehand (e.g. ``x`` is definitely not variable-free).


TimonKnigge:
Also now that I'm looking at this part of the docs anyway, should the calls to ``.is_dcp(dpp=True)`` not be replaced with ``.is_dpp()``?

SteveDiamond:
Thanks! Good catch.

PR #569, opened 2018-08-26
Allow kron to have Variable in first argument

rileyjmurray:
This is a definitive resolution to #457 (native cvxpy implementation, as opposed to numpy-backed implementation).

PR #1643, opened 2022-02-07
Have kron support non-constant expressions in either argument

### LINKED ISSUE #457: Kronecker product with variable matrix as first input
The current implementation only supports Kronecker products with a constant first argument. I would like to construct the following semidefinite constraint which is simple to write when the first argument is a `cvxpy` variable:

`import numpy as np`
`import cvxpy as cvx`
`mu = cvx.Variable(3,3)`
`obj = cvx.Minimize(cvx.trace(mu))`
`constraints = [A + cvx.kron(mu, np.eye(3)) >> 0]`

where `A` is a constant `numpy` matrix. Is there any `cvxpy` function or procedure that makes this possible? Is there any reason `kron` is designed to require a constant first input?

rileyjmurray:
## Description
The Kronecker product atom, "kron", has long had the limitation where the first of its two arguments had to be constant.
This PR extends kron so that either argument can be non-constant (although both must be affine).

I had attempted this before in #569 but failed because I didn't know how to extend cvxcore.
This time I dug into the weeds and edited cvxcore directly.
Tests show the full canonicalization process works correctly.
I also wrote some developer documentation to help with future changes to cvxcore.

Criticism is welcome on coding style or inefficiencies in my cvxcore edits. 

Once merged, this will resolve #457. Subsequent TODO's include web documentation and potentially more unit tests. We specifically might want tests with Hermitian problem data.

## Type of change
- [X] Bug fix

Edit: this PR now also includes a bugfix for ``kron(param, var)``. That bugfix needs to be replicated on the ``release/1.1.x`` branch.

- [X] Write unittests.


cvxpy/cvxcore/src/LinOpOperations.cpp
```diff
@@ -133,8 +134,14 @@ Tensor get_node_coeffs(const LinOp &lin, int arg_idx) {
   case VSTACK:
     coeffs = get_vstack_mat(lin, arg_idx);
     break;
+  case KRON_R:
+    coeffs = get_kronr_mat(lin, arg_idx);
+    break;
+  case KRON_L:
+  	coeffs = get_kronl_mat(lin, arg_idx);
```
SteveDiamond, line 141: indentation

cvxpy/tests/test_atoms.py
```diff
@@ -830,9 +830,79 @@ def test_kron(self) -> None:
...
+        prob.solve()
+        self.assertItemsAlmostEqual(y.value, np.array([[np.max(L / A)]]))
+
+        objective = cp.Maximize(y)
+        prob = cp.Problem(objective, constraints)
+        prob.solve()
+        self.assertItemsAlmostEqual(y.value, np.array([[np.min(U / A)]]))
+        pass
+
+    def test_kron_canon_2(self) -> None:
```
SteveDiamond, line 885: Can you add a test involving parameters so we know that the new cvxcore functionality is valid with parameters? (Must be DPP of course.)

SteveDiamond:
Looks great, just small comments. I'm glad you documented the process better, it's really quite complicated. In general I'm starting to see more people building off/rewriting internals of cvxpy, which suggests a need for better developer documentation.

cvxpy/tests/test_atoms.py
rileyjmurray: Good call. I can say that at this moment, Parameters thoroughly break the implementation. I'll investigate more tomorrow.

rileyjmurray:
Big thanks to @SteveDiamond, @phschiele, and @akshayka for teaming up on this one! All tests are passing and that includes plenty of tests for kron: tests with and without parameters, and tests with ``kron(A, X)`` and ``kron(X, A)``.

Please submit a review on the PR if you can. I'd like to merge today, if possible!

cvxpy/lin_ops/lin_utils.py
```diff
@@ -37,7 +37,7 @@ class Counter:
     """
 
     def __init__(self) -> None:
-        self.count = 0
```
SteveDiamond, line 40: Probably not necessary now? Of course no harm.

setup.py
```diff
@@ -188,7 +188,7 @@ def not_on_windows(s: str) -> str:
     sources=['cvxpy/cvxcore/src/cvxcore.cpp',
              'cvxpy/cvxcore/src/LinOpOperations.cpp',
              'cvxpy/cvxcore/src/Utils.cpp',
-             'cvxpy/cvxcore/python/cvxcore_wrap.cpp'],
+             'cvxpy/cvxcore/python/cvxcore_wrap.cxx'],
```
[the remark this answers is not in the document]
rileyjmurray: Swig automatically generated the file with that extension. I could change back to ``.cpp``, but then whenever someone builds with swig they'd have to manually change the file name. I think it's reasonable to have a different extension since its automatically generated.

cvxpy/lin_ops/lin_utils.py
rileyjmurray: I'd like to keep the change. It might help us with debugging in the future.

cvxpy/cvxcore/src/LinOpOperations.cpp
```diff
@@ -302,43 +310,111 @@ int get_id_data(const LinOp &lin, int arg_idx) {
...
+  //   default order is CSC-like, and when "compressed" the storage
+  //   is actually CSC.
+  int row_offset = 0;
+  int kron_rows = lh_rows * rh_rows;
+  int rh_nnz = rh.nonZeros();
+  std::vector<int> base_row_indices;
+  std::vector<double> vec_rh;
+  base_row_indices.reserve(rh_nnz);
+  vec_rh.reserve(rh_nnz);
+  for (int k = 0; k < rh.outerSize(); ++k) {  // loop over columns
```
phschiele, line 386: For my understanding, `rh.outerSize()` depends on the storage order, right?
I'd assume the default in Eigen is quite stable, just want to check if this is maybe enforced/validated at a different place.
Great that you've added a comment here :+1:

rileyjmurray: Check out the new cvxcore README, section 4 :)

I'm happy to add an enforcement mechanism here if you can figure out how to do that. (The Eigen documentation is hard for me to navigate.)

phschiele: @rileyjmurray Looks like there is `IsRowMajor` [Docs](https://eigen.tuxfamily.org/dox/classEigen_1_1SparseMatrixBase.html) which is also suggested in [this](https://stackoverflow.com/a/25059230) SO post.
So `!M.IsRowMajor` could work.

cvxpy/cvxcore/README.md
```diff
@@ -1,74 +1,85 @@
...
+The former file is generated by SWIG.
+The latter file is written by hand and handles the task of connecting Python ``LinOp`` objects (``cvxpy/lin_ops/lin_op.py:LinOp``)
+to C++ ``LinOp`` objects (``cvxpy/cvxcore/src/LinOpp.hpp:LinOp``).
+That task is actually pretty delicate.
+There is only one ``LinOp`` class in Python; the mathematical nature of a specific Python LinOp ``lin`` is indicated
+with a string stored in ``lin.type``.
+Example LinOp types are "mul", "rmul", "transpose", and "kron_r".
+
+
+## 3. Creating new a LinOp
```
phschiele, line 51:
```suggestion
## 3. Creating a new LinOp
```

setup.py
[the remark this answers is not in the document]
SteveDiamond: oh sure, didn't know that swig did that.

cvxpy/cvxcore/README.md
rileyjmurray: Fixed!

phschiele (approved):
Thanks for including the tests and the assertion. No further comments from my side 👍🏻

Issue #1647, opened 2022-02-09
Feature Request: Warnings or Errors on NumPy functions being called on CvxPy Expressions


PTNobel:
# The Issue

## Existing Errors
Consider:
```python
np.sqrt(cp.Variable())
```

On my computer right now, this code crashes the iPython interpreter and raises the following exception with an unhelpful error message:

```
AttributeError: 'Variable' object has no attribute 'sqrt'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: loop of ufunc does not support argument 0 of type Variable which has no callable sqrt method
```

## Probably wrong code

Consider the code:
```python
np.hstack([cp.Variable(3), cp.Variable(3)])
```

This creates a NumPy array with dtype=object, which is often not what users want or expect.

# My Solution
## Existing Errors

I'd like to either make this work correctly or change this to raise an error with a message along the following lines:

```
NumpyCvxpyCompatibilityError: "You're calling a NumPy function on a CvxPy expression. Please use the appropriate atom from: <link to atoms page>"
```

## Probably wrong code

We should raise a warning or an error in this situation.

# Alternatives

Raise just a warning and have the code actually work if it is an atom we also implement.

# Implementation

I'm happy to do the implementation myself; I implemented a similar thing in [PTNobel/AutoDiff](https://github.com/PTNobel/AutoDiff).

SteveDiamond:
The error would be great. Didn't know it was possible. Can you point to how you did it in your package?

PTNobel:
So AutoDiff actually aims to have these functions to work instead of raise an error.

So NumPy checks for the method `__array_ufunc__` or `__array_function__` and calls them before trying to run its naive implementation of this code. In AutoDiff, I have a dictionary that maps the NumPy functions and ufuncs to my implementation of these methods and then I do a dictionary lookup to effectively have dynamic dispatch. Here's the relevant code:

https://github.com/PTNobel/AutoDiff/blob/caed5e47a408e8668394d3aeabf5db7900551895/auto_diff/vecvalder.py#L60-L73

Here's an example implementation:

https://github.com/PTNobel/AutoDiff/blob/caed5e47a408e8668394d3aeabf5db7900551895/auto_diff/vecvalder_funcs_and_ufuncs.py#L124-L129

If we want to raise an Error we can basically copy these implementations and just write `raise Error`.

[...]

PR #1648, opened 2022-02-09
Adds error for mixing NumPy and CVXPY

PTNobel:
Following up on #1647 

This code is supposed to not raise an error on anything that used to succeed (which was just the `np.hstack`/`np.vstack` equivalents I believe); however, by adding the `__array_function__` method NumPy is now assuming we can handle everything and when we say we can't (i.e. we return `NotImplemented`) it chooses to errors instead of creating an array of objects as it used to. The error message has changed on a number of things to say that we have a `TypeError` instead of an `AttributeError`.

If breaking code that constructs arrays with `np.hstack` and `np.vstack` is not acceptable, (`np.array` will still work though) then we can probably just eliminate the `__array_function__` implementation, but I don't see any easy way to generate the warning while preserving old behavior unless we want to do `AutoDiff`-style manual implementation of each thing we want to keep working.

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -61,6 +61,29 @@ Using ``*`` for matrix multiplication has been deprecated since CVXPY 1.1.
 This code path has been hit %s times so far.
 """
 
+__NUMPY_UFUNC_ERROR__ = """
+You're calling a NumPy function on a CVXPY expression. This is prone to causing
+errors or code that doesn't behave as expected. Consider using one of the
+functions documented here: https://www.cvxpy.org/tutorial/functions/index.html
+"""
+__BINARY_EXPRESSION_UFUNCS__ = {
+        np.add: lambda self, a: self.__radd__(a),
+        np.subtract: lambda self, a: self.__rsub__(a),
+        np.multiply: lambda self, a: self.__rmul__(a),
+        np.divide: lambda self, a: self.__rdiv__(a),
+        np.matmul: lambda self, a: self.__rmatmul__(a),
+        np.power: lambda self, a: self.__rpow__(a),
+        np.left_shift: lambda self, a: self.__rlshift__(a),
+        np.right_shift: lambda self, a: self.__rrshift__(a),
+        np.equal: lambda self, a: self.__eq__(a),
+        # <= and >= are backwards because this is only called for code of the
+        # form ndarray <= Expression
+        np.less_equal: lambda self, a: self.__ge__(a),
+        np.greater_equal: lambda self, a: self.__le__(a),
+        np.less: lambda self, a: self.__gt__(a),
+        np.greater: lambda self, a: self.__lt__(a),
+}
+
 
 class Expression(u.Canonical):
     """A mathematical expression in a convex optimization problem.
@@ -463,6 +486,12 @@ class Expression(u.Canonical):
         """
         return cvxtypes.power()(self, power)
 
+    def __rpow__(self, base: float) -> "Expression":
+        raise NotImplementedError("CVXPY currently does not support variables "
+                                  "on the right side of **. Consider using the"
+                                  " identity that a**x = cp.exp(cp.multiply(np"
+                                  ".log(a), x)).")
+
     # Arithmetic operators.
     @staticmethod
     def cast_to_const(expr: "Expression"):
@@ -678,3 +707,15 @@ class Expression(u.Canonical):
         """Unsupported.
         """
         raise NotImplementedError("Strict inequalities are not allowed.")
+
+    def __array_ufunc__(self, ufunc, method, *args, **kwargs):
+        try:
+            ufunc_handler = __BINARY_EXPRESSION_UFUNCS__[ufunc]
+            if kwargs == {} and \
+                    len(args) == 2 and \
+                    args[1] is self:
+                return ufunc_handler(self, args[0])
+        except KeyError:
+            pass
+
+        raise RuntimeError(__NUMPY_UFUNC_ERROR__)
--- a/cvxpy/tests/test_numpy.py
+++ b/cvxpy/tests/test_numpy.py
@@ -1,0 +1,100 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+import numpy as np
+import pytest
+
+import cvxpy as cp
+from cvxpy.expressions.expression import (__BINARY_EXPRESSION_UFUNCS__,
+                                          __NUMPY_UFUNC_ERROR__,)
+from cvxpy.tests.base_test import BaseTest
+
+
+class TestNumpy(BaseTest):
+    """ Unit tests for using NumPy ufuncs on CVXPY objects should cause errors. """
+
+    def setUp(self) -> None:
+
+        self.x = cp.Variable((2, 2), name='x')
+        self.x.value = [[10., 11], [12, 13]]
+        self.y = cp.Variable(1, name='y')
+
+    def test_np_ufunc_errors(self) -> None:
+        with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+            np.sqrt(self.x)
+
+        with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+            np.log(self.x)
+
+    def test_some_np_ufunc_works(self) -> None:
+        a = np.array([[1., 3.], [3., 1.]])
+        b = np.int64(1)
+
+        for ufunc in __BINARY_EXPRESSION_UFUNCS__:
+            if ufunc is np.multiply:
+                continue  # We don't want to deal with the multiply warnings
+            if ufunc is np.power:
+                continue  # We don't implement __rpow__ yet.
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(self.x, a)
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(a, self.x, out=a)
+
+            if ufunc is np.left_shift or \
+                    ufunc is np.right_shift or \
+                    ufunc is np.equal or \
+                    ufunc is np.less_equal or \
+                    ufunc is np.greater_equal or \
+                    ufunc is np.less or \
+                    ufunc is np.greater:
+                continue
+            self.assertItemsAlmostEqual(ufunc(a, self.x).value, ufunc(a, self.x.value))
+
+        for ufunc in __BINARY_EXPRESSION_UFUNCS__:
+            if ufunc is np.matmul:
+                continue  # matmul doesn't play nice with scalars
+            if ufunc is np.power:
+                continue  # We don't implement __rpow__ yet.
+
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(self.x, b)
+
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(b, self.x, out=b)
+
+            if ufunc is np.left_shift or \
+                    ufunc is np.right_shift or \
+                    ufunc is np.equal or \
+                    ufunc is np.less_equal or \
+                    ufunc is np.greater_equal or \
+                    ufunc is np.less or \
+                    ufunc is np.greater:
+                continue
+
+            self.assertItemsAlmostEqual(ufunc(b, self.x).value, ufunc(b, self.x.value))
+
+    def test_working_numpy_functions(self) -> None:
+        hstack = np.hstack([self.x])
+        self.assertEqual(hstack.shape, (1,))
+        self.assertEqual(hstack.dtype, object)
+        vstack = np.vstack([self.x])
+        self.assertEqual(vstack.shape, (1, 1))
+        self.assertEqual(vstack.dtype, object)
+
+    def test_broken_numpy_functions(self) -> None:
+        with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+            np.linalg.norm(self.x)


SteveDiamond:
Actually, if it's only the ufunc left, it seems fine to raise an error.  Make a new test file ``test_numpy.py``.

SteveDiamond:
@rileyjmurray do we want this in 1.2? Need to finish the PR.

PTNobel:
Hey @SteveDiamond , I can def get it done this week. I had some emergency personal stuff pop up and also took my quals last week, so I got a bit distracted. I'm doing the 364A final on Wednesday, and then should be able to focus on this.

SteveDiamond:
Ok no worries.

PTNobel:
Okay, so I had forgotten that NumPy prefers to call an `__array_ufunc__` method over `__radd__` for code of the form `np.ndarray + cp.Variable`. Worse, if we return NotImplemented NumPy raises a TypeError rather than pass control back to Python so it can try `__radd__`.

Here's my idea for a solution:

 * Have a known set of ufuncs that correspond to binary operations that we don't raise an error for (i.e. `np.left_shift`, `np.right_shift`, `np.add`, `np.subtract`, `np.divide`, `np.multiply`, `np.matmul`, `np.power`, am I missing any?)
 * If args is not an instance of [np.ndarray, cp.Expression] or kwargs is not {} (i.e. we're not in the case triggered by the binary operators), then we raise the exception as normal.
 * If we are in that case, we just manually make the appropriate call, i.e. `self.__radd__(array, expr)`.

Thoughts?

PTNobel:
I just pushed an implementation of my proposal.

I have a tiny test script I'm using right now; I'll write some more careful tests in the CVXPY testing framework, once I have the go-ahead on this plan.

(Also I might be triggering a new warning on `ndarray == expression`, I'm not sure why, since it isn't calling into `__array_ufunc__`)

```
import cvxpy as cp
import numpy as np

x = cp.Variable()
try:
    np.sqrt(x)
except RuntimeError:
    pass
else:
    assert False, "np.sqrt(cp.Variable()) Should have raised an error"

try:
    x + np.array([1])
except RuntimeError:
    assert False, "cp.Variable() + np.ndarray() should not have raised an error"
else:
    pass

try:
    np.array([1]) + x
except RuntimeError:
    print("np.ndarray() + cp.Variable() raised an error")
else:
    pass

try:
    x == np.array([1])
except RuntimeError:
    assert False, "cp.Variable() == np.ndarray() should not have raised an error"
else:
    pass

try:
    np.array([1]) == x
except RuntimeError:
    assert False, "np.ndarray() == cp.Variable() should not have raised an error"
else:
    pass

```

PTNobel:
Ah I figured out the issue. NumPy was calling our `__array_ufunc__` with np.equal, but catching the error, raising a deprecation warning and then falling back to the old behaviour. All we have to do is add `np.equal` to the list and we'll get the behaviour we want.

PTNobel:
Alright, i fixed that bug; since I now handle every binary operator (the only places NumPy might start trying to call this function without the user naming a NumPy function), I think this should have feature parity with the old code!

I'm going through the Contributing checklist and figuring out how to run the tests so I have some more confidence on that though.

PTNobel:
Well, it turns out numpy types other than ndarray (e.g. int64, float64) also call __array_ufunc__ to implement multiplication. My gut feeling is to just remove the "check if the first argument is a np.ndarray" step in my algorithm, so that any type on the left of a binary operator works correctly.

PTNobel:
Okay, I fixed a couple minor bugs and now have:

    751 passed, 182 skipped, 330 warnings in 64.86s (0:01:04) 

I'm a bit nervous there will be a performance impact to have a dictionary lookup on every add and subtract, so I'm going to run those tests now.

PTNobel:
With my patch:

```
===================== test session starts =====================
platform linux -- Python 3.10.2, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/parthnobel/Research/cvxpy, configfile: pyproject.toml
collected 11 items

test_benchmarks.py cone_matrix_stuffing_with_many_constraints: avg=1.339e+00 s , std=0.000e+00 s (1 iterations)
.diffcp_sdp: avg=5.383e+00 s , std=0.000e+00 s (1 iterations)
.least_squares: avg=6.506e-03 s , std=0.000e+00 s (1 iterations)
.sConic canonicalization
(ECOS) solver time:  0.061479839
cvxpy time:  0.6716441787154541
Quadratic canonicalization
(OSQP) solver time:  0.031092942
cvxpy time:  0.7756702301496582
.qp: avg=1.731e-02 s , std=0.000e+00 s (1 iterations)
.small_cone_matrix_stuffing: avg=4.953e-02 s , std=1.317e-02 s (10 iterations)
.small_lp: avg=2.414e-02 s , std=0.000e+00 s (1 iterations)
small_lp_second_time: avg=2.008e-03 s , std=0.000e+00 s (1 iterations)
.sstv_inpainting: avg=2.752e+00 s , std=0.000e+00 s (1 iterations)
.
```

Without my patch:

```
===================================================== test session starts =====================================================
platform linux -- Python 3.10.2, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/parthnobel/Research/cvxpy, configfile: pyproject.toml
collected 11 items

test_benchmarks.py cone_matrix_stuffing_with_many_constraints: avg=1.273e+00 s , std=0.000e+00 s (1 iterations)
.diffcp_sdp: avg=5.556e+00 s , std=0.000e+00 s (1 iterations)
.least_squares: avg=7.001e-03 s , std=0.000e+00 s (1 iterations)
.sConic canonicalization
(ECOS) solver time:  0.060723004
cvxpy time:  0.6589372781350098
Quadratic canonicalization
(OSQP) solver time:  0.030688248
cvxpy time:  0.7772355457927246
.qp: avg=1.679e-02 s , std=0.000e+00 s (1 iterations)
.small_cone_matrix_stuffing: avg=4.559e-02 s , std=1.042e-02 s (10 iterations)
.small_lp: avg=2.286e-02 s , std=0.000e+00 s (1 iterations)
small_lp_second_time: avg=1.629e-03 s , std=0.000e+00 s (1 iterations)
.sstv_inpainting: avg=2.704e+00 s , std=0.000e+00 s (1 iterations)
.

```

PTNobel:
LMK what you think of the performance change here.

PTNobel:
Fixed a formatting issue! Let me know if you all can tolerate the slowdown and like my approach here, I'll squash these commits and rebase onto master and then add a second commit with tests.

cvxpy/expressions/expression.py
```diff
@@ -61,6 +61,29 @@ def cast_op(self, other):
...
+        np.power: lambda self, a: self.__rpower__(a),
+        np.left_shift: lambda self, a: self.__rlshift__(a),
+        np.right_shift: lambda self, a: self.__rrshift__(a),
+        np.equal: lambda self, a: self.__eq__(a),
+        # <= and >= are backwards because this is only called for code of the
+        # form ndarray <= Expression
+        np.less_equal: lambda self, a: self.__ge__(a),
+        np.greater_equal: lambda self, a: self.__le__(a),
+        np.less: lambda self, a: self.__gt__(a),
+        np.greater: lambda self, a: self.__lt__(a),
```
rileyjmurray, line 84: We don't allow strict inequalities, so these could possibly be removed.

rileyjmurray:
@PTNobel can you summarize your changes? I have two specific questions: 
1. What happens with calling ``np.hstack`` and ``np.vstack`` on cvxpy Expressions? I think the current behavior is useful because it provides a way for users to implement sophisticated affine transformations which cvxpy might not natively support.
2.The dictionary lookup is only used when one of the operands in a binary operation is a numpy datatype, right? If so, then I think the performance impact should be relatively modest. We encourage users to build vectorized models, so it's not like they should have several thousand operations which hit that code path.

cvxpy/expressions/expression.py
PTNobel: I put them there so the user gets an error about strict inequalities being prohibited rather than a message that confusingly says they're using NumPy functions when they're not.

SteveDiamond:
I'm in the middle of some stuff and can't read all the comments. However, it appears that this changes may be more complicated than we had thought, with more unexpected consequences. I would suggest we hold off on adding this to 1.2.

rileyjmurray:
@PTNobel I did some digging and it seems there is no meaningful overhead with this approach.

The code below evaluates several types of expressions 100,000 times. The first part of the code shows that the dictionary lookup cost is neglidgable. The bigger overhead comes in an increased number of function calls. That can be seen in the matmul times, which are much larger than time for dict accesses and do increase a bit with your changes. However, the runtime increase from your changes is 2% to 5% on my machine. I'm confident that such an increase won't be the bottleneck operation that affects compilation time from a user perspective; we already know the real source of pain there is cvxcore. 
```
import time
from cvxpy.expressions.expression import __BINARY_EXPRESSION_UFUNCS__, Expression

if __name__ == '__main__':

    trials = int(1e5)

    npufuncs = list(__BINARY_EXPRESSION_UFUNCS__.keys())
    tic = time.time()
    print('\nufunc dictionary lookup times')
    for func in npufuncs:
        tic = time.time()
        for t in range(trials):
            __BINARY_EXPRESSION_UFUNCS__[func]
        toc = time.time()
        print('\t%s' % (toc - tic))

    dim = 5  # runtime doesn't increase even if you make this 500.
    A = np.zeros((dim, dim))
    A_const = cp.Constant(A)
    X = cp.Variable((dim, dim))
    __array_ufunc_impl__ = Expression.__array_ufunc__

    print('\nMatmul times: np.ndarray @ cp.Variable')

    tic = time.time()
    for t in range(trials):
        A @ X
    toc = time.time()
    print('\tnew: %s' % (toc - tic))
    Expression.__array_ufunc__ = None
    tic = time.time()
    for t in range(trials):
        A @ X
    toc = time.time()
    print('\told: %s' % (toc - tic))

    print('\nMatmul times: cp.Constant @ cp.Variable')

    Expression.__array_ufunc__ = __array_ufunc_impl__
    tic = time.time()
    for t in range(trials):
        A_const @ X
    toc = time.time()
    print('\tnew: %s' % (toc - tic))
    Expression.__array_ufunc__ = None
    tic = time.time()
    for t in range(trials):
        A_const @ X
    toc = time.time()
    print('\told: %s' % (toc - tic))
```
Here's the output I get
```
ufunc dictionary lookup times
	0.006108283996582031
	0.008254051208496094
	0.005980253219604492
	0.00611567497253418
	0.0059986114501953125
	0.0060596466064453125
	0.005944013595581055
	0.006208181381225586
	0.005948305130004883
	0.005890846252441406
	0.005970478057861328
	0.006337881088256836
	0.005922555923461914

Matmul times: np.ndarray @ cp.Variable
	new: 2.0843350887298584
	old: 1.866164207458496

Matmul times: cp.Constant @ cp.Variable
	new: 0.5424549579620361
	old: 0.5367915630340576
```

rileyjmurray:
@SteveDiamond I actually think the changes are pretty simple. Parth just documented a pretty complicated development process. My benchmark above shows that it doesn't introduce any meaningful performance hit. I'm in favor of having this in 1.2. @akshayka please offer your thoughts as well.

cvxpy/expressions/expression.py
rileyjmurray: Ignore that comment. Might as well keep it in since dict accesses are so cheap.

PTNobel:
@SteveDiamond I have a bad habit of starting to write too many updates on what I'm doing. The final change is pretty simple. I'm very confident that I caught all the cases that might not have the target behaviour.

@rileyjmurray Here's a summary of the changes. , `function` be any NumPy function , and `bin_op_ufunc` any function in `__BINARY_EXPRESSION_UFUNCS__`.

 * For any NumPy ufunc not in `__BINARY_EXPRESSION_UFUNCS__` (e.g. `np.sqrt`, `np.log`) any CVXPY Expression arguments raises an error telling you to use a CVXPY function instead.
 * For a NumPy function which does not have a fallback implementation that uses a ufunc  (e.g. `np.hstack`, `np.vstack`), it behaves exactly the same as it used to.
 * If a NumPy function which does have a fallback implementation that uses a ufunc (e.g. `np.linalg.norm`), used to raise an error, and now raises an error telling you to use a CVXPY function instead.
 * If a NumPy ufunc is in `__BINARY_EXPRESSION_UFUNCS__` (e.g. `np.add`, `np.mulitply`) it will work if called with arguments (Any, cp.Expression) and no keyword args. Everything else will raise the error.
 * All inline binary operators (e.g. `np.ndarray * cp.Expression`) work the same as they used to.

PTNobel:
I just ran Riley's benchmark code on my computer; results are below. 

So `cp.Constant @ cp.Variable` shouldn't be impacted by these changes at all. So we're just testing noise on that one. Sure enough I see a performance improvement (of 3% i.e. noise) with the new code.

But looks like there's a 5% slowdown (8% - 3% baseline noise from the `cp.Constant @ cp.Expression`) in `np.ndarray @ 
cp.Expression`. We could eliminate the lambda function call pretty easily by just having a bunch of `if ufunc is np.add: ... elif ufunc is np.multiply: ...`, but also I don't think that is really necessary unless cvxcore gets a lot faster soon.

```
ufunc dictionary lookup times
        0.009126424789428711
        0.009150505065917969
        0.009407281875610352
        0.00994253158569336
        0.009477376937866211
        0.009453296661376953
        0.010060787200927734
        0.009533405303955078
        0.009590387344360352
        0.009737014770507812
        0.009598731994628906
        0.009429931640625
        0.010039567947387695

Matmul times: np.ndarray @ cp.Variable
        new: 2.4834892749786377
        old: 2.287794828414917

Matmul times: cp.Constant @ cp.Variable
        new: 0.5890123844146729
        old: 0.6018273830413818
```

SteveDiamond:
I got a chance to look at the code changes, and I retract my earlier concerns. I'm fine with merging this in for 1.2 once there are sufficient tests added.

PTNobel:
I'll aim to write tests tonight.

PTNobel:
Aside, do we not support `cp.Expression` being on the right side of `**`? I just discovered that `cp.Expression` doesn't implement `__rpow__` while writing tests.

PTNobel:
Alright, everything is done, I think?

The only exception is np.array([1])**cp.Variable() currently raises:

```
TypeError: unsupported operand type(s) for ** or pow(): 'numpy.ndarray' and 'Variable'
``` 

And now raises
```
AttributeError: 'Variable' object has no attribute '__rpow__'
```

It is easy to change this to:
```
TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'power'>, '__call__', array([1]), Variable(())): 'ndarray', 'Variable'
```

rileyjmurray:
@PTNobel rather than having an ``AttributeError`` be raised because of a failed dereference in
```
np.power: lambda self, a: self.__rpow__(a)
```
, could you just implement ``Expression.__rpow__`` so that it raises a ``NotImplementedError`` with an informative message?

PTNobel:
Is there a reason I should explain? Or is it just that no one has bothered yet? (Since 2**x is convex, it feels like it could be useful!)

Should I just implement it as `cp.exp(self * cp.log(other))`?

rileyjmurray:
> Aside, do we not support `cp.Expression` being on the right side of `**`? I just discovered that `cp.Expression` doesn't implement `__rpow__` while writing tests.

It does indeed seem that we don't support it! Nothing in our documentation suggests that should be possible. And the existing code raises an error. So I think it's fine to have this PR be deliberate about *what* error is raised.

I'll note that technically ``__rpow__`` could be implemented as follows
```
def expr_rpow(self, other):
    if hasattr(other, 'is_constant'):
        assert other.is_constant()
    if hasattr(other, 'value'):
        other = other.value
    if not isinstance(other, np.ndarray):
        other = np.asarray(other)
    assert np.all(other > 0)
    log_other = np.log(other)
    return cp.exp(cp.multiply(log_other, self))
```
But we're too close to the 1.2 release to talk about adding that kind of significant functionality. We can add this on master after the 1.2 release and then it'll be available in 1.3.

rileyjmurray:
> Is there a reason I should explain? Or is it just that no one has bothered yet? (Since 2**x is convex, it feels like it could be useful!)
> 
> Should I just implement it as `cp.exp(self * cp.log(other))`?

It's definitely just that no one has gotten around to it. Implementing it properly would involve a fair amount of argument validation, which I've given a taste of in my comment above. Let's not rush this and do it poorly. It will be nice to have in the 1.3 release!

The error message (if you haven't already decided on one) can just recommend that people use the identity ``a**b = exp(log(a)*b)`` and use the CVXPY atom for ``exp``.

cvxpy/tests/test_numpy.py
```diff
@@ -0,0 +1,76 @@
...
+        with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+            np.log(self.x)
+
+    def test_some_np_ufunc_works(self) -> None:
+        a = np.array([[1., 3.],[3., 1.]])
+        b = np.int64(1)
+
+        for ufunc in __BINARY_EXPRESSION_UFUNCS__:
+            if ufunc is np.multiply:
+                continue # We don't want to deal with the multiply warnings
```
rileyjmurray, line 48: If you use ``b * self.x`` that should perform element-wise multiplication without raising a warning.

cvxpy/tests/test_numpy.py
```diff
@@ -0,0 +1,76 @@
...
+                ufunc(self.x, b)
+
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(a, self.x, out=a)
+        
+            with pytest.raises(RuntimeError, match=__NUMPY_UFUNC_ERROR__):
+                ufunc(b, self.x, out=b)
+
+            if ufunc is not np.less and ufunc is not np.greater:
+                ufunc(a, self.x)
```
rileyjmurray, line 64: It would be good to have some kind of assertion here. I suggest setting ``x.value`` to something and checking that ``ufunc(a, self.x).value == ufunc(a, self.x.value)`` (up to some tolerance like 1e-10).

rileyjmurray (changes requested):

cvxpy/tests/test_numpy.py
PTNobel: Ah brilliant.

cvxpy/tests/test_numpy.py
PTNobel: Hmm so I added this for everything but the comparison operations, because the expected behaviour of `np.equal(a, self.x).value` is a (function which returns a) boolean and `np.equal(a, self.x.value)` is an array of booleans.

rileyjmurray: That sounds good to me!

cvxpy/tests/test_numpy.py
rileyjmurray: @PTNobel did you make this change? It looks like the latest on this PR still hits a ``continue``.

PTNobel: I split it into two for loops, one for the scalar and one for the matrix. I think they're working?

rileyjmurray (approved):

cvxpy/tests/test_numpy.py
rileyjmurray: Ah, I see it now. Thanks!

SteveDiamond (approved):
LGTM

Issue #1649, opened 2022-02-09
Interface to OR-Tools/PDLP


mlubin:
[OR-Tools](https://github.com/google/or-tools) (master branch) now includes a C++ implementation of the PDLP algorithm that we presented at NeurIPS 2021 ([paper](https://proceedings.neurips.cc/paper/2021/hash/a8fbbd3b11424ce032ba813493d95ad7-Abstract.html)). In contrast to the Julia prototype used the paper, the C++ implementation is multithreaded and a bit faster in serial. Our experience is that it's the fastest LP solver among first-order solvers, especially when solving to high precision (we compared with SCS, haven't benchmarked with OSQP). It's not close to being uniformly better than simplex or barrier solvers, but it does outperform them on occasion. We think this solver would be a useful option for CVXPY users.

Could you provide some guidance on how to create a CVXPY interface?

Some relevant technical points:
- PDLP solves LPs and QPs with diagonal objective matrices (only). The underlying algorithm could be generalized to conic problems, but we're unlikely to implement that anytime soon.
- OR-Tools already has a python interface installable through pip and conda. (Though these releases don't include PDLP yet.)
- OR-Tools has its own solver abstraction. It may make sense to create an "or-tools" wrapper rather than a direct "PDLP" wrapper in order to enable users to access other solvers like Glop. Glop is a production-ready simplex-based solver that's used widely within Google.

CC @dapplegate @lperron @Mizux @bodono

SteveDiamond:
Making an OR-Tools wrapper sounds like the way to go. For reference, here is our documentation on making interfaces: https://www.cvxpy.org/contributing/index.html#solver-interfaces

It's not going to be easy to target a standard form with specialized QP matrices. I would suggest just interfacing as an LP solver.

rileyjmurray:
@mlubin LP solvers follow the conic interface. Support for only diagonal quadratics is interesting. We'd need to think a bit more about how we'd integrate that into CVXPY's reduction system.

If you want this functionality available soon then I suggest you go the LP route. We want to release CVXPY 1.2 before the end of Feb. We're also trying to adopt a stable release cadence, so CVXPY 1.3 won't be available for four to six months after 1.2.

Mizux:
FYI once we release a new version of python `ortools` package with pdlp support, I'll post it here...

mkoeppe:
> * OR-Tools has its own solver abstraction. It may make sense to create an "or-tools" wrapper

[...]

PR #1655, opened 2022-02-13
Glop interface via OR-Tools

mlubin:
Following up on #1649, here's an interface to Glop via ortools. The PDLP interface should share 90% of the code, so when that's available in a public release I'll create a shared parent class following the example of CVXOPT. Until then, this code should allow us to work out most of the kinks in interfacing with ortools.

CC @dapplegate @lperron @Mizux


SteveDiamond:
This looks generally fine. I didn't totally understand how the USER_LIMIT status was being used. Besides the unit tests in test_conic_solvers.py, we have a test of the verbose keyword in test_problem.py. That will work automatically. You probably want to test the other keywords as well to the extent practicable. @rileyjmurray did I miss anything?

cvxpy/reductions/solvers/conic_solvers/glop_conif.py
```diff
@@ -0,0 +1,206 @@
...
+            problem = self.format_constraints(problem, None)
+        data[s.PARAM_PROB] = problem
+        data[self.DIMS] = problem.cone_dims
+
+        constr_map = problem.constr_map
+        inv_data["constraints"] = constr_map[Zero] + constr_map[NonNeg]
+
+        # Min c'x + d such that Ax + b = s, s \in cones.
+        c, d, A, b = problem.apply_parameters()
+        # TODO is csr_matrix necessary? What's the format of A?
```
rileyjmurray, line 69: I believe the default format of ``A`` is CSC.

cvxpy/reductions/solvers/conic_solvers/glop_conif.py
```diff
@@ -0,0 +1,206 @@
...
+                constraint.var_index.append(col_index)
+                constraint.coefficient.append(coeff)
+            if row_index < problem.cone_dims.zero:
+                # a'x + b == 0
+                constraint.lower_bound = -b[row_index]
+                constraint.upper_bound = -b[row_index]
+            else:
+                # a'x + b >= 0
+                constraint.lower_bound = -b[row_index]
+            model.constraint.append(constraint)
```
rileyjmurray, line 98: Can this be vectorized better? Right now it'll be quite slow for large problems. If ``A`` is effectively dense (as might happen in a basis pursuit problem) then this might be extremely slow.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -678,6 +678,37 @@ def test_glpk_mi_lp_4(self) -> None:
...
+    def test_glop_lp_0(self) -> None:
+        StandardTestLPs.test_lp_0(solver='GLOP')
+
+    def test_glop_lp_1(self) -> None:
+        StandardTestLPs.test_lp_1(solver='GLOP')
+
+    def test_glop_lp_2(self) -> None:
+        StandardTestLPs.test_lp_2(solver='GLOP')
+
+    def test_glop_lp_3(self) -> None:
```
rileyjmurray, line 693: I'd prefer to fix this rather than skipping the test.

If it isn't practical to fix this then the comments should be removed and this unit test should have an appropriate decorator (something like  ``@unittest.skip('Known limitation of the GLOP interface.')``).

cvxpy/tests/test_conic_solvers.py
```diff
@@ -678,6 +678,37 @@ def test_glpk_mi_lp_4(self) -> None:
...
+        # StandardTestLPs.test_lp_3(solver='GLOP')
+        pass
+
+    def test_glop_lp_4(self) -> None:
+        StandardTestLPs.test_lp_4(solver='GLOP')
+
+    def test_glop_lp_5(self) -> None:
+        StandardTestLPs.test_lp_5(solver='GLOP')
+
+    def test_glop_lp_6(self) -> None:
```
rileyjmurray, line 705: Same comment as the one I left for ``test_glop_lp_3``.

rileyjmurray:
Thanks for the PR, @mlubin! Can you also please update the web documentation: 

https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#choosing-a-solver

https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#setting-solver-options

mlubin:
Conda is having trouble installing `ortools-python` due to conflicts:

```
UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package numpy conflicts for:
ortools-python -> numpy[version='>=1.18.5,<2.0a0|>=1.19.5,<2.0a0']
scs -> scipy[version='>=0.13.2'] -> numpy[version='1.10.*|1.11.*|1.12.*|1.13.*|>=1.11|>=1.11.3,<2.0a0|>=1.19.4,<2.0a0|>=1.9|>=1.15.1,<2.0a0|>=1.21.2,<2.0a0']
scs -> numpy[version='>=1.14.6,<2.0a0|>=1.16.5,<2.0a0|>=1.16.6,<2.0a0|>=1.17.5,<2.0a0|>=1.18.5,<2.0a0|>=1.21.5,<2.0a0|>=1.19.5,<2.0a0|>=1.21.4,<2.0a0|>=1.19.2,<2.0a0|>=1.18.1,<2.0a0|>=1.9.3,<2.0a0']
osqp -> numpy[version='>=1.14.6,<2.0a0|>=1.16.6,<2.0a0|>=1.18.5,<2.0a0|>=1.19.5,<2.0a0|>=1.21.4,<2.0a0|>=1.18.1,<2.0a0|>=1.9.3,<2.0a0']
numpy=1.16
cvxopt -> mkl-service[version='>=2,<3.0a0'] -> numpy[version='>=1.11.3,<2.0a0']
osqp -> qdldl-python -> numpy[version='1.10.*|1.11.*|1.12.*|1.13.*|>=1.11|>=1.11.3,<2.0a0|>=1.16.5,<2.0a0|>=1.19.4,<2.0a0|>=1.19.2,<2.0a0|>=1.21.5,<2.0a0|>=1.17.5,<2.0a0|>=1.9|>=1.15.1,<2.0a0|>=1.21.2,<2.0a0']
scipy=1.3 -> numpy[version='>=1.11.3,<2.0a0|>=1.14.6,<2.0a0']
```

I think this means that ortools-python requires numpy >=1.18.5 while it's pinned to 1.16 in `install_dependencies.sh`. For reference, the [conda-forge package](https://github.com/conda-forge/or-tools-feedstock/blob/master/recipe/meta.yaml) doesn't fix a numpy version and the original [setup.py.in](https://github.com/google/or-tools/blob/stable/ortools/python/setup.py.in#L47) requires numpy 1.13.3. I'm asked to register for an account when I try to access the source of the conda package: https://anaconda.org/cf-staging/ortools-python.

cvxpy/tests/test_conic_solvers.py
mlubin: Added a working version with custom parameters and added `unittest.skip` for the default test as suggested.

mlubin (commented):
> I didn't totally understand how the USER_LIMIT status was being used.

If a solution is present while the solver didn't claim to solve the problem, I make sure USER_LIMIT is set: https://github.com/cvxpy/cvxpy/blob/master/cvxpy/settings.py#L59.

I updated the docs. The desired ordering wasn't clear. I tried to add GLOP before GLPK for local consistency.

doc/source/tutorial/advanced/index.rst
```diff
@@ -1392,6 +1412,7 @@ on derivatives.
 .. _SCS: http://github.com/cvxgrp/scs
 .. _GLPK: https://www.gnu.org/software/glpk/
 .. _GLPK_MI: https://www.gnu.org/software/glpk/
+.. _GLOP: https://developers.google.com/optimization
```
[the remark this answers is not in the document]
mlubin: Done

SteveDiamond:
Well this all looks great! Are you folks going to release an or-tools conda package compatible with earlier numpy? We could also install or-tools from pip in the CI if that helps.

mlubin:
> We could also install or-tools from pip in the CI if that helps.

I pushed a commit that does this. (CI needs approval to run.)

The OR-Tools devs don't control the conda package, but I'll look into it.

mlubin:
Cool! We do expect to have a new release in the next week or so that includes PDLP, but anyway it seems fine to merge this code and let people try out GLOP in the meantime.

SteveDiamond:
@rileyjmurray let's merge if this looks good to you

rileyjmurray (approved):
Looks good to me!

h-vetinari:
> I think this means that ortools-python requires numpy >=1.18.5 while it's pinned to 1.16 in `install_dependencies.sh`. For reference, the [conda-forge package](https://github.com/conda-forge/or-tools-feedstock/blob/master/recipe/meta.yaml) doesn't fix a numpy version and the original [setup.py.in](https://github.com/google/or-tools/blob/stable/ortools/python/setup.py.in#L47) requires numpy 1.13.3. I'm asked to register for an account when I try to access the source of the conda package: https://anaconda.org/cf-staging/ortools-python.

Hey all, I just saw this PR based on the release announcement. Regarding numpy, conda-forge has a so-called global [pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml) for every library that's somehow being compiled against and ABI-relevant. Since conda-forge is following [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), the oldest-supported numpy (by virtue of being the version that is being compiled against) is 1.18 (though depending on the python version obviously). I would suggest that cvxpy also consider following the same python/numpy deprecation policy.

I'd love to have ortools available for conda-installs unconditionally (or maybe have a separate output à la `cvxpy-ortools` that installs it), but it's currently [only](https://anaconda.org/conda-forge/ortools-python/files) available for linux64, whereas cvxpy is also on linux-aarch64, win-64, osx-64, osx-arm64.

Mizux:
The conda-forge recipe is here: https://github.com/conda-forge/or-tools-feedstock we should ask to @BastianZim directly ;)

BastianZim:
😄 We're working on bringing it to more platforms but are currently stuck on missing dependencies although @h-vetinari has just started on one, I believe. Once those are available, we will add more platforms as well.

h-vetinari:
> The conda-forge recipe is here: https://github.com/conda-forge/or-tools-feedstock we should ask to @BastianZim directly ;)

https://github.com/conda-forge/or-tools-feedstock/issues/9 🙃 

It's a tricky niche of the ecosystem 😅

lperron:
We will build all these archives. All but osx-arm64 next week.
osx-arm64 as soon as we have the hardware :-)
Laurent Perron | Operations Research | ***@***.*** | (33) 1 42 68 53
00



Le jeu. 10 mars 2022 à 09:54, h-vetinari ***@***.***> a
écrit :

Issue #1604, opened 2022-01-12
cvxpy pip package fails to install on M1 with python 3.9


si14:
**Describe the bug**

`cvxpy` pip package fails to install on M1 with python 3.9. The reason seems to be simple: `cvxpy` specifies a version of numpy https://github.com/cvxpy/cvxpy/blob/ac57b1ced831c319ea49de135c4a1c9477ca5292/pyproject.toml#L20 that is not compatible with M1 yet, 1.21.4 being the release they included M1 support in: https://github.com/numpy/numpy/releases/tag/v1.21.4

**To Reproduce**

Try to install `cvxpy` in a clean python 3.9 env on M1.

**Expected behavior**

It shouldn't fail.

**Output**
```
python-3.9.9 ❯ pip install cvxpy
Collecting cvxpy
  Using cached cvxpy-1.1.18.tar.gz (1.3 MB)
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /Users/foobar/python-3.9.9/bin/python /private/var/folders/fs/ygypst711711048hd13tfhnr0000gn/T/pip-standalone-pip-wvesyd8j/__env_pip__.zip/pip install --ignore-installed --no-user --prefix /private/var/folders/fs/ygypst711711048hd13tfhnr0000gn/T/pip-build-env-wqa2kq5g/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- 'numpy>=1.15,<1.16; python_version=='"'"'3.6'"'"'' 'numpy>=1.15,<1.16; python_version=='"'"'3.7'"'"'' 'numpy>=1.16,<1.17; python_version=='"'"'3.8'"'"'' 'numpy>=1.19,<1.20; python_version=='"'"'3.9'"'"'' 'numpy>=1.21,<1.22; python_version=='"'"'3.10'"'"'' 'scipy >= 1.1.0' 'setuptools>=40.8.0' wheel
       cwd: None
  Complete output (910 lines):
  Ignoring numpy: markers 'python_version == "3.6"' don't match your environment
  Ignoring numpy: markers 'python_version == "3.7"' don't match your environment
  Ignoring numpy: markers 'python_version == "3.8"' don't match your environment
  Ignoring numpy: markers 'python_version == "3.10"' don't match your environment
  Collecting numpy<1.20,>=1.19
    Using cached numpy-1.19.5.zip (7.3 MB)
    Installing build dependencies: started
[...]

Issue #1164, opened 2020-10-29
Double Calls to SCS Solver


nickari:
**Describe the bug**
CVXPY version 1.1.7 makes multiple calls to SCS to solve a single SDP, whereas version 1.0.25 does not.

**To Reproduce**
Run the attached python file with SCS version 2.1.2 and CVXPY version 1.1.7 installed, then compare to output from running with SCS version 2.1.2 and CVXPY version 1.0.25 installed.

[cvxpy_compare_test.zip](https://github.com/cvxgrp/cvxpy/files/5460616/cvxpy_compare_test.zip)

**Expected behavior**
A single call to SCS. Additionally that the output from
```
data, _, _ = prob.get_problem_data(cp.SCS)
```
followed by invoking SCS directly should match the output from calling
```
prob.solve(cvxpy.SCS)
```

**Output**
When running with CXVPY 1.0.25 and SCS 2.1.2, the output is:
```
----------------------------------------------------------------------------
	SCS v2.1.2 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
----------------------------------------------------------------------------
Lin-sys: sparse-direct, nnz in A = 210
eps = 1.00e-04, alpha = 1.50, max_iters = 30, normalize = 1, scale = 1.00
acceleration_lookback = 10, rho_x = 1.00e-03
Variables n = 70, constraints m = 138
Cones:	primal zero / dual free vars: 58
	linear vars: 12
	sd vars: 56, sd blks: 3
	exp vars: 12, dual exp vars: 0
Setup time: 3.10e-04s
----------------------------------------------------------------------------
 Iter | pri res | dua res | rel gap | pri obj | dua obj | kap/tau | time (s)
----------------------------------------------------------------------------
     0| 6.18e+19  6.22e+19  1.00e+00 -3.06e+20  2.44e+20  2.79e+20  3.31e-04 
    30| 1.56e-01  1.48e-01  2.82e-02 -3.13e+00 -2.94e+00  4.24e-17  3.46e-03 
----------------------------------------------------------------------------
Status: Solved/Inaccurate
Hit max_iters, solution may be inaccurate, returning best found solution.
Timing: Solve time: 3.47e-03s
	Lin-sys: nnz in L factor: 450, avg solve time: 2.09e-06s
	Cones: avg projection time: 9.04e-05s
	Acceleration: avg step time: 1.17e-05s
----------------------------------------------------------------------------
Error metrics:
[...]

Issue #1654, opened 2022-02-13
Deprecation warnings


erelsgl:
This line:

https://github.com/cvxpy/cvxpy/blob/f93d6b24d2e51033f24e0fff4553e0cbd968d55c/cvxpy/interface/scipy_wrapper.py#L17

raises a warning:

```
cvxpy\cvxpy\interface\scipy_wrapper.py:17: DeprecationWarning: Please use `spmatrix` from the `scipy.sparse` namespace, the `scipy.sparse.base` namespace is deprecated.
```

This line:
https://github.com/cvxpy/cvxpy/blob/0eb50cdfdf1a0590bbb4b6efd8713dbeb52d6c00/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py#L101

raises the following warning:
```
cvxpy\reductions\solvers\conic_solvers\scipy_conif.py:101: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```

I have Windows 10, in a virtual envirnoment based on Python 3.9. Scipy version is 1.8.0.

rileyjmurray:
Resolved by #1665 and #1670.

PR #1670, opened 2022-02-25
WIP: Fix M1 wheels, drop Python 3.6, lots of documentation

rileyjmurray:
## Description

I originally set out to fix issue #1604 for getting wheels to build on M1 macs. I investigated various possibilities (see my comment here: https://github.com/cvxpy/cvxpy/issues/1604#issuecomment-1046051882) and came to the conclusion that a simple use of environment markers in ``pyproject.toml`` should solve the problem. @akshayka @phschiele @h-vetinari please comment on if you think this will be sufficient!

I also resolved the deprecation warnings mentioned in #1654 and took steps to remove support for Python 3.6.

At some point, I got started on a documentation spree. @SteveDiamond @akshayka @bstellato: please review my changes to web documentation concerning the Reductions API. I basically took a stance that nothing in that realm is part of the public API. We need to take a stance there because we need to be able to determine what it means for a change to be backward-compatible. Definitely review the web documentation changes by building locally. I used a lot of ReStructuredText which made for clean web docs at some expense of messier source docs.

Thanks in advance for anyone's comments on this PR!

## Type of change
- [X] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/power.py
+++ b/cvxpy/atoms/elementwise/power.py
@@ -66,7 +66,7 @@ class power(Elementwise):
         is usually *exact*. No such approximation
         is used for DGP problems.
 
-        *UPDATE*: CVXPY supports exponential cone and power cone constraints.
+        CVXPY supports exponential cone and power cone constraints.
         Such constraints could be used to handle the ``power`` atom in DCP problems
         without relying on approximations. Such an approach would also result in
         fewer variables than the current method, even when the current method is
--- a/cvxpy/atoms/elementwise/power.py
+++ b/cvxpy/atoms/elementwise/power.py
@@ -57,16 +57,22 @@ class power(Elementwise):
 
     .. note::
 
-        For DCP problems, ``p`` cannot be represented exactly, so a rational,
-        i.e., fractional, **approximation** must be made. (No such approximation
-        is made for DGP problems.)
-
-        Internally, ``power`` computes a rational approximation
+        For DCP problems, ``power`` assumes ``p`` has a rational representation
+        with a small denominator. **Approximations** are employed when this is not
+        the case. Specifically, ``power`` computes a rational approximation
         to ``p`` with a denominator up to ``max_denom``.
         Increasing ``max_denom`` can give better approximations.
-
         When ``p`` is an ``int`` or ``Fraction`` object, the approximation
-        is usually **exact**.
+        is usually *exact*. No such approximation
+        is used for DGP problems.
+
+        *UPDATE*: CVXPY supports exponential cone and power cone constraints.
+        Such constraints could be used to handle the ``power`` atom in DCP problems
+        without relying on approximations. Such an approach would also result in
+        fewer variables than the current method, even when the current method is
+        an exact reformulation. If you're interested in helping enhance CVXPY with
+        this ability, please get in touch with us and check out
+        `GitHub Issue 1222 <https://github.com/cvxpy/cvxpy/issues/1222>`_!
 
     .. note::
 
--- a/cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
@@ -16,13 +16,12 @@ limitations under the License.
 """
 import time
 
-from packaging.version import Version
-
 import cvxpy.settings as s
 from cvxpy.reductions.solution import Solution, failure_solution
 from cvxpy.reductions.solvers import utilities
 from cvxpy.reductions.solvers.conic_solvers import scs_conif
 from cvxpy.reductions.solvers.conic_solvers.conic_solver import ConicSolver
+from cvxpy.utilities.versioning import Version
 
 
 class DIFFCP(scs_conif.SCS):
--- a/cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
@@ -15,7 +15,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 import time
-from distutils.version import StrictVersion
+
+from packaging.version import Version
 
 import cvxpy.settings as s
 from cvxpy.reductions.solution import Solution, failure_solution
@@ -71,7 +72,7 @@ class DIFFCP(scs_conif.SCS):
         attr = {}
         if solution["solve_method"] == s.SCS:
             import scs
-            if StrictVersion(scs.__version__) < StrictVersion('3.0.0'):
+            if Version(scs.__version__) < Version('3.0.0'):
                 status = scs_conif.SCS.STATUS_MAP[solution["info"]["statusVal"]]
                 attr[s.SOLVE_TIME] = solution["info"]["solveTime"]
                 attr[s.SETUP_TIME] = solution["info"]["setupTime"]
@@ -142,7 +143,7 @@ class DIFFCP(scs_conif.SCS):
 
         if solver_opts["solve_method"] == s.SCS:
             import scs
-            if StrictVersion(scs.__version__) < StrictVersion('3.0.0'):
+            if Version(scs.__version__) < Version('3.0.0'):
                 # Default to eps = 1e-4 instead of 1e-3.
                 solver_opts["eps"] = solver_opts.get("eps", 1e-4)
             else:
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -92,10 +92,11 @@ guide </contributing/index>` and join us `on Discord <https://discord.gg/4urRQeG
 
 **News.**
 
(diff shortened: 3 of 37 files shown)


doc/source/index.rst
```diff
@@ -115,7 +116,7 @@ guide </contributing/index>` and join us `on Discord <https://discord.gg/4urRQeG
     :maxdepth: 3
     :hidden:
 
-    tutorial/index
+    User's Guide <tutorial/index>
```
akshayka, line 119: Thanks! I prefer this over tutorial. This is a very small nit: can you change "User's Guide" to "User Guide" everywhere? The latter is standard. (https://numpy.org/doc/stable/user/, https://en.wikipedia.org/wiki/User_guide)

cvxpy/atoms/elementwise/power.py
```diff
@@ -57,16 +57,22 @@ class power(Elementwise):
...
+        the case. Specifically, ``power`` computes a rational approximation
         to ``p`` with a denominator up to ``max_denom``.
         Increasing ``max_denom`` can give better approximations.
-
         When ``p`` is an ``int`` or ``Fraction`` object, the approximation
-        is usually **exact**.
+        is usually *exact*. No such approximation
+        is used for DGP problems.
+
+        *UPDATE*: CVXPY supports exponential cone and power cone constraints.
```
akshayka, line 69: Nice addition. Is the text "`*UPDATE*:`" necessary?

akshayka:
Riley, this is fantastic. I've left very minor comments.

In the future we should take a pass over the "Advanced" section of the User Guide. It's a hodgepodge right now, some of which might be worth calling out in the non-advanced portion of the User Guide. In particular in might be worth highlighting the solve method options, as well as parametrized programs (especially when Philipp & Michael's massive improvements are merged in!)

cvxpy/atoms/elementwise/power.py
rileyjmurray: Thanks! Not really, no. I'll remove it.

doc/source/index.rst
rileyjmurray: Done!

cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
```diff
@@ -71,7 +72,7 @@ def invert(self, solution, inverse_data):
         attr = {}
         if solution["solve_method"] == s.SCS:
             import scs
-            if StrictVersion(scs.__version__) < StrictVersion('3.0.0'):
+            if Version(scs.__version__) < Version('3.0.0'):
```
SteveDiamond, line 75: Can you comment on the change here? Is this worth adding a new dependency?

doc/source/updates/index.rst
```diff
@@ -8,45 +8,50 @@ CVXPY's project maintainers currently provide support for CVXPY 1.2 and 1.1.
...
- important linear operators in quantum information.
-
+We review those features and the new features in CVXPY 1.2.0 below.
+
+Constraints and atoms
+~~~~~~~~~~~~~~~~~~~~~
+ * 1.2.0: added atoms for `partial trace <https://en.wikipedia.org/wiki/Partial_trace>`_ and partial transpose,
+   which are important linear operators in quantum information
+ * 1.2.0: updated ``kron`` so that either argument in ``kron(A, B)`` can be a non-constant affine Expression,
+   provided the other argument is constant. We previously required that ``A`` was constant.
```
SteveDiamond, line 28: 1.2.0: added xexp atom

SteveDiamond:
Fantastic stuff @rileyjmurray ! You should update the CI to drop python 3.6.

cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
rileyjmurray: Distutils is [deprecated](https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated) and recommends using ``packaging`` to access version comparisons. We already use setuptools though, and the Python foundation suggests using either setuptools or packaging in place of distutils. I'll check if we can get what we need with setuptools.

SteveDiamond: Ok, I didn't know distutils is deprecated.

doc/source/updates/index.rst
rileyjmurray: Good catch! Fixed.

rileyjmurray:
Thanks for the reviews, @SteveDiamond and @akshayka! I've made most of the changes. I just need to investigate setuptools vs packaging per Steven's comment, and I need to fix the tests which are failing for weird error tolerance reasons.

@akshayka, I wholeheartedly agree that the "advanced" section of the tutorial (now User Guide) is due for some reorganization. I have some ideas on that front. Happy to work on them in a month or so.

cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py
rileyjmurray: @SteveDiamond I looked at what ``setuptools`` offers for checking versions and it's not a practical replacement for ``packaging.version``. I just made a very simple version class that works for how we used ``distutils.StrictVersion``. This way we don't need to add ``packaging`` as a dependency.

rileyjmurray:
This also "fixes" #1672 by adding more informative error messages.

SteveDiamond (approved):
LGTM

rileyjmurray:
@akshayka @SteveDiamond @bodono some tests were failing with SCS 3.2. I went back and found that our SCS interface was only attempting re-solves with Anderson acceleration disabled for SCS 2.x. In order to get tests to pass, I had to enable re-solves with all versions of SCS >= 2.0. I added a warning message which is raised when the re-solve occurs, which largely resolves #1164. Note that the warning is raised regardless of whether or not ``verbose=True``. How do each of you feel about this approach?

 I also had to make termination criteria more stringent in some tests that used SCS.

rileyjmurray:
Okay turns out SCS 3.0 disables Anderson acceleration by default. I think I got confused with the source of errors when I initially ran tests with SCS 2.1.4 locally, say SCS 3.0 errors on GH Actions, and then updated my local install to 3.0. Give me some time to figure this out.

h-vetinari:
> @h-vetinari please comment on if you think this will be sufficient!

Thanks for the ping. I'll happily try this in a new release, but this isn't a big issue for conda-forge. Carrying a patch to remove (currently) `pyproject.toml` isn't amazing, but also not terrible either. If it's not needed anymore, all the better, but won't make-or-break the workflow for us.

rileyjmurray:
@bodono 

> SCS 3.x hasn't disabled AA by default, but it should be much, much more stable and cheaper to apply, so I'm surprised to hear you are having issues with it

Okay, I misunderstood the SCS web documentation. It seemed like the default value stated there was ``acceleration_lookback=0``. But I looked at some solver output when cvxpy didn't specify AA lookback and it said lookback=10. 

All in all, it seems I misattributed the various failures to AA. There were two things that went wrong in SCS testing:

1. SCS 3.2 fails some accuracy tests that previous versions of SCS 3.x passed. Minor changes to accuracy parameter ``eps`` in calls to ``prob.solve()`` fixed this.
2. Our test ``TestSCS.test_scs_options`` provided an usual parameter of ``alpha=1e-4``, which caused SCS to diverge. This had nothing to do with AA. Similar divergence was occurring in SCS 3.1 and 3.0, but the iteration limit was so low that SCS wasn't able to diverge "too much" and happened to pass the tests which had very generous tolerances. The solution here was just to fix ``alpha`` to a more reasonable value like 1.0 or 1.4.

Since neither of these was due to AA we have a decision to make: when encountering a solver failure with SCS 3.x and default AA, should cvxpy automatically re-solve the problem with AA turned off? That's the approach we take with 2.x. @bodono since you say AA is much more reliable I'm open to putting this behind us, and only doing re-solves with SCS 2.x is installed.

@SteveDiamond, @phschiele @akshayka your opinions on this would also be appreciated.

phschiele:
@rileyjmurray In my view, since the current behaviour did not lead to issues being reported that were caused by AA, as well as the additional information that it is generally much more stable under 3.x, I think it's ok to only resolve under 2.x.

SteveDiamond:
I agree with Philipp, if we can avoid resolving that's preferable.

rileyjmurray:
That's three in favor! I expect Akshay and Brendan will either agree or have only mild dissent. I'll make the necessary final changes and then merge. Akshay and Brendan: we can of course revisit if you feel strongly about this.

bodono:
> Okay, I misunderstood the SCS web documentation. It seemed like the default value stated there was acceleration_lookback=0. But I looked at some solver output when cvxpy didn't specify AA lookback and it said lookback=10.

Sorry, the documentation was wrong. I fixed that here: https://github.com/cvxgrp/scs/pull/212

rileyjmurray:
Well, shoot. I merged it in without changing the "WIP" title. Ah well. We're done here!

Issue #1678, opened 2022-03-03
cumsum axis bug


SteveDiamond:
**Describe the bug**
``cumsum(x, axis=1)`` breaks when ``x`` has shape ``(n, 1)``. Similarly, ``cumsum(x, axis=0)`` breaks when ``x`` has shape ``(1, n)``. This bug should be easy to fix by adding special cases to https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py

**To Reproduce**
```
import cvxpy as cp

n = 5
axis = 1

if axis == 0:
  x = cp.Variable((1, n))
  expr = cp.cumsum(x, axis=0)

  prob = cp.Problem(cp.Minimize(0),
                   [expr == 1])
  prob.solve()
  print(x.value)
elif axis == 1:
  x = cp.Variable((n, 1))
  expr = cp.cumsum(x, axis=1)

  prob = cp.Problem(cp.Minimize(0),
                   [expr == 1])
  prob.solve()
  print(x.value)
```

**Expected behavior**
The problem should solve without crashing :P

**Output**
```
Traceback (most recent call last):
  File "/Users/sdiamond/Projects/tests/cumsum_test.py", line 20, in <module>
    prob.solve()
  File "/Users/sdiamond/opt/anaconda3/envs/cvxpy/lib/python3.9/site-packages/cvxpy/problems/problem.py", line 473, in solve
    return solve_func(self, *args, **kwargs)
  File "/Users/sdiamond/opt/anaconda3/envs/cvxpy/lib/python3.9/site-packages/cvxpy/problems/problem.py", line 966, in _solve
    data, solving_chain, inverse_data = self.get_problem_data(
  File "/Users/sdiamond/opt/anaconda3/envs/cvxpy/lib/python3.9/site-packages/cvxpy/problems/problem.py", line 627, in get_problem_data
    data, inverse_data = solving_chain.apply(self, verbose)
  File "/Users/sdiamond/opt/anaconda3/envs/cvxpy/lib/python3.9/site-packages/cvxpy/reductions/chain.py", line 76, in apply
    problem, inv = r.apply(problem)
  File "/Users/sdiamond/opt/anaconda3/envs/cvxpy/lib/python3.9/site-packages/cvxpy/reductions/dcp2cone/dcp2cone.py", line 43, in apply
    return super(Dcp2Cone, self).apply(problem)
[...]

PR #1680, opened 2022-03-04
Updated cumsum_canon.py for axis bug

parthb83:
## Description
Fixed the axis bug in cumsum() by adding special cases to https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
Issue link (if applicable): Issue #1678 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
+++ b/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
@@ -26,7 +26,13 @@ def cumsum_canon(expr, args):
     # X = Y[1:,:] - Y[:-1, :]
     Y = Variable(expr.shape)
     if axis == 0:
-        constr = [X[1:] == Y[1:] - Y[:-1], Y[0] == X[0]]
+        if (expr.shape[0] == 1):
+            constr = [Y[0] == X[0]]
+        else:
+            constr = [X[1:] == Y[1:] - Y[:-1], Y[0] == X[0]]
     else:
-        constr = [X[:, 1:] == Y[:, 1:] - Y[:, :-1], Y[:, 0] == X[:, 0]]
+        if (expr.shape[1] == 1):
+            constr = [Y[:, 0] == X[:, 0]]
+        else:
+            constr = [X[:, 1:] == Y[:, 1:] - Y[:, :-1], Y[:, 0] == X[:, 0]]
     return (Y, constr)
--- a/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
+++ b/cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
@@ -27,12 +27,12 @@ def cumsum_canon(expr, args):
     Y = Variable(expr.shape)
     if axis == 0:
         if (expr.shape[0] == 1):
-            constr = [Y[0] == X[0]]
+            return (X, [])
         else:
             constr = [X[1:] == Y[1:] - Y[:-1], Y[0] == X[0]]
     else:
         if (expr.shape[1] == 1):
-            constr = [Y[:, 0] == X[:, 0]]
+            return (X, [])
         else:
             constr = [X[:, 1:] == Y[:, 1:] - Y[:, :-1], Y[:, 0] == X[:, 0]]
     return (Y, constr)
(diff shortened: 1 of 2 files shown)


cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
```diff
@@ -26,7 +26,13 @@ def cumsum_canon(expr, args):
     # X = Y[1:,:] - Y[:-1, :]
     Y = Variable(expr.shape)
     if axis == 0:
-        constr = [X[1:] == Y[1:] - Y[:-1], Y[0] == X[0]]
+        if (expr.shape[0] == 1):
+            constr = [Y[0] == X[0]]
```
SteveDiamond, line 30: I think it would be clearer to just return (X, []) when you have`` axis == 0`` and ``expr.shape[0] == 1``, similarly for ``axis == 1``.

SteveDiamond:
Thanks for doing a PR! There might be a clearer way to do this (see comment). Also, for a PR we need to add unit tests. You can put something in ``test_problem.py``.

parthb83:
```
def test_cumsum_axis(self) -> None:
        """Test the cumsum axis bug with row or column matrix
           See issue #1678
        """
        n = 5

        # Solve for axis = 0
        x1 = cp.Variable((1, n))
        expr1 = cp.cumsum(x1, axis=0)
        prob1 = cp.Problem(cp.Minimize(0),
                           [expr1 == 1])
        prob1.solve()

        # Solve for axis = 1
        x2 = cp.Variable((n, 1))
        expr2 = cp.cumsum(x2, axis=1)
        prob2 = cp.Problem(cp.Minimize(0),
                           [expr2 == 1])
        prob2.solve()
```

Is this test okay? Passes on testing test_problem.py

cvxpy/reductions/eliminate_pwl/atom_canonicalizers/cumsum_canon.py
parthb83:
```
if axis == 0:
        if (expr.shape[0] == 1):
            return (X, [])
        else:
            constr = [X[1:] == Y[1:] - Y[:-1], Y[0] == X[0]]
else:
        if (expr.shape[1] == 1):
            return (X, [])
        else:
            constr = [X[:, 1:] == Y[:, 1:] - Y[:, :-1], Y[:, 0] == X[:, 0]]
```

Final Changes.

SteveDiamond:
This is great, thanks!

cvxpy/tests/test_problem.py
```diff
@@ -1987,3 +1987,23 @@ def test_rmul_param(self) -> None:
         param.value = np.array([1])
         prob.solve()
         assert prob.value == -np.inf
+
+    def test_cumsum_axis(self) -> None:
```
rileyjmurray, line 1991: This unit test has no assertions.

rileyjmurray (commented):
@SteveDiamond @parthb83 the ``test_cumsum_axis`` test doesn't have any assertions. I know it was written to make sure there wasn't an error at canonicalization time, but we do need to make sure canonicalization is performed correctly. I also noticed that there's a ``cummax`` atom which is very similar to ``cumsum``. I'll update the ``cumsum`` test, add a ``cummax`` test, and update ``cummax`` as applicable on the PR for CVXPy 1.1.19. @parthb83 can then copy my changes over into a new PR onto master. (If I don't hear from @parthb83 by 5pm PT today I'll make the PR instead.)

cvxpy/tests/test_problem.py
parthb83: Sorry for not writing assertion statement for `test_cumsum_axis()` but I was pretty confused as to what assertion to use in this case (a little newbie in unit testing) and also I saw that in `test_huber_scs()` in https://github.com/cvxpy/cvxpy/blob/master/cvxpy/tests/test_problem.py/#L1955-L1974 
also doesn't have any assertions so decided not to write them as I thought it was acceptable.

I'm really sorry for this mistake and would be happy to write the assertions for `test_cumsum_axis()`.

parthb83: @rileyjmurray should we add assertions to `test_huber_scs()` too?

rileyjmurray: @parthb83 no worries! I also noticed ``test_huber_scs`` doesn't have assertions, so it's clear you were just following that example. I merged your other PR without changes to ``test_huber_scs`` since it felt out-of-scope. Ideally we should go through all of cvxpy's unit tests and fix ones which don't have assertions. But we have over 900 tests, so that will be a lot of work! We'll need to plan it in advance.

Issue #1619, opened 2022-01-28
GLPK_MI options are not reset


allenlawrence94:
**Describe the bug**
When solving a problem using GLPK's mixed-integer solver, if you set an option when calling `problem.solve`, this option is never reset; it persists for subsequent calls of `problem.solve`. (With the exception of `verbose` / `msg_lev`, because this is set explicitly in the cvxpy interface)

This is because cvxpy's` GLPK_MI` interface sets options via `cvxopt.glpk.options`, but then uses the inherited `GLPK._restore_solver_options` which resets the `cvxopt.solvers.options`.

For similar reasons, the GLPK and GLPK_MI solvers have inconsistent option-setting interfaces. GLPK takes options like `problem.solve(solver=cvxpy.GLPK, glpk={"tm_lim": 10})`, while GLPK_MI takes options like `problem.solve(solver=cvxpy.GLPK_MI, tm_lim=10)`. I'm not sure which behavior is desired.

**To Reproduce**
```python
import cvxpy as cvx
import numpy as np
import cvxopt

dim = 100
sell = cvx.Variable(shape=(dim,), pos=True)
price = np.random.random(dim) - 0.5
problem = cvx.Problem(
    objective=cvx.Maximize(price @ sell),
    constraints=[
        sell <= 1,
        sum(sell) <= dim // 2,
    ]
)

assert "tm_lim" not in cvxopt.glpk.options  # GLPK_MI uses cvxopt.glpk.options
# assert "tm_lim" not in cvxopt.solvers.options  # GLPK uses cvxopt.solvers.options

try:
    problem.solve(solver=cvx.GLPK_MI, tm_lim=10)
    # problem.solve(solver=cvx.GLPK, glpk={"tm_lim": 10})
except:
    pass

assert "tm_lim" not in cvxopt.glpk.options, "tm_lim is still set in glpk.options!"  # this fails
# assert "tm_lim" not in cvxopt.solvers.options, "tm_lim is still set in solvers.options!"  # this passes
```

**Expected behavior**
All solver options should be reset to default between calls of `problem.solve`

**Version**
 - OS: Ubuntu 18.04.6 LTS 
 - CVXPY Version: 1.2.0.dev0+1097.88ff6c9 (master) and 1.1.15

SteveDiamond:
Oops! Is the GLPK (non-MIP) interface bugged in the same way?

allenlawrence94:
Seems the non-MIP interface correctly resets parameters.

[...]

PR #1684, opened 2022-03-07
Fix for resetting solver options for GLPK_MI and muting solver output for GLPK when verbose=False

Michael-git96:
## Description
1) CVXPY was not resetting solver options correctly after running GLPK's mixed-integer solver.
2) When verbose was set to False, the GLPK solver's output was not being muted.
Issue link: https://github.com/cvxpy/cvxpy/issues/1619

## Type of change
Small bug fix


cvxpy/reductions/solvers/conic_solvers/glpk_mi_conif.py
```diff
@@ -110,3 +110,12 @@ def invert(self, solution, inverse_data):
             return Solution(status, opt_val, primal_vars, None, {})
         else:
             return failure_solution(status)
+
+    @staticmethod
+    def _restore_solver_options(old_options) -> None:
+        import cvxopt.glpk
+        for key, value in list(cvxopt.glpk.options.items()):
```
akshayka, line 117: Why is the implementation of this method different from the implementation in the `GLPK` class? The latter iterates over `cvxopt.solvers.options.items()` instead of `cvxopt.glpk.options.items()`. Should `GLPK`s implementation be updated to match this?

@SteveDiamond

akshayka (approved):
LGTM. If it's not too much trouble, can you add a unit test to check that options are restored for both `GLPK` and `GLPK_MI`?

cvxpy/reductions/solvers/conic_solvers/glpk_mi_conif.py
Michael-git96: We could change the GLPK class so that it uses cvxopt.glpk.lp to solve LPs. This implementation would look pretty much the same as the GLPK_MI class which uses cvxopt.glpk.ilp. However there is not much documentation around that and we would lose some of the information that CVXOPT calculates such as primal cost, dual cost, slack etc

Michael-git96:
Sure, I will try and add a unit test.

cvxpy/reductions/solvers/conic_solvers/glpk_mi_conif.py
akshayka: Thanks for explaining. Let's leave it as is for now.

akshayka (approved):
LGTM, thanks for adding the tests!

SteveDiamond (approved):

Issue #1690, opened 2022-03-10
Compiling with assertions causes issues


r-barnes:
**Describe the bug**
Compiling with assertions enabled leads to compile-time and run-time errors!

**To Reproduce**
I modify `setup.py` so that I has the following extension code:
```
canon = Extension(
    '_cvxcore',
    sources=['cvxpy/cvxcore/src/cvxcore.cpp',
             'cvxpy/cvxcore/src/LinOpOperations.cpp',
             'cvxpy/cvxcore/src/Utils.cpp',
             'cvxpy/cvxcore/python/cvxcore_wrap.cxx'],
    include_dirs=['cvxpy/cvxcore/src/',
                  'cvxpy/cvxcore/python/',
                  'cvxpy/cvxcore/include/'],
    extra_compile_args=[
        '-O3',
        '-std=c++11',
        '-Wall',
        '-pedantic',
        not_on_windows('-Wextra'),
        not_on_windows('-Wno-unused-parameter'),
    ],
    extra_link_args=['-O3'],
    undef_macros = [ "NDEBUG" ]  #### ATTENTION!
)
```
setuptools(?) seems to compile code with the `-DNDEBUG` flag on by default. This eliminates all assertions before compilation begins. assertions make us safe, so we should probably include them.

The introduction of the new line `undef_macros = [ "NDEBUG" ]` will cause the C++ code to compiled with assertions included, but this causes issues.

**Expected behavior**
Things compile and run without problems.

**Output**
Compilation issues on current master (TODO):
```
cvxpy/cvxcore/src/LinOpOperations.cpp: In function ‘Tensor get_kronl_mat(const LinOp&, int)’:
cvxpy/cvxcore/src/LinOpOperations.cpp:379:25: error: ‘Eigen::SparseMatrixBase<Eigen::SparseMatrix<double> >::IsRowMajor’ cannot be used as a function
  379 |   assert(!rh.IsRowMajor());
      |                         ^
cvxpy/cvxcore/src/LinOpOperations.cpp: In function ‘Tensor get_vstack_mat(const LinOp&, int)’:
cvxpy/cvxcore/src/LinOpOperations.cpp:430:18: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<const LinOp*>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  430 |   assert(arg_idx <= lin.get_args().size());
[...]

PR #1697, opened 2022-03-11
Fix IsRowMajor assertion

r-barnes:
## Description
Compilation currently fails when assertions are enabled with:
```
cvxpy/cvxcore/src/LinOpOperations.cpp: In function ‘Tensor get_kronl_mat(const LinOp&, int)’:
cvxpy/cvxcore/src/LinOpOperations.cpp:379:25: error: ‘Eigen::SparseMatrixBase<Eigen::SparseMatrix<double> >::IsRowMajor’ cannot be used as a function
  379 |   assert(!rh.IsRowMajor());
      |                         ^
```
Here, we fix this issue by referring to the enum directly. The code now compiles, though I can't test it because master is broken with `AttributeError: module '_cvxcore' has no attribute 'KRON_R'`

Issue link: #1690 

## Type of change
- [X] Bug fix

- [ ] Write unittests.


rileyjmurray:
@r-barnes regarding ``KRON_R``, that failure message is odd. Is it possible that Python is finding an old version of ``_cvxcore``? It would help if you could show the workflow you're using to modify cvxcore (including how you run swig and try to install cvxpy).

phschiele:
@r-barnes I tried this on a fresh clone of your fork. Indeed, installing fails with assertions enabled when your fix is not in place.
Regarding your comment `ATTENTION!`. Does that imply that this change should not be merged? I don't know which downsides it would have to keep assertions permanently enabled, but an alternative could be to set the flag `CFLAGS="-UNDEBUG"` while installing on the CI, so every code will be checked with assertions without making changes to `setup.py`.

r-barnes:
@phschiele - I put the attention there because it's possible `NDEBUG` can affect performance. I would _not_ expect this to be the case unless you have `assert` inside of a tight loop or class accessors.

phschiele:
@cvxpy/maintainers Could you please merge the current master into this branch so we get retrigger the CI and get the benchmarks?

SteveDiamond:
> @cvxpy/maintainers Could you please merge the current master into this branch so we get retrigger the CI and get the benchmarks?

Done.

phschiele:
Since we rarely make changes to cvxcore, I have instead added instructions on how to enable debug checks to the README of CVXCORE.

SteveDiamond (approved):

PR #1703, opened 2022-03-13
CBC.solve_via_data: Work around broken CyLP infeasibility detection

mkoeppe:
## Description
Please include a short summary of the change.
Issue link (if applicable):

https://github.com/coin-or/CyLP/issues/81
@tkralphs

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [X] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #1705, opened 2022-03-14
cvxpy/tests: Add test for infeasible boolean problem

mkoeppe:
## Description

We add a new test `StandardTestLPs.test_mi_lp_5`. This test fails with CBC because of https://github.com/coin-or/CyLP/issues/81

#1703 attempts to fix it.

Issue link (if applicable):

https://trac.sagemath.org/ticket/31962#comment:48

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -573,6 +573,14 @@ class TestCVXOPT(BaseTest):
         StandardTestSDPs.test_sdp_2(solver='CVXOPT')
 
 
+def _cylp_has_isProvenInfeasible():
+    try:
+        from cylp.cy.CyCBCModel import CyCbcModel
+        return hasattr(CyCbcModel, 'isProvenInfeasible')
+    except ImportError:
+        return False
+
+
 @unittest.skipUnless('CBC' in INSTALLED_SOLVERS, 'CBC is not installed.')
 class TestCBC(BaseTest):
 
@@ -640,6 +648,7 @@ class TestCBC(BaseTest):
     def test_cbc_mi_lp_3(self) -> None:
         StandardTestLPs.test_mi_lp_3(solver='CBC')
 
+    @unittest.skipUnless(_cylp_has_isProvenInfeasible(), 'CyLP <= 0.91.4 has no working integer infeasibility detection')
     def test_cbc_mi_lp_5(self) -> None:
         StandardTestLPs.test_mi_lp_5(solver='CBC')
 
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -589,7 +589,6 @@ class TestCBC(BaseTest):
         self.B = cp.Variable((2, 2), name='B')
         self.C = cp.Variable((3, 2), name='C')
 
-    @staticmethod
     def _cylp_has_isProvenInfeasible():
         try:
             from cylp.cy.CyCbcModel import CyCbcModel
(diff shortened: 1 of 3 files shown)


mkoeppe:
> then update `cvxpy/tests/test_conic_solvers.py`

with one method for every solver that can handle integer variables?

rileyjmurray:
Does this problem have a clear qualitative property that existing test MILPs do not? An example of such a property is "the root LP relaxation is infeasible". If this problem does introduce a clear qualitative property compared to existing test cases, then yes, please add the test for every solver that supports MILP.

rileyjmurray:
The new tests are failing in several environments, e.g.,

https://github.com/cvxpy/cvxpy/runs/5549355220?check_suite_focus=true#step:6:1367

https://github.com/cvxpy/cvxpy/runs/5549355386?check_suite_focus=true#step:6:1375

mkoeppe:
> The new tests are failing in several environments, e.g.,
> 
> https://github.com/cvxpy/cvxpy/runs/5549355220?check_suite_focus=true#step:6:1367
> 
> https://github.com/cvxpy/cvxpy/runs/5549355386?check_suite_focus=true#step:6:1375

Yes, this is working as intended. The CBC solver is broken until a working version of CyLP with https://github.com/coin-or/CyLP/pull/150 is released and deployed.

mkoeppe:
> Does this problem have a clear qualitative property that existing test MILPs do not? 

I'd say clear enough, it looks like it is the first tested example of an infeasible ILP for which ILP preprocessing does not already lead to an infeasible relaxation, at least for CBC's preprocessor, with default settings.

So I'm going to add it to the other solvers.

rileyjmurray:
@mkoeppe we don't want to merge changes that result in failing tests. I suggest the new CBC test be modified so it expects an assertion failure when CyLP is at or below a certain version and raises a warning to let the user know that situation was encountered. The test should expect no assertion failure when CyLP is a sufficiently high version.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -573,6 +573,14 @@ def test_cvxopt_sdp_2(self) -> None:
...
 
 
+def _cylp_has_isProvenInfeasible():
+    try:
+        from cylp.cy.CyCBCModel import CyCbcModel
+        return hasattr(CyCbcModel, 'isProvenInfeasible')
+    except ImportError:
+        return False
+
+
```
rileyjmurray, line 583: Please move this into the TestCBC class.

rileyjmurray:
@mkoeppe this looks good to me. You might want to change how CBC status codes are handled per the discussion in https://github.com/cvxpy/cvxpy/commit/d390389cab6a44cd66afff756563b53c4a8aed3a#commitcomment-68775673, which came up after PR #1703 was merged.

mkoeppe:
> You might want to change how CBC status codes are handled per the discussion in [d390389#commitcomment-68775673](https://github.com/cvxpy/cvxpy/commit/d390389cab6a44cd66afff756563b53c4a8aed3a#commitcomment-68775673), which came up after PR #1703 was merged.

I think this needs more discussion, so it's best to leave it for a separate PR.

mkoeppe:
In the CI configurations, it now skips the new test as intended.

I've tested this version locally with https://trac.sagemath.org/ticket/31962 (https://git.sagemath.org/sage.git/commit/?id=6daae939c9ee40f8fa5643a85dc655c067a21525), and it runs and passes the test.

rileyjmurray:
> > You might want to change how CBC status codes are handled per the discussion in [d390389#commitcomment-68775673](https://github.com/cvxpy/cvxpy/commit/d390389cab6a44cd66afff756563b53c4a8aed3a#commitcomment-68775673), which came up after PR #1703 was merged.
> 
> I think this needs more discussion, so it's best to leave it for a separate PR.

Agreed.

rileyjmurray (approved):
LGTM. I'll wait to merge until the end of the day in case @SteveDiamond or @akshayka want to review it.

PR #1719, opened 2022-03-20
pip install cvxpy[GLOP,XPRESS]

mkoeppe:
## Description
Issue link (if applicable):
Add `extras_require` mapping solver names to the pip-installable distribution providing it

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,16 @@
+[options.extras_require]
+# Solver names as in cvxpy.settings = pip-installable distribution providing it
+CBC = cylp
+CVXOPT = cvxopt
+GLOP = ortools
+GLPK = cvxopt
+GLPK_MI = cvxopt
+PDLP = ortools
+SCIP = PySCIPOpt
+SCIPY = scipy
+SDPT3 = sdpt3glue
+XPRESS = xpress
+
 [flake8]
 max-line-length=100
 exclude=build,examples,doc,cvxpy/cvxcore/*,*__init__.py
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,6 @@
 CBC = cylp
 CVXOPT = cvxopt
 DIFFCP = diffcp
-DSDP = cvxopt
 ECOS =
 ECOS_BB =
 GLOP = ortools
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,6 @@ PDLP = ortools
 SCIP = PySCIPOpt
 SCIPY = scipy
 SCS =
-SDPT3 = sdpt3glue
 XPRESS = xpress
 
 [flake8]
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,6 +5,7 @@ CVXOPT = cvxopt
 GLOP = ortools
 GLPK = cvxopt
 GLPK_MI = cvxopt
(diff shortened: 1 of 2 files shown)


SteveDiamond:
This is very interesting! If we make this change, we should update the install guide. Also, mosek can be pip installed (though you need a license to solve anything).

setup.cfg
```diff
@@ -1,3 +1,25 @@
...
+GLPK_MI = cvxopt
+GUROBI = gurobipy
+HIGHS = scipy>=1.6.1
+MOSEK = Mosek
+OSQP =
+PDLP = ortools
+SCIP = PySCIPOpt
+SCIPY = scipy
+SCS =
+SDPT3 = sdpt3glue
```
rileyjmurray, line 20: I don't think it's accurate to say that we support SDPT3. From [the sdpt3glue documentation](https://pythonhosted.org/sdpt3glue/quickguide.html), they clearly structure things so you can call ``prob.solve(solver='SDPT3')`` on a CVXPY Problem object.

setup.cfg
```diff
@@ -1,3 +1,25 @@
+[options.extras_require]
+# Solver names as in cvxpy.settings = pip-installable distribution providing it
+CBC = cylp
+CVXOPT = cvxopt
+DIFFCP = diffcp
+DSDP = cvxopt
```
rileyjmurray, line 6: [CVXOPT's documentation](https://cvxopt.org/userguide/coneprog.html#semidefinite-programming) says that it's DSDP interface can only be called through ``cvxopt.solvers.sdp`` rather than ``cvxopt.solvers.conelp``, so we'd actually need to update the CVXOPT interface to support DSDP.

The documentation also says that CVXOPT's DSDP interface doesn't support problems with equality constraints, so we'd need to figure out how we want to handle that (or if it's appropriate in CVXPY's context).

My suggestion is that we forgo this for the time being.

setup.cfg
```diff
@@ -1,3 +1,25 @@
...
+ECOS =
+ECOS_BB =
+GLOP = ortools
+GLPK = cvxopt
+GLPK_MI = cvxopt
+GUROBI = gurobipy
+HIGHS = scipy>=1.6.1
```
rileyjmurray, line 13: This is a nice touch!

rileyjmurray (changes requested):
Nice work! I left comments about ``sdpt3glue`` and DSDP. Please respond to those in one form or another.

setup.cfg
mkoeppe: I took this from the documentation https://www.cvxpy.org/install/#install-with-sdpt3-support; I don't know anything more specific about it.

rileyjmurray: The language there is a little different with SDPT3 than with other packages. For the main text, it doesn't say "CVXPY supports the SDPT3 solver," which would be analogous to the main text used for other solvers. I think the heading is a little misleading and should be changed. Please remove the SDPT3 line here, and then I'll approve the PR.

mkoeppe: Done.

mkoeppe:
Thanks for the fix for the documentation, which I had forgotten

rileyjmurray:
You're welcome! Thanks for the PR :)

PR #1736, opened 2022-04-03
Fix TypeError in GLOP and PDLP interfaces

### LINKED ISSUE #1734: TypeError in PDLP and GLOP interfaces
Our CI builds are failing because of issues with the PDLP and GLOP interfaces. The problem is that the offset ``d`` in the canonicalized objective ``c @ x + d`` is of type ``ndarray`` with ``d.shape = (,)``, while the most recent PDLP and GLOP interfaces seem to require that ``d`` is a pure numeric type. This can be handled by replacing ``d = d.item()`` in the relevant interfaces.

rileyjmurray:
Fixes #1734


rileyjmurray:
Propagated to 1.2.1.

Issue #1702, opened 2022-03-12
Preferred open source mixed-integer solvers in CVXPY (for quasiconvex problems)


keithbriggs:
https://www.cvxpy.org/tutorial/advanced/index.html says "The preferred open source mixed-integer solvers in CVXPY are [GLPK_MI](https://www.gnu.org/software/glpk/), [CBC](https://projects.coin-or.org/Cbc) and [SCIP](https://scip.zib.de/)".

https://pypi.org/project/cbcpy/ says "pip install cbcpy".   But that command gives:

```
ERROR: Could not find a version that satisfies the requirement cbcpy (from versions: none)
ERROR: No matching distribution found for cbcpy
```

So is CBC really supported?

Moreover, this script:

```
import cvxpy as cp

x = cp.Variable(boolean=True)
y = cp.Variable(boolean=True))
z = cp.Variable(boolean=True))

objective_fn = (7*y-6+z)/(2*x+3)
problem = cp.Problem(cp.Maximize(objective_fn),[x<=y,z<=x])
assert problem.is_dqcp()

for solver in ('ECOS_BB','GLPK_MI','SCIP',):
  print(f'solver={solver}:')
  problem.solve(qcp=True,solver=solver)
  print(f'  Optimal value={problem.value:.4f}')
  print(f'  x={x.value:.4f}')
  print(f'  y={y.value:.4f}')
  print(f'  z={z.value:.4f}')
```

gives 

```
solver=ECOS_BB:
  Optimal value=0.4000
  x=1.0000
  y=1.0000
  z=1.0000
solver=GLPK_MI:
  Optimal value=0.4000
  x=1.0000
  y=1.0000
  z=1.0000
solver=SCIP:
Error encountered when optimising <pyscipopt.scip.Model object at 0x7f6f735f88e0>: getSolObjVal cannot only be called in stage SOLVING without a valid solution (current stage: 10)
Error encountered when optimising <pyscipopt.scip.Model object at 0x7f6f735f8640>: getSolObjVal cannot only be called in stage SOLVING without a valid solution (current stage: 10)
Error encountered when optimising <pyscipopt.scip.Model object at 0x7f6f735f8d60>: getSolObjVal cannot only be called in stage SOLVING without a valid solution (current stage: 10)
Error encountered when optimising <pyscipopt.scip.Model object at 0x7f6f735f8e80>: getSolObjVal cannot only be called in stage SOLVING without a valid solution (current stage: 10)
[...]

PR #1628, opened 2022-02-01
WIP: Build failures: Python 3.6, and SCIP

phschiele:
## Description
This PR tries to fix recent test failures outlined in the issue below 
(second PR as the first branch was not based on the current master).
Issue link (if applicable): #1626 

Python 3.6:
The `dataclasses` module seems to be no longer available. This feature was introduced in Python 3.7, but a backport to 3.6 is available.

pyscipopt:
A new version was released yesterday: https://anaconda.org/conda-forge/pyscipopt/files
Limiting the version to get the tests passing again until the interface has been fixed. 

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

Issue #1709, opened 2022-03-15
Solving a fractional linear boolean problem


keithbriggs:
This issue contains two potential bugs which I think need investigation:

1. Does the grammatically obscure error message `Error encountered when optimising <pyscipopt.scip.Model object at 0x7f772db8dbe0>: getSolObjVal cannot only be called in stage SOLVING without a valid solution (current stage: 10)` coming from SCIP mean that the cvxpy SCIP interface is faulty?
2.  Are the non-boolean results from CBC due to an interface bug, or a bug in CBC itself?

_________________________________________________________________

I'm solving fractional linear boolean quasiconvex problems with a simple bisection method (it's not DQP, so I can't use the dqp=True option).    I would like to report my experiences with the following code (output below):
 - GLPK_MI works perfectly
 - SCIP works in a sense, but is pretty much unusable because of the repeated error messages
 - CBC returns solutions which violate the boolean constraints - looks like a CBC bug to me
 - ECOS_BB crashes badly - I think this is a known problem.

So the overall score is only about 1/4.

```

from sys import exit
from time import time
from itertools import product
import numpy as np
import cvxpy as cp

class Fractional_QUBO:
  '''
   max_x (x.T@A@x)/(x.T@B@x) for A,B both symmetric positive definite
  '''
  def __init__(s,A,B,pp_con=False):
    s.n=A.shape[0]
    assert s.n==A.shape[1]
    assert s.n==B.shape[0]
    assert s.n==B.shape[1]
    assert np.all(A.T==A) # insist that A and B are symmetric
    assert np.all(B.T==B)
    s.A=A
    s.B=B
    s.x=cp.Variable(s.n,name=f'x',boolean=True)
    s.nvars=s.n
    s.xij={}
    s.con=[]
    num=[]
    den=[]
    for i in range(s.n):
      # diagonal terms...
      if A[i,i]!=0.0: num.append(A[i,i]*s.x[i])
      if B[i,i]!=0.0: den.append(B[i,i]*s.x[i])
      for j in range(s.n):
        if i==j: continue
        # off-diagonal terms...
        if A[i,j]!=0.0:
          if (i,j) not in s.xij:
            name=f'x[{s.nvars}]'
            s.xij[i,j]=cp.Variable(name=name,boolean=True)
            s.nvars+=1
          num.append(1*A[i,j]*s.xij[i,j])
          s.con.append(s.xij[i,j]-s.x[i]<=0)
[...]

PR #1739, opened 2022-04-03
Bugfixes for the SCIP interface

rileyjmurray:
This PR makes three changes relating to SCIP:

* it narrows the scope of the "try" step in a try-except block within ``SCIP._solve``.
* it checks for the existence of solutions before trying to access the model objective value (fixing a problem reported in #1702),
* it specifies a version limit that pyscipopt < 4.0.0.

That third point is a follow-up on PR #1628. It turns out that tests fail when pyscipopt 4.x is installed (note: pyscipopt 3 works exclusively with SCIP 7, and pyscipopt 4 works exclusively with SCIP 8). The specific test I saw fail used the problem defined by ``socp_1(axis=0)`` in ``solver_test_helpers.py``. SCIP incorrectly claimed that the problem was infeasible.


keithbriggs:
Maybe #1709 should be linked too?

rileyjmurray:
Propogated bugfix to 1.1.20.

rileyjmurray:
Propagated to 1.2.1.

Issue #1590, opened 2021-12-23
Integer variables in GP


adishavit:
Is it possible to solve a mixed integer geometric program?
Currently, one can set just one attribute: either `integer=True` of `pos=True`. The latter is mandatory for GP problems.
Is there a way of somehow using integer variables in GPs despite this (other than using the as Real and rounding)?

phschiele:
@adishavit This is currently not supported. See #1446 for details.
You can try setting only `integer=True` and add a constraint of the form `x>=0`.
As mentioned in#1446, you could add a `nonneg` wrap if you need the sign for DCP reasons.

SteveDiamond:
@phschiele this is the first situation I've seen where the limitation of one attribute per variable is a real impediment. I'm starting to think that maybe multiple attributes should be allowed. Something to discuss in the discord.

@adishavit this is a very interesting use case! If you have too much trouble making a ``pos`` wrap feel free to DM me.

rileyjmurray:
I remember a discussion on the GPKit repo about this kind of functionality. I'm adapting my comment https://github.com/convexengineering/gpkit/issues/414#issuecomment-582564038 to summarize:

### Constraining a variable to take on specific values in mixed-integer convex programming

Here's a function that takes a scalar variable ``y`` and returns constraints that ensure ``y.value`` is in ``vec``:
```
def varval_in_vec(y, vec):
    vec = np.sort(vec)
    d = np.diff(vec)
    z = Variable(shape=(d.size,), boolean=True)
    cons = [z[i+1] <= z[i] for i in range(d.size)]
    cons.append(y == vec[0] + d @ z)
    return cons
```
The initial constraints in ``cons`` together with 0/1 constraints on ``z`` ensure that ``z`` is always a vector of the form ``[1,...,1,0,...,0]`` -- i.e. consecutive ones, then consecutive zeros. The constraint ``y == vec[0] + d @ z`` ensures that ``y`` takes a value in ``vec``.

### Using ``varval_in_vec`` for mixed-integer geometric programming

In our context, we'll have a vector of positive integers ``a``, and we'll want to constrain a posynomial variable ``x`` to take on a value ``a[i]`` for some ``i``.  From the perspective of the mixed-integer convex programming formulation, we need to enforce constraints on ``y`` where ``y = log(x)``. Of course, ``x`` takes on a desired value as long as ``y`` takes on a value in ``vec = np.log(a)``. 

[...]

PR #1740, opened 2022-04-05
Finite sets

aryamanJgl:
## Description
These changes add support for a `finiteSet` constraint class, that allows for restricting a variable to a given set of arbitrary (finite) lengths, composed of any real quantities. These changes are motivated by the [issue](https://github.com/cvxpy/cvxpy/issues/1590) --- in this PR, only support for DCP problems is offered, I would be open to working on extending the same to DGP problems as well.

A short rundown of the files changed/added:
1. The `cvxpy/constraints/MI_FS.py` file houses the structure of the new constraint class itself.
2. `cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py` consists of the `Canonicalization` class for the `finiteSet` constraint, in addition to the concerned canonicalization method (`exprval_in_vec`, wrapped around `finite_set_canon`)
3. The change to `cvxpy/reductions/solvers/solver_chain.py` is a simple patch-up to ensure that the `Valinvec2mixedint()` reduction is added to the constructed solver chain.
4. The unit tests for the new class reside in `cvxpy/tests/test_valinvec2mixedint.py`

Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/1590

### DIFF
--- a/cvxpy/constraints/finite_set.py
+++ b/cvxpy/constraints/finite_set.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import numpy as np
+
+from cvxpy.constraints.constraint import Constraint
+from cvxpy.expressions import cvxtypes
+from cvxpy.utilities import scopes
+
+
+class FiniteSet(Constraint):
+    def __init__(self, expre, vec, ineq_form: bool=True, constr_id=None) -> None:
+        Expression = cvxtypes.expression()
+        vec = Expression.cast_to_const(vec)
+        self.expre = expre
+        self.vec = vec
+        self.ineq_form = ineq_form
+        super(FiniteSet, self).__init__([expre, vec], constr_id)
+
+    def name(self) -> str:
+        return "%s FS 0" % self.args[0]
+
+    def is_dcp(self, dpp: bool = False) -> bool:
+        """A FiniteSet constraint imposed by exprval_in_vec makes the MICP problem DCP"""
+        if dpp:
+            with scopes.dpp_scope():
+                return self.args[0].is_affine()
+        return self.args[0].is_affine()
+
+    def is_dgp(self, dpp: bool = False) -> bool:
+        return False
+
+    def is_dqcp(self) -> bool:
+        return self.is_dcp()
+
+    @property
+    def size(self):
+        return self.expre.size
+
+    @property
+    def ineq_form(self) -> bool:
+        return self.ineq_form
+
+    @property
+    def shape(self):
+        return self.expre.shape
+
+    @property
+    def residual(self):
+        """
+        The residual of the constraint.
+        Returns
+        -------
+        float
+        """
+        val = self.expre.value
+        res = np.min(np.abs(val - self.vec.value))
+        return res
--- a/cvxpy/constraints/finite_set.py
+++ b/cvxpy/constraints/finite_set.py
@@ -22,14 +22,15 @@ from cvxpy.utilities import scopes
 
 
 class FiniteSet(Constraint):
-    """A class for constraining given expressions to a set of finite size composed of real numbers.
+    """
+    Constrain each entry of an Expression to take a value in a given set of real numbers.
 
     Parameters
     ----------
     expre : Expression
-        The given expression to be constrained. If ``expre`` has multiple elements, then
-        the constraint is applied separtely to each element. I.e., after solving a problem
-        with this constraint, we should have:
+        The given expression to be constrained. This Expression must be affine.
+        If ``expre`` has multiple elements, then the constraint is applied separately to
+        each element. I.e., after solving a problem with this constraint, we should have:
 
         .. code-block:: python
 
@@ -37,19 +38,22 @@ class FiniteSet(Constraint):
                 print(e.value in vec) # => True
 
     vec : Union[Expression, np.ndarray, set]
-        The finite set of values to which the given (affine) expression is to be constrained.
-        
-    ineq_form : bool
-        Controls how this contraint is canonicalized into mixed integer linear constraints.
-        
-        If True, then we use a formulation with ``vec.size - 1`` inequality constraints, one 
-        equality constraint, and ``vec.size - 1`` binary variables for each element of ``expre``.
-        
-        If False, then we use a formuation with ``vec.size`` binary variables and two equality
-        constraints for each element of ``expre``.
+        The finite collection of values to which each entry of ``expre``
+        is to be constrained.
 
-        Defaults to False. The case ``ineq_form=True`` is provided in the hopes that it may 
-        speed up some mixed-integer solvers that use simple branch and bound methods.
+    ineq_form : bool
+        Controls how this constraint is canonicalized into mixed integer linear
+        constraints.
+
+        If True, then we use a formulation with ``vec.size - 1`` inequality constraints,
+        one equality constraint, and ``vec.size - 1`` binary variables for each element
+        of ``expre``.
+
+        If False, then we use a formulation with ``vec.size`` binary variables and two
+        equality constraints for each element of ``expre``.
+
+        Defaults to False. The case ``ineq_form=True`` may speed up some mixed-integer
+        solvers that use simple branch and bound methods.
     """
 
     def __init__(self, expre, vec, ineq_form: bool = False, constr_id=None) -> None:
@@ -57,6 +61,13 @@ class FiniteSet(Constraint):
         if isinstance(vec, set):
             vec = list(vec)
         vec = Expression.cast_to_const(vec).flatten()
+        if not expre.is_affine():
+            msg = """
+            Provided Expression must be affine, but had curvature %s.
+            """ % expre.curvature
+            raise ValueError(msg)
+        # Note: we use the term "expre" rather than "expr" since
+        # "expr" is already a property used by all Constraint classes.
         self.expre = expre
         self.vec = vec
         self._ineq_form = ineq_form
@@ -70,7 +81,7 @@ class FiniteSet(Constraint):
 
     def is_dcp(self, dpp: bool = False) -> bool:
         """
-        A ``FiniteSet`` constraint is DCP if the constrained expression is affine
+        A FiniteSet constraint is DCP if the constrained expression is affine.
         """
         if dpp:
             with scopes.dpp_scope():
@@ -90,8 +101,8 @@ class FiniteSet(Constraint):
     @property
     def ineq_form(self) -> bool:
         """
-        Choose between two constraining methodologies, use ``ineq_form=False`` while working with
-        ``Parameter`` types
+        Choose between two constraining methodologies, use ``ineq_form=False`` while
+        working with ``Parameter`` types.
         """
         return self._ineq_form
 
@@ -108,8 +119,8 @@ class FiniteSet(Constraint):
         -------
         float
         """
-        expre_val = np.array(self.expre.value).flatten()
+        expr_val = np.array(self.expre.value).flatten()
         vec_val = self.vec.value
-        resids = [np.min(np.abs(val - vec_val)) for val in expre_val]
+        resids = [np.min(np.abs(val - vec_val)) for val in expr_val]
         res = max(resids)
         return res
--- a/cvxpy/constraints/finite_set.py
+++ b/cvxpy/constraints/finite_set.py
@@ -22,8 +22,23 @@ from cvxpy.utilities import scopes
 
 
 class FiniteSet(Constraint):
+    """A class for constraining given expressions to a set of finite size composed of real numbers.
 
-    def __init__(self, expre, vec, ineq_form: bool = True, constr_id=None) -> None:
+    Parameters
+    ----------
+    expre : Expression
+        The given expression to be constrained. Note that, ``expre`` can have multiple features, and the
+        constraint is applied element-wise to each feature of the ``Expression`` i.e.:
+
+        .. code-block:: python
+
+            for i in range(expre.size):
+                print(expre[i] in vec) # => True
+
+    vec : NumPy.ndarray/set
+        The finite set of values to which the given (affine) expression is to be constrained.
+    """
+    def __init__(self, expre, vec, ineq_form: bool = False, constr_id=None) -> None:
         Expression = cvxtypes.expression()
         if isinstance(vec, set):
             vec = list(vec)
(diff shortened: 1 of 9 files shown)


SteveDiamond:
This is really quite nice! The only part left is documentation for the website.

aryamanJgl:
> This is really quite nice! The only part left is documentation for the website.

Thank you! I'll get the documentation in order, make the smaller changes and push a commit with all of it as soon as I can.

keithbriggs:
Doesn't it have to be FiniteSet to conform        to the style of capital letters for classes?

aryamanJgl:
> Doesn't it have to be FiniteSet to conform to the style of capital letters for classes?

Yes! I will make that change in the next commit.
Thank you!

phschiele (commented):
Hi @aryamanJgl, thanks for this nice PR!

A few comments/questions from my side:
- I would typically create such finite set constraints as:
```py
z = cp.Variable(len(finite_set), boolean=True)
constraints = [
    cp.sum(cp.multiply(finite_set,z)) == x,
    cp.sum(z) == 1
]
```
This replaces the `n-2` inequality + `1` equality constraints with `2` equality constraints, where `n` is the size of the set. Additionally, (although not currently implemented by CVXPY), some solvers add special support for such [SOS1](https://en.wikipedia.org/wiki/Special_ordered_set) constraints. Did you consider such an implementation as well?
- You can vectorized the inequality constraints as `z[1:] <= z[:-1]`
- Perhaps the changes to the solving chain should go in `_reductions_for_problem_class()` 
- The tests are failing due to code formatting. Check the [docs](https://www.cvxpy.org/contributing/index.html#code-style) for more info. Don't worry if this does not work, though, as it can also be easily fixed at the end.

cvxpy/tests/test_valinvec2mixedint.py
```diff
@@ -0,0 +1,221 @@
...
+    @staticmethod
+    def make_test_1():
+        """
+        vec contains a contiguous range of integers
+        """
+        x = cp.Variable(shape=(4,))
+        expect_x = np.array([0., 7., 3., 0.])
+        vec = np.arange(10)
+        objective = cp.Maximize(x[0] + x[1] + 2 * x[2] - 2 * x[3])
+        constr1 = finiteSet(x[0], vec)
```
SteveDiamond, line 34: I think it would be better to define FiniteSet in a vectorized way, so that each element of x can take on a different value from the set. Then these four lines would be a single set.

aryamanJgl:
@phschiele Thank you for the detailed reply!

> I would typically create such finite set constraints as:
z = cp.Variable(len(finite_set), boolean=True)
constraints = [
    cp.sum(cp.multiply(finite_set,z)) == x,
    cp.sum(z) == 1
]

I just tried adapting this, and it passes all tests! Guess I'll go ahead and add a flag to the implementation of the `Valinvec2mixedint` class for choosing between either of these two methods for now as per @rileyjmurray's suggestion

> You can vectorized the inequality constraints as z[1:] <= z[:-1]

Have made this change to `exprval_in_vec`'s implementation, and checked that it passes all tests.

> Perhaps the changes to the solving chain should go in _reductions_for_problem_class()

Had a conversation about this with @rileyjmurray! I'll move it to `_reductions_for_problem_class()`

> some solvers add special support for such [SOS1](https://en.wikipedia.org/wiki/Special_ordered_set) constraints. Did you consider such an implementation as well?

I haven't! Could you please tell me a bit more about what I ought to do for implementing the same? From what I understand this would be the case when all elements of the constraining `FiniteSet` set are either `0/1`, which as it stands now, would just be treated equivalent to the set `[0,1]` which is definitely not what's intended in the case of _SOS1_ problems if my understanding is correct.

cvxpy/constraints/MI_FS.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
rileyjmurray, line 1: Rename this file to ``finite_set.py``. That's more consistent with the naming conventions of other files in this folder.

cvxpy/constraints/MI_FS.py
```diff
@@ -0,0 +1,67 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from cvxpy.constraints.constraint import Constraint
+from cvxpy.expressions import cvxtypes
+from cvxpy.utilities import scopes
+import numpy as np
+
+
+class finiteSet(Constraint):
```
rileyjmurray, line 23: You've probably made this change locally but haven't pushed: change to ``FiniteSet``.

cvxpy/constraints/MI_FS.py
```diff
@@ -0,0 +1,67 @@
...
+    def shape(self):
+        return self.expre.shape
+
+    @property
+    def residual(self):
+        """The residual of the constraint.
+
+        Returns
+        -------
+        NumPy.ndarray
```
rileyjmurray, line 63: Is this return type correct? I think it returns a float, as written.

cvxpy/tests/test_valinvec2mixedint.py
rileyjmurray: @aryamanJgl I think this suggestion is reasonable. It will slightly complicate the implementation in a few places, but it makes things much nicer for the user. Make sure to write tests that verify intended behavior in this case.

cvxpy/tests/test_valinvec2mixedint.py
```diff
@@ -0,0 +1,221 @@
...
+        sth.solve(solver='GLPK_MI')
+        sth.verify_objective(places=3)
+        sth.verify_primal_values(places=3)
+        pass
+
+    @staticmethod
+    def make_test_4():
+        """
+        Case when vec houses duplicates
+        """
```
rileyjmurray, line 125: Style: one-line docstrings can be formated as
```
def func_name():
    """description of function""""
```

rileyjmurray (commented):
Looks good overall! I think you have local changes which you haven't pushed to your fork yet, or I would be seeing them here. There's also a point raised by @SteveDiamond that I agree with.

aryamanJgl:
@SteveDiamond @rileyjmurray @phschiele Hey! So I made a lot of the suggested changes in the recent commit, namely:
1.  Added support for 'vectorizing' the `FiniteSet` constraint. 
2. Incorporated Philipp's constraining solution as a possible `Canonicalization` path by allowing the user to pass a `flag` while creating the `FiniteSet` constraint (right now it defaults to Riley's solution)
3. Made the construction of Riley's `z` constraints 'vectorized'.
4. Shifted the addition of the `Valinvec2mixedint` reduction to the `_reductions_for_problem_class()` method. 
5. Added tests for all of the above changes wherever relevant (tests 7 and 8 are the new ones)
6. Changed `finiteSet` -> `FiniteSet`
7. Renamed MI_FS.py -> finite_set.py
8. I think style issues now shouldn't persist since I forgot to ran `isort` on the root of the project folder last time, they should hopefully be sorted out now.

I personally think the solutions that I wrote for (1,2) are not very neat and there is definitely scope for improving them.

aryamanJgl:
I just realized that the way I handled the "vectorized" case for Philipp's method had some major inconsistencies -- corrected that in the latest (5th) commit + added a test case for vectorized `FiniteSet` constraint + Philipp's method.

cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
```diff
@@ -0,0 +1,91 @@
...
+            cons += exprval_in_vec1(con.expre[i], vec)
+    else:
+        expr = args[0]
+        cons = exprval_in_vec1(expr, vec)
+    main_con = cons[-1]
+    aux_cons = cons[:-1]
+    return main_con, aux_cons
+
+
+def finite_set_canon2(con, args):
```
phschiele, line 63: Given the similarity of these functions, you could write one factory function that takes either `exprval_in_vec1` or `exprval_in_vec2` as a `Callable` input and returns the corresponding canonicalization function.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -117,6 +119,13 @@ def _reductions_for_problem_class(problem, candidates, gp: bool = False) -> List
                               "(%s)." % candidates)
         else:
             reductions += [Dcp2Cone(), CvxAttr2Constr()]
+
+    constr_types = set()
+    for c in problem.constraints:
+        constr_types.add(type(c))
```
phschiele, line 125:
```suggestion
    constr_types = {type(c) for c in problem.constraints}
```

cvxpy/constraints/finite_set.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+import numpy as np
+
+from cvxpy.constraints.constraint import Constraint
+from cvxpy.expressions import cvxtypes
+from cvxpy.utilities import scopes
+
+
+class FiniteSet(Constraint):
+    def __init__(self, expre, vec, flag: bool = False, constr_id=None) -> None:
```
rileyjmurray, line 25: The ``flag`` argument should have a more informative name. I suggest ``ineq_form``, where ``ineq_form=True``corresponds to the formulation I proposed.

cvxpy/constraints/finite_set.py
```diff
@@ -0,0 +1,68 @@
...
+        Expression = cvxtypes.expression()
+        vec = Expression.cast_to_const(vec)
+        flag = Expression.cast_to_const(flag)
```
rileyjmurray, line 28: Is casting ``flag`` to an Expression really necessary? I think you might be able to do
```
self.flag = flag
...
super(FiniteSet, self).__init__([expre, vec])
```
--- note that ``flag`` isn't used in the ``super`` call. 

Just give that a shot and see if it works. If it doesn't then you can leave the code as-is.

cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
```diff
@@ -0,0 +1,91 @@
...
+        cp.sum(cp.multiply(vec, z)) == expr,
+        cp.sum(z) == 1
+    ]
+    return constraints
+
+
+def finite_set_canon1(con, args):
+    cons = []
+    vec = con.vec.value
+    if con.expre.shape != tuple():
```
rileyjmurray, line 52: You can start by ``expre = expre.flatten()``, then iterate over ``i in range(expre.size)``. (Right now I'm not sure this would work correctly if ``expre`` was a 2d array.

cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
```diff
@@ -0,0 +1,91 @@
...
+    else:
+        expr = args[0]
+        cons = exprval_in_vec2(expr, vec)
+    main_con = cons[0]
+    aux_cons = cons[1:]
+    return main_con, aux_cons
+
+
+def finite_set_canon(con, args):
+    if con.flag.value != 0.0:
```
rileyjmurray, line 78: When you try to change ``flag`` to a bool (per my earlier comment) make sure you change this line as well.

cvxpy/constraints/finite_set.py
aryamanJgl: In my tests, if I did not do that, for some weird reason, `con.ineq_form` ended up containing some bogus values (NOT 0/1 even after the type hinting), and I didn't really understand how to counter the same other than to do this type conversion.

phschiele:
Two more questions came up on my side:

1. Given "Set" is in the name of the constraint, should we support `vec` to be a Python `set` by handling this case explicitly?

Example:
```py
x = cp.Variable()
objective = cp.Maximize(x)
set_vals = {i for i in range(5)}
constraints = [FiniteSet(x, set_vals)]
problem = cp.Problem(objective, constraints)
problem.solve(solver=cp.GLPK_MI)
print(x.value)

>>> fails currently in Expression.cast_to_const(vec)
```

2. What would happen if `vec` is a Parameter? 

Example:
```py
x = cp.Variable()
objective = cp.Maximize(x)
set_vals = cp.Parameter((5,), value=np.arange(5))
constraints = [FiniteSet(x, set_vals)]
problem = cp.Problem(objective, constraints)
problem.solve(solver=cp.GLPK_MI)
print(x.value)

set_vals.value = np.arange(5) + 1
problem = cp.Problem(objective, constraints)
problem.solve(solver=cp.GLPK_MI)
print(x.value)

>>> 4.0
>>> 5.0
```
-> Seems to work out of the box, but perhaps a test should be added.

cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
aryamanJgl: I've been looking into this for a while now but haven't really been able to make progress because of my complete lack of OOP design pattern know-how.

Could you please give me some pointers as to what I could do? Sorry for the trouble.

rileyjmurray: If this is the last remaining issue (and I'm not sure if it is), then I suggest we merged the PR and have @phschiele follow up with a simple PR that implements his suggestion. @aryamanJgl could be selected to review that PR in order to learn about what @phschiele had intended.

SteveDiamond:
Really nice job finishing this PR @aryamanJgl ! I just did a clean install and noticed that the tests require GLPK_MI, which is not a strict dependency. We should alter the tests to skip them if GLPK_MI is not installed.

SanTanBan:
Hi! Please could you elaborate on:-

1) How exactly to use the FiniteSet to make a variable take values from only the predefined set of integer values? Presently I seem to be getting this error:
NameError: name 'FiniteSet' is not defined

2) Does the FiniteSet do something like the following ideas (I have been frantically searching for a way to implement a specific formulation since more than a year and I would link those discussions after implementing it through CVXPY FiniteSet; in case this is what I am looking for...)
   A) https://github.com/coin-or/Cbc/discussions/516#discussioncomment-3007885
   B) https://github.com/coin-or/pulp/discussions/499
   C) https://support.gurobi.com/hc/en-us/community/posts/9160893511185-Requesting-a-new-type-of-variable-introduction
   D) https://stackoverflow.com/questions/73959475/coin-or-cbc-lotsize-variable-creation-and-mps-file-of-formulation
   
3) How is the actual implementation taking place inside the Solvers? Let's say for a Linear Program, every time the LP is solved (say using Simplex), is it checked whether the solution falls within the required integer values; and if not, then new bounds are provided as future constraints (similar to Branch and Bound)?

SteveDiamond:
1. You need to import it as ``from cvxpy.constraints.finite_set import FiniteSet``
2. Finite set imposes the constraint that a variable can only take values from a specified set (e.g., x in {0, 7, 100}).
3. It is implemented via binary variables. It is not sent to the solver as a special constraint, which I believe is what you are looking for.

rileyjmurray:
@SteveDiamond I think it would consider it a bug if users couldn't access ``FiniteSet`` as a top-level import.

@SanTanBan although cvxpy doesn't try to represent these constraints in any fancy way, it is conceivable that we could do as much if one of our solvers supported these constraints natively. So if you know of solvers that *do* support this constraint then you can file an issue asking us to (1) update our interface if we already support the solver or (2) add an interface to the solver in question.

SanTanBan:
According to the Coin-OR community, their CBC solver has the ability to use such types of variables.

https://github.com/coin-or/Cbc/discussions/516#discussioncomment-3019001

I have been trying to utilize this functionality since last year but have faced a lot of problems. I would request you to kindly initiate a project with someone from the Coin-OR community as well. The PuLP community might also be informed. I am waiting to write a paper using this specific functionality in my formulation and such a project would help in comparing solver times as well (different solvers which allow such bucket/lotsize variables without converting them into binary variables).

Further, according to (https://github.com/coin-or/Cbc/discussions/516#discussioncomment-3228702) Gecode also seems to have such a functionality.

PR #1748, opened 2022-04-11
Follow-up FiniteSet

phschiele:
## Description
As discussed, this PR is a follow-up to #1740, containing the following changes:
- ~Using @rileyjmurray 's suggested implementation as the default. I believe this was accidentally changed in 2eab70c974300b8901943994fc50c1c57a389bfa Renamed the functions to improve readability~ This was intentional, see comment below
- Small refactoring to reduce code duplication as suggested 
- Adding a few more tests
- Parametrizing the tests to use both canonicalization functions (+ one test to check that using the default works)
- Skipping the tests if GPLK-MI is not installed as suggested by @SteveDiamond [here](https://github.com/cvxpy/cvxpy/pull/1740#issuecomment-1094513642)
- Special handling for the case where `len(vec) == 2`, as `z[1:] <= z[:-1]` failed in that scenario
- Added a suggestion for the abstractmethod `accepts()`

cc: @aryamanJgl 

## Type of change
- [ ] Bug fix
- [x] Other (Refactoring)

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
+++ b/cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
@@ -6,7 +6,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+    https://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -14,32 +14,33 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from typing import Callable, List, Tuple
+
 import numpy as np
 
 import cvxpy as cp
 from cvxpy.constraints import FiniteSet
+from cvxpy.constraints.constraint import Constraint
 from cvxpy.expressions.variable import Variable
 from cvxpy.reductions.canonicalization import Canonicalization
 
 
-def exprval_in_vec1(expr, vec):
-    assert expr.is_affine()
-    assert expr.size == 1
-    if vec.size == 1:
-        # handling for when vec only has a single element
-        cons = [expr == vec[0]]
-        return cons
+def exprval_in_vec_ineq(expr, vec):
+
     vec = np.sort(vec)
     d = np.diff(vec)
     z = Variable(shape=(d.size,), boolean=True)
     cons = [
-        z[1:] <= z[:-1],
-        expr == vec[0] + d @ z
+        expr == vec[0] + d @ z,
     ]
+    if d.size > 1:
+        cons.append(z[1:] <= z[:-1])
     return cons
 
 
-def exprval_in_vec2(expr, vec):
+def exprval_in_vec_eq(expr, vec):
+    # Reference: https://docs.mosek.com/modeling-cookbook/mio.html#fixed-set-of-values
+
     z = Variable(len(vec), boolean=True)
     constraints = [
         cp.sum(cp.multiply(vec, z)) == expr,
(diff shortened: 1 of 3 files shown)


aryamanJgl:
> Using @rileyjmurray 's suggested implementation as the default. I believe this was accidentally changed.. 

@phschiele If I remember correctly I made that change after Riley discovered that your suggested constraining method works with `Parameter` classes as well  (the underlying reason I believe was that Riley's method sorted the `vec` array, and `Parameter` types don't have a defined notion of ordinality).

phschiele:
> Could you also vectorize FiniteSet so you can pass in a non-scalar argument, and the constraint lets each entry vary independently?

@SteveDiamond I've added a test case for this scenario in https://github.com/cvxpy/cvxpy/pull/1748/commits/36f7ef22964c371b1224a11d957365058ea27ed1

aryamanJgl:
@SteveDiamond I believe in my implementation if a multi-feature variable was passed to the `FiniteSet` constructor, then the same constraint would be applied to all of the features of the variable (i.e. each individual feature of the `Variable` would be constrained to lie in the passed `vec`) which I felt were the intuitive semantics for constructing such a `FiniteSet` constraint.

How would you suggest we could let each feature vary independently when we construct a `FiniteSet` constraint from a non-scalar `Variable` like: `cons= FiniteSet(x, vec) # x is a (n,) variable`. Do you mean we should add support for allowing the user to constrain some arbitrary subset of features of `x` (by perhaps passing a list composed of the respective features they want to constrain that they might've constructed using a list comprehension _etc._)?

SteveDiamond:
> @SteveDiamond I believe in my implementation if a multi-feature variable was passed to the `FiniteSet` constructor, then the same constraint would be applied to all of the features of the variable (i.e. each individual feature of the `Variable` would be constrained to lie in the passed `vec`) which I felt were the intuitive semantics for constructing such a `FiniteSet` constraint.

That is what I mean as well, but the current implementation constrains all features to be exactly the same value. So if I write ``FiniteSet(x, [0,1])`` every entry of ``x`` have the same value. Whereas it should allow the entries to have different values (but still 0 or 1).

aryamanJgl:
@SteveDiamond, on @rileyjmurray's suggestion I just cross-checked with trying out the following problem `min{ x - y : FiniteSet([x, y], {1, 2, 3, 10.5})}` which has the solution `x=1, y=10.5`, which I wrote as:
```
x= Variable(shape= (2,))
objective= Minimize(x[1]-x[0])
constr= FiniteSet(x, {1,2,3,10.5})
prob= Problem(objective, [constr])
```
And it solved out to give the result as expected! So I don't think it is constraining each feature of the passed `Variable` to be the same value.

cvxpy/reductions/discrete2mixedint/valinvec2mixedint.py
```diff
@@ -48,36 +49,33 @@ def exprval_in_vec2(expr, vec):
...
-def finite_set_canon2(con, args):
+def finite_set_canon(con, _args) -> Tuple[Constraint, List]:
     cons = []
     vec = con.vec.value
+    if vec.size == 1:
+        # handling for when vec only has a single element
+        return con.expre == vec[0], []
+
     expre = con.expre.flatten()
     for i in range(expre.size):
```
SteveDiamond, line 67: Can you vectorize this code? I hadn't noticed that you were iterating over each entry of the expression, I see now that it's not constraining all entries to be the same. But for a large expression this will be very inefficient compared to a vectorized approach.

phschiele: Done in [7e1c7c3](https://github.com/cvxpy/cvxpy/pull/1748/commits/7e1c7c3540f6d2c58b70f3694fb68c59dc93102f)

aryamanJgl: Ahh! I see what you mean!
@rileyjmurray gave me an idea about what can be done here, I will try working on it!

aryamanJgl: Oh, I see @phschiele already took care of it in [7e1c7c3](https://github.com/cvxpy/cvxpy/pull/1748/commits/7e1c7c3540f6d2c58b70f3694fb68c59dc93102f)!

cvxpy/tests/test_valinvec2mixedint.py
```diff
@@ -237,71 +244,91 @@ def make_test_7():
...
+        objective = cp.Maximize(cp.sum(x))
+        set_vals = {0, 1, 2}
+        constraints = [FiniteSet(x, set_vals, ineq_form=ineq_form),
+                       x <= np.arange(4).reshape(shape)]
+        problem = cp.Problem(objective, constraints)
+        problem.solve(solver=cp.GLPK_MI)
+        assert np.allclose(x.value, np.array([[0, 1], [2, 2]]))
+
+
+@solver_installed
```
SteveDiamond, line 326: Why is this test separated out from the others, i.e., not in the class?

SteveDiamond (approved):
LGTM

cvxpy/tests/test_valinvec2mixedint.py
phschiele: This is because the class is parametrized to test both canonicalization methods explicitly. This test just makes sure that it also works when not specifying this argument.

Issue #1631, opened 2022-02-01
Infrastructure to help users to submit benchmarks


rileyjmurray:
Speeding up cvxcore is a big priority for us. In order for us to know what works "in practice," it's important for us to have access to a wide range of benchmarks. It's not practical for the cvxpy project maintainers (or a summer intern) to generate those benchmarks themselves. I think we should expand on the ``SolverTestHelper`` framework we use in ``test_conic_solvers.py``. Here are three things we could do:

1. At the very least we can advertise that framework and solicit contributions.
2. We might make another repo within the cvxpy GitHub organization:  "cvxpy/benchmarks", so we can handle contributions of test cases which involve large amounts of metadata.
3. We could try to take this even farther and create some automated tools (e.g., a web app on cvxpy.org) where users can submit problem data and we automatically generate code and a PR so the test case meets formatting requirements for our benchmarking infrastructure.

Ping @akshayka, @SteveDiamond, @bstellato, @phschiele, @adishavit.

SteveDiamond:
I 100% agree with 1. and 2. 3. could be great to given additional manpower.

akshayka:
3 is super cool!

phschiele:
@Michael-git96 and I have collected a handful of (still mostly synthetic) benchmarks that were posted in various recent issues. This was done as part of the current initiative of writing a cvxcore prototype in pure Python. 

I think it would be great to formalise a framework for such benchmarks and make them easy to extend.

adishavit:
This sounds awesome.
Most of my models have very very complex construction code - not something I can submit.
I am wondering if, at the point where you (finally) create your `cp.Problem()`, there would be some way to write out (and anonymize) the objective and the constraints in a format that can be submitted as a benchmark.  

For the community this would make submissions much easier to make (just one method call) and also safer (no sharing of code). For the CVXPY devs, it would mean no need to run custom example scripts per benchmark.
In addition, it would also be compatible with any submitted benchmark code, since that would eventually create the `cp.Problem()`.

Transurgeon:
I'll close this issue since we do have a https://github.com/cvxpy/benchmarks now and also have already added quite a few test problems from various places, thanks to @phschiele.

PR #1749, opened 2022-04-11
WIP: Benchmark setup

phschiele:
## Description
While working on CVXCORE, it became apparent that a benchmarking setup was required to assess the performance more easily. This PR adds a basic framework, which can be extended going forward (both in terms of features, and benchmarks).

Features (to be discussed which ones are relevant):
- Timing the current branch only
- Interactively compare two branches
- Run benchmarks multiple times to get the best, worst, mean, and standard deviation
- Include information about the local machine (CPU model, available RAM) -> I would remove this, as it adds two additional dependencies

Open points:
- Where should the data for the benchmarks be stored? (@rileyjmurray mentioned it might go into a new repo)
- What to do with the existing `tests/test_benchmarks` -> potentially could be kept small versions as unit tests and larger versions as benchmarks
- CI integration

Possible extension:
- Allow comparison of multiple solvers for compilation + solve time

Related issue: #1631

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ tags
 
 # Things specific to this project
 cvxpy/version.py
+cvxpy/benchmarks/benchmark_report.md
--- a/cvxpy/benchmarks/benchmark.py
+++ b/cvxpy/benchmarks/benchmark.py
@@ -1,0 +1,78 @@
+import time
+import tracemalloc
+from abc import ABC, abstractmethod
+from typing import Callable
+
+import cvxpy as cp
+
+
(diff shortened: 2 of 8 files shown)


SteveDiamond:
Would it make any sense to just start a new benchmarking repo, since we don't want to add any dependencies to cvxpy? You could put the data there too.

SteveDiamond:
Sounds good. I made a repo here: https://github.com/cvxpy/benchmarks

PR #1751, opened 2022-04-14
Benchmark job update

yasirroni:
## Description
This PR add bots to run on pull request to do the benchmarking (designed not in push to reduce work). This approach also means that the benchmarking is done on cloud and independent from the developer who write the commit.

After PR is made, we can see in `changes` to compare the benchmarking result before and after pull request.

This PR is based on previously closed PR https://github.com/cvxpy/cvxpy/pull/1749

Slight modification on `run_bencmark.py` is needed due to error run on my laptop.

This PR is better suited on this repository instead of the separate repository #IMHO. 

To separate user and developer, dependency for test only be used by the GitHub workflow, located at reuirements_benchmark.txt

Also note that the benchmark actions is located at separate file. This is more like a demonstration and later can be migrated into proper build that using proprietary solver (gurobi, cplex, etc).

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -44,4 +44,5 @@ jobs:
       uses: ad-m/github-push-action@master
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: ${{ github.head_ref }}
         force: true
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,47 @@
+name: benchmark
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+  # build
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.8"
+        cache: 'pip' # cache for faster build
+    - name: Update pip
+      run: |
+        python -m pip install --upgrade pip
+    - name: Install package
+      run: |
+        pip install -e .
+    - name: Install dependencies
+      run: |
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+
+  # benchmark:
+    - name: Install benchmark package
+      run: |
+        pip install -r requirements_benchmark.txt
+    - name: Benchmark test
+      run: |
+        python cvxpy/benchmarks/run_benchmarks.py
+
+  # commit:
+    - name: Commit files
+      run: |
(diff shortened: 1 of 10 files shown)


yasirroni:
Due to permission issue, the push by GitHub bot is failing.

But, the commit is success.

Maybe still need some tweaks so that the push is done not on the 'https://github.com/cvxpy/cvxpy.git/' but on the committer repository.

SteveDiamond:
This is an interesting PR but please don't make any more changes without discussing with us. I don't want a bot committing to the repository.

yasirroni:
> This is an interesting PR but please don't make any more changes without discussing with us. I don't want a bot committing to the repository.

Sure. I find a solution to make the bot committing to the source branch. Currently testing in this PR. Is it okay?

SteveDiamond:
> > This is an interesting PR but please don't make any more changes without discussing with us. I don't want a bot committing to the repository.
> 
> @SteveDiamond Sure. I find a solution to make the bot committing to the source branch. Currently testing in this PR. Is it okay?

No, please do not do this. You can develop the PR on a fork.

yasirroni:
> > > This is an interesting PR but please don't make any more changes without discussing with us. I don't want a bot committing to the repository.
> > 
> > 
> > @SteveDiamond Sure. I find a solution to make the bot committing to the source branch. Currently testing in this PR. Is it okay?
> 
> No, please do not do this. You can develop the PR on a fork.

I've been testing on my fork, but everything works there. I will close this for now and test make a PR on other repository.

phschiele:
@yasirroni @SteveDiamond Running the benchmarks on CI is indeed a nice thing to have. How about this course of action:

1. benchmarking setup is developed in the other repo 
2. add action in in main repo that simply installs the benchmark setup as a package 
3. run branch and master 
4. instead of committing, the action can simply comment the comparison table in the PR

This would also have the benefit that both versions are run on the same machine. If the results are committed, you might get a different server on the next run, complicating the comparison.

yasirroni:
> @yasirroni @SteveDiamond Running the benchmarks on CI is indeed a nice thing to have. How about this course of action:
> 
> 1. benchmarking setup is developed in the other repo
> 2. add action in in main repo that simply installs the benchmark setup as a package
> 3. run branch and master
> 4. instead of committing, the action can simply comment the comparison table in the PR
> 
> This would also have the benefit that both versions are run on the same machine. If the results are committed, you might get a different server on the next run, complicating the comparison.

I'm agreed to all the points. Since in my test workflows, it is very complicated to make a bit that push to the original PR. Need to find a workflow example that able to comments on PR.

Issue #1695, opened 2022-03-11
Flatten and vec don't support the "order" argument


SteveDiamond:
**Describe the bug**
``expr.flatten()`` and ``vec(expr)`` should support the ``order`` argument, to choose C versus Fortran ordering.

**To Reproduce**
```
import cvxpy as cp

x = cp.Variable((3, 2))
expr = cp.vec(x, order='C')
expr = x.flatten(order='C')
```

**Expected behavior**
The behavior should be the same as ``reshape(expr, (expr.size), order=...)``.

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

**Version**
 - OS: Mac 10.14
 - CVXPY Version: 1.2

**Additional context**

rileyjmurray:
Actually @SteveDiamond, I'm not sure if I'd classify this as a bug. Is there documentation that says that users can expect ``cp.vec`` or ``expr.flatten`` to accept an order argument? I ask because if there isn't, then the expected behavior you described would technically be a new feature, rather than a bugfix. This is relevant for deciding whether something can be updated in cvxpy 1.1.X and 1.2.X or whether it would need to released in 1.3.X

SteveDiamond:
It's not a bug from that viewpoint.

pat749:
in documentation there is nothing said about ,so it is a new feature and in (expr = cp.vec(x, order='C')) line vec() got an unexpected keyword argument 'order',so it is unevaluated expression and here it is not used in  colloquial expression so its new features or bug.

rileyjmurray:
Resolved by #1754.

PR #1754, opened 2022-04-19
Added feature mentioned in issue #1695

Brown2345:
## Description
vec and flatten both gain the option order as an optional argument, allowing a change of ordering from Fortran to C, if desired.
Issue link: https://github.com/cvxpy/cvxpy/issues/1695

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
This is good! You should add tests as well, maybe in test_atoms.py.

rileyjmurray (changes requested):
@Brown2345 I know this PR is simple and you opened it a while ago, but are you interested in finishing it? We'd need a couple of simple unit tests and to update the function docstrings. Let us know how you'd like to proceed.

rileyjmurray:
ping @Brown2345

SteveDiamond:
@rileyjmurray I added tests and doc-strings.

phschiele (commented):
Thanks for adding the test cases @SteveDiamond.
They cover both orders + the default argument for constants and variables, which is the full scope of the PR.
I agree with the notion in the issue that this is not a bug, as it was not mentioned before.

SteveDiamond:
Thanks, merging in!

Issue #1750, opened 2022-04-14
Consider adding `autoflake` or `autoflake8`?


yasirroni:
**Is your feature request related to a problem? Please describe.**

Nope

**Describe the solution you'd like**

Since we use `flake8` in pre-commit to warn user about style and use `isort` to modify in place,, we can consider using `autoflake` (archived) or `autoflake8` (maintained) to add some auto fix raised by `flake8`. Some, because no tools is available to autofix all problem raised by `flake8` (even `autotpep8` + `autoflake8` can't auto fix everything).

https://github.com/cvxpy/cvxpy/blob/ffc05bb6237fbf70644998635e4e24e792c12a30/.pre-commit-config.yaml#L1-L14

Some fix that will be fixed by `autoflake`/`autoflake8` including: unused import and unused variable.

See  [`autoflake`](https://github.com/PyCQA/autoflake) or [`autoflake8`](https://github.com/fsouza/autoflake8)

**Describe alternatives you've considered**

Nope

**Additional context**

I can make a PR if the maintainer want to implement it.

Of course, adding more pre-commit hook will add some minor time when commit.

PR #1756, opened 2022-04-21
Add autoflake pre-commit; remove unused variable; remove unused pass

yasirroni:
## Description
Add autoflake pre-commit; remove unused variable; remove unused pass

Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/1750

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,17 @@
 repos:
+  # remove unused import and variable using autoflake
+  - repo: https://github.com/PyCQA/autoflake
+    rev: v1.4
+    hooks:
+      - id: autoflake
+        args: [
+          '--in-place',
+          '--remove-all-unused-imports',
+          '--ignore-init-module-imports',
+          '--remove-unused-variables',
+          '--'
+        ]
+
   # import sorting with isort
   - repo: https://github.com/pycqa/isort
     rev: 5.8.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,9 +6,12 @@ repos:
       - id: autoflake
         args: [
           '--in-place',
+          '--recursive',
           '--remove-all-unused-imports',
           '--ignore-init-module-imports',
           '--remove-unused-variables',
+          '--exclude',
+          'test_*',
           '--'
         ]
 
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -446,7 +446,7 @@ def set_linC_data(linC, linPy) -> None:
     """Sets numerical data fields in linC."""
     assert linPy.data is not None
     if isinstance(linPy.data, tuple) and isinstance(linPy.data[0], slice):
-        slice_data = set_slice_data(linC, linPy)
+        set_slice_data(linC, linPy)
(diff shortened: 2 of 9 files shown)


rileyjmurray (changes requested):
Hi @yasirroni. Wow, you made a lot of PR's at once!

I'll let @phschiele comment on if these changes to pre-commit make sense. He's the one who suggested we use pre-commit in the first place.

I would prefer that the ``pass`` statements appear in the unit tests. Those make it easier to place break points to investigate test values that are "weird" even when the test fails outright.

yasirroni:
> I would prefer that the pass statements appear in the unit tests. Those make it easier to place break points to investigate test values that are "weird" even when the test fails outright.

Okay. Shoul i git reset and force, add new commit that revert the removed pass, or you prefer just cherry-pick?

___

Anyway, this PR will add extra ms to run autoflake. But, it is still better than manually edit after violating pep8.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,17 @@
...
+  - repo: https://github.com/PyCQA/autoflake
+    rev: v1.4
+    hooks:
+      - id: autoflake
+        args: [
+          '--in-place',
+          '--remove-all-unused-imports',
+          '--ignore-init-module-imports',
+          '--remove-unused-variables',
+          '--'
```
phschiele, line 12: Nit: Is this line required?

phschiele (commented):
@yasirroni Thanks for this PR! 

In general, I like the idea of autoflake and it does not seem to have negative side effects judging by the changes. One comment would be that, as you even mention in the PR description, even relatively small convenience improvements like this do add a (development) dependency, which can go unmaintained etc. 
-> Overall in favor of adding it

I don't have a strong view on the `pass` statements, but @rileyjmurray's comment regarding debugging makes sense.

.pre-commit-config.yaml
yasirroni: Those are setting to set how autoflake work. Default mode, it will not do those three. `--ignore-init-module-imports` only relevant if we use `--remove-all-unused-imports`.

phschiele: Understood, I was referring only to the trailing `--` in l.12, though, which seems redundant.

phschiele:
@yasirroni It would be great if you could incorporate @rileyjmurray's suggested change so we can get this PR merged.

.pre-commit-config.yaml
yasirroni: No, it was not redundant and explained by pre-commit author in their issue to be used for ending arguments. Need to dig the issue if want to find the specific post.

yasirroni:
> @yasirroni It would be great if you could incorporate @rileyjmurray's suggested change so we can get this PR merged.

So, the only modification requested is to not remove the pass? Okay then. Will work on that soon.

.pre-commit-config.yaml
yasirroni: Note that autoflake did not support pyproject.toml

yasirroni:
Since the main functionality of autoflake is to remove unused pass, the new arguments will skip all check in tests.

SteveDiamond:
LGTM

SteveDiamond (approved):

Issue #858, opened 2019-10-31
Add condition number as quasiconvex atom


sbarratt:
**Is your feature request related to a problem? Please describe.**
Currently, lambda_max / lambda_min is not quasiconvex in CVXPY:
```python
import cxvpy as cp

A = cp.Variable((2, 2))
objective = cp.lambda_max(A) / cp.lambda_min(A)
objective.is_quasiconvex() # False
```

**Describe the solution you'd like**
I'd like to add a (quasiconvex) atom, `cp.condition_number(A)`.

**Describe alternatives you've considered**
Make `cp.lambda_max / cp.lambda_min` work.

rileyjmurray:
Technically, condition number is only quasi-convex on the set of positive definite matrices. Refer to MOSEK's modeling cookbook for an example showing how to do this. (Printed page 57, PDF page 60, https://docs.mosek.com/MOSEKModelingCookbook-letter.pdf)

akshayka:
Thanks for sharing, @rileyjmurray. I hadn't previously read that cookbook; it's excellent!

I loved this bit:

> In other words, we present simple convex building blocks, which can then be
combined into more elaborate convex models. We call this approach *extremely disciplined
modeling*. With the advent of more expressive and sophisticated tools like conic optimization,
we feel that this approach is better suited.

rileyjmurray:
Oh yeah, I love their modeling cookbook. I think it should be required reading for anyone who uses  convex optimization in their research. It's super short, shows you a huge number of tricks, and covers a large range of applications. I also got a laugh out of the "extremely disciplined modeling" bit!

parthb83:
Hey @rileyjmurray! I would like to work on this issue as was suggested by you. Kindly guide me through the process of working on this issue.

rileyjmurray:
Hi Parth. Sorry, my bandwidth is actually pretty limited right now. Maybe @phschiele could help?

phschiele:
@rileyjmurray Sure, we'll figure out some details on Discord. 
Perhaps besides the atom itself, we could leverage this to add (at least high-level) instructions for adding an atom.

SteveDiamond:
completed

PR #1768, opened 2022-05-08
Add condition_number as a quasi-convex atom

parthb83:
## Description
Added a new atom function `condition_number()` as a quasi-convex atom to compute the condition number.
Added the atom file, canonicalization file and written unittests for the same.
`condition_number(A) = lambda_max(A) / lambda_min(A)`

Issue link (if applicable): #858 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


parthb83:
@rileyjmurray @phschiele I am currently facing issues in writing test for `test_constant_atoms.py`

cvxpy/atoms/con_num.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def sign_from_args(self) -> Tuple[bool, bool]:
+        """Returns sign (is positive, is negative) of the expression.
+        """
+        return (False, False)
+
+    def is_atom_convex(self) -> bool:
+        """Is the atom convex?
+        """
+        return True
```
SteveDiamond, line 79: Should be False

cvxpy/reductions/dcp2cone/atom_canonicalizers/con_num_canon.py
```diff
@@ -0,0 +1,42 @@
+"""
```
SteveDiamond, line 1: This reduction is in the wrong place. It's a dqcp2dcp reduction.

SteveDiamond:
Thanks for getting started on this! DQCP atoms are fairly different from DCP atoms like xexp. I would suggest discussing further in the Discord about how exactly to do this.

cvxpy/reductions/dcp2cone/atom_canonicalizers/con_num_canon.py
parthb83: But the dqcp2dcp folder has no subfolder for atom_canonicalization. Should we make a new folder for this or just add the canon file in the dqcp2dcp folder?

SteveDiamond: Hmm you're right, dqcp2dcp pulls from dcp atom_canonicalization. That's very strange. @akshayka is their a reason for not having a separate dqcp2dcp atom_canonicalization folder?

rileyjmurray:
Chiming in: I think it should be ``cond_num`` rather than ``con_num``. The prefix "con" is often used for other purposes in building constrained optimization models :)

keithbriggs:
I would go further and say it should be `condition_number`.

This would fit the general principle that the less frequently used a function is, the longer its name should be.  It makes code more readable.

rileyjmurray:
@keithbriggs makes a good point. This atom will only be used in very specific situations. It's unlikely that it will be used in more than one place in a given optimization model. @akshayka @SteveDiamond @phschiele @bstellato what do you all think?

SteveDiamond:
I vote for ``condition_number``

rileyjmurray:
Hi @parthb83! Just checking in on this. I see that our CI tests are failing due to code style issues. You can prevent that by following the instructions here: https://www.cvxpy.org/contributing/index.html#code-style.

parthb83:
> Hi @parthb83! Just checking in on this. I see that our CI tests are failing due to code style issues. You can prevent that by following the instructions here: https://www.cvxpy.org/contributing/index.html#code-style. It would also be good if you could push changes to resolve Steven Diamond's comments.

Hi @rileyjmurray! I've just pushed the desired changes but I think the main problem is that the `condition_number_canon.py` reduction is in the wrong place and it should be in `dqcp2dcp` reduction. Since that folder doesn't have a subfolder for `atom_canonicalization`, I cannot move the reduction there. I discussed the same with Steven on Discord and he also said that there is little we can do as @akshayka knows more about the `dqcp2dcp` stuff, so I'm still waiting for his reply.

akshayka:
A quasiconvex atom is replaced with a parametrized family of its sub/superlevel sets here:

cvxpy/reductions/dqcp2dcp/sets.py

If I recall correctly the value of the parameter can for some atoms change its set in a way that I was unable to capture using the way we canonicalize DCP atoms, so the the sets essentially function as lazy constraints that are evaluated/constructed when the parameter t is known.

If on the other hand you're just rewriting cond_number to be DQCP compliant, then you shouldn't have to write a custom sub/superlevel set for it. I don't have a place for such lowerings but we could make one.

parthb83:
Interesting! `SCS == 2.1.2` was able to solve the test for `1 <= A[0], A[1] <= 2` but couldn't solve it accurately for `0 <= A[0], A[1] <= 2`, although in my opinion, both the test are almost identical and there shouldn't have been any problem for SCS 2.1.2 to solve the latter one (The answer for both the constraints is the lower bound of A[0] and A[1]). 

Anyways, leaving that aside, I guess the only thing left now is to update the docs. Other than that, can you please review the changes @SteveDiamond, @rileyjmurray?

cvxpy/atoms/condition_number.py
```diff
@@ -0,0 +1,98 @@
...
+        Returns:
+            A list of SciPy CSC sparse matrices or None.
+        """
+        raise NotImplementedError
+
+    def validate_arguments(self) -> None:
+        """Verify that the argument A is square.
+        """
+        if not self.args[0].ndim == 2 or self.args[0].shape[0] != self.args[0].shape[1]:
+            raise ValueError("The argument '%s' to con_num must resolve to a square matrix."
```
SteveDiamond, line 62: con_num -> condition_number

cvxpy/atoms/condition_number.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def shape_from_args(self) -> Tuple[int, ...]:
+        """Returns the (row, col) shape of the expression.
+        """
+        return tuple()
+
+    def sign_from_args(self) -> Tuple[bool, bool]:
+        """Returns sign (is positive, is negative) of the expression.
+        """
+        return (False, False)
```
SteveDiamond, line 73: I believe this is positive so should return (True, False)

cvxpy/reductions/dqcp2dcp/sets.py
```diff
@@ -124,13 +128,30 @@ def gen_lambda_max_sub(expr, t):
...
+    n = A.shape[0]
+    u = Variable(pos=True)
+
+    prom_ut = promote(u * t, (n,))
+    prom_u = promote(u, (n,))
+    tmp_expr1 = A - diag_vec(prom_u)
+    tmp_expr2 = diag_vec(prom_ut) - A
+
+    return [A == A.T,
+            A >> 0,
```
SteveDiamond, line 142: A >> 0 -> PSD(A)

cvxpy/tests/test_dqcp.py
```diff
@@ -494,6 +494,20 @@ def test_gen_lambda_max_matrix_completion(self) -> None:
...
+        constr = [
+            A[0][0] == 2.0,
+            A[1][1] == 3.0,
+            A[0][1] <= 2, A[0][1] >= 1,
+            A[1][0] <= 2, A[1][0] >= 1,
+        ]
+        prob = cp.Problem(cp.Minimize(con_num), constr)
+        self.assertTrue(prob.is_dqcp())
+        # smoke test
+        prob.solve(cp.SCS, qcp=True)
```
SteveDiamond, line 509: can we test something about the output? Ideally we would know the optimal value and check that the variable attains it. Right now you aren't testing the correctness of the canonicalization.

cvxpy/atoms/condition_number.py
rileyjmurray: @parthb83 in general it's good to do some digging into an atom to figure out the appropriate properties. For example, you can learn about condition numbers here: https://en.wikipedia.org/wiki/Condition_number#Matrices.

cvxpy/atoms/condition_number.py
keithbriggs: Also, shouldn't we be using f-strings rather than the old %-formatting?

cvxpy/tests/test_dqcp.py
parthb83: So I tested this for the output and the output isn't accurate enough. For the above constraints, the output was
```
[[2.00004109, 0.97275377]
 [0.97275377, 2.99999668]]
```
which is close to the actual output but not accurate enough. Do you have any idea how can we increase the accuarcy of canonicalization?

SteveDiamond: You can increase the accuracy by setting solver options (https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options). But it would be better to just reduce the accuracy of the assert. Search the codebase for ``assertItemsAlmostEqual`` and use the ``places`` argument to control accuracy.

parthb83: But is the following accuracy acceptable for the canonicalization?

SteveDiamond: I see what you mean, the accuracy is pretty bad. Try solving it with ``prob.solve(solver=cp.SCS, qcp=True, eps=1e-8, max_iters=20_000)``. If that doesn't improve things it's ok. We can still merge the PR and investigate it more on our side.

parthb83: > I see what you mean, the accuracy is pretty bad. Try solving it with `prob.solve(solver=cp.SCS, qcp=True, eps=1e-8, max_iters=20_000)`. If that doesn't improve things it's ok. We can still merge the PR and investigate it more on our side.

This didn't change the answer much so I'm checking the answer in `test_dqcp.py` for just 1 place.

SteveDiamond (approved):
LGTM

rileyjmurray:
I'll review this weekend, and if it looks good I'll merge.

SteveDiamond:
@rileyjmurray where did we land with this PR?

cvxpy/atoms/condition_number.py
```diff
@@ -0,0 +1,99 @@
...
+
+from scipy import linalg as LA
+
+from cvxpy.atoms.atom import Atom
+from cvxpy.constraints.constraint import Constraint
+
+
+class condition_number(Atom):
+    """ Condition Number; :math:`\\lambda_{\\max}(A) / \\lambda_{\\min}(A)`.
+    """
```
rileyjmurray, line 26: State that ``A`` needs to be PSD.

cvxpy/reductions/dqcp2dcp/sets.py
```diff
@@ -124,13 +128,30 @@ def gen_lambda_max_sub(expr, t):
...
+
+    prom_ut = promote(u * t, (n,))
+    prom_u = promote(u, (n,))
+    tmp_expr1 = A - diag_vec(prom_u)
+    tmp_expr2 = diag_vec(prom_ut) - A
+
+    return [A == A.T,
+            PSD(A),
+            PSD(tmp_expr1),
+            PSD(tmp_expr2)]
```
rileyjmurray, line 144: ``A == A.T`` is an inefficient way to enforce symmetry. You only need ``n(n-1)/2`` of these ``n**2`` constraints. We should also check if ``A`` is declared as symmetric before bothering to enforce symmetry here.

cvxpy/tests/test_dqcp.py
rileyjmurray: Let me check with another solver.

rileyjmurray (changes requested):
Sorry for the delay on this. There's a canonicalization inefficiency I'd like addressed. I also have a minor comment on the docstring. Lastly, I'll check out this code locally and run the test case. I'm curious if using a solver other than SCS can lead to better results.

SteveDiamond:
ping. You can use the ``upper_tri`` atom to reduce the number of constrains (i.e., ``upper_tri(A) == upper_tri(A.T)``).

SteveDiamond:
@rileyjmurray all the requested changes have been made now. could you review/approve?

SteveDiamond:
Test failures are unrelated to PR. Some kind of package change with one of the MIP solvers...

rileyjmurray:
Approved!

PR #1769, opened 2022-05-10
Add SDPA Interface

usamamuneeb:
## Description

This is an interface for the SDPA solver.

[Official Website](http://sdpa.sourceforge.net)
[Unofficially Maintained Python 3 Frontend](https://sdpa-python.github.io)

## Type of change
- [ ] Bug fix

- [x] Write unittests.


rileyjmurray:
Thanks for making this PR, @usamamuneeb! I'll review this over the weekend.

rileyjmurray (changes requested):
Hi @usamamuneeb! Thanks again for making this PR.

The main thing that jumps out to me is the need for more documentation. You can look at PR #1682 to see that you need to update ``doc/source/install/index.rst`` and ``doc/source/tutorial/advanced/index.rst``. When you update those files, make sure to provide links to the [sdpa-python GitHub repo](https://github.com/sdpa-python/sdpa-python) and its [proper web documentation](https://sdpa-python.github.io/).

The other thing that stands out to me is the difficulty of installing this SDPA support. This is relevant because we try to have all solver interfaces tested in our continuous integration pipelines. PR #1655 added support for the GLOP solver and updated our CI infrastructure by changing a single line in ``continuous_integration/install_dependencies.sh``. I think Installing SDPA by building from source would take prohibitively long for us to run it on all platforms. But it would be nice for us to test it on Linux for some version of Python.

That said, rather than updating CVXPY's CI in this PR, I'd prefer if you could make a GitHub Actions pipeline *on the sdpa-python repo* that demonstrates how to successfully build sdpa-python with OpenBLAS and run its examples. If you make such a pipeline then I can port your code to CVXPY's CI with appropriate conventions/style.

usamamuneeb:
Hi @rileyjmurray, thanks for reviewing!

I agree with the difficulty of building from source. Fortunately, `pip` and `conda` packages are on the way and I guess that should serve as a better solution for the CI infrastructure.

Once they are up, I will add the relevant lines in `continuous_integration/install_dependencies.sh` as well as the documentation files. Thanks again!

rileyjmurray:
That's awesome news! It'll be great to have another interior-point SDP solver accessible from CVXPY :)

rileyjmurray:
@usamamuneeb any reason why you closed this?

usamamuneeb:
I tried temporarily renaming `master` to `main` as it's more convenient while working. GitHub thought the branch is deleted and closed the PR. Anyways, seems like renaming it back has restored it. Is it possible to reopen this PR?

usamamuneeb:
`sdpa-python` is now `pip` installable for Linux, Windows and macOS for `x86_64`. For other architectures, it's untested but I have improved the [source building instructions](https://sdpa-python.github.io/docs/installation/) in case someone would be willing to test on other architectures.

I have added in all the documentation, and have updated `continuous_integration/install_dependencies.sh` so SDPA can be checked in the CI pipeline.

usamamuneeb:
The failing test seems to be `test_sdpa_lp_5`. While developing the interface, I had to [tweak the parameters](https://github.com/cvxpy/cvxpy/blob/9055db9e40e0c7ee1fac7945bb19e00c0e6caa9b/cvxpy/tests/test_conic_solvers.py#L597) for it to make it pass. I was using the reference BLAS on Linux.

While building wheels, I used OpenBLAS for Windows and Linux and the default BLAS (i.e. provided by Accelerate framework) for macOS.

In the tests, the `AssertionError` on macOS stems from the `check_complementarity` method of LP5, while the `AssertionError` on Windows and Linux stems from `verify_objective` method of LP5. I remember while tuning that I had to keep tuning it until each would pass (LP5 seems to be quite sensitive when using SDPA).

I could retune the parameters for OpenBLAS and that would solve the problem for Windows and Linux, however, I guess to make LP5 pass for all platforms, I would have to use OpenBLAS on macOS as well (I decided to use the default BLAS on macOS so I don't need to bundle BLAS with the macOS wheels). On Windows and Linux, there's no other option as I use MinGW for building on Windows (and if not statically linked, user will have to download the right `libblas` and `liblapack`). On Linux, `auditwheel` on `manylinux` decides what to bundle along with the wheel.

SteveDiamond:
> The failing test seems to be `test_sdpa_lp_5`. While developing the interface, I had to [tweak the parameters](https://github.com/cvxpy/cvxpy/blob/9055db9e40e0c7ee1fac7945bb19e00c0e6caa9b/cvxpy/tests/test_conic_solvers.py#L597) for it to make it pass. I was using the reference BLAS on Linux.
> 
> While building wheels, I used OpenBLAS for Windows and Linux and the default BLAS (i.e. provided by Accelerate framework) for macOS.
> 
> In the tests, the `AssertionError` on macOS stems from the `check_complementarity` method of LP5, while the `AssertionError` on Windows and Linux stems from `verify_objective` method of LP5. I remember while tuning that I had to keep tuning it until each would pass (LP5 seems to be quite sensitive when using SDPA).
> 
> I could retune the parameters for OpenBLAS and that would solve the problem for Windows and Linux, however, I guess to make LP5 pass for all platforms, I would have to use OpenBLAS on macOS as well (I decided to use the default BLAS on macOS so I don't need to bundle BLAS with the macOS wheels). On Windows and Linux, there's no other option as I use MinGW for building on Windows (and if not statically linked, user will have to download the right `libblas` and `liblapack`). On Linux, `auditwheel` on `manylinux` decides what to bundle along with the wheel.

IMO it's fine in this case to just mark the test as failing. @rileyjmurray what do you think?

rileyjmurray:
I think it's fine to skip ``lp_5`` for SDPA. It's deliberately constructed to be a degenerate case and is mostly there to determine if we can set solver parameters properly.

usamamuneeb:
> I think it's fine to skip `lp_5` for SDPA. It's deliberately constructed to be a degenerate case and is mostly there to determine if we can set solver parameters properly.

Should I remove/comment out `test_sdpa_lp_5` in `cvxpy/tests/test_conic_solvers.py` or is that something to be done in the CI infrastructure?

rileyjmurray (changes requested):
Looking good! My one suggestion would be to have the web docs contain a remark that CVXPY could be extended to convert SOCPs to SDPs for purposes of using SDPA, and that we would welcome a contribution to that effect.

rileyjmurray:
> > I think it's fine to skip `lp_5` for SDPA. It's deliberately constructed to be a degenerate case and is mostly there to determine if we can set solver parameters properly.
> 
> Should I remove/comment out `test_sdpa_lp_5` in `cvxpy/tests/test_conic_solvers.py` or is that something to be done in the CI infrastructure?

You can add an annotation like ``@pytest.mark.skip(reason="Known limitation of SDPA for degenerate LPs.")`` before the ``def`` statement for the test.

usamamuneeb:
> Looking good! My one suggestion would be to have the web docs contain a remark that CVXPY could be extended to convert SOCPs to SDPs for purposes of using SDPA, and that we would welcome a contribution to that effect.

I actually put it on the [Contributing](https://sdpa-python.github.io/docs/contributing/#large-scope-projects) page of SDPA for Python website to add SOCP support. It seems like that was something initially planned but not completed (while the SDPA manuals mention SOCP support, in the SDPA code, the relevant functions raise an error `current version does not support SOCP`).

Since conversion of SOCPs to SDPs looks something non-standard (and SDPA specific), it might be better to keep it there. What do you think?

And I just recalled that we can now remove the link to [1419](https://github.com/cvxpy/cvxpy/discussions/1419) from the CVXPY [Contributing](https://www.cvxpy.org/contributing/index.html#wishlist) page. I can do that in the same commit.

rileyjmurray:
> Since conversion of SOCPs to SDPs looks something non-standard (and SDPA specific), it might be better to keep it there. What do you think?

I absolutely agree with you from a performance perspective. Maybe say that (1) adding native support for SOCPs to SDPA may be an interesting project on the numerical solver side and (2) adding functionality for CVXPY to convert SOCPs to SDPs is of potential pedagogical value.

@SteveDiamond what do you think?

SteveDiamond:
> > Since conversion of SOCPs to SDPs looks something non-standard (and SDPA specific), it might be better to keep it there. What do you think?
> 
> I absolutely agree with you from a performance perspective. Maybe say that (1) adding native support for SOCPs to SDPA may be an interesting project on the numerical solver side and (2) adding functionality for CVXPY to convert SOCPs to SDPs is of potential pedagogical value.
> 
> @SteveDiamond what do you think?

I agree with @rileyjmurray, the proper place for the conversion is in SDPA but there would be some value in having a reduction from SOCP to SDP in CVXPY.

usamamuneeb:
It indeed does make sense to have a reduction that can map a problem from a smaller set (e.g. SOCPs) to a superset (SDPs in this case). Not just for SOCP to SDP but perhaps all the other such reductions as well (i.e. LP to QP, QP to SOCP, etc).

For now SDPA may be the only solver that does not specifically support SOCPs (but does support a superset of them), but there can potentially be other such solvers as well.

If a solver can theoretically solve an SOCP (because it can solve a superset), CVXPY should be able to convert SOC constraints to PSD.

SteveDiamond:
Looks like test_verbose still fails. That checks that the ``verbose`` keyword silences all output from the solver.

usamamuneeb:
> Looks like test_verbose still fails. That checks that the `verbose` keyword silences all output from the solver.

Sorry I didn't realize it was also because of this interface (I only looked at `test_conic_solvers.py` during development).

Below is the output that would be printed by the solver with `verbose=False`.

```
Entering DMUMPS driver with JOB, N, NZ =  -2           0              0
Converted to SDPA internal data / Starting SDPA main loop
Converting optimal solution to CLP format
Start: getCLPresult
Making result infomation...
```

The second and subsequent lines come from `sdpa-python`. They can be fixed by printing them only if `verbose=True`. The first line is printed by [MUMPS](http://mumps.enseeiht.fr), a library imported by SDPA. A makeshift solution would be to comment it out in the code and rebuild the wheels. Even though MUMPS is statically linked on all platforms (and it will not be a problem if another MUMPS is present on the system), this would silence it even if `verbose=True`.

This might be a temporary solution until I find a proper solution to it (multi-level output logging is one of the things I have requested on the [SDPA for Python contributing page](https://sdpa-python.github.io/docs/contributing/#small-scope-projects)).

SteveDiamond:
> > Looks like test_verbose still fails. That checks that the `verbose` keyword silences all output from the solver.
> 
> Sorry I didn't realize it was also because of this interface (I only looked at `test_conic_solvers.py` during development).
> 
> Below is the output that would be printed by the solver with `verbose=False`.
> 
> ```
> Entering DMUMPS driver with JOB, N, NZ =  -2           0              0
> Converted to SDPA internal data / Starting SDPA main loop
> Converting optimal solution to CLP format
> Start: getCLPresult
> Making result infomation...
> ```
> 
> The second and subsequent lines come from `sdpa-python`. They can be fixed by printing them only if `verbose=True`. The first line is printed by [MUMPS](http://mumps.enseeiht.fr), a library imported by SDPA. A makeshift solution would be to comment it out in the code and rebuild the wheels. Even though MUMPS is statically linked on all platforms (and it will not be a problem if another MUMPS is present on the system), this would silence it even if `verbose=True`.
> 
> This might be a temporary solution until I find a proper solution to it (multi-level output logging is one of the things I have requested on the [SDPA for Python contributing page](https://sdpa-python.github.io/docs/contributing/#small-scope-projects)).

That sounds a bit dramatic. Can you fix the lines from `sdpa-python`, but leave MUMPS alone? There's a list of solvers that aren't evaluated on line 276 of ``test_problem.py``.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -573,6 +573,40 @@ def test_cvxopt_sdp_2(self) -> None:
...
+
+    def test_sdpa_lp_3(self) -> None:
+        StandardTestLPs.test_lp_3(solver='SDPA')
+
+    def test_sdpa_lp_4(self) -> None:
+        StandardTestLPs.test_lp_4(solver='SDPA')
+
+    @unittest.skip('Known limitation of SDPA for degenerate LPs.')
+    def test_sdpa_lp_5(self) -> None:
+        # this also tests the ability to pass solver options
```
SteveDiamond, line 596: Since this test isn't running anymore, maybe try a solver option in a different test?

SteveDiamond (approved):
LGTM, congrats on passing the tests! I had one minor comment but otherwise good to go.

cvxpy/tests/test_conic_solvers.py
usamamuneeb: Good idea! The `verbose` parameter also translates to a solver option so that's one verification. Another approach would be to just ship OpenBLAS/reference BLAS with macOS wheels as well and just run LP5. It was able to pass using the interface provided parameters on reference BLAS. The best verification for solver options is to pass a test which otherwise won't pass.

Once the unimplemented routines are finished, I plan to create another pull request to add additional options to the documentation, so maybe we could plan for this to go with that future PR?

rileyjmurray (approved):
Looks good!

PR #1789, opened 2022-05-28
Atom for the Von Neumann Entropy

aryamanjeendgar:
In this PR I add support for the Von Neumann Entropy as an `Atom`. The `vn_entr` class itself lies in `cvxpy/atoms`, and its respective Canonicalization method lies in `cvxpy/reductions/dcp2cone/atom_canonicalizers`, finally, the tests reside in `cvxpy/tests/test_vn_entr.py`. I chose to omit the `_grad` method for this PR as per @SteveDiamond's advice.

The implementation of the function itself is wrapped in the `numeric` procedure in `vn_entr.py`--- $S(\rho)=-\sum_x \lambda_x\log\lambda_x$ --- this form can be found in equation (11.40) of Neilsen & Chuang's QIC text. 

The Canonicalization procedure is based off of  *Proposition-4* of [THIS](https://link.springer.com/article/10.1007/s10107-016-0998-2) paper.


keithbriggs:
"it's respective Canonicalization method" should be "its respective Canonicalization method".

rileyjmurray:
The failing tests seem to be for versions of Python where SCS 2.1.2 is installed, rather than SCS 3. I ran the tests on my laptop with SCS 3 and they passed, but the return code was "inaccurate" even after I increased ``max_iters`` to 500,000. So I think we'll need MOSEK for these tests to pass.

cvxpy/tests/test_vn_entr.py
```diff
@@ -0,0 +1,150 @@
...
+class Test_vn_entr:
+    @staticmethod
+    def make_test_1():
+        """(2,2) matrix, 100 largest ev, 1e-3 off-diagonal element"""
+        n = 2
+        N = cp.Variable(shape=(n, n), PSD=True)
+        expect_N = np.array([[0.36787973, 0.00099987],
+                             [0.00099987, 0.36787973]])
+        eps = 1e2
+        cons1 = N >> 0
```
rileyjmurray, line 33: The constraint that ``N >> 0`` is redundant, since you already declared ``PSD = True`` for ``N``. Same comment goes for other tests.

cvxpy/tests/test_vn_entr.py
```diff
@@ -0,0 +1,150 @@
...
+        N = U.T@N@U
+        objective = cp.Maximize(vn_entr(N))
+        obj_pair = (objective, 1.1033501770078251)
+        con_pairs = [
+            (cons1, None),
+            (cons2, None),
+            (cons3, None)
+        ]
+        var_pairs = [
+            (N, expect_N)
```
rileyjmurray, line 109: Here you're reassigning ``N``. That should cause tests to fail since ``expect_N`` is supposed to be with respect to the original variable.

rileyjmurray (commented):
@aryamanJgl here are some things to be aware of. I'm investigating other stuff and hopefully will provide more comments soon.

cvxpy/tests/test_vn_entr.py
```diff
@@ -0,0 +1,150 @@
...
+        expect_N = np.array([[3.70615097e-01, -6.50164940e-15, -2.51416162e-13],
+                             [-6.50164940e-15, 3.66498027e-01, -9.99852741e-03],
+                             [-2.51416162e-13, -9.99852741e-03, 3.66498027e-01]])
+        U = cp.Constant(np.array([[0, -1, 0],
+                                  [1, 0, 0],
+                                  [0, 0, 1]]))
+        eps = 1e3
+        cons1 = N >> 0
+        cons2 = N << eps*np.eye(N.shape[0])
+        cons3 = N[2][0] == 1e-2
```
rileyjmurray, line 99: This is inconsistent with ``expect_N``.

rileyjmurray:
@aryamanJgl after running the tests with MOSEK I'm getting that they all fail. Something is going wrong in ``vn_entr`` canonicalization, since MOSEK is reporting that the problem is unbounded even though the feasible set should be compact.

rileyjmurray:
@aryamanJgl here is a tiny example of a test that's failing:
```
N = cp.Variable(shape=(2, 2), PSD=True)
constraints = [cp.trace(N) <= 1,
               N[0, 1] == -0.1,
               ]
objective = cp.Maximize(cp.vn_entr(N))
prob.solve(verbose=True)  # infeasible
```
The solver (SCS or MOSEK) says the problem is infeasible, but ``N = [[0.5, -0.1],[-0.1, 0.5]]`` is feasible (its eigenvalues are 0.4 and 0.6).

keithbriggs:
Isn't this a another case where a long name is needed? vn_entr is completely cryptic. I would go for von_neumann_entropy.

rileyjmurray:
> Isn't this a another case where a long name is needed? vn_entr is completely cryptic. I would go for von_neumann_entropy.

That's a good point. Although I'd vote for ``von_neumann_entr``, since basic (Shannon) entropy is just ``entr``. In the near future we'll have a bunch of other atoms where we'll want to continue to use existing abbreviations, like ``quantum_rel_entr`` and ``operator_rel_entr``.

@aryamanJgl if you don't object, how about you change it to ``von_neumann_entr``?

cvxpy/atoms/von_neumann_entr.py
```diff
@@ -25,11 +25,11 @@
 from cvxpy.constraints.constraint import Constraint
 
 
-class vn_entr(Atom):
+class von_neumann_entr(Atom):
 
     def __init__(self, N) -> None:
         # TODO: add a check that N is symmetric/Hermitian.
-        super(vn_entr, self).__init__(N)
+        super(von_neumann_entr, self).__init__(N)
```
rileyjmurray, line 32: This function needs a docstring.

rileyjmurray, line 32: This function needs a docstring.

cvxpy/tests/test_von_neumann_entr.py
```diff
@@ -0,0 +1,196 @@
...
+                             [0.04393948, -0.0242426, 0.30757584]])
+        v1 = np.array([[-0.12309149],
+                       [-0.49236596],
+                       [-0.86164044]])
+        v2 = np.array([[0.90453403],
+                       [0.30151134],
+                       [-0.30151134]])
+        lambda1 = 0.3
+        lambda2 = 0.2
+        trMax = 0.9
```
rileyjmurray, line 88: This should be ``trMin``.

cvxpy/tests/test_von_neumann_entr.py
```diff
@@ -0,0 +1,196 @@
...
+                       [5.11915, 4.42217, 2.72919]])
+
+        b = np.array([19.16342, 17.62551])
+
+        N = cp.Variable(shape=(3, 3), PSD=True)
+        expect_N = np.array([[0.82564479, 0., 0.],
+                             [0., 1.27395299, 0.],
+                             [0., 0., 0.64292239]])
+        objective = cp.Minimize(-von_neumann_entr(N))
+        obj_pair = (objective, -0.13372929515039528)
```
rileyjmurray, line 176: You should provide the code used to generate this test data. (At least, the code to generate ``expect_N``.)

rileyjmurray (changes requested):
Mostly looks good! The main thing that stands out is the need for documentation. The docstring for ``von_neumann_entr`` is missing, and the web docs need to be updated to mention this new atom. I also left minor comments on the tests.

rileyjmurray (approved):
Looks good to me!

rileyjmurray:
@SteveDiamond can you review this PR? We need to project maintainers to approve before merging, since this affects the public API.

cvxpy/atoms/von_neumann_entr.py
```diff
@@ -0,0 +1,123 @@
...
+        A PSD matrix
+    """
+
+    def __init__(self, X) -> None:
+        # TODO: add a check that N is symmetric/Hermitian.
+        super(von_neumann_entr, self).__init__(X)
+
+    def numeric(self, values):
+        N = values[0]
+        if hasattr(N, 'value'):
```
SteveDiamond, line 58: Normally we do ``isinstance(N, np.ndarray)``.

cvxpy/reductions/dcp2cone/atom_canonicalizers/von_neumann_entr_canon.py
```diff
@@ -0,0 +1,59 @@
...
+    n = N.shape[0]
+    x = Variable(shape=(n,))
+    t = Variable()
+
+    # START code that applies to all spectral functions #
+
+    constrs = []
+    for r in range(1, n+1):
+        # lambda_sum_largest(N, r) <= sum(x[:r])
+        expr_r = lambda_sum_largest(N, r)
```
SteveDiamond, line 36: This is going to create a massive internal representation, right? Are we sure we this is how we want to do the canonicalization? For instance, doesn't lambda_sum_largest apply ``N >> 0``? That should only happen once. Please ignore this comment if I'm totally off.

SteveDiamond (approved):
Approved subject to my comments. Also shouldn't we add it to the table of atoms?

cvxpy/reductions/dcp2cone/atom_canonicalizers/von_neumann_entr_canon.py
rileyjmurray: @SteveDiamond the representation will indeed be large, but ``lambda_sum_largest`` doesn't impose any redundant constraints across these function calls. One thing @aryamanJgl can do is use ``range(1, n)`` instead of ``range(1, n+1)`` and then simplify ``lambda_sum_largest(N, n) == trace(N)``.

rileyjmurray:
@aryamanJgl please address my response to Steven's comment. Also, there are two things to do for the docs.

1. ``von_neumann_entr`` isn't an elementwise atom. It should be documented [here](https://github.com/cvxpy/cvxpy/blob/master/doc/source/api_reference/cvxpy.atoms.other_atoms.rst), not at ``doc/source/api_reference/cvxpy.atoms.elementwise.rst``.
2. Add ``von_neumann_entr`` to the table of "Scalar Functions" [here](https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/functions/index.rst).

cvxpy/atoms/von_neumann_entr.py
rileyjmurray: Is ``values[0]``guaranteed to be a certain type? This check makes sure Expression objects work and plain numeric numpy ndarrays work.

SteveDiamond: values[0] is supposed to be an ndarray, float, or possibly a scipy sparse matrix. You can leave the attribute check, it was just something that stuck out to me.

PR #1796, opened 2022-06-10
COPT interface

SteveDiamond:
## Description
This is a commit to add an interface to the Cardinal solver.
Docs: https://guide.coap.online/copt/en-doc/index.html#
Install: https://github.com/COPT-Public/COPT-Release

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
The COPT solver interface was contributed by the solver developers.

SteveDiamond:
I think this is done. @rileyjmurray @phschiele

phschiele (approved):
LGTM! A few observations/comments
- I have tried adding COPT to `test_mip_vars.py` (even though SOCP MIP are not supported) -> it does not raise an error but returns a non-integer solution. Perhaps there should be an extra check to prevent this case.
- In the last paragraph of Advanced Features / Mixed-integer programs CPLEX and XPRESS are mentioned to work with 1000 and 5000 variables/constraints, respectively. COPT seems to work with 2000 each.
- Whenever the solver is invoked, it prints some build information.

NB: Tried the solver on some larger problems and it is indeed quite quick! Great to have it available!

SteveDiamond:
Thanks @phschiele! I added a reference to COPT in that section of the docs. I think the other issues are ok for this PR. We will need to update the overall logic for choosing a solver to support solvers that only solve MIP versions of a subset of the convex problems they support. Actually MOSEK already falls into this category, and will just be wrong if you try and solve a MI-SDP.

The build message is unfortunate. I'll ask the devs to remove it. But we can still merge and address that later IMO.

SteveDiamond:
The mosek CI license has expired. Will have a new one soon.

doc/source/tutorial/advanced/index.rst
```diff
@@ -561,21 +571,7 @@ You can change the solver called by CVXPY using the ``solver`` keyword argument.
...
-    optimal value with GLOP: 6.0
-    optimal value with GLPK: 6.0
-    optimal value with GLPK_MI: 6.0
-    optimal value with GUROBI: 6.0
-    optimal value with MOSEK: 6.0
-    optimal value with CBC: 6.0
-    optimal value with CPLEX: 6.0
-    optimal value with NAG: 6.000000003182365
-    optimal value with PDLP: 6.0
-    optimal value with SCIP: 6.0
```
rileyjmurray, line 578: This list is getting really long so I support removing it. But if we do remove it then the code snippets should be changed accordingly.

rileyjmurray (commented):
Mostly looks good! I have one comment on the web documentation. Once we converge on that then it'll be ready to merge.

doc/source/tutorial/advanced/index.rst
SteveDiamond: Thanks! I don't quite understand what you mean. I truncated the output so it shows the beginning and the end. It's not a literal code output anymore. Are you suggesting shortening the code snippet?

rileyjmurray: Ah okay. Truncating so that only the beginning and end are shown is fine. I'd have a line of ellipsis, like
 ``
 ...
 ```
 to indicate that we're artificially hiding most of the output.

SteveDiamond:
> Benchmarks that have stayed the same:
> 
> ```
>    before           after         ratio
>  [e7bce89f]       [e54a70e8]
>       32.3±0s          33.6±0s     1.04  cvar_benchmark.CVaRBenchmark.time_compile_problem
>       24.5±0s          25.2±0s     1.03  qp_1611_benchmark.QP1611Benchmark.time_compile_problem
>       48.1±0s          48.8±0s     1.01  sdp_segfault_1132_benchmark.SDPSegfault1132Benchmark.time_compile_problem
>       24.8±0s          25.0±0s     1.00  simple_LP_benchmarks.SimpleLPBenchmark.time_compile_problem
>       4.49±0s          4.49±0s     1.00  simple_LP_benchmarks.SimpleFullyParametrizedLPBenchmark.time_compile_problem
>       1.86±0s          1.84±0s     0.99  simple_LP_benchmarks.SimpleScalarParametrizedLPBenchmark.time_compile_problem
> ```

Amazing job @parthb83 !!!

SteveDiamond:
@phschiele @rileyjmurray  I went ahead and added functionality to specify which cones a MI conic solver supports (as opposed to the non-MI version of that solver, which might support different cones). If one of you has a chance to review those changes that would be great!

doc/source/tutorial/advanced/index.rst
```diff
@@ -460,7 +464,9 @@ The table below shows the types of problems the supported solvers can handle.
 | `SCIPY`_       | X  |    |      |     |     |     |     |
 +----------------+----+----+------+-----+-----+-----+-----+
 
-(*) Except mixed-integer SDP.
+(*) Mixed-integer LP only.
+
+(**) Except mixed-integer SDP.
```
rileyjmurray, line 469: I would switch which of these has one asterisk and which has two.

rileyjmurray (approved):
Looks good to me! I did leave one tiny comment about web docs though.

doc/source/tutorial/advanced/index.rst
SteveDiamond: Could you say more? I did * for the first item in the table with an asterisk and ** for the second.

PR #1798, opened 2022-06-10
Added benchmarks workflow to benchmark the master branch of cvxpy and upload results

parthb83:
## Description
Added a `benchmarks.yml` file in `.github/workflows` to trigger benchmarks on master branch of cvxpy whenever a PR?commit is made. The workflow works as follows:

- Checkout the `main` branch of `cvxpy/benchmarks` and install dependencies.
- Checkout the `gh-pages` branch of `cvxpy/benchmarks` and clone the `results` and `docs` folder inside the workflow.
- Benchmarks are run on the latest commit on `master` branch in cvxpy.
- Upload the `results` and `docs` folder created back to `gh-pages` branch of `cvxpy/benchmarks` to store the results and deploy the docs folder to see results in an interactive website.

## Things to do
Create a **Personal Access Token(PAT)** with minimum rights so that it is sufficient to push the changes back to `gh-pages` branch on `cvxpy/benchmarks`. Store the token in secrets as `BENCHMARK_DEPLOY_TOKEN`.

## Type of change
- [] New feature (backwards compatible)
- [ ] Bug fix

- [ ] Write unittests.


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,61 @@
+name: benchmarks
+
+on:
+    pull_request:
+    push:
+        branches:
+            - master
+        tags:
+          - '*'
```
phschiele, line 9:
```suggestion
    push:
        branches:
            - master
```

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,61 @@
...
+          mv results pages/
+          mv docs pages/
+
+      - name: Deploy HTML and Results Artifacts
+        uses: JamesIves/github-pages-deploy-action@v4.2.2
+        with:
+          folder: pages/
+          repository-name: cvxpy/benchmarks
+          branch: gh-pages
+          token: "${{ secrets.BENCHMARK_DEPLOY_TOKEN }}"
```
phschiele, line 61: It would be better to use a deploy key over a personal access token, as the former can have permissions for only the benchmark repository. 
If a PAT got leaked, all repos (public and private) of the person that generated the token would be exposed. The blast radius of the deploy keys would be very small.
I'll look into how this can be done and post an update here.

phschiele (commented):
Had one small comment and one security consideration above.
Meanwhile, if anyone wants to see what this PR enables, here is a live version over at @parthb83 's fork: https://parthb83.github.io/benchmarks/ (still very empty of course, but the idea should be clear).

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          mkdir docs
+          mkdir results
+          cp -a deployed/results/. results/
+          cp -a deployed/docs/. docs/
+
+      - name: Run benchmarks
+        run: |
+          asv machine --yes
+          asv run --quick
```
phschiele, line 46: We should run the current commit and also the previous to get a comparison on the same machine (GH action uses different servers so there would be a lot of noise in the timeline otherwise.

parthb83: I was thinking that maybe we could use `asv machine --machine <Machine_Name>` to give a common name to all the machines across the different workflows.

phschiele:
Update: 
[It worked](https://cvxpy.github.io/benchmarks/docs)

PR #1802, opened 2022-06-13
Add check to pypi uploads

SteveDiamond:
## Description
This will not upload files if they already exist on pypi. As a result, the upload jobs will pass rather than fail.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


phschiele (approved):
LGTM!

PR #1803, opened 2022-06-13
[New atom] dotsort

phschiele:
## Description
This PR implements the `dotsort` atom.
While @ericlux and I were looking at a weighted _generalization_ of `sum_largest`, @stephenpboyd pointed out that this is in fact a _special case_ of `dotsort`, i.e., $\langle sort\left(vec(X)\right), sort\left(vec(W)\right) \rangle$.

A derivation for `sum_largest` is given e.g. in example 2.7 [here](https://docs.mosek.com/modeling-cookbook/linear.html#weak-and-strong-duality). I'll note that the implementation of `sum_largest` uses a different notation in the dual, which makes translating from this reference to the implementation a bit confusing. In the PR description and implementation, I will stick to the notation used in `sum_largest`. A reference for the implementation (in Julia) can be found [here](https://github.com/jump-dev/Convex.jl/blob/e3ca4249495b836e437043f37f34d86163297ef2/src/atoms/lp_cone/dotsort.jl).

One can think of extending the primal form of `sum_largest` to 

$$ \begin{split}\begin{array}{ll}
\mbox{maximize}      &  x^TZw \\
\mbox{subject to}    &  0\leq Z_{ij}\leq 1, & i=1,\ldots,n, j=1,\ldots,k,\\
                     &  \sum_i Z_{ij} = 1, & i=1,\ldots,n,\\
                     &  \sum_j Z_{ij} \leq 1, & j=1,\ldots,k
\end{array}\end{split} $$

When this problem is solved, $Z$ is used to assign the largest value of $w$ to the largest value of $x$, etc.. I.e., it will look something like this:
image

The dual looks like this and thus allows to have non-constant $x$:

$$ \begin{split}\begin{array}{ll}
\mbox{minimize}   & \sum_j q_j +\sum_i t_i &      & \\
\mbox{subject to }& t_i+q_j         & \geq & w_j * x_i, & i=1,\ldots,n, j=1,\ldots,k\\
                  & t_i           & \geq & 0, & i=1,\ldots,n,
\end{array}\end{split} $$

I have left some implementation details as comments in the code below.

## Tasks
- [x] main code
- [x] tests
- [x] docs
- [x] grad implementation  

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/atoms/dotsort.py
```diff
@@ -0,0 +1,110 @@
...
+    where :math:`X` is an expression and :math:`W` is constant.
+
+    | Both arguments are flattened, i.e., we define :math:`x=vec(X)`, :math:`w=vec(W)`.
+    | If the length of :math:`w` is less than the length of :math:`x`,
+     it is conceptually padded with zeros.
+    | When the length of :math:`w` is larger than the length of :math:`x`, an exception is raised.
+
+    `dotsort` is a generalization of `sum_largest` and `sum_smallest`:
+
+    | `sum_largest(x, 3)` is equivalent to `dotsort(x,[1,1,1])`
```
phschiele, line 35: Theoretically, one could rewrite `sum_largest` and `sum_smallest` in terms of `dotsort` now with about the same performance (see the other comment).
I did not do some for two reasons:
1. this would change `sum_largest` from a class to a function, which one might consider a breaking change.
2. `sum_largest` has `k` as `data`, which changes the interface for the `atom.copy()` function.

cvxpy/reductions/eliminate_pwl/atom_canonicalizers/dotsort_canon.py
```diff
@@ -0,0 +1,39 @@
...
+from cvxpy import reshape
+from cvxpy.atoms.affine.sum import sum
+from cvxpy.expressions.variable import Variable
+
+
+def dotsort_canon(expr, args):
+    x = args[0]
+    w = args[1]
+
+    w_unique, w_counts = np.unique(w.value, return_counts=True)
```
phschiele, line 27: This change allows to only add `n * len(unique(k))` instead of `n * k` constraints. 
-> E.g. `dotsort(x, [1,1,1])` exactly reduces to `sum_largest(x,3)` here, as `w_unique` would be `[1]` and `w_counts` would be `[3]`.

cvxpy/tests/test_atoms.py
```diff
@@ -782,6 +782,174 @@ def test_sum_smallest(self) -> None:
...
+        self.assertEqual(str(cm.exception),
+                         "The W argument must be constant.")
+
+        # swapped arguments
+        with self.assertRaises(Exception) as cm:
+            cp.dotsort([1, 2, 3], self.x)
+        self.assertEqual(str(cm.exception),
+                         "The W argument must be constant.")
+
+        with self.assertRaises(Exception) as cm:
```
phschiele, line 949:
```suggestion
        # non-dcp composition
        with self.assertRaises(Exception) as cm:
```

adishavit:
Could this somehow be used without the sum to get a sorted vector?

phschiele:
> Could this somehow be used without the sum to get a sorted vector?

@adishavit I suppose you can use this approach to sort a constant vector, e.g. as shown [here](https://folk.ntnu.no/staal/programming/algorithms/convex-sorting/), but that's probably not what you mean.

Writing a vector as `y = [largest value of x, second largest value of x, ...]` where x is a variable is not convex (see [here](https://math.stackexchange.com/a/3133505) for a related SO post).

cvxpy/atoms/dotsort.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def is_atom_concave(self) -> bool:
+        """Is the atom concave?
+        """
+        return False
+
+    def is_incr(self, idx) -> bool:
+        """Is the composition non-decreasing in argument idx?
+        """
+        return np.all(self.args[1].value >= 0)
```
SteveDiamond, line 115: is_incr can't look at values, it can only look at symbolic attributes like nonneg, nonpos

cvxpy/atoms/dotsort.py
```diff
@@ -0,0 +1,135 @@
...
+    assigned to the largest entry in :math:`x`, etc..
+    """
+
+    def __init__(self, X, W) -> None:
+        super(dotsort, self).__init__(X, W)
+
+    def validate_arguments(self) -> None:
+        if not self.args[1].is_constant():
+            raise ValueError("The W argument must be constant.")
+        if self.args[0].size < self.args[1].size:
```
SteveDiamond, line 51: So if W is smaller than X any missing element of W is assumed to be 0? What if W is 0D or 1D and X is 2D?

cvxpy/atoms/dotsort.py
SteveDiamond: actually sorry this might be fine if you also guarantee that W is not a parameter

adishavit:
I was asking because I’ve used a sorting matrix in an LP before. I don’t remember the formulation ATM but it works fine with CVXPY. 
I thought maybe there was a more specialize atom…

> @adishavit I suppose you can use this approach to sort a constant vector, e.g. as shown [here](https://folk.ntnu.no/staal/programming/algorithms/convex-sorting/), but that's probably not what you mean.
> 
> Writing a vector as `y = [largest value of x, second largest value of x, ...]` where `x` is a variable is not convex

SteveDiamond (approved):
LGTM

cvxpy/atoms/dotsort.py
phschiele: @SteveDiamond I've added parameter support in [7fc9ce7](https://github.com/cvxpy/cvxpy/pull/1803/commits/7fc9ce745baa25887e6ff9d9643193bc315f94a5), but noticed that it would classify some non-DPP compliant formulations as DPP. Instead of restricting `W` to parameter instances as in [af44e3b](https://github.com/cvxpy/cvxpy/pull/1803/commits/af44e3b9db86386ddfa1abee3fc1b4c3e80de448), 
I've now ensured that if `W` is parametrized, it must be parameter-affine, see [8d74114](https://github.com/cvxpy/cvxpy/pull/1803/commits/8d7411460d9ee67f67c5a580d28ab8aab7e02f44).

phschiele:
PR is complete from my side. It needs another reviewer as it changes the public API.
@SteveDiamond @rileyjmurray @akshayka @bstellato

rileyjmurray (approved):
LGTM!

Issue #1805, opened 2022-06-20
Suppressing SCIP warning


keithbriggs:
Is there any way to suppress this warning? 

```
(CVXPY) Jun 20 04:37:03 PM: Encountered unexpected exception importing solver SCIP:
NotImplementedError("PySCIPOpt (SCIP's Python wrapper) is installed and itsversion is 4.0.0. 
CVXPY only supports PySCIPOpt < 4.0.0.")
```

I'm not using SCIP, so I don't want to see this over and over again.  Also, there's a typo "itsversion".

Alternatively, to actually support SCIP 4.0.0 would be nice!

SteveDiamond:
Could you checkout PR #1807 and see if that fixes the issue?

PR #1807, opened 2022-06-24
Hide SCIP warning.

SteveDiamond:
## Description
This PR addresses fixes a typo and hides the warning for having the wrong version of SCIP on import.
Issue link (if applicable): #1805

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
@@ -91,9 +91,9 @@ class SCIP(ConicSolver):
         import pyscipopt
         v = pyscipopt.__version__
         if Version(v) >= Version('4.0.0'):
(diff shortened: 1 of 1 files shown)


SteveDiamond:
@keithbriggs can you try this PR and see if the code works for you?

keithbriggs:
Steve: thanks very much - the warning has gone away!   Good to see that the US is actually moving forward on at least one issue.   I see that SCIP is now at  4.0.1.   It don't need SCIP at the moment, but it would be nice to have this supported eventually.

phschiele:
In case anyone else is wondering, this works because of this Exception around the solver imports:
https://github.com/cvxpy/cvxpy/blob/f788cce686fcecf705b52f35a154526fea65362a/cvxpy/reductions/solvers/solver.py#L62-L67

PR #1818, opened 2022-07-06
Assume PSD argument to quad_form

SteveDiamond:
## Description
Adds an argument to quad_form to skip checking that P is PSD.
Issue link (if applicable): 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/quad_form.py
+++ b/cvxpy/atoms/quad_form.py
@@ -35,7 +35,16 @@ class CvxPyDomainError(Exception):
 class QuadForm(Atom):
     _allow_complex = True
 
-    def __init__(self, x, P) -> None:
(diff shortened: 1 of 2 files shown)


SteveDiamond:
I wasn't sure whether there should be an ``assume_NSD`` argument for symmetry.

phschiele (commented):
That's very handy! I have done `psd_wrap(P)` many times, e.g. when P was already asserted to be psd beforehand, or when it is psd by construction. Reusing this functionality is very clean :+1: 
Based on previous issues, psd seems to be the more common application, so I don't think the nsd equivalent needs to be added right away for symmetry, especially as this is just a short hand for existing functionality.

SteveDiamond:
Thanks for taking a look! I'll go ahead and merge then.

PR #1820, opened 2022-07-08
Follow up with assume PSD

SteveDiamond:
## Description
I made a small mistake in the tests.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_quad_form.py
+++ b/cvxpy/tests/test_quad_form.py
@@ -200,3 +200,15 @@ class TestNonOptimal(BaseTest):
         prob = cvxpy.Problem(cvxpy.Minimize(expr),
                              [A @ x <= b])
         prob.solve(solver=cvxpy.SCS)
+
+    def test_assume_psd(self) -> None:
+        """Test assume_PSD argument.
+        """
+        x = cvxpy.Variable(3)
+        A = np.eye(3)
+        expr = cvxpy.quad_form(x, A, assume_PSD=True)
+        assert expr.is_convex()
+
+        A = np.eye(3)
+        expr = cvxpy.quad_form(x, A, assume_PSD=True)
+        assert expr.is_convex()
--- a/cvxpy/tests/test_quad_form.py
+++ b/cvxpy/tests/test_quad_form.py
@@ -209,6 +209,6 @@ class TestNonOptimal(BaseTest):
         expr = cvxpy.quad_form(x, A, assume_PSD=True)
         assert expr.is_convex()
 
-        A = np.eye(3)
+        A = -np.eye(3)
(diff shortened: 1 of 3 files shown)


phschiele:
@SteveDiamond Good spot, I also missed this. 
I actually even tried the NSD case locally, just to see what happens if one would try to solve such a problem.
For OSQP, one would get the following printout + subsequent exception:
```
ERROR in LDL_factor: Error in KKT matrix LDL factorization when computing the nonzero elements. The problem seems to be non-convex
ERROR in osqp_setup: KKT matrix factorization.
The problem seems to be non-convex.
...
ValueError: Workspace allocation error!
```
Perhaps this should be caught and reraised as a cvxpy `SolverError`. This would be interface specific, of course, but for OSQP as the default QP solver, it might make sense to add it alongside here. Could be a problem if anyone catches it based on being a `ValueError` right now.

cvxpy/tests/test_quad_form.py
```diff
@@ -77,138 +78,144 @@ def test_sparse_quad_form(self) -> None:
...
-        A = np.eye(3)
-        expr = cvxpy.quad_form(x, A, assume_PSD=True)
+        A = -np.eye(3)
+        expr = cp.quad_form(x, A, assume_PSD=True)
         assert expr.is_convex()
+
+        prob = cp.Problem(cp.Minimize(expr))
+        # Transform to a SolverError.
+        with pytest.raises(cp.SolverError,
+                           match="Workspace allocation error!"):
```
rileyjmurray, line 220: Can we be confident that the error message will contain this text in future releases of OSQP?

rileyjmurray: @SteveDiamond

SteveDiamond: I don't know. But what would be the alternative?

phschiele: @rileyjmurray if this test starts to fail because of this, it would be as simple as updating the string, right? Is your main concern to make as few assumption about solver outputs in general for maintenance reasons?

rileyjmurray: I'm just wondering if there's a different (sub)string that we might check. ``Error in KKT matrix LDL factorization`` seems more reasonable here. If the tests start to fail in the future it's true that this would be an easy fix.

SteveDiamond: For whatever reason ``match`` is only catching "Workspace allocation error!" and ignoring the rest of the visible error message. So other substrings don't work here.

rileyjmurray: Gotcha. In that case I'm in favor of merging as-is.

Issue #1747, opened 2022-04-11
Incorrect documentation for CBC interface


tkralphs:
The documentation related to [Cbc](https://github.com/coin-or/Cbc) [here](https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options) is incorrect. None of the options listed are supported. Also, all links to COIN-OR projects ([Cbc](https://github.com/coin-or/Cbc), [Cgl](https://github.com/coin-or/Cgl), [Clp](https://github.com/coin-or/Clp), etc.) should be to https://github.com/coin-or not to https://projects.coin-or.org.

SteveDiamond:
That's surprising and unfortunate that cylp doesn't pass the arguments through to Cbc. Is there another python interface you would suggest?

tkralphs:
To be precise, CyLP doesn't have a single uniform method for passing _arbitrary_ Cbc parameters. It does have purpose-built methods for setting _certain_ parameters, but part of the problem is also that the CVXPY interface doesn't call what methods do exist. The interface only attempts to pass parameters like this.
https://github.com/cvxpy/cvxpy/blob/ffc05bb6237fbf70644998635e4e24e792c12a30/cvxpy/reductions/solvers/conic_solvers/cbc_conif.py#L165-L166 
But this isn't the way parameters are set in CyLP currently. 

What should probably be done is simply to modify CyLP so that one can pass arbitrary parameters through its solve call to the solve call of Cbc. This is one of the ways to set parameters when calling Cbc's solve method through the C/C++ interface and extending this to the CyLP solve call should be relatively easy. 

There are plenty of other Python interfaces to Cbc and there's been a lot of discussion about alternatives that could be used in CVXPY, but I guess that CyLP has been retained mainly because it is more efficient than other options. It would seem a bit gratuitous to use one high-level modeling language inside another one. Honestly, I don't know why one would use CVXPY to solve a MIP in the first place, given how many good alternatives there are.

rileyjmurray:
Thanks for opening this issue and providing that detailed explanation, @tkralphs. We need to do a better job of testing that CVXPY's attempts at setting solver parameters actually result in the intended behavior. I agree that using a modeling platform to access CBC options would be gratuitous. We'll correct the documentation soon.

As a secondary point, when you refer to alternatives to CVXPY for solving MILPs, which platforms do you have in mind within the Python ecosystem? I know of PICOS, Python-MIP, and (to some extent) Google OR-Tools.

[...]

PR #1821, opened 2022-07-09
CBC interface: add ability to configure the underlying CLP solver

bkurtz:
## Description
We found a need for one of our models to be able to configure the CLP solver used for initial solves of CBC models and non-integer problems when using CBC.  We're curious if cvxpy would be open to a change along the lines of what we're proposing here to allow configuring the solver better.  If so, happy to go through the contribution checklist and make sure I'm following the rules, but hoping for some high-level guidance before I invest too much effort!

(We were initially concerned that "CBC" seemed not to be responding to any of our config settings, then realized that CBC wasn't actually being run at all since our model solved entirely with CLP.  Might be related to #1747?)

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
This looks fine to me. @rileyjmurray what do you think?

rileyjmurray:
@bkurtz we'd love a contribution to this effect! It would be good to get @tkralphs opinion on any details.

tkralphs:
This seems to make sense, we would just need to verify that `initialSolve()` does properly process the parameters and pass them through. It looks to me like it should and I would assume @bkurtz has already verified this, at least empirically.

bkurtz:
I'll take a look at this next week and see what I can make of it.  I have to admit to only having tested that changing both the `primalTolerance` and `dualTolerance` settings together had an impact on the solution, but it seemed like while I was here I might as well add all the things that looked configurable.

I like the idea of setting parameters and verifying an impact on the solution better than trying to parse useful things out of the verbose output, but it also seems like it might be harder to construct meaningful problems, so I'll have to see what I can come up with.

rileyjmurray:
@bkurtz would you be willing to finish this PR?

bkurtz:
@rileyjmurray Do still intend to finish this, but it's been bumped down my list of priorities a bit for a couple of reasons - guessing more on the order of 4-6 weeks to get it back to the top of my queue and done.  No objections if someone else wants to finish this, or if you want to close the PR until I've actually got a finished patch.

rileyjmurray:
Thanks for the reply, @bkurtz. I think it would be very helpful to have a CBC / CLP user contribute this patch. Since you fit that bill leave this PR open until you can get to it.

rileyjmurray:
@bkurtz, is there any chance you can get around to finishing this PR?

bkurtz:
Tentatively slotted it into next week's schedule; will keep you posted if it looks like it's going to slip further.

cvxpy/reductions/solvers/conic_solvers/cbc_conif.py
```diff
@@ -160,10 +160,18 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
 
         # Build model & solve
         status = None
+        clp_model_options = {"dualTolerance", "primalTolerance", "maxNumIteration", "logLevel", "automaticScaling", "scaling", "infeasibilityCost", "optimizationDirection"}
```
bkurtz, line 163: `logLevel` is technically an option for both `clp` and `cbc`, but it's also adjustable already via the `verbose` argument here.  Any strong opinions on how this should be implemented?  If I don't hear anything further, I'll tweak the implementation to support `logLevel` as an option on both, even though that duplicates functionality with passing `verbose` as a solver arg.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -607,6 +610,120 @@ def test_sdpa_sdp_2(self) -> None:
...
+        pytest.param({"primalTolerance": 1.0}, StandardTestLPs.test_lp_4, id="primalTolerance"),
+        pytest.param({"maxNumIteration": 1}, StandardTestLPs.test_lp_1, id="maxNumIteration"),
+        # pytest.param({"automaticScaling": True}, StandardTestLPs.test_lp_1, id="automaticScaling"),
+        # pytest.param({"scaling": 0}, StandardTestLPs.test_lp_2, id="scaling"),
+        # pytest.param({"infeasibilityCost": 1e-5}, StandardTestLPs.test_lp_1, id="infeasibilityCost"),
+        pytest.param({"optimizationDirection": "max"}, StandardTestLPs.test_lp_3, id="optimizationDirection"),
+        # presolve option is easier to detect via changes to log messages
+    ]
+)
+def test_cbc_lp_options(opts: dict, sth: Callable[[], sths.SolverTestHelper]) -> None:
```
bkurtz, line 627: mocked out a couple of different ways to test some of these configs:
* several were easy to test by solving existing sample problems with/without the option and verifying different results each way (this function)
* I also tried enabling more verbose logging and comparing solver string outputs (didn't do anything fancy looking for options, just an assumption that the log output should be *different* with/without the option) - this is `test_cbc_lp_options_via_logs` below
* `logLevel` needs to be tested directly via looking at logs

Of note, I added all these as top-level functions so as to be able to use pytest features (parameterizations and `capfd` for capturing log output).  If there's a better/preferred way to do that, let me know.

I didn't manage to get this fully working for all parameters, and unfortunately I'm about at the limit of time I'm willing to put into this.  In particular:
* I didn't manage to find any evidence that `automaticScaling` is impacting a problem, although the code indicates that it's being passed through, so I assume this is due to me not having a sample problem where it makes a difference
* Same for `infeasibilityCost`
* Unfortunately, though I was able to verify them in a debugger, at least in my local test environment, none of the tests that capture logs actually pick up logs unless I first catch them in a debugger (which does something magic around disabling then re-enabling capturing of stdout), so I assume the tests won't work in CI.

I'm happy to take little bits of time to clean this up further (e.g. clean up code, drop whichever test approach we're less happy about, etc.) but I probably don't have time to keep trying to find other cases/problems that would show off those last two config parameters, and I'm probably not going to put a lot more time into getting log capturing to work either.

cvxpy/reductions/solvers/conic_solvers/cbc_conif.py
rileyjmurray: I think supporting ``logLevel`` for both is fine (in fact, good).

cvxpy/tests/test_conic_solvers.py
rileyjmurray: @bkurtz I'm not familiar with the pytest features that you're trying to use here to capture output. I do see that tests are failing due to difficulties of capturing logs programmatically. How can we help move this forward? Are there some failing tests that look like they'd be easier to resolve than others? It's okay if some functionality that you've implemented isn't tested, but I want to make sure the relevant code paths are hit.

bkurtz: They were relatively new to me too.  I'm sure you can look up docs too, but to save you time if you're interested, [`capfd` docs](https://docs.pytest.org/en/6.2.x/capture.html#accessing-captured-output-from-a-test-function).  I also tried [`contextlib.redirect_stdout`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout), without additional luck, and spent quite a lot of time poking around in the pytest/pdb internals to see if I could figure out why the capturing works when I catch it in a debugger and not otherwise, but without luck.

Unless you have any brilliant ideas on how to get log output capturing to work, I think the path forward is probably to disable or delete all the test cases that are currently failing.  That would still leave us with good coverage of most of the config parameters, and certainly the standard *paths*, but there would be a couple of config parameters that would be un-tested.  Code paths that might be of concern (esp. if we delete the log-based tests entirely) would be:
* `logLevel` is hard to test reliably if you can't capture logs
* I did find a case or two where `presolve` had an impact on the actual solving, but it feels a little brittle to me to rely on that (e.g. a future upgrade to the solver might change that), so I didn't write the test that way.  `presolve` is relevant because it's passed as an argument to the solve function rather than applied as an option to the model, i.e. it's a different code path that wouldn't be tested other ways.

If you're okay with tests that *hit* all the code paths, but don't necessarily provide hard evidence of them having been hit (i.e. we hit them to make sure they don't crash, but we don't necessarily verify that they had an impact on the solving), that's a fairly trivial adjustment to make.

rileyjmurray: Great, thanks for those pointers. I'll start playing with this and get back to you (/update this PR).

rileyjmurray: Okay I've figured out how to capture logs. The problem is that ``contextlib``'s ``redirect_stdout`` and ``redirect_stderr`` don't apply to subprocesses. I found someone's custom ``contextmanager`` on StackExchange that is intended to capture subprocess output (see [here](https://stackoverflow.com/questions/4675728/redirect-stdout-to-a-file-in-python)) and got it to work, albeit in a hacky way. I'll report back here once I've figured out if this solution can be cleaned up a bit.

bkurtz: I'd be happy to help clean up if you can post the working example; I poked around on that stackexchange page a little more and wasn't able to get a working example still.

rileyjmurray: Okay, here's the code
```
from cvxpy.tests.solver_test_helpers import StandardTestLPs
from contextlib import contextmanager
import tempfile
import sys
import os


def fileno(file_or_fd):
    fd = getattr(file_or_fd, 'fileno', lambda: file_or_fd)()
    if not isinstance(fd, int):
        raise ValueError("Expected a file (`.fileno()`) or a file descriptor")
    return fd


@contextmanager
def stdout_redirected(to=os.devnull, stdout=None):
    if stdout is None:
        stdout = sys.stdout

    stdout_fd = fileno(stdout)
    # copy stdout_fd before it is overwritten
    #NOTE: `copied` is inheritable on Windows when duplicating a standard stream
    with os.fdopen(os.dup(stdout_fd), 'wb') as copied:
        stdout.flush()  # flush library buffers that dup2 knows nothing about
        try:
            os.dup2(fileno(to), stdout_fd)  # $ exec >&to
        except ValueError:  # filename
            with open(to, 'wb') as to_file:
                os.dup2(to_file.fileno(), stdout_fd)  # $ exec > to
        try:
            yield stdout # allow code to be run with the redirected stdout
        finally:
            # restore stdout to its previous value
            #NOTE: dup2 makes stdout_fd inheritable unconditionally
            stdout.flush()
            os.dup2(copied.fileno(), stdout_fd)

# here's the workflow we need. Should encapsulate this in a function.
myfile = tempfile.NamedTemporaryFile('w', delete=False)
with stdout_redirected(myfile):
        StandardTestLPs.test_lp_0(solver='CBC', duals=False, logLevel=1)
        sys.stdout.flush()
myfile.close()
myfileopened = open(myfile.name, 'r')
lines = myfileopened.readlines()
print(lines)

```

rileyjmurray: Working on this now!

rileyjmurray: Aaaannnd it's actually not working when called directly from terminal. It only works when launched from PyCharm. No idea what's going on there. I'll play around a bit more and probably push my code just so we can all see what I've got.

bkurtz: Oh interesting.  That's consistent with what I found today, which is that it didn't work for me.  I think the approach with calling `os.dup` or `os.dup2` on the raw file descriptors is the same thing that pytest does under the hood with `capfd` (as opposed to `capsys` which just replaces `sys.stdout`).  I also tried setting the "inheritable" flag on the file handle with no success.

Interesting that running in pycharm gets it working.  I might test that tomorrow briefly and see if it helps at all with the original tests - might give some idea of what's going on; maybe there's some tty parameter or something that's causing the magic (I found dropping into `pdb` in the shell also triggered something inexplicable that magically made the capturing work).

rileyjmurray: Okay so here's my last update for the day. It's easier to just paste here than push broken changes to a branch.

The first thing I did was make a new file, ``cvxpy/utilities/io.py``, with the following contents
```
import tempfile
from contextlib import contextmanager
import sys
import os


def fileno(file_or_fd):
    fd = getattr(file_or_fd, 'fileno', lambda: file_or_fd)()
    if not isinstance(fd, int):
        raise ValueError("Expected a file (`.fileno()`) or a file descriptor")
    return fd


@contextmanager
def stdout_to_tempfile(to=None):
    if to is None:
        to = tempfile.NamedTemporaryFile(mode='w', delete=False)
    else:
        assert hasattr(to, 'file')
        to.file.writeable()
        # ^ will raise an OSError if not writeable.
    stdout = sys.stdout
    stdout_fd = fileno(stdout)
    # copy stdout_fd before it is overwritten
    # NOTE: `copied` is inheritable on Windows when duplicating a standard stream
    with os.fdopen(os.dup(stdout_fd), 'wb') as copied:
        stdout.flush()  # flush library buffers that dup2 knows nothing about
        try:
            os.dup2(fileno(to), stdout_fd)  # $ exec >&to
        except ValueError:  # filename
            with open(to, 'wb') as to_file:
                os.dup2(to_file.fileno(), stdout_fd)  # $ exec > to
        try:
            yield to  # allow code to be run with the redirected stdout
        finally:
            # restore stdout to its previous value
            # NOTE: dup2 makes stdout_fd inheritable unconditionally
            stdout.flush()
            os.dup2(copied.fileno(), stdout_fd)
            to.close()
```

Then I test it with the following code (which behaves as expected when launched from PyCharm, but not from terminal):
```
from cvxpy.utilities.io import stdout_to_tempfile
from cvxpy.tests.test_conic_solvers import StandardTestLPs

with stdout_to_tempfile() as tf:
    StandardTestLPs.test_lp_0(solver='CBC', duals=False, logLevel=1)
tf = open(tf.name, 'r')
print(tf.readlines())
```

rileyjmurray: @bkurtz let me know if you plan on experimenting more, or if you'd like to "call it" and proceed with weaker tests that don't actually inspect logs.

bkurtz: I'd like to poke one more thing to see if I can find any obvious differences between situations that work and those that don't, but otherwise it might be time to call it.  Should have an update by this evening.

bkurtz: The key here turns out to be that python and `libc` maintain their own buffers for each file handle, and `Clp` (via `CyLP`) is using the C buffers, so you have to call libc's `fflush` before trying to read the logs.  (I ended up finding https://github.com/pytest-dev/pytest/issues/8753 which describes essentially the issue and the solution nicely.)

Not 100% sure how that works on Windows, so might need to add a conditional on the OS before calling `fflush` or something.

rileyjmurray: Tests pass!!! I'll mark this conversation as resolved.

rileyjmurray (changes requested):
We're on the home stretch, @bkurtz! I do have a couple of requests. First, please move the test functions in ``test_conic_solvers.py`` underneath the ``TestCBC`` class. This will produce errors, but you can resolve the errors if you simply declare the class as
```
class TestCBC:
    ...
```
rather than
```
class TestCBC(BaseTest):
...
```
It would also be nice if you could revisit some of the TODOs you left yourself (for example, you left some notes about the ``duals=False`` option when calling SolverTestHelpers).

bkurtz:
Did some cleanup.  I ended up cleaning these up under a second class for now, because dropping `BaseTest` from the main `TestCBC` class resulted in one of the other tests there failing.  I also dropped one of the test functions I'd written entirely, because it didn't feel like it added incremental value.

rileyjmurray (commented):
I have a question about when you tried to use a single ``TestCBC`` class that didn't inherit from ``BaseTest``.

Did you remove the ``test_options`` and ``setUp`` functions? It's abundantly clear that the former function hasn't tested anything meaningful in a long time, and your tests subsume it by a significant margin, so it should be fine to remove.

I ask because I tried dropping ``BaseTest`` as the parent class after removing ``test_options`` and ``setUp`` functions, and that seemed to work on my Linux machine.

bkurtz:
ah, okay, if I can kill the old `test_options`, this should be easy.  I'm not sure my new tests really cover the same things it was intended to test, but as you say it's abundantly clear that it isn't doing anything interesting anymore anyway.  Will push an update tomorrow.

rileyjmurray (approved):
The failing builds are all due to installation issues on Windows, which are not related to this PR.

bkurtz:
Thanks for wrapping this up!  You pushed the skip fix about half an hour before I sat down to work on it myself!

rileyjmurray:
No, thank you! It was a pleasure figuring the I/O out together, and we really appreciate you reaching out and doing the hard work.

PR #1830, opened 2022-07-18
Add support for solving mixed-integer programs via HiGHS/SciPy interface

### LINKED ISSUE #1443: Add an interface to the highs MIP solver
Apparently [Highs](https://github.com/ERGO-Code/HiGHS/blob/87ca783572a0ad8ee8a8b98f73585ae58d2e0032/examples/call_highs_from_python.py) includes an MIP solver. I tested the LP solver which is really fast and scalable. As MIP solving is based on LP problems, I assume that it will be also efficient for MIP solving. As efficient open source MIP solvers are rare, I thought it may be interesting to add it to cvxpy.

tasseff:
## Description
This pull request will close #1443 by adding support for solving mixed-integer programs via SciPy functionality that will be available in version 1.9.0. Please see [this discussion](https://github.com/scipy/scipy/issues/16609#issuecomment-1186266704) about HiGHS/SciPy updates that should resolve test instances that currently do not pass but are marked with `TODO`.

## Type of change
- [ ] Bug fix

- [x] Write unittests.
  - Two tests fail for me that are not related to these changes:
```
FAILED test_grad.py::TestGrad::test_huber - ValueError: Inexact indices into sparse matrices are not allowed
FAILED test_grad.py::TestGrad::test_partial_problem - ValueError: Inexact indices into sparse matrices are not allowed
```
  - I don't think these changes affect the benchmarks.


cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -97,20 +120,44 @@ def apply(self, problem):
...
+        if problem_is_a_mip:
+            integrality = [0] * data[s.C].shape[0]
+
+            for index in data[s.BOOL_IDX] + data[s.INT_IDX]:
+                integrality[index] = 1
+
+            bounds = [(None, None)] * data[s.C].shape[0]
+
+            for index in data[s.BOOL_IDX]:
+                bounds[index] = (0, 1)
```
tasseff, line 148: Should these binary variable bounds be specified in the CVXPY canonicalized problem, or should they be specified via the SciPy `bounds` option, as is currently being done?

ghackebeil: Seems like this might be the first interface where the solver doesn't accept a "binary" identifier which would implicitly carry those bounds, so I can see why the canonicalization process would leave binary bounds out of the specification so far. I think what you are doing here makes sense.

cvxpy/utilities/versioning.py
```diff
@@ -26,6 +26,7 @@ class Version:
 
     def __init__(self, v: Union[str, Tuple[int, int, int]]):
         if isinstance(v, str):
+            v = v.split('rc')[0] if 'rc' in v else v
```
SteveDiamond, line 29: Highlighting this change for @rileyjmurray.

SteveDiamond (approved):

SteveDiamond:
This LGTM. @rileyjmurray can you take a look?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -55,6 +59,20 @@ def name(self):
         """
         return s.SCIPY
 
+    def accepts(self, problem) -> bool:
+        """Can HiGHS solve the problem?
+        """
+        # TODO check if is matrix stuffed.
```
rileyjmurray, line 65: What is this comment here for?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -55,6 +59,20 @@ def name(self):
...
+        if not problem.objective.args[0].is_affine():
+            return False
+        for constr in problem.constraints:
+            if type(constr) not in self.SUPPORTED_CONSTRAINTS:
+                return False
+            for arg in constr.args:
+                if not arg.is_affine():
+                    return False
+        return True
```
rileyjmurray, line 74: This seems like pretty boilerplate code. I looked at ``ConicSolver`` and it includes similar checks. @SteveDiamond, would it be reasonable for @tasseff to update [this function](https://github.com/tasseff/cvxpy/blob/c4ba1ab85ae1dbb6967df2ec1e54f843479bd330/cvxpy/reductions/solvers/conic_solvers/conic_solver.py#L105) so that it includes a check that the objective is affine, and then have the ``SCIPY`` class just use the parent class implementation?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -97,20 +120,44 @@ def apply(self, problem):
     def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
         from scipy import optimize as opt
 
+        # Check if the problem is a MIP and for SciPy version support.
+        problem_is_a_mip = data[s.BOOL_IDX] or data[s.INT_IDX]
+        scipy_supports_mip = not (Version(scipy.__version__) < Version('1.9.0'))
+
+        if problem_is_a_mip and not scipy_supports_mip:
+            raise ValueError("Solving MIPs requires SciPy version >= 1.9")
+
```
rileyjmurray, line 129: This should be handled by ``accepts()``.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -141,28 +188,52 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
         else:
 
             warnings.warn("It is best to specify the 'method' parameter "
                           "within scipy_options. The main advantage "
-                          "of this solver, is its ability to use the "
+                          "of this solver is its ability to use the "
                           "HiGHS LP solvers via scipy.optimize.linprog() "
                           "which require a SciPy version >= 1.6.1 ."
                           "\n\nThe default method '{}' will be"
                           " used in this case.\n".format(meth))
```
rileyjmurray, line 227: I don't like that this warning is here by default. I know it wasn't added in this PR but I think it might be good to remove here. @SteveDiamond what do you think?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -141,28 +188,52 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
-                                   A_eq=data[s.A], b_eq=data[s.B], method=meth,
-                                   bounds=(None, None))
+            if problem_is_a_mip:
+                solution = opt.linprog(data[s.C], A_ub=data[s.G], b_ub=data[s.H],
+                                       A_eq=data[s.A], b_eq=data[s.B], method=meth,
+                                       integrality=integrality, bounds=bounds)
+            else:
+                solution = opt.linprog(data[s.C], A_ub=data[s.G], b_ub=data[s.H],
+                                       A_eq=data[s.A], b_eq=data[s.B], method=meth,
+                                       bounds=bounds)
```
rileyjmurray, line 236: This code block seems to be redundant with lines 207 -- 216.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
SteveDiamond: Sorry I don't fully follow. The warning is because they're using the scipy interface but not HiGHS.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
SteveDiamond: Sure, good idea!

cvxpy/utilities/versioning.py
rileyjmurray: Conceptually this change is fine. But there's no need to check if ``'rc'`` is in ``v`` before calling ``v.split('rc')``. If the string isn't there then ``res = v.split('rc')`` will be a list of length 1 with ``v == res[0]``.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -1703,6 +1703,31 @@ def test_scipy_lp_4(self) -> None:
...
+
+    @unittest.skipUnless('SCIPY' in INSTALLED_MI_SOLVERS, 'SCIPY version cannot solve MILPs')
+    def test_scipy_mi_lp_4(self) -> None:
+        StandardTestLPs.test_mi_lp_4(solver='SCIPY')
+
+    # TODO: Remove the below "skip" once SciPy HiGHS is updated.
+    @pytest.mark.skip(reason="Known bug in SCIPY")
+    def test_scipy_mi_lp_5(self) -> None:
+        StandardTestLPs.test_mi_lp_5(solver='SCIPY')
+
```
rileyjmurray, line 1730: I just realized that all of the test MILPs have integer values at optimality (even if the variables aren't constrained to integers). @SteveDiamond we should add a simple test where the optimal solution has some fractional variables. Would you like that to be added in this PR or should we leave that for later?

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
rileyjmurray: But if they have a recent version of SciPy then they'll have access to HiGHS, and we'll select that for them automatically. My reservation about this is that CVXPY users like to call ``prob.solve()`` and just have things work.

In this situation if they have up-to-date SciPy and no separate MILP solver, then they'd get a warning message when calling ``prob.solve()``. There would be no context for what "this solver" means to a novice user.

cvxpy/tests/test_conic_solvers.py
SteveDiamond: Good catch! We should add that later IMO.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
SteveDiamond: Ah ok. Definitely HiGHS should be used by default, good point. Then it's not clear what this warning is about and it should be removed I gues.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
tasseff: How should the version checks be handled if you want me to update and use the `ConicSolver` implementation of `accepts`? It seems like a custom `accepts` implementation will be required if you want to retain backward compatibility with older versions of SciPy (which cannot solve MIPs). Same concern with removing the warning in your comment below this one: do you prefer backward compatibility, or do you want to assume compatible SciPy versions? (FYI: SciPy 1.9.0 was released yesterday.)

rileyjmurray: ``ConicSolver.accepts`` can use the following method in the ``ParamConeProg`` class to check if the problem is mixed-integer:
https://github.com/tasseff/cvxpy/blob/c4ba1ab85ae1dbb6967df2ec1e54f843479bd330/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py#L155-L158
The ``SCIPY(ConicSolver)`` class will be responsible for correctly setting ``self.MIP_CAPABLE`` (which it does in the current PR's code). Assuming that value is correctly set then a generic solver-independent implementation of ``ConicSolver.accepts`` should work fine. Does that sound about right to you?

rileyjmurray: Regarding the error message, my objection is that the message is raised even for SciPy versions 1.9.0 if someone doesn't explicitly specify the solve method. I'm of the opinion that if SciPy >= 1.9.0 is installed then we should silently default to HiGHS.

CVXPY has a logging feature that operates on top of the logging features of numerical solvers. If this message is displayed then I suggest it be part of that log. That way someone would see the message if they solve with ``prob.solve(verbose=True)``.

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
tasseff: The current `accepts` method for a `ConicSolver` checks if `isinstance(problem, ParamConeProg)`. According to the `ParamConeProg` documentation, the objective is `minimize   c'x  + d`, so are we already guaranteed the objective will be affine? The reason I ask is that I added an additional check like:
```python
if len(problem.objective.args) > 0:
    objective_is_affine = problem.objective.args[0].is_affine()
else:
    objective_is_affine = True
```
but it turns out that `problem` does not always have an `objective` member, e.g., in the tests, I see
```
FAILED cvxpy/tests/test_linear_cone.py::TestLinearCone::test_scalar_lp - AttributeError: 'ParamConeProg' object has no attribute 'objective'
FAILED cvxpy/tests/test_linear_cone.py::TestLinearCone::test_socp - AttributeError: 'ParamConeProg' object has no attribute 'objective'
FAILED cvxpy/tests/test_linear_cone.py::TestLinearCone::test_vector_lp - AttributeError: 'ParamConeProg' object has no attribute 'objective'
```
What are your thoughts?

rileyjmurray: Since the objective is affine in ``ParamConeProg`` we can remove that check. Although it's odd that some ``ParamConeProg`` instances don't have an objective attribute ... @SteveDiamond any idea how that could happen?

tasseff:
@rileyjmurray I think I've addressed all your comments. Thanks for the great and quick suggestions!

Two tests still fail that are unrelated to this PR but are a consequence of the new version of SciPy:
```
FAILED cvxpy/tests/test_grad.py::TestGrad::test_huber - ValueError: Inexact indices into sparse matrices are not allowed
FAILED cvxpy/tests/test_grad.py::TestGrad::test_partial_problem - ValueError: Inexact indices into sparse matrices are not allowed
```
I'm unsure how these should be fixed.

rileyjmurray (changes requested):
The code looks good to me! I've launched the GitHub Actions tests. If all tests pass except those mentioned in https://github.com/cvxpy/cvxpy/pull/1830#issuecomment-1200461496 then I'd say the code is good to merge. The one thing that still needs to be updated is the web documentation source files. Changes are needed ...

[Here, for mention of SciPy 1.9.0 and the MILP functionality.](https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst#install-with-scipy-support)

and

[Here, indicating MILP support when SciPy version 1.9.0 or higher is installed.](https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#choosing-a-solver)

and 

[Farther down on this list, mentioning that ``integrality`` isn't allowed in ``scipy_options``.](https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#setting-solver-options)

cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
SteveDiamond: ``ParamConeProg`` doesn't have an ``objective`` attribute ever. It stores ``c``, ``A``, and other info, but there's no field called ``objective``.

tasseff:
@rileyjmurray, okay, I think this should be good to go! Let me know if you have any other suggestions.

rileyjmurray (approved):

rileyjmurray:
Looks good to me! Merging.

PR #1833, opened 2022-07-21
Atom to approximate the exponential cone: `ExpConeQuad`

aryamanjeendgar:
In this PR, I add support for the scalar relative entropy cone via the approximation method that was defined in Theorem-3 of [THIS paper](https://arxiv.org/abs/1705.00812). 

The sREC also happens to be directly connected (under a permutation and sign change of the arguments) to the exponential cone, and @rileyjmurray added a method for conversion from `ExpCone` to `ExpConeQuad` to that end. 

As it stands now, the biggest shortcoming in the current state of the PR, are two things:
1. Lack of support for elementwise application of `ExpConeQuad` to vector variables --- this is due to the usage of the `quad_over_lin` atom for implementing the semidefinite approximation (currently, one of the test cases has been explicitly modified to make it pass)
2. Lack of proper experimentation for arriving at a sensible set of defaults for the parameters $(m,k)$ in the approximation --- while $m$ is related to the quadrature approximation used in the procedure, $k$ doesn't lend itself to any kind of an obvious intuitive interpretation.

There's also some documentation missing, but I will add that in the next few commits.

Thank you!


cvxpy/reductions/cone2cone/common2common.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Common2Common(Canonicalization):
+
+    CANON_METHODS = {
+        ExpConeQuad: ExpConeQuad_canon
+    }
+
+    def __init__(self, problem=None) -> None:
+        super(Common2Common, self).__init__(
```
SteveDiamond, line 107: why do we have the new common2common for ``ExpConeQuad``?

cvxpy/reductions/cone2cone/common2common.py
```diff
@@ -0,0 +1,108 @@
...
+    w, t = gauss_legendre(m)
+    T = Variable(m)
+    lead_con = Zero(w @ T + con.z/2**k)
+    constrs = [Zero(Z[0] - y)]
+
+    for i in range(k):
+        # The following matrix needs to be PSD.
+        #     [Z[i]  , Z[i+1]]
+        #     [Z[i+1], x     ]
+        expr = quad_over_lin(Z[i+1], x)
```
SteveDiamond, line 82: The two options here are to implement vectorized ``quad_over_lin`` or do a vectorized ``SOC`` representation directly in this block.

cvxpy/reductions/cone2cone/common2common.py
rileyjmurray: @SteveDiamond this was my suggestion. The idea is that ``common2common`` could handle things like "power cone to exponential cone" and "2-by-2 SDP to SOCP". Although, strictly speaking, this is an approximation and not an exact reformulation. So it might make sense to make an ``approximations`` canonicalization path and put this there instead. We'll have more cones like ``OperatorRelativeEntropyCone`` that could go in such a future canonicalization path.

cvxpy/constraints/exponential.py
```diff
@@ -14,13 +14,15 @@
...
 """
+from __future__ import annotations
 
 from typing import List, Tuple
 
 import numpy as np
 
 from cvxpy.constraints.constraint import Constraint
 from cvxpy.expressions import cvxtypes
+# from cvxpy.expressions.constants.constant import Constant
```
rileyjmurray, line 25: Why is this commented out? Remove it if it isn't needed.

cvxpy/reductions/cone2cone/common2common.py
```diff
@@ -0,0 +1,108 @@
+"""
+Copyright 2021 the CVXPY developers
```
rileyjmurray, line 2: 2022

cvxpy/constraints/exponential.py
```diff
@@ -136,3 +141,114 @@ def save_dual_value(self, value) -> None:
...
+        return self.is_dcp()
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        s = (3,) + self.x.shape
+        return s
+
+    def save_dual_value(self, value) -> None:
+        # TODO: implement me.
+        return
```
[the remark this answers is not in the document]
rileyjmurray: The math associated with this task would be nontrivial for an undergrad. If you want to work on this question then I can give you some stepping stones that would make building up to this task less daunting.

aryamanjeendgar:
Just wanted to kick off an important discussion regarding where `ExpConeQuad` should be placed within the web documentation.

So, as it stands right now, semantically, `ExpConeQuad`  pretty much offers an alternative (which can be made arbitrarily accurate by tweaking $(m,k)$) canonicalization pathway to the standard `ExpCone`; hence, I felt that it might be a bit awkward to place it within all the other constraint classes (like `ExpCone`, `SOC`, PowConeND` etc.) since it is _functionally_ redundant to `ExpCone` from a modeling perspective, and instead, expose it through the `as_expconequad(. , .)` method that has been written for converting from `ExpCone` to `ExpConeQuad`.

I would like to hear everyone's thoughts on this! @SteveDiamond @rileyjmurray

rileyjmurray:
> Just wanted to kick off an important discussion regarding where `ExpConeQuad` should be placed within the web documentation.
> 
> So, as it stands right now, semantically, `ExpConeQuad` pretty much offers an alternate (which can be made arbitrarily accurate by tweaking (m,k) ) canonicalization pathway to the standard `ExpCone`; hence, I felt that it might be a bit awkward to place it within all the other constraint classes (like `ExpCone`, `SOC`, `PowConeND` etc.) since it is _functionally_ redundant to `ExpCone` from a modeling perspective, and instead, expose it through the `as_expconequad(. , .)` method that has been written for converting from `ExpCone` to `ExpConeQuad`.

I'd be in favor of retaining an ``ExpConeQuad`` class but not really suggesting that users work with that class. Rather, they should use the ``ExpCone`` class and then call ``con.as_expconequad(m k)``. How about this:

- ``ExpConeQuad`` gets a disclaimer at the top of its docstring saying that we don't promise stability of its API. However, we do promise stability of the API for ``ExpCone.as_expconequad``.
-  The docstring for ``ExpCone.as_expconequad`` contains a meaningful amount of information as to how the approximation works, but it wouldn't be as detailed as the documentation for ``ExpConeQuad``.

If we go in that direction then I think we would expose ``ExpConeQuad`` in [this part of the web docs](https://www.cvxpy.org/api_reference/cvxpy.constraints.html) like every other Constraint class.

rileyjmurray:
@aryamanjeendgar the style check with ``isort`` is failing. Take a look and fix when you get a chance.

cvxpy/constraints/exponential.py
```diff
@@ -136,3 +140,114 @@ def save_dual_value(self, value) -> None:
...
+    Semidefinite Approximations of the Matrix Logarithm.
+
+    Parameters
+    ----------
+    x : Variable
+        x in the exponential cone.
+    y : Variable
+        y in the exponential cone.
+    $\\tau$ : Variable
+        $\\tau$ in the exponential cone.
```
rileyjmurray, line 165: Two points.
1. These might be Expression objects, not necessarily Variable objects.
2. These expressions aren't constrained to lay in the exponential cone. They're constrained to lay in "this approximation of" [the scalar relative entropy cone].

cvxpy/reductions/cone2cone/common2common.py
```diff
@@ -0,0 +1,140 @@
...
+from cvxpy.atoms.quad_over_lin import quad_over_lin
+from cvxpy.constraints.constraint import Constraint
+from cvxpy.constraints.exponential import ExpConeQuad
+from cvxpy.constraints.zero import Zero
+from cvxpy.expressions.variable import Variable
+from cvxpy.reductions.canonicalization import Canonicalization
+
+COMMON_CONES = {
+    ExpConeQuad: {quad_over_lin}
+}
```
rileyjmurray, line 31: Can you remind me what purpose this serves?

cvxpy/reductions/cone2cone/common2common.py
```diff
@@ -0,0 +1,140 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from typing import List, Tuple
+
+import numpy as np
+
+import cvxpy as cp
+from cvxpy.atoms.quad_over_lin import quad_over_lin
```
rileyjmurray, line 22: I don't think we need ``quad_over_lin`` anymore.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -219,12 +221,18 @@ def construct_solving_chain(problem, candidates,
...
+    com_cos = [ct for ct in constr_types if ct in COMMON_CONES]
     # ^ The way we populate "ex_cos" will need to change if and when
     # we have atoms that require exotic cones.
     for co in ex_cos:
         sim_cos = EXOTIC_CONES[co]  # get the set of required simple cones
         constr_types.update(sim_cos)
         constr_types.remove(co)
+
+    for co in com_cos:
+        sim_cos = COMMON_CONES[co]  # get the set of required simple cones
```
rileyjmurray, line 233: Remark: I don't like this naming convention of ``COMMON_CONES``.

cvxpy/tests/test_cone2cone.py
```diff
@@ -483,3 +483,81 @@ def test_pcp_4b(self):
...
+        c = 0.75
+        t = cp.Variable(name='t')
+        x = cp.Variable(shape=(3,), name='x')
+        s = cp.Variable(shape=(3,), name='s')
+        e = cp.Constant(np.ones(3, ))
+        objective = cp.Minimize(t - c * e @ s)
+        con1 = cp.norm(L.T @ x, p=2) <= t
+        con2 = ExpCone(s[0], e[0], x[0]).as_expconequad(5, 5)
+        con3 = ExpCone(s[1], e[1], x[1]).as_expconequad(5, 5)
+        con4 = ExpCone(s[2], e[2], x[2]).as_expconequad(5, 5)
```
rileyjmurray, line 543: The test should use a vectorized constraint.

rileyjmurray (changes requested):
I think we should rename several things:
1. ``common2common.py`` should be renamed to ``approximations.py``.
2. ``ExpConeQuad`` should be renamed to ``RelEntrQuad``. 
3. ``ExpCone.as_expconequad`` should be ``ExpCone.as_quad_approx``.

When making these changes, be sure to update docstrings of functions in the current ``ExpConeQuad`` class accordingly. You should also update variable names used in your changes to ``solving_chain.py`` to reflect the new semantics of ``approximations.py``.

There were a couple of other small points but I think that'll do it for now.

cvxpy/reductions/cone2cone/common2common.py
aryamanjeendgar: This is simply for maintaining a list of `COMMON_CONES`, for instance, we would add `OpRelCone` to this list next --- this list is used for checking membership while constructing the solving chain, like so:

```
approx_cos = [ct for ct in constr_types if ct in APPROX_CONES]
for co in approx_cos:
     # do something
```

cvxpy/reductions/solvers/solving_chain.py
aryamanjeendgar: I was thinking of changing the name to `APPROX_CONES` and `QuadApprox` (from `Common2Common`) --- comments?

cvxpy/tests/test_cone2cone.py
```diff
@@ -538,9 +538,7 @@ def expcone_socp_1(self) -> STH.SolverTestHelper:
         e = cp.Constant(np.ones(3, ))
         objective = cp.Minimize(t - c * e @ s)
         con1 = cp.norm(L.T @ x, p=2) <= t
-        con2 = ExpCone(s[0], e[0], x[0]).as_expconequad(5, 5)
-        con3 = ExpCone(s[1], e[1], x[1]).as_expconequad(5, 5)
-        con4 = ExpCone(s[2], e[2], x[2]).as_expconequad(5, 5)
+        con2 = ExpCone(s, e, x)
```
rileyjmurray, line 541: ``con2 = ExpCone(s, e, x).as_quad_approx(5, 5)``.

cvxpy/tests/test_cone2cone.py
```diff
@@ -549,9 +547,10 @@ def expcone_socp_1(self) -> STH.SolverTestHelper:
         ]
         con_pairs = [
             (con1, 1.0),
-            (con2, np.array([-0.75, -0.75, -0.75])),
-            (con3, np.array([-1.16363, -1.20777, -1.70371])),
-            (con4, np.array([1.30190, 1.38082, 2.67496]))
+            (con2, [np.array([-0.75, -0.75, -0.75]),
+                    np.array([-1.16363, -1.20777, -1.70371]),
+                    np.array([1.30190, 1.38082, 2.67496])]
+             )
```
rileyjmurray, line 553: Once ``con2`` is approximated we won't be able to use a dual variable test here. (You'll put ``None`` instead of this list of arrays.)

rileyjmurray (changes requested):
Mostly looks good! Small changes needed in ``TestRelEntrQuad``.

rileyjmurray (approved):

rileyjmurray:
Approved! @SteveDiamond or @akshayka please review as soon as convenient. Since this changes CVXPY's public API we need more than one maintainer to approve. (We will have follow-up PRs though, so these changes can easily be modified.)

cvxpy/constraints/exponential.py
```diff
@@ -14,6 +14,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from __future__ import annotations
```
SteveDiamond, line 17: Is this necessary?

SteveDiamond (approved):
LGTM

cvxpy/constraints/exponential.py
rileyjmurray: It seems as though it's needed for Python 3.7 - 3.10, at least: https://stackoverflow.com/questions/61544854/from-future-import-annotations.

SteveDiamond: I don't think they're saying it's necessary, unless you have very complex annotations. I know you can do annotations without importing this. Not a big deal of course.

Issue #1834, opened 2022-07-21
diff fails on 2-D arrays with 1 row


sunnyxxchen:
**Describe the bug**
The `diff` command raises the ValueError `"Must have k >= 0 and X must have < k elements along "axis"` when attempting to calculate discrete differences for a 1-by-N array across the row.

**To Reproduce**
```
import cvxpy as cp

x1 = np.array([[1, 2, 3, 4, 5]])
x2 = cp.Variable((1, 5))

cp.diff(x1, axis=1)
cp.diff(x2, axis=1)
```

**Expected behavior**
`cp.diff` should be able to calculate the discrete difference for these examples. The corresponding numpy command works correctly on these examples.

**Version**
 - OS: macOS 12.4
 - CVXPY Version: 1.2.1

SteveDiamond:
This is fixed on master.

PR #1835, opened 2022-07-21
Fix bug with diff

SteveDiamond:
## Description
There was a bug in diff that made it impossible to use on some arrays with dimension 1.
Issue link (if applicable): #1834 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/diff.py
+++ b/cvxpy/atoms/affine/diff.py
@@ -39,7 +39,8 @@ def diff(x, k: int = 1, axis: int = 0):
     elif axis == 1:
         x = x.T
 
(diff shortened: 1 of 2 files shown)


cvxpy/atoms/affine/diff.py
```diff
@@ -39,7 +39,8 @@ def diff(x, k: int = 1, axis: int = 0):
     elif axis == 1:
         x = x.T
 
-    if k < 0 or k >= x.shape[axis]:
+    # Always test shape[0] because if axis == 1 x is transposed.
+    if k < 0 or k >= x.shape[0]:
```
phschiele, line 43: Makes sense :+1:

PR #1836, opened 2022-07-22
Add new QpSolver and ConicSolver for COPT under CVXPY

wujianjack:
## Description
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.


wujianjack:
Hi @SteveDiamond 

This is `wujian` from COPT developer, we added new QpSolver plugin to solve LP/MILP/QP problems under CVXPY, and enhanced ConicSolver plugin to solve LP/MILP/SOCP/QCP/SDP problems under CVXPY.

The minimal requirements for COPT version will be 5.0.2, which is the latest release.

Please review the changes. Thanks

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
```diff
@@ -64,6 +90,59 @@ def accepts(self, problem):
                     return False
         return True
 
+    @staticmethod
+    def psd_format_mat(constr):
```
SteveDiamond, line 94: This is the same as in the SCS interface, right? Can we factor this out into some common code? @phschiele @rileyjmurray where do you think this common code should go?

SteveDiamond:
Looks great overall! The main thing missing is tests of COPT in ``test_qp_solvers.py``.

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
phschiele: Same for `tri_to_full()`. Could go into `conic_solver.py`.

wujianjack:
> Looks great overall! The main thing missing is tests of COPT in `test_qp_solvers.py`.

I tried `pytest test_qp_solvers.py` and works, what should I do for testing COPT QpSolver? Thanks.

SteveDiamond:
> > Looks great overall! The main thing missing is tests of COPT in `test_qp_solvers.py`.
> 
> I tried `pytest test_qp_solvers.py` and works, what should I do for testing COPT QpSolver? Thanks.

Sorry, actually test_qp_solvers.py is fine. I had forgotten that it automatically tests all qp solvers. So the only remaining question in the PR is whether ``psd_format_mat`` and ``tri_to_full`` should be put in conic_solvers, with an argument to distinguish between the SCS usage and the COPT usage. I think that's something we can deal with later potentially.

SteveDiamond (approved):

SteveDiamond:
Actually, @wujianjack it looks like ``test_verbose`` is failing. Did you make it so copt doesn't print anything when ``verbose = False``? If not that's ok too, we can fix that later.

wujianjack:
> Actually, @wujianjack it looks like test_verbose is failing. Did you make it so copt doesn't print anything when verbose = False? If not that's ok too, we can fix that later.

Thanks! @SteveDiamond

We will investigate on this, looks like the problem data was not loaded correctly, and so we print out `coptcore.CoptError: Fail to load cone problem by matrix`, even when `verbose=False`.

wujianjack:
> Actually, @wujianjack it looks like `test_verbose` is failing. Did you make it so copt doesn't print anything when `verbose = False`? If not that's ok too, we can fix that later.

The SDP problem tested in `test_verbose` has PSD variable with dimension of 1, and currently we will check if dimension of PSD variable is at least 2, that's why we throw error `coptcore.CoptError: Fail to load cone problem by matrix`, maybe we should change the way we deal with dimension of PSD variable.

SteveDiamond:
> > Actually, @wujianjack it looks like `test_verbose` is failing. Did you make it so copt doesn't print anything when `verbose = False`? If not that's ok too, we can fix that later.
> 
> The SDP problem tested in `test_verbose` has PSD variable with dimension of 1, and currently we will check if dimension of PSD variable is at least 2, that's why we throw error `coptcore.CoptError: Fail to load cone problem by matrix`, maybe we should change the way we deal with dimension of PSD variable.

From our perspective, it would be easier if you changed your SDP interface. Otherwise we'll need to add special logic on our side to catch such problems. But again we can sort this all out later. For the moment we can add COPT to the list of solvers that don't pass the verbose test and then merge.

wujianjack:
> > > Actually, @wujianjack it looks like `test_verbose` is failing. Did you make it so copt doesn't print anything when `verbose = False`? If not that's ok too, we can fix that later.
> > 
> > 
> > The SDP problem tested in `test_verbose` has PSD variable with dimension of 1, and currently we will check if dimension of PSD variable is at least 2, that's why we throw error `coptcore.CoptError: Fail to load cone problem by matrix`, maybe we should change the way we deal with dimension of PSD variable.
> 
> From our perspective, it would be easier if you changed your SDP interface. Otherwise we'll need to add special logic on our side to catch such problems. But again we can sort this all out later. For the moment we can add COPT to the list of solvers that don't pass the verbose test and then merge.

Yes, I agree. We should take care of this on the solver side. We will fix the underlying issues in next patch release. The interface will stay the same, so if it's okay, we can merge it.

Thanks again! @SteveDiamond

SteveDiamond:
@wujianjack the latest COPT release is failing our tests: 
https://github.com/cvxpy/cvxpy/actions/runs/4708770183/jobs/8351417182?pr=2105

phschiele:
@wujianjack @SteveDiamond I narrowed it down to the recovery of the dual solution.

For example, for `test_copt_socp_1`, the previous COPT version returned
```
[-0.779, -2.865, -2.865, -2.220, -1.220, 1.338]
```
from
https://github.com/cvxpy/cvxpy/blob/8439878589565be7574cd811f592eefb2e385187/cvxpy/reductions/solvers/conic_solvers/copt_conif.py#L339
whereas the new version returns:
```
[-0.779, 0.0, 0.0, -2.220, -1.220, 1.337]
```

rileyjmurray:
Does the new solution satisfy complementary slackness?

phschiele:
> Does the new solution satisfy complementary slackness?

@rileyjmurray No, that's exactly the assertion that is failing.
https://github.com/cvxpy/cvxpy/blob/8439878589565be7574cd811f592eefb2e385187/cvxpy/tests/solver_test_helpers.py#L133

wujianjack:
> @wujianjack the latest COPT release is failing our tests: https://github.com/cvxpy/cvxpy/actions/runs/4708770183/jobs/8351417182?pr=2105

Thanks @SteveDiamond , we will check and fix it soon. Look like a bug internally, not the cvxpy interface side.

wujianjack:
> @wujianjack @SteveDiamond I narrowed it down to the recovery of the dual solution.
> 
> For example, for `test_copt_socp_1`, the previous COPT version returned
> 
> ```
> [-0.779, -2.865, -2.865, -2.220, -1.220, 1.338]
> ```
> 
> from
> 
> https://github.com/cvxpy/cvxpy/blob/8439878589565be7574cd811f592eefb2e385187/cvxpy/reductions/solvers/conic_solvers/copt_conif.py#L339
> 
> whereas the new version returns:
> 
> ```
> [-0.779, 0.0, 0.0, -2.220, -1.220, 1.337]
>          ^^^^^^^^
> ```

Thanks @phschiele, we will check and fix it soon.

wujianjack:
@SteveDiamond @phschiele Thanks for reporting this, we have fixed the issue in our latest COPT 6.5.2 release and verified no more failures for the test sets of cvxpy, please let me know if there are any further issues.

Thanks again!

Issue #1521, opened 2021-10-29
Slow compilation of standard form LP (time increases linearly in number of variables)


akshayka:
Compilation time of a standard form LP (minimize c @ x, subject to x >= 0) increases linearly with the number of variables (for both OSQP and SCS).

The slowdown occurs regardless of whether c is sparse (indeed, empty) or dense.

**To Reproduce**
```python3
import time                                                                      
                                                                                 
import cvxpy as cp                                                               
import matplotlib.pyplot as plt                                                  
import numpy as np                                                               
import scipy.sparse as sparse                                                    
from tqdm.notebook import tqdm                                                   
                                                                                 
                                                                                 
def construct_lp(n):                                                             
    c = sparse.csr_matrix((n, 1))  # timings are similar if c is a dense numpy array
    x = cp.Variable(n)                                                           
    return cp.Problem(cp.Minimize(c.T @ x), [x >= 0])                            
                                                                                 
n_range = list(map(int, np.logspace(1, 8, 8)))                                   
solvers = {"OSQP": [], "SCS": []}                                                
for solver in solvers:                                                           
    compile_times = solvers[solver]                                              
    for n in tqdm(n_range):                                                      
        problem = construct_lp(n)                                                
        start = time.time()                                                      
        problem.get_problem_data(solver, verbose=True)                           
        end = time.time()                                                        
        print(end - start)                                                       
        compile_times.append(end - start)                                        
                                                                                 
for solver in solvers:                                                           
    compile_times = solvers[solver]                                              
    plt.plot(n_range, compile_times, label=solver, marker='o')                   
    plt.xscale('log')                                                            
    plt.yscale('log')                                                            
plt.legend()                                                                     
plt.show()  
```

**Expected behavior**
Should compilation time really be linear in the number of variables? As a user, I would think that the data (c) should just be copied in. It seems strange that compilation is taking much longer than the time needed to generate the problem data (given that this problem is essentially in standard form).

More generally we should strive for standard form problems to compile quickly.

**Output**
![image](https://user-images.githubusercontent.com/1994308/139471547-6019dc15-7772-464a-8ac1-ac4748cf21a9.png)
x-axis is the size of the variable, y-axis is seconds taken to compile.

**Version**
 - CVXPY Version: 1.1.15

**To get started with helping out**
We should profile this script to understand why compilation is slow.

phschiele:
@akshayka I did a quick profile of the script, see below.

[...]

PR #1840, opened 2022-07-22
WIP: New SciPy backend

phschiele:
## Description
This PR attempts to bring a new backend to CVXPY that no longer relies on CVXCORE.
I.e., the code in `lin_ops/parser.py` should contain all functionality currently in the C++ code.
Preliminary testing shows that it is usually faster when no/few parameters are involved. 
For problems with many parameters, the stacking of 2D sparse matrices is slow. A follow-up with ND-sparse libraries is in progress.
The main idea is to make this feature `opt-in` now to get more feedback on its performance in different applications.
Issue link (if applicable): #1521

Some questions @cvxpy/maintainers:
1. What's the best way to pipe the backend choice through? Obviously, the current implementation is just for demonstration
2. What about testing? For development, I added a special flag that would test all backends and ensure the result is the same across backends, but I was not sure if we would like to have such a flag exposed.

TODOs:
- [x] Documentation
- [x] Remove some code duplication
- [x] Unit tests
- [x] Regression tests (i.e., running all tests under the new backend)

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
I would expose the backend similarly to how we expose solvers, as an argument to solve.

Maybe you could add another test file that focuses on comparing backends?

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,743 @@
...
+    @property
+    @abstractmethod
+    def rows(self) -> int:
+        pass
+
+    @abstractmethod
+    def get_A_b(self):
+        """
+        Returns [A b]
+        """
```
SteveDiamond, line 613: should this be ``pass``?

SteveDiamond:
The algorithm looks fine to me. The code is very concise, cleaner but still similar to what's in cvxcore. I wonder if we should run all the tests against each of the backends rather than writing new tests that explicitly vary the backend?

phschiele:
@SteveDiamond We can have the full python/os matrix run on the default backend, and have one config run the other backends.

cvxpy/lin_ops/parser.py
phschiele, line 613:
```suggestion
        """
        pass
```

phschiele: Done

phschiele: Actually, @SteveDiamond, this was automatically removed by `autoflake`, because the function has a docstring.
Flake8 itself did not complain, however, so we might reconsider using that tool.

SteveDiamond: yeah, maybe best to remove autoflake.

phschiele:
For problems not containing any parameters, the improvements are:
cvar_benchmark: 15%
optimal_advertising: 74%
sdp_segfault_1132_benchmark: 85%
SimpleLPBenchmark: 5%

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,776 @@
...
+
+    @classmethod
+    def combine(cls, tensors: list[TensorRepresentation]):
+        parameter_offset, row, col, data = np.array([]), np.array([]), np.array([]), np.array([])
+        for t in tensors:
+            parameter_offset = np.append(parameter_offset, t.parameter_offset)
+            row = np.append(row, t.row)
+            col = np.append(col, t.col)
+            data = np.append(data, t.data)
+        return cls(parameter_offset, row, col, data)
```
rileyjmurray, line 48: It might be faster to use lists instead of numpy arrays. A numpy array needs to be contiguous in-memory, while a list doesn't have that requirement. (Of course you'd convert the lists to arrays before returning.)

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,776 @@
...
+        view.select_rows(triu_indices)
+        return view
+
+    @staticmethod
+    def diag_mat(lin: LinOp, view: TensorView) -> TensorView:
+        # diagonal matrix to vector
+        rows = lin.shape[0]
+        diag_indices = np.arange(rows) * rows + np.arange(rows)
+        view.select_rows(diag_indices)
+        return view
```
rileyjmurray, line 281: Are these implementations supposed to be used for "all" backends?

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,776 @@
...
+    def get_param_tensor(self, shape, parameter_id):
+        pass
+
+
+class ScipyBackend(Backend):
+
+    @staticmethod
+    def reshape_constant_data(constant_data: dict[int, sp.csr_matrix], new_shape: tuple[int]):
+        return {k: [v_i.reshape(new_shape[1:], order="F")
+                    for v_i in v] for k, v in constant_data.items()}
```
rileyjmurray, line 322: Can this be made more "flat" instead of nested?

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,776 @@
...
+
+    def concatenate_tensors(self, tensors: list[tuple[TensorRepresentation, int]]):
+        for tensor, row_offset in tensors:
+            tensor.row += row_offset
+        return TensorRepresentation.combine([t[0] for t in tensors])
+
+    def reshape_tensors(self, tensor: TensorRepresentation, total_rows: int) -> sp.csc_matrix:
+        # Windows uses int32 by default at time of writing, so we need to enforce int64 here
+        rows = (tensor.col.astype(np.int64) * np.int64(total_rows) + tensor.row.astype(np.int64))
+        cols = tensor.parameter_offset.astype(int)
```
rileyjmurray, line 332: Is there a reason why ``rows`` needs to be ``int64`` but ``cols`` can just be ``int``?

cvxpy/lin_ops/parser.py
```diff
@@ -0,0 +1,776 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def reshape_constant_data(constant_data, new_shape: tuple[int]):
+        pass
+
+    @abstractmethod
+    def concatenate_tensors(self, tensors: list[tuple[TensorView, int]]) -> TensorView:
+        """
+        Takes list of tensors and stacks them "row-wise"
```
rileyjmurray, line 135: What's meant by "row-wise"? If I have tensors of shape ``(x, y, z)`` and ``(p, q, ell)``, do we require ``(y, z) == (q, ell)`` and return a tensor of shape ``(x + p, y, z)``?

cvxpy/problems/problem.py
```diff
@@ -610,6 +610,11 @@ def get_problem_data(
         cvxpy.error.DPPError
             Raised if DPP settings are invalid.
         """
+
+        # TODO: remove
```
rileyjmurray, line 614: Please make this TODO clearer.

rileyjmurray (changes requested):
I left a few specific questions. Overall, I think it's important that nontrivial documentation be written for these functions. I'll ask that classes and their main methods to be documented at a similar level of reductions in this file: https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/cone2cone/affine2direct.py. In my ideal world the individual functions like ``ScipyBackend.mul`` would describe the mathematical form of their output.

cvxpy/lin_ops/parser.py
rileyjmurray: To clarify: I'd like to be able to use the debugger in this function. Right now I can't set any useful breakpoints. Having explicit for-loops makes it much easier to debug.

cvxpy/lin_ops/parser.py
phschiele: @rileyjmurray Great spot! I had written as you described initially for this very reason, but did some benchmarks as it became one of the bottlenecks. To my surprise, the above method was the fasted one in my testing.
I used actual problem data back then, but here is an example of what I tested:
```py
import numpy as np
A = [np.arange(i,i+1000) for i in range(20)]
```
Lists:
```py
%%timeit
res = []
for a in A:
    res.extend(a)
np.array(res)  
>>> 1.29 ms ± 14.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
`np.append`
```py
%%timeit
res = np.array([])
for a in A:
    np.append(res, a)
>>> 76.3 µs ± 181 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
Perhaps a comment could be added that, for our use case, this is faster than using lists?

cvxpy/lin_ops/parser.py
phschiele: Precisely, given `select_rows` is an abstract method that all backends should support, this approach resulted in the least duplication (when I added additional backends locally).

cvxpy/lin_ops/parser.py
phschiele: Fair point, I only changed this after tests failed. Generally, I try using native python types.
This only failed under windows btw, so I had to be specific here([reason](https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine)).
`rows` is of the order `A.size`, whereas `cols` is of the order `parameters.size`.
I will change it to be consistent within this function, though.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+        parameter_offset, row, col, data = np.array([]), np.array([]), np.array([]), np.array([])
+        for t in tensors:
+            parameter_offset = np.append(parameter_offset, t.parameter_offset)
+            row = np.append(row, t.row)
+            col = np.append(col, t.col)
+            data = np.append(data, t.data)
+        return cls(parameter_offset, row, col, data)
+
+
+class CanonBackend(ABC):
```
akshayka, line 59: Can you add a short docstring that describes what functionality this class provides, and explains the semantics of its attributes?

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+    @abstractmethod
+    def kron_r(self, lin: LinOp, view: TensorView) -> TensorView:
+        pass
+
+    @abstractmethod
+    def kron_l(self, lin: LinOp, view: TensorView) -> TensorView:
+        pass
+
+    @abstractmethod
+    def get_variable_tensor(self, shape: tuple[int], variable_id: int):
```
akshayka, line 313: Can you document this method, as well as `get_data_tensor` and `get_param_tensor`.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+    def get_param_tensor(self, shape, parameter_id):
+        shape = int(np.prod(shape))
+        slices = []
+        for idx in np.arange(shape):
+            slices.append(sp.csr_matrix(((np.array([1.])), ((np.array([idx])),
+                                                            (np.array([0])))), shape=(shape, 1)))
+        return {parameter_id: slices}
+
+
+class TensorView(ABC):
```
akshayka, line 597: Document the functionality of this class and the semantics of its attributes, please.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+                                             if i is not None])
+
+    def select_rows(self, rows: np.ndarray) -> None:
+
+        def func(x):
+            return x[rows, :]
+
+        self.apply_all(func)
+
+    def apply_all(self, func: Callable) -> None:
```
akshayka, line 692: Can you document the arguments to and semantics of this method, as well as `accumulate_over_variables` and `apply_to_parameters`?

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+        if self.constant_data is not None:
+            self.constant_data = {k: [func(v_i).tocsr() for v_i in v]
+                                  for k, v in self.constant_data.items()}
+
+    def create_new_tensor_view(self, variable_ids, tensor, is_parameter_free: bool):
+        return ScipyTensorView(variable_ids, tensor, is_parameter_free, self.param_size_plus_one,
+                               self.id_to_col, self.param_to_size, self.param_to_col,
+                               self.var_length)
+
+    def accumulate_over_variables(self, func, is_param_free_function: bool):
```
akshayka, line 706: Are this method and the function `apply_to_parameters` specific to `ScipyTensorView`? I noticed that they are not declared as members of `TensorView`.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+            return None
+        elif a is not None and b is None:
+            return a
+        elif a is None and b is not None:
+            return b
+        else:
+            return ScipyTensorView.add_dicts(a, b)
+
+    @staticmethod
+    def add_dicts(a, b):
```
akshayka, line 768: Docstring, and add type annotations?

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+        shape = int(np.prod(shape))
+        slices = []
+        for idx in np.arange(shape):
+            slices.append(sp.csr_matrix(((np.array([1.])), ((np.array([idx])),
+                                                            (np.array([0])))), shape=(shape, 1)))
+        return {parameter_id: slices}
+
+
+class TensorView(ABC):
+    def __init__(self, variable_ids: set[int] | None, tensor, is_parameter_free: bool,
```
akshayka, line 598: `tensor` is missing type annotation

akshayka (commented):
This looks great! Like Riley, I've asked for documentation in a few key places. It's great that functions and methods have type annotations, though I noticed type annotations are missing in just a few places; it would help legibility to annotate all arguments and return types.

akshayka:
> I would expose the backend similarly to how we expose solvers, as an argument to solve.

If we do this, we should have a plan to deprecate the `CPP` backend in the future. It would be painful to have to maintain two implementations of `build_matrix`.

SteveDiamond:
> > I would expose the backend similarly to how we expose solvers, as an argument to solve.
> 
> If we do this, we should have a plan to deprecate the `CPP` backend in the future. It would be painful to have to maintain two implementations of `build_matrix`.

The two backends have different performance characteristics so unfortunately we have to roll with both for a while. It's definitely not possible to switch the default backend without much more testing.

cvxpy/lin_ops/parser.py
phschiele: @rileyjmurray Take a look at cfef17d, specifically `test_transpose()`, `test_upper_tri()`, and `test_diag_mat()`. They all use the row selection logic from the base class, so the SciPy backend only has to implement `select_rows`.
I will add a shortened version of the test description as a doc string for the functions.

phschiele:
The SDP example is significantly slower under the new backend, even though it does not use parameters.
This is worth looking into, but as the current suggestion is to add the new backend as experimental/non-default, this should not be a blocker.

```
 before           after         ratio
     [d15197bf]       [e3576451]
+      7.27±0s          41.3±0s     5.69  semidefinite_programming.SemidefiniteProgramming.time_compile_problem
```

Thanks @parthb83 for this recent addition to the benchmark suite!

cvxpy/lin_ops/canon_backend.py
```diff
@@ -0,0 +1,785 @@
...
+    Sparse representation of a 3D Tensor. Semantically similar to COO format, with one extra
+    dimension. Here, 'row' is axis 0, 'col' axis 1, and 'parameter_offset' axis 2.
+    """
+    parameter_offset: np.ndarray
+    row: np.ndarray
+    col: np.ndarray
+    data: np.ndarray
+
+    @classmethod
+    def combine(cls, tensors: list[TensorRepresentation]):
```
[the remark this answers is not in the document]
phschiele: This is possible now, thanks to `from __future__ import annotations` (i.e. adding the return type `TensorRepresentation` within the class `TensorRepresentation`.

cvxpy/lin_ops/parser.py
rileyjmurray: Wonderful! Thanks for clarifying.

cvxpy/problems/problem.py
```diff
@@ -589,6 +591,10 @@ def get_problem_data(
         ignore_dpp : bool, optional
             When True, DPP problems will be treated as non-DPP,
             which may speed up compilation. Defaults to False.
+        canon_backend : str, optional
+            specifies which backend to use for canonicalization, which can affect
+            performance. Defaults to None, i.e., selecting the default
```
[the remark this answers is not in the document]
rileyjmurray: Sounds good to me!

cvxpy/problems/problem.py
phschiele: TODO has been removed.

cvxpy/lin_ops/canon_backend.py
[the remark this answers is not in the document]
phschiele: Done.

cvxpy/lin_ops/canon_backend.py
phschiele: Added doc strings to all functions in the file.

cvxpy/lin_ops/canon_backend.py
phschiele: Done.

cvxpy/lin_ops/canon_backend.py
phschiele: Done.

cvxpy/lin_ops/canon_backend.py
phschiele: Done.

cvxpy/lin_ops/canon_backend.py
phschiele: Done.

cvxpy/lin_ops/canon_backend.py
phschiele: Thanks, added it.

cvxpy/lin_ops/canon_backend.py
phschiele: Great spot. Indeed, this function is not called by any superclass function, so I did not add it as an abstractmethod. 
For example, I have a local experimental "NumpyTensorView" (for debugging purposes only), which does not have this function.

cvxpy/problems/problem.py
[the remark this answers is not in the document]
phschiele: Done.

cvxpy/problems/problem.py
```diff
@@ -940,6 +952,10 @@ def _solve(self,
...
+            backend.
```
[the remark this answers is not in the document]
phschiele: Done.

cvxpy/lin_ops/parser.py
phschiele: Added a `# noqa` for now to not pollute this PR with a potential removal of autoflake.

cvxpy/lin_ops/parser.py
phschiele: Done.

cvxpy/lin_ops/parser.py
phschiele: I have clarified row-wise to mean "axis 0". Your example is exactly what this function does, so it aligns with familiar concatenation functions from numpy, etc. .

cvxpy/lin_ops/parser.py
phschiele: I tried to use dict / list comprehensions due to their slight performance benefits. I agree that this makes it slightly harder to debug, though. I did set a breakpoint at the return statement and called `next()` on the dict to get the first key, value pair. Then I applied the inner function to the first element of the value (list) for debugging. One can also quickly break up the comprehensions for debugging, if necessary. 
Since performance is a major consideration for this part of the code, I would suggest keeping the comprehensions, even if it makes debugging slightly more cumbersome. If you have a strong view otherwise, we can also change it.

phschiele:
@cvxpy/maintainers Many thanks for your detailed reviews! I have addressed all of your comments above. 
In particular, I have added:
- Extensive tests, with unit tests covering all major functions of the backend, and regression tests covering every line containing logic. Note: the coverage will not show correctly on sonar, as the coverage run uses the CPP backend.
- A new workflow that runs all tests on the new backend. This workflow is not executed as part of the PR yet, so we only see if it works as expected when we merge and potentially need to readjust if it doesn't.
- All functions have short docstrings and complete type annotations now. The docstrings are always in the base class to not repeat them for each subclass. They should get passed down on most IDEs e.g. when looking up a function.
- Added a very short web documentation section just stating the semantics. We can expand this and give more details if you think it makes sense. We can also consider adding it to the "News" section of the docs, since the performance impact can be major.

doc/source/tutorial/advanced/index.rst
```diff
@@ -1581,3 +1581,14 @@ you should set the class variable ``MIP_CAPABLE`` to ``True``. If your solver is
 and a conic solver (as opposed to a QP solver), you should set the class variable ``MI_SUPPORTED_CONSTRAINTS`` 
 to the list of cones supported when solving mixed integer problems. Usually ``MI_SUPPORTED_CONSTRAINTS`` 
 will be the same as the class variable ``SUPPORTED_CONSTRAINTS``.
+
+Canonicalization backends
+------------------------------------
+Users can select from multiple canonicalization backends by adding the ``canon_backend``
+keyword argument to the ``.solve()`` call, e.g. ``problem.solve(canon_backend="SCIPY")`` (Introduced in CVXPY 1.3).
```
SteveDiamond, line 1588: Usually we add a global in settings.py for any string like "SCIPY"

SteveDiamond:
Thanks for wrapping this up! Definitely put something in the news section!

doc/source/tutorial/advanced/index.rst
phschiele: I did, and I use that string internally as `s.SCIPY_CANON_BACKEND`. I will also add it to the top level `__init__` to be consistent with the solver options.

phschiele: Added here [118ee3c](https://github.com/cvxpy/cvxpy/pull/1840/commits/118ee3c5e1cd39effabe7f0715b3e7a2a56d7f7b)

akshayka (approved):

akshayka:
Looks great! What a Herculean PR. Congrats Philipp for getting this through.

SteveDiamond:
Thanks @phschiele !!!! This is going to make a huge difference for cvxpy users, and in general open up a whole new area of investigation. I looked through everything again and it all looks great! The tests are very helpful for documenting the details of how all the different linops are handled.

PR #1847, opened 2022-07-29
Perspective transform of Huber function 

jcbedoyam:
## Description

Added new atom for perspective transform of Huber function. This is equivalent to CVX Huber function with three arguments [#1377](https://github.com/cvxpy/cvxpy/issues/1377)

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/atoms/elementwise/huber_pers.py
```diff
@@ -0,0 +1,124 @@
+"""
+Copyright
```
rileyjmurray, line 2: Copyright 2022, the CVXPY authors

cvxpy/reductions/dcp2cone/atom_canonicalizers/huber_pers_canon.py
```diff
@@ -0,0 +1,46 @@
+"""
+Copyright 2013 Steven Diamond
```
rileyjmurray, line 2: Copyright 2022, the CVXPY authors

cvxpy/reductions/qp2quad_form/atom_canonicalizers/huber_pers_canon.py
```diff
@@ -0,0 +1,46 @@
+"""
+Copyright 2022 CVXPY.
```
rileyjmurray, line 2: Copyright 2022, the CVXPY authors

cvxpy/tests/test_atoms.py
```diff
@@ -684,6 +684,66 @@ def test_vec_to_upper_tri(self) -> None:
         A_actual = vec_to_upper_tri(a, strict=True).value
         assert np.allclose(A_actual, A_expect)
 
+    def test_huber_pers(self) -> None:
```
rileyjmurray, line 687: I'd rename this to ``test_huber_pers_properties`` and create a new function called ``test_huber_pers_canon``. The latter function would provide system-level tests to verify that cvxpy is correctly canonicalizing the 3-argument Huber function. I can suggest system-level tests if you'd like.

rileyjmurray (changes requested):
Hi @jcbedoyam. Thanks for this contribution! You were very thorough in implementing your intended functionality and integrating it with cvxpy. However, I think if we have  a "perspective of Huber function" atom then we should be able to have the "perspective parameter" ``t`` be a non-constant expression. If we restrict ``t`` to be constant then we can trivially produce perspective transforms using code like the following.
```
def perspective(atom, var_args, t):
    # var_args should be a list with positional arguments suitable for calling ``atom(*var_args)``
    t = cp.Expression.cast_to_const(t)
    assert t.is_constant()
    assert t.size == 1
    if isinstance(t, cp.Parameter):
        assert t.is_pos()
    else:
        assert t.value > 0
    scaled_var_args = [arg / t for arg in var_args]
    expr = t*atom(*scaled_var_args)
    return expr
```

rileyjmurray:
@jcbedoyam I don't think it's possible to implement the perspective of the Huber function using the ``qp2quad_form`` reduction path. That is -- I think that the perspective of the Huber function can only be supported by solvers that allow second-order-cone constraints.

cvxpy/tests/test_atoms.py
jcbedoyam: Thanks, I'll gladly receive your suggestions of the system-level tests.

rileyjmurray:
My bandwidth is pretty limited right now so I can't guide this PR process in a timely way. @aryamanjeendgar, could you help with this? You have quite a bit of experience writing new atoms :)

rileyjmurray:
Ping @aryamanjeendgar; see my comment above.

rileyjmurray:
Actually, this will mostly be handled by #1867. It's best to wait until that's done before working on this any further, since once it's done this will just be a matter of making a small wrapper.

ericlux:
While #1867 introduces the perspective of an expression with respect to its *variables*, it is not pre-composable. That is if you form the perspective of some expression $f(x)$, $p(x,s) = sf(x/s)$, you will not be able to compose $p(g(x),s)$ for example.

rileyjmurray:
This is subsumed by #1867, a feature that will appear in CVXPY 1.3.

Issue #1827, opened 2022-07-14
deepcopy leads to inconsistent behavior (RuntimeError)


codeterm:
**Describe the bug**
Hi team, when I make a deep copy of the constraints (and/or the objective function) and use the copy the create the problem,  RuntimeError is raised in a simple case.

**To Reproduce**
The snippet is the first example on the CVXPY main website with a one-line change:

import cvxpy as cp

x = cp.Variable()
y = cp.Variable()

constraints = [x + y == 1,
               x - y >= 1]

#************************************************
import copy
constraints = copy.deepcopy(constraints)
#************************************************

obj = cp.Minimize((x - y)**2)

prob = cp.Problem(obj, constraints)
prob.solve()  # Returns the optimal value.
print("status:", prob.status)
print("optimal value", prob.value)
print("optimal var", x.value, y.value)

**Expected behavior**
deepcopy shouldn't change the behavior, so the result should have been the same as the original example (with minor floating-point error):
status: optimal
optimal value 0.999999999761
optimal var 1.00000000001 -1.19961841702e-11

**Output**
RuntimeError                              Traceback (most recent call last)                                                                                                                                                                                             
Input In [167], in <cell line: 22>()                                                                                                                                                                                                                                    
     20 # Form and solve problem.                                                                                                                                                                                                                                       
     21 prob = cp.Problem(obj, constraints)                                                                                                                                                                                                                             
---> 22 prob.solve()  # Returns the optimal value.                                                                                                                                                                                                                      
     23 print("status:", prob.status)                                                                                                                                                                                                                                   
     24 print("optimal value", prob.value)                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                        
File /opt/anaconda3/envs/qm/lib/python3.10/site-packages/cvxpy/problems/problem.py:481, in Problem.solve(self, *args, **kwargs)                                                                                                                                         
    479 else:                                                                                                                                                                                                                                                           
    480     solve_func = Problem._solve                                                                                                                                                                                                                                 
--> 481 return solve_func(self, *args, **kwargs)                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                        
File /opt/anaconda3/envs/qm/lib/python3.10/site-packages/cvxpy/problems/problem.py:1007, in Problem._solve(self, solver, warm_start, verbose, gp, qcp, requires_grad, enforce_dpp, ignore_dpp, **kwargs)                                                                
   1004         self.unpack(chain.retrieve(soln))
[...]

PR #1852, opened 2022-07-31
Allow deecopy of constraints

phschiele:
## Description
Always compare `expression.id` instead of (implicitly) `hash(expression)` for variables and parameters.
Issue link (if applicable): #1827 

TODO:
A question remains if cvxpy should always be two instances to be the same if they have the same `.id`.
E.g. what is the expected behavior of the following (a similar example can be constructed for parameters):

```py
    x = cp.Variable(nonneg=True)
    y = copy.deepcopy(x)

    obj = cp.Minimize((x + y))
    prob = cp.Problem(obj)
    prob.solve()
```
There are also the `.copy()` functions, which do not deepcopy the leaves, presumably for this very reason.
Should `deepcopy(expression)` be generally discouraged? 

## Type of change
- [x] Bug fix

- [x] Write unittests.


SteveDiamond (approved):

SteveDiamond:
I would say in your example x and y are the same variable. cvxpy expressions are supposed to be immutable, so I can't see any benefit or purpose of deepcopy over regular copy, but if people want to deepcopy for some reason it should work. So the PR is valuable.

phschiele:
@SteveDiamond Thanks, I agree that there should not be a reason to deepcopy. Just to prevent introducing subtle bugs here, the following problem:
```py
    x = cp.Variable(nonneg=True)
    y = copy.copy(x)

    obj = cp.Minimize((x + y))
    prob = cp.Problem(obj, [y >= 0.75])
    prob.solve()
```
Would result in 
`obj.value = 1.75`, `x.value = 0.75`, `y.value = 0.75` before this PR, but
`obj.value = None`, `x.value = 0.75`, `y.value = None` after this PR, irrespective if copy or deepcopy was used, which is probably not good.

phschiele:
An alternative would be to override `__copy__` and `__deepcopy__` to fail and point to the `.copy()` function instead.

SteveDiamond:
> An alternative would be to override `__copy__` and `__deepcopy__` to fail and point to the `.copy()` function instead.

That sounds much better. We put a lot of work into `.copy()`. I agree that getting obj.value = None, x.value = 0.75, y.value = None above is not desirable behavior.

adishavit:
Cc: @michael-123123 
IIRC, he had done some work on cloning variables.

rileyjmurray:
Ping @phschiele what's the status on this?

phschiele:
@SteveDiamond @rileyjmurray Changed to forward `__copy__` to `.copy()` and fail on `__deepcopy__`.

SteveDiamond (approved):

phschiele:
@SteveDiamond are you waiting for another review, or is this PR good to merge?

SteveDiamond:
We can merge.

Issue #1605, opened 2022-01-14
Target quadratic objectives when calling SCS


bodono:
SCS has native support for [quadratic objectives](https://www.cvxgrp.org/scs/index.html) now. If a problem has a quadratic objective then solving it directly is often much, much faster than converting to a second-order cone program.

rileyjmurray:
Very much in favor of this. It would require creating a new standard form. Right now we have cone programs (linear) and QPs (with only LP-representable constraints). Perhaps a GSoC project, @akshayka @SteveDiamond, @bstellato?

jakeknigge:
Hi @rileyjmurray and @SteveDiamond, I'd traded a couple emails with @bodono about contributing to SCS (I’ve benefitted from using CVX{PY, R, ""} [where CVX"" = CVX] and the related ecosystem for years and I'd like to start paying my CVXGRP dues). He suggested that this issue could be an interesting place to start—and I agree! It looks like a fun one, touching a variety of places in the stack. Let me know what you have in mind and how we can proceed. Thanks!

SteveDiamond:
Hi @jakeknigge ! Can you join the discord (https://discord.gg/4urRQeGBCr) and reach out to us there? Then we can discuss approaches in more detail.

SteveDiamond:
@jakeknigge could you email me? I'd like to chat about how things are going with this, since we have a new solver that requires this interface.

SteveDiamond:
Here is an outline for how this might be implemented:

1. Modify ``ConeMatrixStuffing`` so that it can take a quadratic objective.
2. Modify ``ParamConeProg`` so that it can take a quadratic objective (like ``ParamQuadProg``).
3. Add a ``quad_objective`` boolean flag to ``ConicSolver``. This will be false by default and true for SCS and other conic solvers that support quadratic objectives.
4. Modify ``dcp2cone`` so that it will preserve/canonicalize quadratic objectives if the ``quad_objective`` flag is true.

Basically I think it's better to extend the definition of a conic solver to include a quadratic objective, rather than write a new set of reductions. We'll have to be careful that there isn't a performance regression for the standard (non-quadratic objective) conic reduction path, but we can worry about that after the basic functionality is working.

If we go this route, then the QP reduction path might become superfluous. We can think about that later.

PR #1856, opened 2022-08-04
conic constraints + quadratic objective

SteveDiamond:
## Description
This PR adds support for the quadratic objective + conic constraints standard form currently supported by SCS.

Main changes:
- Updated cone_matrix_stuffing and dcp2cone to handle quadratic objectives.
- Added ``has_quadratic_term`` to check if an objective expression is compatible with the quadratic objective + conic constraints standard form.
- Added some logic to construct_solving_chain to follow the new canonicalization path when appropriate.

Canonicalization is only altered for problems that use SCS (or another conic solver that supports quadratic objectives) and have an objective with a quadratic term. In theory all conic problems could use the new canonicalization path, but I didn't want to risk performance regressions. I also added a new solver argument for SCS that can force cvxpy to use the old canonicalization path.

Issue link (if applicable): #1605 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/reductions/qp2quad_form/atom_canonicalizers/__init__.py
```diff
@@ -25,6 +25,7 @@
 from cvxpy.transforms.indicator import indicator
 
 CANON_METHODS = {}
+QUAD_CANON_METHODS = {}
```
rileyjmurray, line 28: If these variable names are going to be used then there should be comments that disambiguate them.

cvxpy/reductions/qp2quad_form/atom_canonicalizers/__init__.py
```diff
@@ -42,7 +43,9 @@
...
-CANON_METHODS[quad_over_lin] = quad_over_lin_canon
-CANON_METHODS[power] = power_canon
-CANON_METHODS[huber] = huber_canon
-CANON_METHODS[QuadForm] = quad_form_canon
+QUAD_CANON_METHODS[quad_over_lin] = quad_over_lin_canon
+QUAD_CANON_METHODS[power] = power_canon
+QUAD_CANON_METHODS[huber] = huber_canon
+QUAD_CANON_METHODS[QuadForm] = quad_form_canon
+
+CANON_METHODS.update(QUAD_CANON_METHODS)
```
rileyjmurray, line 51: This makes it look like ``QUAD_CANON_METHODS`` doesn't actually need to be defined. Do you reference it later?

cvxpy/reductions/utilities.py
```diff
@@ -80,3 +82,74 @@ def group_constraints(constraints):
...
+    """Utility class for condensing the mapping from parameters to problem data.
+
+    Attributes
+    ----------
+    matrix_data : SciPy sparse matrix
+       A matrix representing the mapping from parameter to problem data.
+    var_len : int
+       The length of the problem variable.
+    quad_form: (optional) if True, consider quadratic form matrix P
+    """
```
rileyjmurray, line 97: I'd like for this to be more thoroughly documented. Why is it needed? What shortcoming does it address? We should be clear about these things since they'll come up when rewriting cvxcore (ping @phschiele).

rileyjmurray (changes requested):
Impressive work! It's surprising that this functionality could be added in such little code. Although, I'd like for us to think carefully about whether it makes sense to modify the ``ParamConeProg`` and ``Dcp2Cone`` classes or subclass them. Here are some thoughts:

* I'm not a fan of changing the ``ParamConeProg`` class so that it allows nonlinear objectives. I'd prefer to define a new class called ``ParamQuadConeProg`` that subclasses ``ParamConeProg``. If it seems like doing so would require a lot of code duplication then I think that's an argument to simplify some of the functions in the ``ParamConeProg`` class to handle smaller units of functionality a piece.

* I also think that ``Dcp2Cone`` should be clearer with its scope as to affine versus quadratic representations. I can see how defining a new canonicalization class might be overkill, but perhaps a name change is in order.

There is also a helper for which I've requested additional documentation.

cvxpy/problems/problem.py
```diff
@@ -510,7 +510,8 @@ def get_problem_data(
         gp: bool = False,
         enforce_dpp: bool = False,
         ignore_dpp: bool = False,
-        verbose: bool = False
+        verbose: bool = False,
+        **kwargs
```
akshayka, line 514: Instead of `**kwargs`, what do you think about explicitly taking a `dict`? Otherwise we could run into an unfortunate edge case in which a solver has an option whose name is the same as an argument to `get_problem_data`, making it impossible to forward that argument to the solver. I know we take var args in `solve`, and I guess that was done to support custom solve methods ...  but here I think there's no reason to do so.

So my suggestion would be to change the signature to:

```python3
    def get_problem_data(
        self, solver,
        gp: bool = False,
        enforce_dpp: bool = False,
        ignore_dpp: bool = False,
        verbose: bool = False
        solver_args: Optional[dict] = None):
```

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -194,14 +187,13 @@ def param_value(idx):
...
-        return c, d, A, np.atleast_1d(b)
+        A, b = self.reduced_A.get_matrix_from_tensor(param_vec, with_offset=True)
+        if quad_obj:
+            self.reduced_P.cache(keep_zeros)
+            P, _ = self.reduced_P.get_matrix_from_tensor(param_vec, with_offset=False)
+            return P, c, d, A, np.atleast_1d(b)
+        else:
+            return c, d, A, np.atleast_1d(b)
 
     def apply_param_jac(self, delc, delA, delb, active_params=None):
```
akshayka, line 198: Does this function need to be updated to handle quadratic objectives? If it's not supported, we can just raise a `ValueError` if this method  is called when the `ParamConeProg` has a quadratic objective.

akshayka (commented):
Like Riley said, this is way fewer LOC than I anticipated. Very readable. Nice!

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
SteveDiamond: Done.

cvxpy/problems/problem.py
SteveDiamond: Done.

cvxpy/reductions/utilities.py
SteveDiamond: Done.

cvxpy/reductions/qp2quad_form/atom_canonicalizers/__init__.py
SteveDiamond: Yes, added a comment about that.

cvxpy/reductions/qp2quad_form/atom_canonicalizers/__init__.py
SteveDiamond: Done.

cvxpy/reductions/dcp2cone/dcp2cone.py
```diff
@@ -26,9 +35,11 @@ class Dcp2Cone(Canonicalization):
     them into problems with affine objectives and conic constraints whose
     arguments are affine.
     """
```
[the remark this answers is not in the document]
SteveDiamond: Fixed.

cvxpy/atoms/affine/affine_atom.py
```diff
@@ -74,6 +74,11 @@ def is_decr(self, idx) -> bool:
     def is_quadratic(self) -> bool:
         return all(arg.is_quadratic() for arg in self.args)
 
+    def has_quadratic_term(self) -> bool:
+        """Does the affine head of the expression contain a quadratic term?
```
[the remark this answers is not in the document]
SteveDiamond: Added explanation.

SteveDiamond:
Thanks for reviewing! Let me explain more why I added functionality to dcp2cone, cone_matrix_stuffing, and ParamConeProg rather than defining new classes. Since quadratic cone programs are a superset of both cone programs and quadratic programs, I see adding quadratic cone program support as an opportunity to simplify the canonicalization logic. We can deprecate the QP canonicalization path and reduce the QP and conic canonicalizations to a single pathway. That was too much for me to bite off right now, but I wanted to lay the ground work for that simplification by enhancing the conic canonicalization pathway.

I'm happy to rename various classes to more clearly emphasize that they now handle quadratics if that's desired.

cvxpy/reductions/utilities.py
```diff
@@ -87,6 +87,14 @@ def group_constraints(constraints):
...
     """Utility class for condensing the mapping from parameters to problem data.
 
+    For maximum efficiency of representation and application, the mapping from
+    parameters to problem data must be condensed. It begins as a CSC sparse matrix,
+    such that multiplying by a parameter vector gives the problem data. The row index
+    array and column pointer array are saved, and a dense matrix that when multiplied
+    by a parameter vector gives the values array. The ReducedMat class caches the
+    condensed representation and provides a method for multiplying by a parameter
+    vector.
+
```
rileyjmurray, line 97: Thanks for adding this! I don't quite follow the phrase:

> and a dense matrix that when multiplied by a parameter vector gives the values array

What's the dense matrix?

cvxpy/reductions/utilities.py
```diff
@@ -80,3 +82,82 @@ def group_constraints(constraints):
...
+    """
+
+    def __init__(self, matrix_data, var_len: int, quad_form: bool = False) -> None:
+        self.matrix_data = matrix_data
+        self.var_len = var_len
+        self.quad_form = quad_form
+        # Computed attributes.
+        self.reduced_mat = None
+        self.problem_data_index = None
+        self.mapping_nonzero = None
```
rileyjmurray, line 114: Could you specify the data types and interpretations of these variables? I know it's a pain but I really don't want to add more matrix stuffing code that isn't clearly documented.

cvxpy/reductions/utilities.py
```diff
@@ -80,3 +82,82 @@ def group_constraints(constraints):
     for c in constraints:
         constr_map[type(c)].append(c)
     return constr_map
+
+
+class ReducedMat:
```
rileyjmurray, line 87: Consider adding a remark in the docstring that this functionality was originally duplicated in the ``ParamQuadProg`` and ``ParamConeProg`` classes, and that this class consolidates that functionality.

cvxpy/reductions/utilities.py
rileyjmurray: I see now that this class is consolidating very similar code that appears in ``ParamConeProg`` and ``ParamQuadProg``. I'm very much in favor of this approach. I still think that now that it's in a standalone class it's a good opportunity to thoroughly document.

rileyjmurray (changes requested):
Just a few more questions / requests.

SteveDiamond:
Thanks @rileyjmurray for the further comments! I added documentation as requested.

PR #1861, opened 2022-08-09
Update README to reflect Python 3.7+ requirement

chrisyeh96:
cvxpy requires Python 3.7+, as stated on the [Installation Guide](https://www.cvxpy.org/install/index.html) and in [setup.py](https://github.com/cvxpy/cvxpy/blob/v1.2.1/setup.py#L227).


SteveDiamond:
Good catch!

PR #1862, opened 2022-08-09
Typing annotation improvements

chrisyeh96:
This pull requests improves typing annotations. The main changes are:

- adds `int` as a valid type for shape argument in `cp.Parameter()` and `cp.Variable()`
- using [`__future__.annotations`](https://peps.python.org/pep-0563/), we can simplify the typing annotations

### DIFF
--- a/cvxpy/expressions/constants/parameter.py
+++ b/cvxpy/expressions/constants/parameter.py
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-from typing import List, Optional, Tuple
+from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


SteveDiamond (approved):

SteveDiamond:
Thanks, I didn't know annotations could be done in this way!

chrisyeh96:
Glad to help! And yeah, `from __future__ import annotations` ([https://peps.python.org/pep-0563/](https://peps.python.org/pep-0563/)) really makes typing annotations much easier. This is only available starting in Python 3.7, which is why I had to double check CVXPY's minimum supported Python version, and submitted #1861.

rileyjmurray:
@SteveDiamond quick note: this ``from __future__`` business is being used here in a similar way as in the PR for the quadrature-based approximation of the relative entropy cone.

PR #1859, opened 2022-08-06
Fix time_limit_sec for GLOP and PDLP

mlubin:
Fixes #1858

I looked for examples of tests that check that the time limit is actually hit but didn't find any. These are potentially flaky, but still useful to check that the time limit is actually enforced.

PR #1866, opened 2022-08-14
Make log_det robust to numerica

### LINKED ISSUE #1816: log_det for complex psd matrix
**Describe the bug**
Hi cvxpy team,

I found that in the log_det function, the input matrix is treated as a real positive semidefinite (PSD) matrix. If the input is a complex matrix, the result turns out to be -np.inf, due to these lines:
```
symm = (values[0] + values[0].T)/2
sign, logdet = LA.slogdet(symm)
if sign == 1:
    return logdet
else:
    return -np.inf
```

In addition, as far as I know, PSD matrices are always considered as real matrics in cvxpy. It might lead to unexpected results if they are as well specified as complex. Since complex PSD matrices are very commonly used in many areas like wireless communication, it would be great if cvxpy could handle it in the future.

**To Reproduce**
A small example for the problem of log_det function:

```
import cvxpy as cp

X = cp.Variable(shape = (3,3),
            complex = True)


cons = [X>>0, cp.real(cp.trace(X))<=9]
obj = cp.Maximize(cp.log_det(X))

prob = cp.Problem(objective=obj, constraints=cons)

prob.solve()
print(prob.value)
print(X.value)

```

and it gives -np.inf as the optimal objective value although `X` is the solved correctly, which is diag{3, 3, 3}. 

**Output**
```
-inf
[[3.00000000e+00+0.00000000e+00j 1.41069594e-15-5.53050530e-16j
  3.72971957e-16-6.41441562e-16j]
 [1.41069594e-15+5.53050529e-16j 3.00000000e+00+0.00000000e+00j
  2.88571453e-15+1.42317559e-15j]
 [3.72971957e-16+6.41441563e-16j 2.88571453e-15-1.42317559e-15j
  3.00000000e+00+0.00000000e+00j]]
```

**Version**
 - OS: Windows 11
 - CVXPY Version: 1.2.1

rileyjmurray:
…"sign" from numpys slogdet is real.

This fixes #1816. The current ``log_det`` implementation can return wrong results if (1) the matrix was complex but not diagonal (2) the matrix is complex and diagonal, but numpy's ``slogdet`` function returns a sign with an extremely small imaginary part. The solution is simple since the atom is supposed to project onto the set of Hermtian matrices as a first step: Hermitian matrices always have real eigenvalues, so we should just check the real part of the sign returned from numpy's ``slogdet``.

### DIFF
--- a/cvxpy/tests/test_complex.py
+++ b/cvxpy/tests/test_complex.py
@@ -314,11 +314,26 @@ class TestComplex(BaseTest):
         P = np.arange(9) - 2j*np.arange(9)
         P = np.reshape(P, (3, 3))
         P = np.conj(P.T).dot(P)/100 + np.eye(3)*.1
-        value = cp.log_det(P).value
+        logdet_value = cp.log_det(P).value
         X = Variable((3, 3), complex=True)
-        prob = Problem(cp.Maximize(cp.log_det(X)), [X == P])
+        objective = cp.log_det(X)
+        prob = Problem(cp.Maximize(objective), [X == P])
         result = prob.solve(solver=cp.SCS, eps=1e-6)
-        self.assertAlmostEqual(result, value, places=2)
+        self.assertAlmostEqual(result, logdet_value, places=2)
+        objective_value = objective.value
+        s, ld = np.linalg.slogdet(P)
+        self.assertAlmostEqual(objective_value, ld)
+
+        # Test case for Issue 1816.
+        #   The optimal solution is the identity matrix scaled by 3.
+        #   NumPy's slogdet function returns a sign "s" with a tiny complex
+        #   part which causes 1 == s to fail.
+        cons = [X >> 0, cp.real(cp.trace(X)) <= 9]
+        obj = cp.Maximize(cp.log_det(X))
+        prob = cp.Problem(objective=obj, constraints=cons)
+        prob.solve(solver=cp.SCS, eps=1e-6)
+        self.assertAlmostEqual(obj.value, 3*np.log(3))
+        pass
 
     def test_eigval_atoms(self) -> None:
         """Test eigenvalue atoms.
--- a/cvxpy/atoms/log_det.py
+++ b/cvxpy/atoms/log_det.py
@@ -39,10 +39,10 @@ class log_det(Atom):
         For PSD matrix A, this is the sum of logs of eigenvalues of A
         and is equivalent to the nuclear norm of the matrix logarithm of A.
         """
-        # take symmetric part of the input.
-        symm = (values[0] + values[0].T)/2
+        # take Hermitian part of the input.
+        symm = (values[0] + np.conj(values[0].T))/2
         sign, logdet = LA.slogdet(symm)
-        if sign == 1:
+        if np.real(sign) == 1:
             return logdet
         else:
             return -np.inf


rileyjmurray:
@mlubin unit tests are failing due to type errors with PDLP. See https://github.com/cvxpy/cvxpy/runs/7823099292?check_suite_focus=true#step:6:1023. Can you comment on what the issue might be, and maybe contribute a fix? This could be related to your recent PR #1859.

cvxpy/tests/test_complex.py
```diff
@@ -314,11 +314,26 @@ def test_log_det(self) -> None:
...
+        # Test case for Issue 1816.
+        #   The optimal solution is the identity matrix scaled by 3.
+        #   NumPy's slogdet function returns a sign "s" with a tiny complex
+        #   part which causes 1 == s to fail.
+        cons = [X >> 0, cp.real(cp.trace(X)) <= 9]
+        obj = cp.Maximize(cp.log_det(X))
+        prob = cp.Problem(objective=obj, constraints=cons)
+        prob.solve(solver=cp.SCS, eps=1e-6)
+        self.assertAlmostEqual(obj.value, 3*np.log(3))
+        pass
```
SteveDiamond, line 336: Remove pass.

SteveDiamond (approved):
LGTM other than tiny comment.

rileyjmurray:
For reasons unknown to me, the test meaningfully fails on Ubuntu and macOS Python 3.8.
```
self = <cvxpy.tests.test_complex.TestComplex testMethod=test_log_det>

    def test_log_det(self) -> None:
        """Test log det.
        """
        P = np.arange(9) - 2j*np.arange(9)
        P = np.reshape(P, (3, 3))
        P = np.conj(P.T).dot(P)/100 + np.eye(3)*.1
        logdet_value = cp.log_det(P).value
        X = Variable((3, 3), complex=True)
        objective = cp.log_det(X)
        prob = Problem(cp.Maximize(objective), [X == P])
        result = prob.solve(solver=cp.SCS, eps=1e-6)
>       self.assertAlmostEqual(result, logdet_value, places=2)

cvxpy/tests/test_complex.py:322: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cvxpy/tests/base_test.py:39: in assertAlmostEqual
    super(BaseTest, self).assertAlmostEqual(a, b, places=places, delta=delta)
E   AssertionError: -inf != -1.692819521373[149](https://github.com/cvxpy/cvxpy/runs/7868109908?check_suite_focus=true#step:6:150)6 within 2 places (inf difference)
```

SteveDiamond:
> For reasons unknown to me, the test meaningfully fails on Ubuntu and macOS Python 3.8, and macOS Python 3.7. Maybe it has something to do with how these cases use SCS 2.1.2? Will investigate.
> 
> EDIT: the three failing configurations all use SCS 2.1.2. I selected two passing configurations at random and they used SCS 3.2.0. So maybe the old SCS is inaccurately reporting that the problem is infeasible? Could change to CVXOPT and see if that fixes things.
> 
> EDIT THE SECOND: It's not just SCS 2.1.2 vs 3.2.0. I also checked passing tests for Python 3.7 on Windows and Ubuntu and they use SCS 2.1.2.
> 
> ```
> self = <cvxpy.tests.test_complex.TestComplex testMethod=test_log_det>
> 
>     def test_log_det(self) -> None:
>         """Test log det.
>         """
>         P = np.arange(9) - 2j*np.arange(9)
>         P = np.reshape(P, (3, 3))
>         P = np.conj(P.T).dot(P)/100 + np.eye(3)*.1
>         logdet_value = cp.log_det(P).value
>         X = Variable((3, 3), complex=True)
>         objective = cp.log_det(X)
>         prob = Problem(cp.Maximize(objective), [X == P])
>         result = prob.solve(solver=cp.SCS, eps=1e-6)
> >       self.assertAlmostEqual(result, logdet_value, places=2)
> 
> cvxpy/tests/test_complex.py:322: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> cvxpy/tests/base_test.py:39: in assertAlmostEqual
>     super(BaseTest, self).assertAlmostEqual(a, b, places=places, delta=delta)
> E   AssertionError: -inf != -1.692819521373[149](https://github.com/cvxpy/cvxpy/runs/7868109908?check_suite_focus=true#step:6:150)6 within 2 places (inf difference)
> ```

Weird, wonder why 2.1.2 is behaving differently on different setups. In any case no issue with using CVXOPT, though you'll need to add a check that will skip the test if CVXOPT is not installed. No need to stress over what's going on with SCS.

rileyjmurray:
So apparently the failures aren't due to SCS. What's happening is that ``prob.value`` is being set to ``prob.objective.value``. I changed the implementation of ``logdet(X).value`` so the fact that it's returning ``-np.inf`` on some platforms does suggest that my implementation wasn't kosher. I'll fix.

SteveDiamond:
@cvxpy/maintainers this is ready for review. The benchmarks fail but I think we can ignore that, unrelated to the PR.

cvxpy/tests/test_complex.py
```diff
@@ -314,11 +314,25 @@ def test_log_det(self) -> None:
...
         P = np.conj(P.T).dot(P)/100 + np.eye(3)*.1
-        value = cp.log_det(P).value
+        logdet_value = cp.log_det(P).value
         X = Variable((3, 3), complex=True)
-        prob = Problem(cp.Maximize(cp.log_det(X)), [X == P])
-        result = prob.solve(solver=cp.SCS, eps=1e-6)
-        self.assertAlmostEqual(result, value, places=2)
+        objective = cp.log_det(X)
+        prob = Problem(cp.Maximize(objective), [X == P])
+        result = prob.solve(solver=cp.SCS, eps=1e-6, verbose=True)
```
phschiele, line 321:
```suggestion
        result = prob.solve(solver=cp.SCS, eps=1e-6)
```

cvxpy/tests/test_complex.py
```diff
@@ -314,11 +314,25 @@ def test_log_det(self) -> None:
...
+        self.assertAlmostEqual(objective_value, ld, places=3)
+
+        # Test case for Issue 1816.
+        #   The optimal solution is the identity matrix scaled by 3.
+        #   NumPy's slogdet function returns a sign "s" with a tiny complex
+        #   part which causes 1 == s to fail.
+        cons = [X >> 0, cp.real(cp.trace(X)) <= 9]
+        obj = cp.Maximize(cp.log_det(X))
+        prob = cp.Problem(objective=obj, constraints=cons)
+        prob.solve(solver=cp.SCS, eps=1e-6, verbose=True)
```
phschiele, line 334:
```suggestion
        prob.solve(solver=cp.SCS, eps=1e-6)
```

phschiele (approved):
The change makes sense to me, added a tiny suggestion to remove the verbose keyword from the test case.

SteveDiamond:
Thanks!!!

Issue #420, opened 2017-12-14
Generic perspective function support


b4be1:
I would like to implement an [f-divergence](https://en.wikipedia.org/wiki/F-divergence) in cvxpy. How can I do that? Let's say for a fixed function such as f(x) = x**2 - 1.

rileyjmurray:
An f-divergence is just a sum of perspective functions. I don't know if cvxpy has native support for general perspective functions (its documentation does not say so), but it *does* have the perspective function for quadratics "quad_over_lin". So if you really only need f-divergences built off of quadratics, then you can get by with quad_over_lin.

b4be1:
Hi @rileyjmurray, thanks a lot for the info. I want an implementation for any convex f, because I am in particular interested in using the [alpha-divergence](www.mdpi.com/1099-4300/12/6/1532/pdf), and the quadratic was just an example.

SteveDiamond:
Perspective functions are actually possible generically, but never implemented.

rileyjmurray:
@SteveDiamond , can explain what you mean by your last comment? It sounds like you're saying that in principle, it would be possible to have a meta-atom that transforms existing cvxpy atoms into their perspective function variants. If that's the case such a feature would be very useful. For one: it would apply not only allow things like f-divergences, but it could also unify seemingly distinct functions like kl_div, quad_over_lin, etc.... The benefit of putting these functions through a common parser is that for general nonlinear solvers (e.g. cvxopt), work spent improving performance on one type of perspective function might carry over to another.

moehle:
Not sure if this is helpful, but see Section 2 of [this paper](http://web.stanford.edu/~boyd/papers/pdf/sw_aff_ctrl.pdf) for an idea of how this *could* be done.  I wouldn't worry too much about the technical conditions, like closedness, which only really apply if you're doing integer stuff.  It basically says that if f(x) has a conic representation with cone K, then so does its perspective.  Basically you just take any affine terms in the epigraph representations and scale them by a new scalar variable.

rileyjmurray:
@moehle , that looks awesome!  @SteveDiamond , if someone was willing to implement the procedure in Nicholas's paper, would you be open to including it in cvxpy 1.0?

(Under normal circumstances I would volunteer for such an implementation, but at this point it would be pretty far back in my queue of work.)

ericlux:
[...]

PR #1867, opened 2022-08-14
WIP: Generalized perspective support

ericlux:
## Description
Support for generic handling of the perspective transform has been a longstanding issue dating back to pre 1.0 days. We (@phschiele  and I) use the fact that the perspective transform of a function with a given graph form is trivially obtained. We implement this functionality as a new atom called `perspective` which takes in a (convex) scalar expression `f ` and a nonnegative affine scalar `s`. Under the hood, the canonicalization computes `f`'s graph form, adds the new variable `s`, and represents the perspective of `f` by returning the epigraph variable `t`.

For further reference on the graph form perspective rule, see @moehle's [paper](https://www-sciencedirect-com.stanford.idm.oclc.org/science/article/pii/S0167691115001747).

This is still a work in progress and we hope to get some feedback on the general implementation idea. We provide extensive tests for problems involving the second order cone, the exponential cone, the power cone, and the psd cone. There is still some minor additions: dealing with concave functions, grad, elementwise capability, but we think the core functionality is there.

Issue link (if applicable): #420

## Type of change
- [ ] Bug fix

- [x] Write unittests.


adishavit:
The link to the paper requires a password.  
Can you provide an alternate?  
Is [this](https://web.stanford.edu/~boyd/papers/sw_aff_ctrl.html) the one?

ericlux:
Yes, sorry about that! That’s the correct paper.

cvxpy/atoms/elementwise/elementwise.py
```diff
@@ -60,7 +60,7 @@ def elemwise_grad_to_diag(value, rows, cols):
         """
         if not np.isscalar(value):
             value = value.ravel(order='F')
-        return sp.dia_matrix((value, [0]), shape=(rows, cols)).tocsc()
+        return sp.dia_matrix(([value], [0]), shape=(rows, cols)).tocsc()
```
SteveDiamond, line 63: why is this changed?

cvxpy/atoms/elementwise/exploration.ipynb
```diff
@@ -0,0 +1,72 @@
+{
```
SteveDiamond, line 1: Is this file intended to be committed?

cvxpy/atoms/elementwise/elementwise.py
ericlux: This fails under scipy1.9, nothing to do with the perspective. Philipp and I just ID'd it along the way.

cvxpy/atoms/elementwise/exploration.ipynb
ericlux: Nope, I'm bad at git! Clearing now

cvxpy/reductions/dcp2cone/atom_canonicalizers/perspective_canon.py
```diff
@@ -0,0 +1,92 @@
...
+    constraints.append(-c@x_canon + t - s*d >= 0)
+
+    # recover initial variables
+
+    end_inds = sorted(prob_canon.var_id_to_col.values()) + [x_canon.shape[0]]
+
+    for var in expr.f.variables():
+        start_ind = prob_canon.var_id_to_col[var.id]
+        end_ind = end_inds[end_inds.index(start_ind)+1]
+        if var.is_psd():
```
SteveDiamond, line 80: Should be var.is_symmetric()?

ericlux: I had this before but ran into a failed test because a scalar variable is symmetric. But thanks for catching this; the matrix totally could be NSD also. Any thoughts on how to handle this? In general wasn’t sure the best generic way to set these “outer” variables (that the user constructing the perspective created) equal to their underlying counterpart which is exposed in the canonicalization (which causes this outer matrix to engage in an equality constraint with a different shape variable in x_canon).

cvxpy/tests/test_perspective.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+def test_rel_entr():
+    x = cp.Variable()
+    s = cp.Variable(nonneg=True)
+    f = -cp.log(x)
+    obj = cp.perspective(f, s)
+    constraints = [1 <= s, s <= 2, 1 <= x, x <= 2]
+    prob = cp.Problem(cp.Minimize(obj), constraints)
+    prob.solve(solver=cp.MOSEK)
```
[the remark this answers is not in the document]
ericlux: Moved to default solvers. Resolved by [b9f1fc4](https://github.com/cvxpy/cvxpy/pull/1867/commits/b9f1fc4a1244738ad3f7d8bbd3db4de90f87657d)

SteveDiamond:
@ericlux we had a discussion about this PR, and it's almost ready to be integrated for 1.3. The only remaining TODO is to add an assert that no parameters are involved, since that case isn't handled.

ericlux:
Oh awesome! In that case we will defer parameters to later, and add the assert!

ericlux:
@SteveDiamond updated the documentation as discussed
Screenshot 2022-11-19 at 2 11 59 PM

ericlux:
@SteveDiamond tests passing after diffcp fix; should be good to go.

SteveDiamond (approved):
LGTM. Could you clarify how parameters can be used with parameters? It looks like there are no restrictions.

ericlux:
We added `chain = aux_prob._construct_chain(solver_opts=solver_opts, ignore_dpp=True)` in the canon method. So any parameter will just be evaluated without getting the dpp benefits for repeated solves.

SteveDiamond:
> We added `chain = aux_prob._construct_chain(solver_opts=solver_opts, ignore_dpp=True)` in the canon method. So any parameter will just be evaluated without getting the dpp benefits for repeated solves. We'll double check to make sure is_dpp() returns false for a parameterized perspective expression.

Got it, thanks! If the PR looks good to @phschiele we can merge then. Thanks for all your hard work on this!

PR #1868, opened 2022-08-14
Upper bound or-tools to 9.3

mlubin:
Re https://github.com/cvxpy/cvxpy/pull/1866#issuecomment-1214264182, or-tools 9.4 was recently released with breaking API changes. Upper bounding should get tests passing again.

CC @mizux

### DIFF
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -28,7 +28,7 @@ elif [[ "$PYTHON_VERSION" == "3.10" ]]; then
     python -m pip install sdpa-python
 fi
 
-python -m pip install ortools coptpy
+python -m pip install ortools<9.4 coptpy
 
 # CBC comes with wheels for windows and needs coin-or-cbc to compile otherwise
 # conda-forge in progress: https://github.com/conda-forge/staged-recipes/pull/14950
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -28,7 +28,7 @@ elif [[ "$PYTHON_VERSION" == "3.10" ]]; then
     python -m pip install sdpa-python
 fi
 
-python -m pip install ortools<9.4 coptpy
+python -m pip install "ortools>=9.3,<9.4" coptpy
 
 # CBC comes with wheels for windows and needs coin-or-cbc to compile otherwise
 # conda-forge in progress: https://github.com/conda-forge/staged-recipes/pull/14950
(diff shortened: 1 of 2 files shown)


rileyjmurray:
Thanks for looking into this, @mlubin!

Although, putting an upper bound on the or-tools version is pretty undesirable long-term. Is the plan to patch the 9.4 release so that it un-breaks the API changes? If not, we should have code paths to support 9.3 and 9.4.

mlubin:
I doubt 9.4 will be patched to un-break the API changes. or-tools doesn't follow semantic versioning. For a better user experience and to avoid breaking cvxpy's CI, we probably do need a permanent defensive upper bound on or-tools.

I will look into updating the code to support 9.4. My preference would be to update the code to 9.4 and drop support for 9.3 as soon as it's practical, unless it's trivial to support both simultaneously.

rileyjmurray:
@mlubin that's a bit problematic for us. We don't want to regularly drop support for older versions of solver interfaces. Here are places in existing solver interface code where we've taken steps to ensure backward compatibility: [[1]](https://github.com/cvxpy/cvxpy/blob/a74689b5bee6a9d7e5fdc51fd31814f23597832b/cvxpy/reductions/solvers/conic_solvers/cplex_conif.py#L77-L86), [[2]](https://github.com/cvxpy/cvxpy/blob/a74689b5bee6a9d7e5fdc51fd31814f23597832b/cvxpy/reductions/solvers/conic_solvers/diffcp_conif.py#L80-L87), [[3]](https://github.com/cvxpy/cvxpy/blob/a74689b5bee6a9d7e5fdc51fd31814f23597832b/cvxpy/reductions/solvers/conic_solvers/scs_conif.py#L31-L37), [[4]](https://github.com/cvxpy/cvxpy/blob/a74689b5bee6a9d7e5fdc51fd31814f23597832b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py#L81-L90), [[5]](https://github.com/cvxpy/cvxpy/blob/a74689b5bee6a9d7e5fdc51fd31814f23597832b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py#L205-L222). The SCIP interface is a notable place where we do enforce a version limit. We only do that because the underlying numerical solver returns incorrect results in the latest release, and we've filed a bug report to that effect (see [here](https://github.com/scipopt/PySCIPOpt/issues/583)).

If this sort of API breaking change is rare, then I'd be open to dropping support for or-tools 9.3. Otherwise we'll need a bigger discussion on this.

Ping @cvxpy/maintainers and @phschiele. What are your thoughts?

SteveDiamond:
I would vote for a fix that supports both 9.3 and 9.4. If it's quite complicated we can drop 9.3, but we normally would support both versions in this situation.

Issue #1865, opened 2022-08-13
Norm operator with string input on a one dimensional array results in cryptic error


KerimovEmil:
**Describe the bug**
Norm operator with string input on a one dimensional array results in cryptic error. Specifically the norm atom doesn't correctly handle the cases of when `x` is a one dimensional vector with the options of `fro` or `nuc`. It passes `p` directly into the `pnorm` atom which tries to then compare the string against the number 0, resulting in a type error. 

If it makes sense I can create a PR to handle those two cases, if `x` is not a matrix. For the frobenius norm at least I think it should default to the 2-norm if `x` is a vector. 

**To Reproduce**

```
import cvxpy
w = cvxpy.Variable(10)
t = cxvpy.norm(w, p='fro')
```

**Expected behavior**
I expected either an error message specifying that the frobenius option is not available for vector inputs or even better for it to default to the 2 norm after casting `x` as a vector. 

**Output**
![image](https://user-images.githubusercontent.com/8174342/184509338-c910f0d0-688b-4c42-867a-1700e1373134.png)

**Version**
 - OS: windows
 - CVXPY Version: 1.2

**Additional context**
This issue was found when I tried upgrading some existing code from the older 0.4 cvxpy version to the latest 1.2. In the older code since it didn't check for the dimension of `x`, the code resulted in just defaulting to `pnorm(vec(x), 2)`.

rileyjmurray:
Hi @KerimovEmil, thanks for the report. I'm not sure what our semantics should be here. The ``pnorm`` function is supposed to be restricted to vectors or applied along an axis to a matrix. Based on that, strings specifying matrix norms (like the Frobenius or nuclear norms) wouldn't be acceptable arguments. @SteveDiamond @phschiele @akshayka do you suppose we should raise an error here?

phschiele:
@rileyjmurray Raising an informative error message seems like a good idea.

SteveDiamond:
To be clear this isn't calling ``pnorm`` externally. It's about the behavior of ``norm``. I would vote for @KerimovEmil's idea of ``fro`` mapping to the L2 norm and ``nuc`` raising an error. It's pretty clear what the user meant for ``fro`` norm even if it's not strictly correct semantics.

SteveDiamond:
Resolved by #1871

PR #1871, opened 2022-08-14
Adding edge case handling for string inputs into norm

KerimovEmil:
Signed-off-by: KerimovEmil 

## Description
Raising an error for non-supported string arguments into norm atom, and handling the fro norm.
Addressing issue: https://github.com/cvxpy/cvxpy/issues/1865

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/norm.py
+++ b/cvxpy/atoms/norm.py
@@ -67,9 +67,9 @@ def norm(x, p: Union[int, str] = 2, axis=None):
     else:
         if p == 1 or x.is_scalar():
             return norm1(x, axis=axis)
-        elif p in [np.inf, "inf", "Inf"]:
+        elif str(p).lower() == "inf":
             return norm_inf(x, axis)
-        elif p == 'fro':
+        elif str(p).lower() == "fro":
             return pnorm(vec(x), 2, axis)
         elif isinstance(p, str):
             raise RuntimeError(f'Unsupported norm option {p} for non-matrix.')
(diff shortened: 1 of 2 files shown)


phschiele:
@KerimovEmil thanks for taking the time to make this PR!
Generally, for a bug fix, it is nice to provide a test case that would fail before your PR, and pass after.
In your case, you can add a test `test_norm_exceptions()` to `test_atoms.py`.
You can check `test_list_input()` in the same file for an example of such an exception test.

KerimovEmil:
> @KerimovEmil thanks for taking the time to make this PR!
> Generally, for a bug fix, it is nice to provide a test case that would fail before your PR, and pass after.
> In your case, you can add a test `test_norm_exceptions()` to `test_atoms.py`.
> You can check `test_list_input()` in the same file for an example of such an exception test.

Thanks! I'll try to add that later tonight.

cvxpy/atoms/norm.py
```diff
@@ -69,6 +69,10 @@ def norm(x, p: Union[int, str] = 2, axis=None):
             return norm1(x, axis=axis)
         elif p in [np.inf, "inf", "Inf"]:
             return norm_inf(x, axis)
+        elif p == 'fro':
```
phschiele, line 72:
```suggestion
        elif str(p).lower == "inf":
            return norm_inf(x, axis)
        elif str(p).lower == "fro":
```
`str(np.inf) == "inf"`

SteveDiamond:
Looks good, thanks @KerimovEmil !

PR #1875, opened 2022-08-16
Constraint class for the operator relative entropy cone, `OpRelCone`, with symmetric matrices

aryamanjeendgar:
In this PR, I add support for the "Full" operator relative entropy cone (specifically, the approximation scheme for the same as defined in [FSP17](https://arxiv.org/abs/1705.00812)), which is the epigraph associated with the _operator relative entropy_, which in turn, is defined as the _non-commutative perspective_ of the negative logarithm

The _Operator Relative Entropy_ is defined as so:

$$
D_{\text{op}}(X||Y)=-X^{1/2} \log(X^{-1/2} Y X^{-1/2}) X^{1/2} 
$$

The corresponding "Full" Operator Relative Entropy Cone is defined as follows:

$$
K_{\text{re}}^{n}=\text{cl}\\{(X,Y,T)\in {\mathbb{H}}_{\scriptsize{++}}^{n}\times \mathbb{H}_{\scriptsize{++}}^{n}\times \mathbb{H}^{n}: D_{\text{op}}(X||Y)\preceq T\\}
$$

The PR implements a well motivated approximation to the above set, where instead of working with the non-commutative perspective of the negative logarithm, we operate over an approximate proxy for the same (which is defined in terms of two key ideas i.e. an integral representation of the logarithm and the functional equation: $\log(x)=\frac{1}{h}\log(x^h)$ ), and hence; instead, we observe the epigraph of the non-commutative perspective of this approximate construction.

The current implementation is largely complete. Web documentation is still needed and the implementation needs to support complex (i.e., actually Hermitian, rather than symmetric) inputs.

Thank you!


cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +560,121 @@ def test_expcone_socp_1(self):
...
+        con_pairs = [(con1, None),
+                     (con2, None),
+                     (con3, None)]
+        k = 3
+        obj = cp.Minimize(cp.atoms.lambda_sum_largest(T, k))
+        obj_pair = (obj, -1.6554572363409088)
+        expect_T = np.array([[-5.51808473e-01, -3.22477062e-13, -1.53851070e-13, 9.32509529e-14],
+                             [-3.22476879e-13, -5.51808473e-01, 5.93445801e-15, 4.76722578e-13],
+                             [-1.53851100e-13, 5.93442592e-15, -5.51808473e-01, 4.09022045e-14],
+                             [9.32509585e-14, 4.76722648e-13, 4.09021549e-14, -5.51808473e-01]])
```
rileyjmurray, line 627: Comments:
1. There's no need to provide numerical values on the order of 1e-12, since we only run tests to 2 or 6 decimal places. So the small values here should be explicitly zeroed out.
2. If you're testing nontrivial functionality then you should explain how you got the answer ``expect_T``. Only a small portion of our tests actually do this, but I think it's important.

Other points
1. Your constraints on ``a_diag`` and ``b_diag`` should make sure they don't all take the same values at optimality. Right now that's happening, and it's an uninteresting test since if a symmetric matrix only has one distinct eigenvalue then that matrix must be a multiple of the identity.
2. You can try computing compute the reference value ``T_expect`` by having ``T_expect = U @ diag(t_expect) @ U.T``, and then using scalar relative entropy (exact, not quad approx) to compute the optimal value of ``t_expect``.

aryamanjeendgar:
The non-commutative test cases are now working!

As it turned out, we were getting a case of non-unique solutions due to us not explicitly imposing that every involved matrix in the program be symmetric (which is the case in the actual reduction that we implement). The latest commit takes care of that and also adds in a warning to the user if they construct a `OpRelCone` constraint by passing in matrices that haven't been explicitly declared to be symmetric beforehand (which is anyways imposed internally later on, in a less efficient manner).

We were able to in part, figure this out by running through the same test case in `CVXQUAD`, where the symmetric condition was explicitly imposed internally

rileyjmurray:
@cvxpy/maintainers this is ready for review!

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
         sth.verify_primal_values(places=2)
         sth.verify_objective(places=2)
+
+
+class TestOpRelCone(BaseTest):
+
+    @staticmethod
+    def Dop_commute_eig(a, b):
+        val = a.value * np.log(a.value/b.value)
+        return np.sum(val)
```
rileyjmurray, line 571: This is really the trace of operator relative entropy, with operators that have spectra ``(a, b)``.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        """
+        Computes Operator relative entropy of matrices A and B, where A is specified
+        in terms of its eigendecomposition.
+        Dop is the non-commutative perspective of the negative logarithm, following is the defn.
+        of the noncommutative perspective of a function `g`:
+        $P_g(X,Y)=Y^{1/2}g(Y^{-1/2}XY^{-1/2})Y^{1/2}$
+        We reproduce the same for the negative logarithm below:
+        """
+        flank = U1 @ np.diag(a)**(0.5) @ U1.T
+        # ^ changed to be more efficient
```
rileyjmurray, line 588: Remove comments.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        in_flank = U1 @ np.diag(a**(-0.5)) @ U1.T
+        # ^ changed to be more efficient
```
rileyjmurray, line 590: Remove comment

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        return -(flank @ sp.linalg.logm(in_flank @ B @ in_flank) @ flank)
+
+    @staticmethod
+    def Dop_eig(a, U1, B):
+        return sum(sp.linalg.eigh(TestOpRelCone.Dop(a.value, U1, B.value))[0])
```
rileyjmurray, line 595: Again, this is specifically the trace of operator relative entropy.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+    @staticmethod
+    def Dop(a, U1, B):
+        """
+        Computes Operator relative entropy of matrices A and B, where A is specified
+        in terms of its eigendecomposition.
+        Dop is the non-commutative perspective of the negative logarithm, following is the defn.
+        of the noncommutative perspective of a function `g`:
+        $P_g(X,Y)=Y^{1/2}g(Y^{-1/2}XY^{-1/2})Y^{1/2}$
+        We reproduce the same for the negative logarithm below:
+        """
```
rileyjmurray, line 586: This docstring doesn't mention ``U1`` and it seems to mention ``a`` as a matrix, rather than a vector. Explain.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        # ^ changed to be more efficient
+        in_flank = U1 @ np.diag(a**(-0.5)) @ U1.T
+        # ^ changed to be more efficient
+        return -(flank @ sp.linalg.logm(in_flank @ B @ in_flank) @ flank)
+
+    @staticmethod
+    def Dop_eig(a, U1, B):
+        return sum(sp.linalg.eigh(TestOpRelCone.Dop(a.value, U1, B.value))[0])
+
+    def oprelcone_1(self) -> STH.SolverTestHelper:
```
rileyjmurray, line 597: Describe the mathematical nature of this test with a docstring. Don't just rely on in-line comments in the body of the function.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        b_diag = cp.Variable(shape=(n,), pos=True)
+        # the below constraint ensures that `A` and `B` are
+        # defined in terms of their eigendecomposition and have the same eigenvectors
+        # i.e. they commute
+        A = U @ cp.diag(a_diag) @ U.T
+        B = U @ cp.diag(b_diag) @ U.T
+        T = cp.Variable(shape=(n, n))
+        # constrains A,B,T \in OpRelCone
+        con1 = cp.constraints.OpRelCone(A, B, T, 5, 5)
+        # imposing some non-trivial constraints to ensure feasibility
```
rileyjmurray, line 611: These constraints weren't just for feasibility.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,182 @@ def test_expcone_socp_1(self):
...
+        sth = STH.SolverTestHelper(obj_pair, var_pairs, con_pairs)
+        return sth
+
+    def test_oprelcone_1(self):
+        sth = self.oprelcone_1()
+        sth.solve(solver='SCS')
+        sth.verify_primal_values(places=2)
+        sth.verify_objective(places=2)
+
+    def oprelcone_2(self) -> STH.SolverTestHelper:
```
rileyjmurray, line 647: As with my comment on ``oprelcone_1``: provide a mathematical explanation of how this test works in a docstring.

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,107 @@ def shape(self) -> Tuple[int, ...]:
...
+        tup = (self.X, self.Y, self.Z, self.m, self.k)
+        return "OpRelCone(%s, %s, %s, %s, %s)" % tup
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    @property
+    def residual(self):
+        # TODO: implement me
+        return
```
rileyjmurray, line 314: You should ``raise NotImplementedError()``

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,107 @@ def shape(self) -> Tuple[int, ...]:
...
+
+    def __init__(self, X, Y, Z, m, k, constr_id=None) -> None:
+        Expression = cvxtypes.expression()
+        self.X = Expression.cast_to_const(X)
+        self.Y = Expression.cast_to_const(Y)
+        self.Z = Expression.cast_to_const(Z)
+        if (not X.is_symmetric()) or (not Y.is_symmetric()) or (not Z.is_symmetric()):
+            msg = ("One of the input matrices to the program has not explicitly been declared"
+                   " to be symmetric, we reccomend setting 'symmetric=True' for the same, or it"
+                   " will be done in a less efficient way internally")
```
rileyjmurray, line 290: The end of this sentence should have a period.

cvxpy/tests/test_cone2cone.py
```diff
@@ -566,9 +566,8 @@ def test_expcone_socp_1(self):
 class TestOpRelCone(BaseTest):
 
     @staticmethod
-    def Dop_commute_eig(a, b):
-        val = a.value * np.log(a.value/b.value)
-        return np.sum(val)
+    def Dop_commute_eig(a, b, U):
+        return trace(TestOpRelCone.Dop_commute(a, b, U)).value
```
rileyjmurray, line 570: While I appreciate the clear semantics of this, if you're going to return the trace of operator relative entropy with commuting matrices, then your original implementation suffices. You should just rename this to ``tr_Dop_commute(a, b)``. Also, you shouldn't use a cvxpy Atom unless you need to. In this case I believe ``np.linalg.trace`` or ``scipy.linalg.trace`` would be preferable.

cvxpy/tests/test_cone2cone.py
```diff
@@ -578,23 +577,28 @@ def Dop_commute(a, b, U):
...
         flank = U1 @ np.diag(a)**(0.5) @ U1.T
-        # ^ changed to be more efficient
         in_flank = U1 @ np.diag(a**(-0.5)) @ U1.T
-        # ^ changed to be more efficient
         return -(flank @ sp.linalg.logm(in_flank @ B @ in_flank) @ flank)
 
     @staticmethod
     def Dop_eig(a, U1, B):
-        return sum(sp.linalg.eigh(TestOpRelCone.Dop(a.value, U1, B.value))[0])
+        return trace(TestOpRelCone.Dop(a.value, U1, B.value)).value
```
rileyjmurray, line 594: Rename this function to ``tr_Dop(a, U, B)``, and use ``scipy.linalg.trace`` instead of the cvxpy trace Atom.

cvxpy/tests/test_cone2cone.py
```diff
@@ -578,23 +577,28 @@ def Dop_commute(a, b, U):
...
 
     def oprelcone_1(self) -> STH.SolverTestHelper:
+        """
+        These tests construct two matrices that commute (imposing all eigenvectors equal)
+        and then use the fact that: T=Dop(A, B) for (A, B, T) in OpRelCone i.e. T >> Dop(A, B)
+        (for an objective that is an increasing function of the eigenvalues).
+        """
```
rileyjmurray, line 601: Explain why this test is valid. This is good given the discussion that you and I have had, but it won't be a useful description if someone has to read this code years from now.

cvxpy/tests/test_cone2cone.py
```diff
@@ -695,6 +699,15 @@ def test_oprelcone_2(self):
...
     def oprelcone_3(self) -> STH.SolverTestHelper:
+        """
+        This test uses the same idea from the tests with commutative
+        matrices, instead, here, we make the input matrices to Dop, non-commutative,
+        the same condition as before i.e. T=Dop(A, B) for (A, B, T) in OpRelCone
+        (for an objective that is an increasing function of the eigenvalues) holds,
+        the difference here then, is in how we compute Dop, which is done by
+        faithfully following it's definition as the non-commutative perspective of
+        the negative logarithm as implemented in the Dop_eig/Dop methods
+        """
```
rileyjmurray, line 710: Clarify how this difference affects test code. I believe that in this case we had to rely on CVXQUAD's correctness to get the reference solution. Compare to the commutative tests, where we could reformulate using standard CVXPY atoms that we already trust as correct.

rileyjmurray (changes requested):
An improvement! Just a bit more to go :)

aryamanjeendgar:
@SteveDiamond I had a question regarding what would be a consistent naming convention for these approximation classes that we are implementing, so, we named the Scalar Relative Entropy: `RelEntrQuad`, and this current class has been named `OpRelCone` --- I think it makes a lot of sense to rename the latter to include `Quad` to explicitly indicate the approximate nature of this class's construction, but I was then not sure about whether it should be `OpRelConeQuad` or `OpRelQuad` (if we go with the former, then I think it's only fair to rename `RelEntrQuad` to `RelEntrConeQuad`)

Something also relevant to the above discussion is something that @rileyjmurray mentioned, which was that a recent paper by Fawzi talked about how it is possible to implement the _exact_ operator relative entropy cone, and so we would probably want to have that construction as a part of CVXpy at some point as well (i.e. an actual `OpRelCone` class would also probably exist down the road)

SteveDiamond:
My vote is for ``OpRelConeQuad`` and ``RelEntrConeQuad``, leaving room for ``OpRelCone`` as you said.

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,107 @@ def shape(self) -> Tuple[int, ...]:
...
+    X : Expression
+        x in the (approximate) operator relative entropy cone
+    Y : Expression
+        y in the (approximate) operator relative entropy cone
+    T : Expression
+        T in the (approximate) operator relative entropy cone
+    m: Parameter directly related to the number of generated nodes for the quadrature
+    approximation used in the algorithm
+    k: Another parameter controlling the approximation
+    Number of semidefinite constraints used in constructing this approximation: $(m+k)$
```
rileyjmurray, line 279: Docstrings render with ReStructured Text. So you should use
```
:math:`m+k`
```

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,206 @@ def test_expcone_socp_1(self):
...
+
+class TestOpRelConeQuad(BaseTest):
+
+    @staticmethod
+    def tr_Dop_commute(a, b, U):
+        return np.trace(TestOpRelConeQuad.Dop_commute(a, b, U))
+
+    @staticmethod
+    def Dop_commute(a, b, U):
+        return U @ np.diag(a.value * np.log(a.value/b.value)) @ np.linalg.inv(U)
```
rileyjmurray, line 574: Since ``U`` is orthogonal, it's better to use ``U.T`` instead of ``inv(U)``.

cvxpy/tests/test_cone2cone.py
```diff
@@ -560,3 +561,206 @@ def test_expcone_socp_1(self):
...
+                       [0.02778609, 0.16051504, 0.46046874, 0.87260036]])
+        a_diag = cp.Variable(shape=(n,), pos=True)
+        b_diag = cp.Variable(shape=(n,), pos=True)
+        A = U1 @ cp.diag(a_diag) @ U1.T
+        B = U2 @ cp.diag(b_diag) @ U2.T
+        T = cp.Variable(shape=(n, n), symmetric=True)
+        a_lower = np.array([0.49964878, 1.11493702, 1.69806999, 2.39632391])
+        a_upper = np.array([0.50111539, 1.14133115, 1.69967363, 2.43768103])
+        b_lower = np.array([0.33998621, 1.18669717, 1.43276673, 2.01425798])
+        b_upper = np.array([0.38687005, 1.18908653, 1.43546562, 2.01528889])
```
rileyjmurray, line 734: These upper and lower bounds are really close to one another. Can you try making the gap larger? My reasoning is this: if we only test that the solution matches what we expect up to 2 decimal places, then we should make sure the problem has feasible solutions which differ by more than 2 decimal places.

rileyjmurray (approved):
@cvxpy/maintainers this is ready for review.

cvxpy/tests/test_cone2cone.py
rileyjmurray: Thanks for fixing!

cvxpy/tests/test_cone2cone.py
rileyjmurray: Thanks for fixing!

cvxpy/constraints/exponential.py
rileyjmurray: I went in and tweaked the docstring a bit more.

rileyjmurray:
@aryamanjeendgar flake8 is failing. Please do a ``git pull`` to get my latest commit on this branch. Then fix the style issues that flake8 doesn't like.

rileyjmurray (approved):

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,114 @@ def shape(self) -> Tuple[int, ...]:
...
+        Must be positive. Controls the number of quadrature nodes used in a local
+        approximation of the matrix logarithm. Increasing this value results in
+        better local approximations, but does not significantly expand the region
+        of inputs for which the approximation is effective.
+    k: int
+        Must be positive. Sets the number of scaling points about which the
+        quadrature approximation is performed. Increasing this value will
+        expand the region of inputs over which the approximation is effective.
+
+    This approximation uses :math:`m + k` semidefinte constraints.
```
SteveDiamond, line 286: semidefinite

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,114 @@ def shape(self) -> Tuple[int, ...]:
...
+    """
+
+    def __init__(self, X, Y, Z, m, k, constr_id=None) -> None:
```
SteveDiamond, line 289: should we do types here for the arguments?

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,114 @@ def shape(self) -> Tuple[int, ...]:
...
+        Expression = cvxtypes.expression()
+        self.X = Expression.cast_to_const(X)
+        self.Y = Expression.cast_to_const(Y)
+        self.Z = Expression.cast_to_const(Z)
+        if (not X.is_symmetric()) or (not Y.is_symmetric()) or (not Z.is_symmetric()):
+            msg = ("One of the input matrices to the program has not explicitly been declared"
+                   " to be symmetric, we recommend setting 'symmetric=True' for the same, or it"
+                   " will be done in a less efficient way internally.")
+            warnings.warn(msg)
```
SteveDiamond, line 298: Nice!

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,114 @@ def shape(self) -> Tuple[int, ...]:
...
+
+        Returns
+        -------
+        list
+            A list of the sizes of the elementwise cones.
+        """
+        return [3]*self.num_cones()
+
+    def is_dcp(self, dpp: bool = False) -> bool:
+        """An operator relative conic constraint is DCP when (A, B, C)is affine
```
SteveDiamond, line 345: (A, b, C) is affine

cvxpy/constraints/exponential.py
```diff
@@ -251,3 +252,114 @@ def shape(self) -> Tuple[int, ...]:
...
+        return self.is_dcp()
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        s = (3,) + self.X.shape
+        return s
+
+    def save_dual_value(self, value) -> None:
+        # TODO: implement me.
+        return
```
SteveDiamond, line 365: should be ``pass`` instead of return?

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -273,8 +276,11 @@ def construct_solving_chain(problem, candidates,
                 and (has_constr or not solver_instance.REQUIRES_CONSTR)):
             if ex_cos:
                 reductions.append(Exotic2Common())
-            if approx_cos:
+            if RelEntrQuad in approx_cos:
                 reductions.append(QuadApprox())
+            if OpRelConeQuad in approx_cos:
+                reductions.append(OpRelConeQuadApprox())
```
SteveDiamond, line 282: Why are OpRelConeQuadApprox and QuadApprox separate reductions?

SteveDiamond:
Looks good other than minor comments! Feel free to merge once those are addressed.

cvxpy/constraints/exponential.py
rileyjmurray: We should! @aryamanjeendgar please add type hints.

cvxpy/constraints/exponential.py
rileyjmurray: Not sure, but probably. @aryamanjeendgar please compare to the SOC, PSD, and ExpCone classes for consistency.

cvxpy/reductions/solvers/solving_chain.py
rileyjmurray: This is a good point. @aryamanjeendgar I think it's possible for the ``QuadApprox`` to handle the scalar and operator cases. You can just update
https://github.com/cvxpy/cvxpy/blob/04af387a3e1cf1e4df3c23f02708301d779cbf86/cvxpy/reductions/cone2cone/approximations.py#L131-L139
so that 
```
CANON_METHODS = {
    RelEntrQuad: RelEntrQuad_canon,
    OpRelEntrConeQuad : OpRelEntrConeQuad_canon
}
```.

File: cvxpy/reductions/solvers/solving_chain.py [hunk 25]
```diff
@@ -276,10 +274,8 @@ def construct_solving_chain(problem, candidates,
                 and (has_constr or not solver_instance.REQUIRES_CONSTR)):
             if ex_cos:
                 reductions.append(Exotic2Common())
-            if RelEntrQuad in approx_cos:
+            if RelEntrQuad or OpRelConeQuad in approx_cos:
```
rileyjmurray, line 277: This line isn't doing what you think it should be doing. For example,
```
'' or '1' in ['']
```
is False. The check should be
```
RelEntrQuad in approx_cos or OpRelConeQuad in approx_cos
```

SteveDiamond:
Great job @aryamanjeendgar!!!

PR #1877, opened 2022-08-23
Initial values for Gurobi

SteveDiamond:
## Description
This adds the ability to specify an initial guess for Gurobi from #1664.
Issue link (if applicable): #1664 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/utilities.py
+++ b/cvxpy/reductions/solvers/utilities.py
@@ -21,13 +21,14 @@ import numpy as np
 import cvxpy.interface as intf
 
 
-def stack_vals(variables: list, default: float) -> np.ndarray:
+def stack_vals(variables: list, default: float, order: str = "C") -> np.ndarray:
     """Stacks the values of the given variables.
 
     Parameters
     ----------
     variables: list of cvxpy variables.
     default: value to use when variable value is None.
+    order: unravel values in C or Fortran ("F") order
 
     Returns
     -------
@@ -36,7 +37,7 @@ def stack_vals(variables: list, default: float) -> np.ndarray:
     value = []
     for variable in variables:
         if variable.value is not None:
-            value.append(np.ravel(variable.value, 'C'))
+            value.append(np.ravel(variable.value, order))
         else:  # unknown values.
             value.append(np.full(variable.size, default))
     return np.concatenate(value)
--- a/cvxpy/reductions/solvers/utilities.py
+++ b/cvxpy/reductions/solvers/utilities.py
@@ -21,7 +21,7 @@ import numpy as np
 import cvxpy.interface as intf
 
 
-def stack_vals(variables: list, default: float, order: str = "C") -> np.ndarray:
+def stack_vals(variables: list, default: float, order: str = "F") -> np.ndarray:
     """Stacks the values of the given variables.
 
     Parameters
--- a/cvxpy/reductions/solvers/utilities.py
+++ b/cvxpy/reductions/solvers/utilities.py
@@ -21,6 +21,18 @@ import numpy as np
 import cvxpy.interface as intf
 
 
+def stack_vals(variables, default):
+    """Stacks the values of the given variables.
+    """
+    value = []
+    for variable in variables:
+        if variable.value is not None:
+            value.append(np.ravel(variable.value, 'C'))
+        else:  # unknown values.
+            value.append(np.full(variable.size, default))
+    return np.concatenate(value)
+
+
 def expcone_permutor(n_cones, exp_cone_order) -> np.ndarray:
     order = np.tile(np.array(exp_cone_order), n_cones)  # e.g. [1,0,2, 1,0,2, 1,0,2,...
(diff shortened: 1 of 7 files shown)


SteveDiamond:
FYI @aszekMosek two of the mosek tests are failing due to the new v10 release.

SteveDiamond:
This is ready for review. The test failures are due to a new mosek release.

rileyjmurray:
@SteveDiamond the failures for MOSEK are rather small precision issues; nothing too crazy. We should look into it but there's a small chance that there's an issue on our end.

SteveDiamond:
@phschiele @rileyjmurray can we review and merge this? It has a fix for the failing mosek tests.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -1165,6 +1165,13 @@ def test_gurobi_warm_start(self) -> None:
             self.assertEqual(result, 4)
             self.assertItemsAlmostEqual(self.x.value, [1, 2])
 
+            # Try creating a new problem and setting x.value.
+            self.x.value = np.array([1, 2])
+            prob = cp.Problem(objective, constraints)
+            result = prob.solve(solver=cp.GUROBI, warm_start=True)
+            self.assertEqual(result, 4)
+            self.assertItemsAlmostEqual(self.x.value, [1, 2])
+
```
rileyjmurray, line 1174: I don't think we can rely on this test to verify correctness of the implementation. At minimum I'd try setting ``self.x.value`` to something that's very far from optimal and see if the solver takes more iterations. Then you can try initializing at the optimal value and seeing if the solver takes fewer iterations.

cvxpy/tests/test_qp_solvers.py
```diff
@@ -445,7 +446,21 @@ def test_warm_start(self) -> None:
...
+            m = 20
+            n = 10
+            np.random.seed(1)
+            A = np.random.randn(m, n)
+            b = np.random.randn(m)
+
+            x = Variable(n, boolean=True)
+            prob = Problem(Minimize(sum_squares(A @ x - b)))
+            x.value = np.ones(n)
+            prob.solve(solver=cp.GUROBI, warm_start=True)
```
rileyjmurray, line 463: I have a similar feeling about this test and the last test; they should both be structured so that we're confident that options are getting passed correctly. I know it's a pain to write that kind of test, but I want to avoid situations like with our CBC interface (where it turns out we've been silently failing to set solver options for a long time).

In this case I would generate a problem where ``A`` is wide, ``b = A @ np.random.randint(-10, 10, n)``, and ``x.value = np.ones(n)``. If ``A`` is wide enough then I think this problem will be hard to solve, since most convex relaxations encountered in branch-and-bound will be able to achieve $Ax=b$, but probably not with integer solutions.

cvxpy/reductions/solvers/utilities.py
```diff
@@ -21,6 +21,27 @@
...
+    default: value to use when variable value is None.
+
+    Returns
+    -------
+    An initial guess for the solution.
+    """
+    value = []
+    for variable in variables:
+        if variable.value is not None:
+            value.append(np.ravel(variable.value, 'C'))
```
phschiele, line 39: Do all solvers require to ravel the variables in 'C' order?

rileyjmurray (changes requested):
I left feedback for the tests.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -1165,6 +1165,13 @@ def test_gurobi_warm_start(self) -> None:
             self.assertEqual(result, 4)
             self.assertItemsAlmostEqual(self.x.value, [1, 2])
 
+            # Try creating a new problem and setting x.value.
+            self.x.value = np.array([1, 2])
+            prob = cp.Problem(objective, constraints)
+            result = prob.solve(solver=cp.GUROBI, warm_start=True)
+            self.assertEqual(result, 4)
+            self.assertItemsAlmostEqual(self.x.value, [1, 2])
```
phschiele, line 1173: It looks like this test is using the qp interface.

cvxpy/reductions/solvers/utilities.py
```diff
@@ -21,6 +21,28 @@
 import cvxpy.interface as intf
 
 
+def stack_vals(variables: list, default: float, order: str = "C") -> np.ndarray:
```
phschiele, line 24: @SteveDiamond Thanks for adding the order argument!
Just wondering: Where does the C order for Gurobi come from? I would have thought to match the columns of A, F, and P, we would need to ravel in Fortran order, too.

cvxpy/tests/test_conic_solvers.py
phschiele: @rileyjmurray Good point. I played around with your suggestion, but the initial value did not affect the iteration count in my (short) testing. 

https://www.gurobi.com/documentation/9.5/refman/start.html suggests that `Start` only affects MIP problems.

SteveDiamond:
Are we good with this PR? I accidentally started the new SCIP interface from this PR, so need to merge this in first.

rileyjmurray (approved):

PR #1885, opened 2022-08-31
Fix tests that started failing with Mosek 10

SteveDiamond:
## Description
Fixes tests that started failing once Mosek 10 was deployed.
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/solver_test_helpers.py
+++ b/cvxpy/tests/solver_test_helpers.py
@@ -491,7 +491,7 @@ def pcp_1() -> SolverTestHelper:
                    x[0] + x[1] + 3 * x[2] >= 1.0,
                    y_square <= 25]
     obj = cp.Minimize(3 * x[0] + 2 * x[1] + x[2])
-    expect_x = np.array([-3.874621860638774, -2.129788233677883, 2.33480343377204])
+    expect_x = np.array([-3.87468642, -2.12968504, 2.33479049])
     expect_epis = expect_x ** 2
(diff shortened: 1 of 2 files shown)


phschiele (commented):
LGTM!

rileyjmurray:
@SteveDiamond one thing stands out to me. This power cone test has the same optimal solution as this SOCP test:
https://github.com/cvxpy/cvxpy/blob/83b684ccae0c4f514eacb2ffd6f4d9b7af597ba9/cvxpy/tests/solver_test_helpers.py#L268-L295
So actually changing the expected optimal values doesn't seem appropriate.

SteveDiamond:
I wasn't aware of that. If I keep the old optimal values, I need to do
```
    def test_mosek_pcp_1(self) -> None:
        StandardTestPCPs.test_pcp_1(solver='MOSEK', places=2)
```
That might be better, not sure.

rileyjmurray:
Hmm. As an initial step how about we use MOSEK to solve ``socp_1`` and ``pcp_1`` with verbose=True, and see how the outputs differ. I think that the optimal solution is unique since there's a binding 2-norm constraint. But if both solver logs look good then we can try visualizing the feasible set; it might be that there's a neighborhood of very-close-to-optimal solutions for this problem. If that's the case then it might make sense to change the problem data to avoid that.

Ping @aszekMosek, in case you have thoughts.

PR #1888, opened 2022-09-08
Interface to Clarabel.rs solver

goulart-paul:
## Description
Implements cvxpy interface to the Clarabel solver.   
Main repo for the solver:  https://github.com/oxfordcontrol/Clarabel.rs
Documentation, including native python interface : https://oxfordcontrol.github.io/ClarabelDocs/stable/

## Type of change
- [x] New feature (backwards compatible)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

Supports nonnegative, second order, exponential and power cones with linear or quadratic objectives.   

Binaries are available here: https://pypi.org/project/clarabel/

PR #1893, opened 2022-09-18
Add clarabel to cvxpy/__init__.py

phschiele:
## Description
Tiny follow up to #1888 , allowing the usage `cp.CLARABEL`.

Also just a small note on adding CLARABEL at index 0 of `CONIC_SOLVERS` in `solvers/defines.py`:
This implicitly makes CLARABEL (if installed) the default solver for supported problems. As soon as CLARABEL is being 
added as a dependency, this _could_ cause some regressions for users simply calling `.solve()`, which we should be aware of.

Again, many thanks to @goulart-paul for the great contribution.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
Good catch about the position of clarabel in defines. Could you move it to after ECOS?

SteveDiamond (approved):
Thanks for catching this!

PR #1894, opened 2022-09-19
Interface to proxqp solver

fabinsch:
## Description
Implements cvxpy interface to the proxqp solver.
Main repo for the solver: https://github.com/simple-robotics/proxsuite
Documentation: https://simple-robotics.github.io/proxsuite/

## Type of change
- [ ] Bug fix

- [x] Write unittests (unittests for proxqp solver are run by `test_qp_solvers.py`).


cvxpy/reductions/solvers/qp_solvers/proxqp_qpif.py
```diff
@@ -0,0 +1,159 @@
...
+
+class PROXQP(QpSolver):
+    """QP interface for the PROXQP solver"""
+
+    MIP_CAPABLE = False
+
+    # Map of Proxqp status to CVXPY status.
+    STATUS_MAP = {"PROXQP_SOLVED": s.OPTIMAL,
+                  "PROXQP_MAX_ITER_REACHED": s.SOLVER_ERROR,
+                  "PROXQP_PRIMAL_INFEASIBLE": s.SOLVER_ERROR,
```
SteveDiamond, line 34: Shouldn't this be INFEASIBLE? Similarly DUAL_INFEASIBLE should be UNBOUNDED

SteveDiamond:
This PR looks great! Seems like the CI failures are due to install issues.

fabinsch:
> This PR looks great! Seems like the CI failures are due to install issues.

Hello, thank you. Yes, currently it is possible to install proxqp via pip only for macOS and Linux with python >+= 3.8, and for windows we are using conda. Therefor, we have some pipelines that are failing for the install. 

The first issue py>=3.8 should be solved in the next days, we will be able to get wheels for python3.7 quite easily. However, for windows it can take more time with our current setup..

cvxpy/reductions/solvers/qp_solvers/proxqp_qpif.py
fabinsch: You're right, thanks. I changed it accordingly.

cvxpy/reductions/solvers/qp_solvers/proxqp_qpif.py
```diff
@@ -31,8 +31,8 @@ class PROXQP(QpSolver):
     # Map of Proxqp status to CVXPY status.
     STATUS_MAP = {"PROXQP_SOLVED": s.OPTIMAL,
                   "PROXQP_MAX_ITER_REACHED": s.SOLVER_ERROR,
```
SteveDiamond, line 33: This should be USER_LIMIT. sorry I missed this.

fabinsch:
Hey @SteveDiamond, this PR is ready now. All the checks are passing.

doc/source/tutorial/advanced/index.rst
```diff
@@ -745,6 +751,29 @@ Here is the complete list of solver options.
 
 For others see `OSQP documentation <https://osqp.org/docs/interfaces/solver_settings.html>`_.
 
+`PROXQP`_ options:
+
+``'backend'``
+    solver backend [dense, sparase] (default: dense).
```
SteveDiamond, line 757: typo: sparse

cvxpy/reductions/solvers/qp_solvers/proxqp_qpif.py
```diff
@@ -0,0 +1,159 @@
...
+    def name(self):
+        return s.PROXQP
+
+    def import_solver(self) -> None:
+        import proxsuite
+        proxsuite
+
+    def invert(self, solution, inverse_data):
+        attr = {s.SOLVE_TIME: solution.info.run_time}
+        attr[s.EXTRA_STATS] = solution
```
SteveDiamond, line 54: Could we store this as ``{"solution": solution}``? This will make it easier to update EXTRA_STATS in the future.

SteveDiamond:
Nice job @fabinsch! Looks good to me other than minor comments.

fabinsch:
> Nice job @fabinsch! Looks good to me other than minor comments.

Perfect, I pushed the changes, thanks for reviewing.

cvxpy/reductions/solvers/qp_solvers/proxqp_qpif.py
```diff
@@ -51,7 +51,7 @@ def import_solver(self) -> None:
 
     def invert(self, solution, inverse_data):
         attr = {s.SOLVE_TIME: solution.info.run_time}
-        attr[s.EXTRA_STATS] = solution
+        attr["solution"] = solution
```
SteveDiamond, line 54: Sorry this is not quite what I meant. I meant you should change it to this:
``attr[s.EXTRA_STATS] = {"solution": solution}``

fabinsch: I'm sorry, I missunderstood. I fixed it.

SteveDiamond (approved):

jcarpent:
Thanks @fabinsch for adding this interface to ProxQP.
@SteveDiamond Do you have a timeline in mind to merge this PR and release a new version of CVXPY including this contribution?

rileyjmurray:
@cvxpy/maintainers if someone else approves this then we can merge it.

@jcarpent it will be in CVXPY 1.3.0, which we hope to release very soon.

phschiele:
@fabinsch Thanks for this great contribution!

I was able to install it without any problems, solve some test problems, set the options, and build the docs locally.
Also double-checked that the tests are being run.

One question I had was why the backend is set to 'dense' as a default.

Bambade:
Thanks a lot @fabinsch! 
@phschiele, thanks also for the question. We plan to release soon another sparse backend. So we think it is better to set it to "dense" by default, while both can of course be used.

rileyjmurray:
@Bambade just a heads up: cvxpy always represents problems with sparse data. It seems from [this example of yours](https://github.com/Simple-Robotics/proxsuite/blob/main/examples/python/init_dense_qp_with_other_options.py) that you internally convert sparse data to dense data when using the dense backend. Is that true? If so, that's very cool!

Separately: if proxqp is going to use the dense backend by default then we should be very clear about this. I would state as much on the page with installation instructions.

Bambade:
Thanks for this precision @rileyjmurray. If cvxpy does only sparse matrices we will convert them ourselves in the wrapping of cvxpy when using proxqp dense backend. 

To be more precise about it: we won't be able anymore to convert sparse to dense inside proxsuite dense backend because we found it’s slowing down the performance. Thus we expect the user to use it properly: dense matrices for dense backend and sparse matrices for sparse backend. It will be precised in proxqp documentation at its next release. It is indeed important to mention as you point out !

SteveDiamond:
@Bambade is the new proxqp interface out? I'm seeing CI failures in the ProxQP tests that weren't there before.

fabinsch:
@SteveDiamond I adapted the proxqp interface according to the changes that @Bambade explained (introduced in [proxsuite 0.2.0](https://github.com/Simple-Robotics/proxsuite/releases/tag/v0.2.0)). Now the CI is passing again without any failures.

fabinsch:
Hello @SteveDiamond and cvxpy maintainers, are we good to merge? Thanks a lot for taking the time to consider this PR.

SteveDiamond:
Yes we are good. Sorry for the delay. We also hope to have a new release of cvxpy out before too long.

PR #1896, opened 2022-09-23
Allow setting use_quad_obj=False for Clarabel

phschiele:
## Description
When setting `use_quad_obj=False` a `TypeError` is raised for Clarabel. 
Like in the SCS interface, we need to delete the key before parsing the options.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond (approved):
Good catch!

PR #1897, opened 2022-09-23
New example: Kurtosis portfolio optimization

dcajasn:
## Description
This pull request adds a new example that shows how to optimize the kurtosis of an investment portfolio using DCP rules as shown in __[Cajas (2022)](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=4202967)__.

## Type of change
- [ x] Other (Example)


SteveDiamond:
Looks good! My only request is that the matrix M3 be assembled with ``bmat`` instead of in pieces with multiple vstacks. Otherwise looks great!

dcajasn:
Hi @SteveDiamond, I made the change using `bmat` instead of `vstack` and `hstack`.

SteveDiamond:
Great, thanks!

Issue #1889, opened 2022-09-10
Unable to link cvxpy to SCIP


Mark531:
**Describe the bug**
I need to solve a non-linear MIP problem, so I installed pyscipopt. Though, after importing the package, cvxpy still cannot find the SCIP solver.

**To Reproduce**
```
import pyscipopt
import numpy as np
import cvxpy as cp

x = np.arange(100)
y = 5 * x**2 - 3 * x + 4
v = cp.Variable(integer=True)
problem = cp.Problem(cp.Minimize(cp.sum_squares(y - v)))
problem.solve(solver=cp.SCIP)
```

**Expected behavior**
The problem should be solved by SCIP.

**Output**
> SolverError: 
> 
>                     CVXPY needs additional software (a `mixed-integer solver`) to handle this model.
>                     The web documentation
>                         https://www.cvxpy.org/tutorial/advanced/index.html#mixed-integer-programs
>                     reviews open-source and commercial options for mixed-integer solvers.
> 
>                     Quick fix: if you install the python package CVXOPT (pip install cvxopt),
>                     then CVXPY can use the open-source mixed-integer solver `GLPK`.

**Version**
 - OS: Windows 10 64-bits
 - CVXPY Version: 1.1.3

rileyjmurray:
What version of pyscipopt are you using? Per our documentation, we don't support pyscipopt 4 or higher: https://www.cvxpy.org/install/index.html#install-with-scip-support. (Our reason for this is that there is a correctness bug in SCIP 8, which is the only version of SCIP available with pyscipopt.

Mark531:
You're right, I have installed the 4.2.0 version. I will downgrade the package, I just hope we won't lose important bug corrections...

Mark531:
So, I installed the version 3.5.0 of pyscipopt and running the program above gives the very same error...

SteveDiamond:
This is the code being run by CVXPY to verify that pyscipopt is installed:
```
from cvxpy.utilities.versioning import Version
import pyscipopt
v = pyscipopt.__version__
if Version(v) >= Version('4.0.0'):
    msg = 'PySCIPOpt (SCIP\'s Python wrapper) is installed and its ' \
          'version is %s. CVXPY only supports PySCIPOpt < 4.0.0.' % v
    raise ModuleNotFoundError(msg)
```

Mark531:
Running your code triggers this error:

[...]

PR #1898, opened 2022-09-26
New SCIP interface

SteveDiamond:
## Description
This interface works with the latest pyscipopt and SCIP. It does not provide dual variables for SOCPs, only for LPs.
Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/1889

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -31,7 +31,7 @@ fi
 
 # SCIP only works with scipy >= 1.5 due to dependency conflicts when installing on Linux/macOS
 if [[ "$PYTHON_VERSION" == "3.9" ]] || [[ "$RUNNER_OS" == "Windows" ]]; then
-  conda install pyscipopt  # TODO: update interface https://github.com/cvxpy/cvxpy/pull/1628
+  conda install pyscipopt
 fi
 
 # Only install Mosek if license is available (secret is not copied to forks)
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -31,7 +31,7 @@ fi
 
 # SCIP only works with scipy >= 1.5 due to dependency conflicts when installing on Linux/macOS
 if [[ "$PYTHON_VERSION" == "3.9" ]] || [[ "$RUNNER_OS" == "Windows" ]]; then
-  conda install pyscipopt"<4.0"  # TODO: update interface https://github.com/cvxpy/cvxpy/pull/1628
+  conda install pyscipopt  # TODO: update interface https://github.com/cvxpy/cvxpy/pull/1628
 fi
 
 # Only install Mosek if license is available (secret is not copied to forks)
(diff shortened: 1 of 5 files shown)


continuous_integration/install_dependencies.sh
```diff
@@ -31,7 +31,7 @@ fi
 
 # SCIP only works with scipy >= 1.5 due to dependency conflicts when installing on Linux/macOS
 if [[ "$PYTHON_VERSION" == "3.9" ]] || [[ "$RUNNER_OS" == "Windows" ]]; then
-  conda install pyscipopt"<4.0"  # TODO: update interface https://github.com/cvxpy/cvxpy/pull/1628
+  conda install pyscipopt  # TODO: update interface https://github.com/cvxpy/cvxpy/pull/1628
```
phschiele, line 34:
```suggestion
  conda install pyscipopt
```
This comment is resolved by this PR.

phschiele (approved):
- Tested that interface works with `pyscipopt 3.5` (`scip 7`) as well as `pyscipopt 4.2` (`scip 8`).
- Had to change to `places=2` for `test_scip_socp_1` when using `4.2`. Perhaps we should loosen the tolerance to ensure its passing for all users (can't make the change here directly short of creating a separate PR pointing to this one). 
- docs are building
- coverage is looking good (97%), but not shown here because we can't install `pysciopt` on the mac os/3.8, which is running the coverage report.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -265,16 +260,20 @@ def _set_params(
...
-        model.setPresolve(SCIP_PARAMSETTING.OFF)
-        model.setHeuristics(SCIP_PARAMSETTING.OFF)
-        model.disablePropagation()
+        is_mip = data[s.BOOL_IDX] or data[s.INT_IDX]
+        has_soc_constr = len(dims[s.SOC_DIM]) > 1
+        if not (is_mip or has_soc_constr):
+            # These settings are needed  to allow the dual to be calculated
+            model.setPresolve(SCIP_PARAMSETTING.OFF)
+            model.setHeuristics(SCIP_PARAMSETTING.OFF)
+            model.disablePropagation()
```
phschiele, line 276: There could be an edge case where these settings are overwritten by the user later.

rileyjmurray: @phschiele could we raise a warning if the user tries to change either of these values?

SteveDiamond: Isn't the model object created fresh each time we solve the problem though? Or are these settings globals?

phschiele: I was referring to the case where the user specifies these options as kwargs which are parsed a few lines below. Raising a warning as Riley suggested would be good imo.

PR #1899, opened 2022-09-27
Approximate canonicalization for `von_neumann_entr`

aryamanjeendgar:
This PR implements an alternate, approximate canonicalization pathway for the `von_neumann_entr` atom. 
I adapted the same series of semi-definite approximations that I have been implementing so far for the previous `Constraint` classes to get a representation for `von_neumann_entr`. 

The API for `von_neumann_entr` remains unchanged --- if the user wants to employ this approximate formulation in their program, then they can just pass in the usual parameters for the approximation along with the input matrix, like so: `atom=von_neumann_entr(N, (m, k))`.

Some documentation in `von_neumann_entr`'s docstring still remains, but other than that, this is mostly complete!

Thanks!

### DIFF
--- a/cvxpy/atoms/von_neumann_entr.py
+++ b/cvxpy/atoms/von_neumann_entr.py
@@ -101,7 +101,7 @@ class von_neumann_entr(Atom):
         return False
 
     def get_data(self):
-        return self.quad_approx
+        return [self.quad_approx]
 
     def _grad(self, values):
         """Gives the (sub/super)gradient of the atom w.r.t. each argument.
--- a/cvxpy/atoms/von_neumann_entr.py
+++ b/cvxpy/atoms/von_neumann_entr.py
@@ -100,6 +100,9 @@ class von_neumann_entr(Atom):
         """
         return False
 
+    def get_data(self):
+        return self.quad_approx
+
     def _grad(self, values):
         """Gives the (sub/super)gradient of the atom w.r.t. each argument.
 
--- a/cvxpy/atoms/von_neumann_entr.py
+++ b/cvxpy/atoms/von_neumann_entr.py
@@ -49,8 +49,9 @@ class von_neumann_entr(Atom):
         A PSD matrix
     """
 
-    def __init__(self, X) -> None:
+    def __init__(self, X, quad_approx: Tuple[int, int] = ()) -> None:
         # TODO: add a check that N is symmetric/Hermitian.
+        self.quad_approx = quad_approx
         super(von_neumann_entr, self).__init__(X)
 
     def numeric(self, values):
--- a/cvxpy/reductions/cone2cone/approximations.py
+++ b/cvxpy/reductions/cone2cone/approximations.py
@@ -167,10 +167,11 @@ def OpRelConeQuad_canon(con: OpRelConeQuad, args) -> Tuple[Constraint, List[Cons
 def von_neumann_entr_QuadApprox(expr, args):
     N, m, k = args[0], expr.quad_approx[0], expr.quad_approx[1]
     n = N.shape[0]
-    t = Variable(shape=N.shape)
+    t = Variable(shape=N.shape, symmetric=True)
     con = OpRelConeQuad(N, cp.Constant(np.eye(n)), t, m, k)
     lead_con, cons = OpRelConeQuad_canon(con, con.args)
-    return -cp.trace(con.Z), ([lead_con] + cons)
+    cons.append(lead_con)
+    return -cp.trace(con.Z), cons
 
 
 def von_neumann_entr_canon_dispatch(expr, args):
--- a/cvxpy/reductions/cone2cone/approximations.py
+++ b/cvxpy/reductions/cone2cone/approximations.py
@@ -25,6 +25,8 @@ from cvxpy.constraints.exponential import OpRelConeQuad, RelEntrQuad
 from cvxpy.constraints.zero import Zero
 from cvxpy.expressions.variable import Variable
 from cvxpy.reductions.canonicalization import Canonicalization
+from cvxpy.reductions.dcp2cone.atom_canonicalizers.von_neumann_entr_canon import (
+    von_neumann_entr_canon,)
 
 APPROX_CONES = {
     RelEntrQuad: {cp.SOC},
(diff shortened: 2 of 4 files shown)


cvxpy/atoms/von_neumann_entr.py
```diff
@@ -49,8 +49,9 @@ class von_neumann_entr(Atom):
         A PSD matrix
     """
 
-    def __init__(self, X) -> None:
+    def __init__(self, X, quad_approx: Tuple[int, int] = ()) -> None:
         # TODO: add a check that N is symmetric/Hermitian.
+        self.quad_approx = quad_approx
```
SteveDiamond, line 54: I think you need a get_data function that returns quad_approx, like L208 of pnorm.py

SteveDiamond: https://github.com/cvxpy/cvxpy/blob/03fe861fd3450d67445d657cb9a27bcc496e5030/cvxpy/atoms/pnorm.py#L208

aryamanjeendgar: Ah, sorry for missing that! Have taken care of it in the most recent commit!

cvxpy/reductions/cone2cone/approximations.py
```diff
@@ -162,6 +164,22 @@ def OpRelConeQuad_canon(con: OpRelConeQuad, args) -> Tuple[Constraint, List[Cons
     return lead_con, constrs
 
 
+def von_neumann_entr_QuadApprox(expr, args):
+    N, m, k = args[0], expr.quad_approx[0], expr.quad_approx[1]
+    n = N.shape[0]
+    t = Variable(shape=N.shape)
```
rileyjmurray, line 170: My dude, you personally wrote the warning message that we should provide symmetric input to ``OpRelConeQuad``!

Declare ``t = cp.Variable(shape=(n, n), symmetric=True)``.

cvxpy/reductions/cone2cone/approximations.py
```diff
@@ -162,6 +164,22 @@ def OpRelConeQuad_canon(con: OpRelConeQuad, args) -> Tuple[Constraint, List[Cons
...
+    con = OpRelConeQuad(N, cp.Constant(np.eye(n)), t, m, k)
+    lead_con, cons = OpRelConeQuad_canon(con, con.args)
+    return -cp.trace(con.Z), ([lead_con] + cons)
```
rileyjmurray, line 173: do
```
cons.append(lead_con)
return -cp.trace(con.Z), cons
```

cvxpy/tests/test_von_neumann_entr.py
```diff
@@ -203,3 +203,91 @@ def test_4(self):
         sth.verify_objective(places=3)
         sth.verify_primal_values(places=3)
         sth.check_primal_feasibility(places=3)
+
+    @staticmethod
+    def make_test_5():
+        """Test#1 for the approximate canonicalization"""
```
rileyjmurray, line 209: Describe the changes you made relative to the original test. How did you compute the reference solution in this case? If the only difference is that you provide ``quad_approx=(5,5)`` I suggest that you create a helper function like 
```
    def build_test1(quad_approx=False):
        n = 3
        N = cp.Variable(shape=(n, n), PSD=True)
        expect_N = np.array([[0.14523781, 0.02381009, 0.10238067],
                             [0.02381009, 0.28095125, 0.13809581],
                             [0.10238067, 0.13809581, 0.37380924]])
        v1 = np.array([[-0.26726124],
                       [-0.53452248],
                       [-0.80178373]])
        v2 = np.array([[0.87287156],
                       [0.21821789],
                       [-0.43643578]])
        lambda1 = 0.5
        lambda2 = 0.1
        trMax = 0.8
        cons1 = N @ v1 == lambda1 * v1
        cons2 = N @ v2 == lambda2 * v2
        cons3 = trace(N) <= trMax
        if quad_approx is not None:
            quad_approx_params = (5, 5)
            objective = cp.Maximize(von_neumann_entr(N, quad_approx_params))
        else:
            objective = cp.Maximize(von_neumann_entr(N))
        obj_pair = (objective, 0.8987186478352693)
        con_pairs = [
            (cons1, None),
            (cons2, None),
            (cons3, None)
        ]
        var_pairs = [
            (N, expect_N)
        ]
        sth = STH.SolverTestHelper(obj_pair, var_pairs, con_pairs)
        return sth
```
Then call this function in two different contexts for running actual tests.

cvxpy/tests/test_von_neumann_entr.py
```diff
@@ -203,3 +203,91 @@ def test_4(self):
...
+        return sth
+
+    def test_5(self):
+        sth = Test_von_neumann_entr.make_test_5()
+        sth.solve(**self.SOLVE_ARGS)
+        sth.verify_objective(places=3)
+        sth.verify_primal_values(places=3)
+
+    @staticmethod
+    def make_test_6():
```
rileyjmurray, line 247: My first comment here is that my instinct is to modify this along the lines of my suggestions for ``make_test_5``. However, I worry that this test might be unnecessarily expensive since it uses (8, 8) for the quad_approx parameters. Can you use smaller values, like ``(3, 2)``, and compute the reference solution using CVXQUAD?

rileyjmurray (changes requested):

cvxpy/reductions/cone2cone/approximations.py
aryamanjeendgar: Big oopsie! Took care of this!

cvxpy/tests/test_von_neumann_entr.py
aryamanjeendgar: These test cases have readily computable solutions, so we don't need CVXQUAD for the references -- so I stuck with the same values as before for these.

cvxpy/tests/test_von_neumann_entr.py
aryamanjeendgar: The `(8,8)` was mostly just out of curiosity XD, I changed it to a much more reasonable `(3,2)`.

cvxpy/reductions/cone2cone/approximations.py
aryamanjeendgar: Changed!

cvxpy/tests/test_von_neumann_entr.py
rileyjmurray: @aryamanjeendgar can you please make this change?

aryamanjeendgar: @rileyjmurray Done!

rileyjmurray (approved):
@cvxpy/maintainers this is ready for review

cvxpy/atoms/von_neumann_entr.py
```diff
@@ -99,6 +100,9 @@ def is_decr(self, idx) -> bool:
         """
         return False
 
+    def get_data(self):
+        return self.quad_approx
```
SteveDiamond, line 104: return [self.quad_approx]

SteveDiamond (approved):
LGTM other than minor issue with get_data

PR #1909, opened 2022-10-13
Import more constraints at top level

SteveDiamond:
## Description
There was some discussion recently about how FiniteSet should be importable at the top level. I was making a PR to address that, and noticed many other constraints are not importable at the top level. This PR imports more constraints in the top level ``__init__`` file.

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


phschiele:
@SteveDiamond IMO the `Constraint` class itself should also be importable directly.
I often find myself writing `from cvxpy.constraints.constraint import Constraint` (e.g. for type hinting).

phschiele:
@SteveDiamond One thing that IMO should be "public API" is the retrieval of CVXPY cone form, which should neither require specifying a solver nor using private functions (as currently done in #1867).
My reasoning is that:
1. It's a commonly used and powerful feature (e.g. used for perspective, upcoming saddle point package, @TobiaMarcucci 's research)
2. It's already documented: https://www.cvxpy.org/tutorial/advanced/index.html#getting-the-standard-form

It is not a constraint, so it maybe should be a separate PR, happy to contribute to it.

(@PTNobel)

SteveDiamond:
> @SteveDiamond One thing that IMO should be "public API" is the retrieval of CVXPY cone form, which should neither require specifying a solver nor using private functions (as currently done in #1867). My reasoning is that:
> 
> 1. It's a commonly used and powerful feature (e.g. used for perspective, upcoming saddle point package, @TobiaMarcucci 's research)
> 2. It's already documented: https://www.cvxpy.org/tutorial/advanced/index.html#getting-the-standard-form
> 
> It is not a constraint, so it maybe should be a separate PR, happy to contribute to it.
> 
> (@PTNobel)

Sounds good, it should be a separate PR though. I would probably do this by making a function that produces a ``ParamConeProg`` or ``ParamQuadProg`` as the final output. If it's possible I'd like this to not be more of an experimental API, meaning we can change it without doing a new major version. These intermediate objects haven't been designed with the idea that they'll be part of an API.

rileyjmurray (approved):

rileyjmurray:
> > @SteveDiamond One thing that IMO should be "public API" is the retrieval of CVXPY cone form, which should neither require specifying a solver nor using private functions (as currently done in #1867). My reasoning is that:
> > 
> > 1. It's a commonly used and powerful feature (e.g. used for perspective, upcoming saddle point package, @TobiaMarcucci 's research)
> > 2. It's already documented: https://www.cvxpy.org/tutorial/advanced/index.html#getting-the-standard-form
> > 
> > It is not a constraint, so it maybe should be a separate PR, happy to contribute to it.
> > (@PTNobel)
> 
> Sounds good, it should be a separate PR though. I would probably do this by making a function that produces a `ParamConeProg` or `ParamQuadProg` as the final output. If it's possible I'd like this to be an experimental API, meaning we can change it without doing a new major version. These intermediate objects haven't been designed with the idea that they'll be part of an API.

I like the idea of using a ``cvxpy.experimental`` namespace for these sorts of things.

SteveDiamond:
Merging. Test failures are due to a known mosek issue.

PR #1917, opened 2022-10-20
nonnegative check for harmonic mean and small variable/string fixes

xhulianoThe1:
## Description
Harmonic mean should always have positive entry so added error if it doesn't. If there's a better way to check based on the current code structure let me know. Also, small variable/string fixes.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
Thanks! Atoms that require non-negative inputs still work with expressions of unknown sign. They just impose the constraint that their input is >= 0. So I'm not sure the approach here is what we want.

xhulianoThe1:
Closing for now, will re-vist when I have time

PR #1936, opened 2022-11-07
Fix backend workflow

phschiele:
## Description
Get the new backend workflow working (it only executed the first time after merging the backend PR).

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #1937, opened 2022-11-07
[CI] Exclude failing SCIPY MIP test

phschiele:
## Description
When running the test suite with SciPy 1.9, the `TestMIPVariable` test fails, blocking #1936.
~We should probably investigate the root cause beyond excluding it from the tests, happy to create an issue for tracking this investigation.~ See #1938

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_mip_vars.py
+++ b/cvxpy/tests/test_mip_vars.py
@@ -40,7 +40,7 @@ class TestMIPVariable(BaseTest):
         for solver in self.solvers:
             self.bool_prob(solver)
             if solver != cp.SCIPY:
-                self.int_prob(solver)
+                self.int_prob(solver)  # issue #1938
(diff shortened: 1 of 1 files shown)


cvxpy/tests/test_mip_vars.py
```diff
@@ -39,7 +39,8 @@ def setUp(self) -> None:
     def test_all_solvers(self) -> None:
         for solver in self.solvers:
             self.bool_prob(solver)
-            self.int_prob(solver)
+            if solver != cp.SCIPY:
+                self.int_prob(solver)
```
phschiele, line 43:
```suggestion
                self.int_prob(solver)  # issue #1938
```

SteveDiamond (approved):

PR #1951, opened 2022-11-21
Pin setuptools version to fix installation with pip

### LINKED ISSUE #1950: Package no longer installable from pip
**Describe the bug**
cvxpy fails to build when running `pip install cvxpy` on a platform that doesn't supply binaries for the scs dependency.

**To Reproduce**
On an arm mac (M1/M2) in a fresh virtual environment run `pip install cvxpy`

**Expected behavior**
Pip is expected to either download or build the dependencies.

**Output**
We're greeted with

```
Processing /Users/user/dependencies/cvxpy
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
....
Collecting scs>=1.1.6
  Using cached scs-3.2.2.tar.gz (663 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [62 lines of output]
....
          from numpy.distutils import log
        File "/private/var/folders/pm/97sf8nsj4vz8kvg14d9ymn3r0000gn/T/pip-build-env-kgudujw0/overlay/lib/python3.10/site-packages/numpy/distutils/log.py", line 4, in <module>
          from distutils.log import Log as old_Log
      ImportError: cannot import name 'Log' from 'distutils.log' (/private/var/folders/pm/97sf8nsj4vz8kvg14d9ymn3r0000gn/T/pip-build-env-kgudujw0/overlay/lib/python3.10/site-packages/setuptools/_distutils/log.py)
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

**Version**
 - OS: macOS Ventura
 - CVXPY Version: 1.2.2

**Additional context**
The `pip install` approach worked on Friday and was broken due to this https://github.com/pypa/setuptools/issues/3693 issue.
A possible solution would be to pin the setuptools version to <60 per [this comment](https://github.com/numpy/numpy/issues/22623#issuecomment-1321012633)

piiq:
## Description
This PR pins the setuptools version to the latest version that's used in the project's CI.
Resolves #1950 that was introduced due to https://github.com/pypa/setuptools/issues/3693

## Type of change
- [x] Bug fix

- ~~[ ] Add our license to new files.~~
- ~~[ ] Write unittests.~~

Test procedure:
- Install package from source with `python setup.py develop && pip install .`
- Run unit tests with `cd cvxpy/tests && pytest`
- Run benchmarks with ``

Test results:
- Unit tests: `852 passed, 289 skipped, 543 warnings in 41.25s`
- Benchmark: `9 passed, 3 skipped, 1 warning in 6.43s`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ PDLP = ortools>=9.3,<9.5
 PROXQP = proxsuite
 SCIP = PySCIPOpt
 SCIPY = scipy
-SCS =
+SCS = setuptools<64.0.2
 XPRESS = xpress
 
 [flake8]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,6 @@ requires = [
     "numpy>=1.21,<1.22; python_version=='3.10' and (platform_machine!='arm64' or platform_system!='Darwin')",
     "numpy>=1.21.4,<1.22; python_version=='3.10' and (platform_machine=='arm64' and platform_system=='Darwin')",
     "scipy >= 1.1.0",
-    "setuptools>=40.8.0",
+    "setuptools>=40.8.0,<=64.0.2",
     "wheel"
 ]
--- a/setup.py
+++ b/setup.py
@@ -231,6 +231,7 @@ setup(
         "scs >= 1.1.6",
         "numpy >= 1.15",
         "scipy >= 1.1.0",
+        "setuptools <= 64.0.2",
     ],
     setup_requires=["numpy >= 1.15"],
 )


piiq:
@phschiele @SteveDiamond I would appreciate a review and let me know if there is anything I can do to have this patch released to pypi

SteveDiamond:
@piiq can you clarify more what happened? I saw your link, but is this bug in setuptools going to be fixed soon, or is this a permanent problem due to deprecation?

piiq:
Thanks for the reply @SteveDiamond 

If i understand correctly it's a permanent problem due to deprecation on the setuptools side that the numpy maintainers are not planning to adjust to. From [this comment](https://github.com/numpy/numpy/issues/22623#issuecomment-1321012633) in a numpy issue thread I understood that numpy does not and will not support setuptools>60, but all versions up to the 65.5.0 were ok.

For cvxpy the problem is that it breaks the build process of SCS when installing with pip on a platform for which wheels with pre-built SCS binaries are not provided.

I've explicitly pinned the versions in `pyproject.toml`, `setup.py` and `setup.cfg` to be lower than the highest version that's installed onto the build system in `continuous_integration/install_dependencies.sh`. This solves the failing build issue because this pinned setuptools version is lower than 65.5.0.

For a test that the patch works as expected on an Apple Silicon machine I run `pip install cvxpy` or `pip install git+https://github.com/cvxpy/cvxpy.git@master` and see the SCS build fail and I run `pip install git+https://github.com/piiq/cvxpy.git@bugfix/pin-setuptools` and see that the build succeeds

setup.cfg
```diff
@@ -17,7 +17,7 @@ PDLP = ortools>=9.3,<9.5
 PROXQP = proxsuite
 SCIP = PySCIPOpt
 SCIPY = scipy
-SCS =
+SCS = setuptools<64.0.2
```
phschiele, line 20: It feels like this change should rather go into SCS itself?

piiq: Thanks for the review @phschiele 
It was actually added to the scs repo yesterday with [this PR](https://github.com/bodono/scs-python/pull/65). I have added it here to be on the safe side (and because the scs-python package on pip was not yet updated).
Please advice how to move forward

SteveDiamond: The change to setup.cfg is not strictly necessary right? I'd be ok merging without this change.

piiq: The change to setup.cfg is what actually made it work out of the box.

I see 2 options 
- wait for @bodono or someone else from the scs maintainers to cut a new scs-python release and if that resolves the issue we can close this PR because it's not needed
- merge this PR as is and then cvxpy will be able to build scs by itself (even the older versions)

It's you call. I support any option as long as it makes cvxpy pip installable

If you have time to answer, can you explain me why you oppose changes in setup.cfg? Just for me to understand the details

piiq: If I understand correctly pyproject.toml and setup.py define build requirements for the cvxpy package itself and in the setup.cfg we can specify dependencies for the dependencies. It's a bit convoluted and before this PR I was assuming that the settings of the build system defined in the pyproject.toml override the ones in that the dependencies have.
But what I have discovered is that pip installing cvxpy with changed pyproject.toml and setup.py does not bring in a required version of setuptools on the scs build stage (specifically when it's installing build dependencies) thus the default setuptools (65.6.0) is used. So specifying the build dependencies in the setup.cfg helped.

SteveDiamond: Ok interesting, I didn't know that either. I'm happy to merge the PR then.

SteveDiamond (approved):

PR #1976, opened 2022-12-17
Skew-symmetric property for Expression objects. Improve Hermitian atom canonicalization.

rileyjmurray:
### The problem
Here is CVXPY's current code for canonicalizing a complex Variable into real and imaginary parts:
```
def variable_canon(expr, real_args, imag_args, real2imag):
    if expr.is_real():
        return expr, None

    imag = Variable(expr.shape, var_id=real2imag[expr.id])
    if expr.is_imag():
        return None, imag
    elif expr.is_complex() and expr.is_hermitian():
        return Variable(expr.shape, var_id=expr.id, symmetric=True), (imag - imag.T)/2
    else:  # Complex.
        return Variable(expr.shape, var_id=expr.id), imag
```
The parameterization of Hermitian matrices is higher-dimensional than necessary, as explained below.

> The standard representation an $n \times n$ Hermitian matrix in terms of real quantities consists of an $n \times n$ symmetric part and an $n \times n$ skew-symmetric part. (Recall that a square matrix $A$ is called skew-symmetric if $A^T = -A$.)  The dimension of this representation is exactly $n^2$. However, the implementation above uses an unstructued complex variable for the imaginary part of the Hermitian matrix, which adds $n(n+1)/2$ unnecessary free parameters.

This PR resolves the above inefficiency and its downstream effects. (It's a subset of the changes originally proposed in PR #1969. That PR will be closed once this PR and a couple others are merged.)

### The solution

The first step in addressing the inefficiency above is to define ``imag = Variable(expr.shape[0] * (expr.shape[0]-1) // 2)`` when ``expr.is_hermitian()``. Then, instead of returning ``(imag - imag.T)/2`` for the imaginary part of ``expr``, we return the skew-symmetric Expression ``strict_upper_tri(imag) - strict_upper_tri(imag).T``.  This solution requires modest changes to the ``Complex2Real`` reduction defined in ``reductions/complex2real.py``.

This PR goes a step further. Its main user-facing change is a new skew-symmetric property for Expression objects.  By default, an expression ``expr`` will have ``expr.is_skew_symmetric() == False``. However, if ``expr`` is a Constant then ``expr.is_skew_symmetric()`` will be determined by inspecting ``expr.value``. One can also assert that an Expression is skew-symmetric by using a new wrap atom: ``skew_symmetric_wrap``.

The ``skew_symmetric_wrap`` atom helps with canonicalization of matrix atoms that accept complex input. When many such atoms are applied to a complex matrix $C = X + iY$ (with $X$ and $Y$ real), they canonicalize by applying a real version of the atom to the lifted representation

$$Z(X, Y) = \\begin{bmatrix} X & Y \\\ -Y & X \\end{bmatrix}.$$

This lifted representation is important because it is easy to express its singular values and singular vectors in terms of those of $C$. (This is true even when $C$ is not square!) In the special case when $C = X + i Y$ is Hermitian, we must have that $X$ is symmetric and $Y$ is skew-symmetric. It is easy to show that when this happens, $Z(X, Y)$ is symmetric, and in fact one can represent its eigenvalues and eigenvectors in terms of the eigenvalues of eigenvectors of $C$. 

This PR includes a new ``expand_complex`` function in ``complex2real/canonicalizers/matrix_canon.py`` that preserves the facts above for purposes of DCP checks. This new function replaces the old ``hermitian_canon`` function. Its main change is to check that if ``X.is_symmetric()`` and ``Y.is_skew_symmetric()``, then it forms ``Z(X, Y)`` and returns ``symmetric_wrap(Z(X,Y))``.

### Smaller changes
This PR also includes minor changes to other wrap atoms:
* It introduces a ``symmetric_wrap`` atom.
* It requires that certain DCP properties hold in order to use a given wrap. For example: ``symmetric_wrap(M)``, ``hermitian_wrap(M)``, and ``psd_wrap(M)`` now explicitly check that the provided matrix ``M`` is square.


cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
```diff
@@ -0,0 +1,45 @@
...
+        return expr, None
+    elif expr.is_imag():
+        # Purely imaginary
+        imag = Variable(expr.shape, var_id=real2imag[expr.id])
+        return None, imag
+    elif expr.is_complex() and expr.is_hermitian():
+        n = expr.shape[0]
+        real = Variable((n, n), var_id=expr.id, symmetric=True)
+        if n > 1:
+            imag_var = Variable(shape=n*(n-1)//2, var_id=real2imag[expr.id])
```
SteveDiamond, line 35: At some point it would be logical to add skew_symmetric as a possible property for variables and parameters in the constructor.

cvxpy/expressions/expression.py
```diff
@@ -351,6 +351,11 @@ def is_symmetric(self) -> bool:
         # Defaults to false unless scalar.
         return self.is_scalar()
 
+    def is_skew_symmetric(self) -> bool:
+        """Is this Expression, X, a real matrix that satisfies X + X.T == 0?
+        """
+        return False
```
SteveDiamond, line 357: Should we add more logic for finding skew symmetric expressions in atoms? I'm not sure how much is there for symmetric expressions.

SteveDiamond:
This is a big and welcome PR! Thanks for being so thorough!

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
rileyjmurray: Agreed. I'd say we should do that now, but I don't want to slow this down. There's still _plenty_ more work in getting the quantum information atoms working properly with complex input.

cvxpy/expressions/expression.py
rileyjmurray: ``is_symmetric()`` just defaults to ``is_scalar()``. We can't do that for skew-symmetry since a 1-by-1 skew-symmetric matrix can only be the zero matrix.

I don't know of many functions / mappings that are certain to produce skew-symmetric output when evaluated on some domain. The main example would be

$$ A \\mapsto \\begin{bmatrix} 0 & A^T \\\ -A & 0 \\end{bmatrix}$$

which takes any matrix $A$ to a skew-symmetric matrix with the same singular values. I'm sure there are examples where tensor products preserve skew-symmetry as well. I think it would be okay to add those features later if users actually need them.

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
SteveDiamond: OK let's leave it for later then. I just wanted to mention it.

cvxpy/expressions/expression.py
SteveDiamond: You could at least add it for transpose.

rileyjmurray: Ah! Excellent point :)

SteveDiamond (approved):

h-vetinari:
Is it intentional that this seems to be the only place where the renaming `atom_canonicalizers` -> `canonicalizers` was done?

There's still a lot of `*/atom_canonicalizers` [left](https://github.com/search?q=repo%3Acvxpy%2Fcvxpy+path%3A*%2Fatom_canonicalizers%2F*&type=code).

rileyjmurray:
> Is it intentional that this seems to be the only place where the renaming `atom_canonicalizers` -> `canonicalizers` was done?
> 
> There's still a lot of `*/atom_canonicalizers` [left](https://github.com/search?q=repo%3Acvxpy%2Fcvxpy+path%3A*%2Fatom_canonicalizers%2F*&type=code).

See my responses here: https://github.com/conda-forge/cvxpy-feedstock/pull/75#discussion_r1061136534.

PR #1978, opened 2022-12-18
Operator relative entropy and von Neumann entropy: allow Hermitian input

rileyjmurray:
@SteveDiamond @phschiele this is ready for review. Once this PR is merged we will have all the functionality I wanted ready for CVXPY 1.3. (There's still more to do, but it's reasonable to hold until CVXPY 1.4.)

### Main changes
The quadrature-based approximation of the operator relative entropy cone: ``OpRelEntrConeQuad``.
* This constraint class now accepts Hermitian input. Previously, it only accepted symmetric input.
* The ``TestOpRelConeQuad`` class  in ``tests/test_cone2cone.py`` has been rewritten.

The ``von_neumann_entr`` atom.
* This now accepts Hermitian input. Previously it only accepted symmetric input.
* The ``Test_von_neumann_entr`` class in ``tests/test_von_neumann_entr.py`` has been rewritten.
* Since its introduction into cvxpy's master branch,``von_neumann_entr``'s default canonicalization codepath has used PSD + ExpCone constraints. It also had the option of using a quadrature-based approximation technique to remove the ExpCone constraints and replace them with more PSD constraints. This reformulation was very inefficient. This PR makes it so that if the user provides approximation parameters to ``von_neumann_entr``, then the ExpCone constraints will be replaced with SOC constraints.

### Bigger incidental changes

I changed the ``canonicalize_expr()`` function in ``complex2real/complex2real.py``, see [here](https://github.com/cvxpy/cvxpy/pull/1978/files#diff-06ac9467275a4f02950ff75e8cc0f196fae9c2475e68ea88beb0b80606966a5aR180-R188). The change makes sure that any Constraint objects are wrapped by a list before being returned. Previously this hasn't been necessary because most other Constraint objects had canonicalizers in ``complex2real/canonicalizers`` that were called _even when their inputs were exclusively real-valued_. However, that existing approach has a longstanding bug where someone using ``ExpCone`` or ``PowCone3D`` _anywhere_ in a model with complex-valued expressions would result in an error.

That change isn't sufficient to address all bugs. I added a ``REAL_DUALS_IMPLEMENTED`` attribute to the ``Constraint`` class. It defaults to True, but I've set it to False in a handful of classes: ``PowConeND`` and both quadrature-based approximation cones. This change should be propagated to earlier CVXPY releases, since it amounts to a bugfix for models that use ``PowConeND` alongside complex expressions.

### Small incidental changes
* The trace atom now recognizes that the trace of a Hermitian matrix (or a real matrix) is real. Making this change involved moving the complex2real canonicalization path for trace from ``separable_canon()`` in ``complex2real/canonicalizers/aff_canon.py`` into a new ``trace_canon()`` function in ``complex2real/canonicalizers/matrix_canon.py``.
* MOSEK encounters a solver failure when testing the composite atom ``xexp(pos(expr))`` in ``test_constant_atoms.py``. I've added that atom-solver pair to the list of known solver errors, so that test won't run. It's an esoteric test, but the folks at @aszekMosek might want to investigate if the latest release of MOSEK has some numerical problems.
* I renamed the file ``utilities/eigvals.py`` to ``utilities/linalg.py``, and added some functions to the renamed file. One new function computes an orthonormal basis for the range of an input matrix. Another function computes an orthonormal basis for the orthogonal complement of the range of an input matrix.

### DIFF
--- a/cvxpy/atoms/affine/trace.py
+++ b/cvxpy/atoms/affine/trace.py
@@ -65,6 +65,9 @@ class trace(AffAtom):
     def is_real(self) -> bool:
         return self.args[0].is_real() or self.args[0].is_hermitian()
 
+    def is_complex(self) -> bool:
+        return not self.is_real()
+
     def is_atom_log_log_convex(self) -> bool:
         """Is the atom log-log convex?
         """
--- a/cvxpy/atoms/affine/trace.py
+++ b/cvxpy/atoms/affine/trace.py
@@ -62,6 +62,9 @@ class trace(AffAtom):
         """
         return tuple()
 
+    def is_real(self) -> bool:
+        return self.args[0].is_real() or self.args[0].is_hermitian()
+
     def is_atom_log_log_convex(self) -> bool:
         """Is the atom log-log convex?
         """
--- a/cvxpy/constraints/constraint.py
+++ b/cvxpy/constraints/constraint.py
@@ -40,6 +40,8 @@ class Constraint(u.Canonical):
 
     __metaclass__ = abc.ABCMeta
 
+    REAL_DUALS_IMPLEMENTED = True
+
     def __init__(self, args, constr_id=None) -> None:
         # TODO cast constants.
         # self.args = [cvxtypes.expression().cast_to_const(arg) for arg in args]
--- a/cvxpy/constraints/constraint.py
+++ b/cvxpy/constraints/constraint.py
@@ -40,8 +40,6 @@ class Constraint(u.Canonical):
 
     __metaclass__ = abc.ABCMeta
 
-    REAL_DUALS_IMPLEMENTED = True
-
     def __init__(self, args, constr_id=None) -> None:
         # TODO cast constants.
         # self.args = [cvxtypes.expression().cast_to_const(arg) for arg in args]
--- a/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
+++ b/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
@@ -19,8 +19,9 @@ from typing import List, Optional, Union
 import numpy as np
 
 from cvxpy.atoms import (bmat, lambda_sum_largest, normNuc, reshape,
-                         symmetric_wrap, vstack,)
+                         symmetric_wrap, von_neumann_entr, vstack,)
 from cvxpy.atoms.affine.wraps import psd_wrap
+from cvxpy.constraints.exponential import OpRelEntrConeQuad
 from cvxpy.expressions.constants.constant import Constant
 from cvxpy.expressions.expression import Expression
 
@@ -28,7 +29,7 @@ from cvxpy.expressions.expression import Expression
 def expand_complex(real_part: Optional[Expression],
                    imag_part: Optional[Expression]):
     """
-    We expand_complex the matrix A to B = [[Re(A), -Im(A)], [Im(A), Re(A)]].
+    We expand the matrix A to B = [[Re(A), -Im(A)], [Im(A), Re(A)]].
 
     The resulting matrix has special structure if A is Hermitian.
     Specifically, if x is an eigenvector of A, then [Re(x), Im(x)]
@@ -69,6 +70,20 @@ def hermitian_canon(expr: Expression,
     return expr_canon, None
 
 
+def trace_canon(expr: Expression,
+                real_args: List[Union[Expression, None]],
+                imag_args: List[Union[Expression, None]], real2imag):
+    if real_args[0] is None:
+        real_part = None
+    else:
+        real_part = expr.copy([real_args[0]])
+    if (imag_args[0] is None) or expr.is_hermitian():
+        imag_part = None
+    else:
+        imag_part = expr.copy([imag_args[0]])
+    return real_part, imag_part
+
+
 def norm_nuc_canon(expr: normNuc,
                    real_args: List[Union[Expression, None]],
                    imag_args: List[Union[Expression, None]], real2imag):
@@ -92,6 +107,37 @@ def lambda_sum_largest_canon(expr: lambda_sum_largest,
     if imag_args[0] is not None:
         real /= 2
     return real, imag
+
+
+def von_neumann_entr_canon(expr: von_neumann_entr,
+                           real_args: List[Union[Expression, None]],
+                           imag_args: List[Union[Expression, None]], real2imag):
+    """
+    The von Neumann entropy of X is sum(entr(eigvals(X)).
+    Each eigenvalue of X appears twice as an eigenvalue of the Hermitian dilation of X.
+    """
+    args = expr.args
+    canon_expr = expand_and_reapply(args[0], real_args[0], imag_args[0])
+    if imag_args[0] is not None:
+        canon_expr /= 2
+    return canon_expr, None
+
+
+def op_rel_entr_cone_canon(expr: OpRelEntrConeQuad,
+                           real_args: List[Union[Expression, None]],
+                           imag_args: List[Union[Expression, None]], real2imag):
+    """Transform Hermitian input for OpRelEntrConeQuad into equivalent
+    symmetric input for OpRelEntrConeQuad
+    """
+    must_expand = any(a is not None for a in imag_args)
+    if must_expand:
+        X_dilation = expand_complex(real_args[0], imag_args[0])
+        Y_dilation = expand_complex(real_args[1], imag_args[1])
+        Z_dilation = expand_complex(real_args[2], imag_args[2])
+        canon_expr = expr.copy([X_dilation, Y_dilation, Z_dilation])
+    else:
+        canon_expr = expr.copy(real_args)
+    return [canon_expr], None
 
 
 def at_least_2D(expr: Expression):
--- a/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
+++ b/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
@@ -116,8 +116,7 @@ def von_neumann_entr_canon(expr: von_neumann_entr,
     The von Neumann entropy of X is sum(entr(eigvals(X)).
     Each eigenvalue of X appears twice as an eigenvalue of the Hermitian dilation of X.
     """
-    args = expr.args
-    canon_expr = expand_and_reapply(args[0], real_args[0], imag_args[0])
+    canon_expr = expand_and_reapply(expr, real_args[0], imag_args[0])
     if imag_args[0] is not None:
         canon_expr /= 2
     return canon_expr, None
--- a/cvxpy/reductions/complex2real/complex2real.py
+++ b/cvxpy/reductions/complex2real/complex2real.py
@@ -18,7 +18,7 @@ from cvxpy import problems
 from cvxpy import settings as s
 from cvxpy.atoms.affine.upper_tri import vec_to_upper_tri
 from cvxpy.constraints import (PSD, SOC, Equality, Inequality, NonNeg, NonPos,
-                               OpRelEntrConeQuad, Zero,)
+                               OpRelEntrConeQuad, PowConeND, RelEntrConeQuad, Zero,)
 from cvxpy.constraints.constraint import Constraint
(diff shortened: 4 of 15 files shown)


cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
```diff
@@ -94,6 +109,37 @@ def lambda_sum_largest_canon(expr: lambda_sum_largest,
...
+                           imag_args: List[Union[Expression, None]], real2imag):
+    """
+    The von Neumann entropy of X is sum(entr(eigvals(X)).
+    Each eigenvalue of X appears twice as an eigenvalue of the Hermitian dilation of X.
+    """
+    args = expr.args
+    canon_expr = expand_and_reapply(args[0], real_args[0], imag_args[0])
+    if imag_args[0] is not None:
+        canon_expr /= 2
+    return canon_expr, None
```
rileyjmurray, line 123: Not tested yet.

aszekMosek:
I will check the Mosek problem after new year.

cvxpy/reductions/complex2real/complex2real.py
```diff
@@ -171,4 +176,13 @@ def canonicalize_expr(self, expr, real_args, imag_args, real2imag, leaf_map):
             return result
         else:
             assert all(v is None for v in imag_args)
-            return expr.copy(real_args), None
+            real_out = expr.copy(real_args)
+            if isinstance(expr, Constraint):
```
SteveDiamond, line 180: Can you explain this more? I don't really follow the bug description. If the arguments are real shouldn't that be fine, the reduction is a no-op? Is the issue that certain constraints can't provide duals when the complex2real reduction is present? And why is it important that things be in a list?

SteveDiamond:
I'm a bit confused by the test failure. Looking at install_dependencies.sh, it appears that for windows 3.8 numpy should be installed with version 1.16. It's not clear to me why 1.24 is being installed.

cvxpy/reductions/complex2real/complex2real.py
rileyjmurray: @SteveDiamond if you look at these lines: https://github.com/cvxpy/cvxpy/pull/1978/files/fa6ed23435bac36e7805273ec4230398982625c9#diff-06ac9467275a4f02950ff75e8cc0f196fae9c2475e68ea88beb0b80606966a5aL61-L64, you'll see that there's an assumption that constraint canonicalization returns a list of Constraint objects, not an individual Constraint. Right now that's satisfied because almost all Constraint classes have explicit complex2real canonicalization methods. However, classes like ExpCone, PowCone3D, and PowConeND don't have these methods.

For example, consider running the following code on master. The model includes complex variables, but it uses an ExpCone constraint on an unadorned real variable.
```
import cvxpy as cp
import numpy as np
x = cp.Variable(shape=(3,), complex=True)
y = cp.Variable(shape=(3,))
a = np.random.randn(3) + 1j * np.random.randn(3)
objective = cp.Minimize(cp.norm(x - a) + cp.norm(y))
constraints = [cp.real(x) == y, cp.constraints.ExpCone(y[0], y[1], y[2])]
prob.solve()
```
It produces the following output:
```
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/problems/problem.py", line 493, in solve
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/problems/problem.py", line 1054, in _solve
    data, solving_chain, inverse_data = self.get_problem_data(
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/problems/problem.py", line 681, in get_problem_data
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/reductions/chain.py", line 76, in apply
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/reductions/complex2real/complex2real.py", line 62, in apply
    constrs.extend(real_constrs)
TypeError: 'ExpCone' object is not iterable
```
So this is currently a bug on master.

rileyjmurray:
> I'm a bit confused by the test failure. Looking at install_dependencies.sh, it appears that for windows with Python 3.8 numpy should be installed with version 1.16. It's not clear to me why 1.24 is being installed.

I also don't know why 1.24 got installed. Maybe @phschiele has thoughts?

cvxpy/reductions/complex2real/complex2real.py
SteveDiamond: Ah ok, I see. The core issue is you're adding a catch all no-op transform. Thanks for clarifying!

phschiele:
@rileyjmurray It looks like this is because of a mismatch in the windows runners.
Can you try changing all `windows-2019` to `windows-2022` in https://github.com/cvxpy/cvxpy/blob/master/.github/workflows/build.yml ?

cvxpy/atoms/affine/trace.py
```diff
@@ -62,6 +62,12 @@ def shape_from_args(self) -> Tuple[int, ...]:
         """
         return tuple()
 
+    def is_real(self) -> bool:
+        return self.args[0].is_real() or self.args[0].is_hermitian()
+
+    def is_complex(self) -> bool:
+        return not self.is_real()
```
PTNobel, line 69: Apologies if I don't know something about cvxpy conventions, but this surprised me since $\mathbb R \subseteq \mathbb C$.

Should I be reading `self.is_complex` as "do we need two scalars to store this?" and not a math statement?

cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
```diff
@@ -69,6 +70,20 @@ def hermitian_canon(expr: Expression,
     return expr_canon, None
 
 
+def trace_canon(expr: Expression,
+                real_args: List[Union[Expression, None]],
+                imag_args: List[Union[Expression, None]], real2imag):
+    if real_args[0] is None:
+        real_part = None
+    else:
+        real_part = expr.copy([real_args[0]])
```
PTNobel, line 79: Why do you throw away everything in the List except the first entry? Do we know that it only has length 1?

cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
```diff
@@ -69,6 +70,20 @@ def hermitian_canon(expr: Expression,
...
+    if (imag_args[0] is None) or expr.is_hermitian():
+        imag_part = None
+    else:
+        imag_part = expr.copy([imag_args[0]])
```
PTNobel, line 83: Same as above.

cvxpy/reductions/complex2real/complex2real.py
```diff
@@ -18,7 +18,8 @@
 from cvxpy import settings as s
 from cvxpy.atoms.affine.upper_tri import vec_to_upper_tri
 from cvxpy.constraints import (PSD, SOC, Equality, Inequality, NonNeg, NonPos,
-                               Zero,)
+                               OpRelEntrConeQuad, Zero,)
+from cvxpy.constraints.constraint import Constraint
```
PTNobel, line 22: After #1909 this could be `from cvxpy import Constraint`; not sure if the simplification matters.

cvxpy/constraints/constraint.py
```diff
@@ -40,6 +40,8 @@ class Constraint(u.Canonical):
 
     __metaclass__ = abc.ABCMeta
 
+    REAL_DUALS_IMPLEMENTED = True
```
PTNobel, line 43: I'm not sure what this means; maybe a comment explaining it could be useful implementing other constriants in the future?

PTNobel (commented):
Not sure if this is a useful review, but I had a couple questions.

cvxpy/atoms/affine/trace.py
rileyjmurray: I'm actually not 100% certain about the semantics of ``is_complex``. My instinct tells me that ``is_complex`` should be an indicator of if we need to go through the ``complex2real`` canonicalization code path, which would apply even when an Expression has zero real part. @SteveDiamond @akshayka can you comment on this?

cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
rileyjmurray: So this isn't documented anywhere, but canonicalization ensures that the lengths of the ``real_args`` and ``imag_args`` lists are equal to the number of arguments to the atom / constraint in question. Since trace only accepts a single argument, these lists are guaranteed to have length 1. @SteveDiamond please correct me if I'm wrong.

cvxpy/atoms/affine/trace.py
SteveDiamond: @PTNobel that's right, it's about what information is needed. Complex means there is an imaginary component. There's also is_imag in case there's no real component.

rileyjmurray: @SteveDiamond based on that, it seems to me that all Expression objects should be able to implement ``is_complex()`` as ``return not self.is_real()``, and then ``is_real()`` would need atom-specific / constraint-specific implementations. Would you agree with that? If so, maybe we should create a GitHub issue with plans to make that change.

SteveDiamond: That might be cleaner but it's fairly similar to how things work now. Right now is_real default to ``not is_complex()`` and ``is_complex()`` is False by default for all atoms.

SteveDiamond:
> @rileyjmurray It looks like this is because of a mismatch in the windows runners. Can you try changing all `windows-2019` to `windows-2022` in https://github.com/cvxpy/cvxpy/blob/master/.github/workflows/build.yml ?

@phschiele I changed the runners to windows-2022 but we're getting the same failure.

SteveDiamond:
Looks like the build is passing now. I basically special cased Python 3.8 on windows to make sure it used the right version of numpy. I think we can merge this once @parthb83 's review is addressed.

PTNobel:
(@SteveDiamond wrong Parth)

cvxpy/constraints/constraint.py
rileyjmurray: My latest commit actually removes this field. It didn't make sense to include in such a prominent place since it was only used in ``complex2real.py``.

cvxpy/reductions/complex2real/complex2real.py
rileyjmurray: I'm going to leave it as-is so that isort doesn't get mad at me.

cvxpy/reductions/complex2real/complex2real.py
rileyjmurray: @SteveDiamond just a heads up: I moved this case-check into ``Complex2Real.apply()``, since having it there reduces the likelihood that it gets hit repeatedly in recursive calls. (I'm not sure if recursive calls actually happen in my original change, but this new change is clearly safer from a performance standpoint.)

rileyjmurray: @SteveDiamond just a heads up: I moved this case-check into ``Complex2Real.apply()``, since having it there reduces the likelihood that it gets hit repeatedly in recursive calls. (I'm not sure if recursive calls actually happen in my original change, but this new change is clearly safer from a performance standpoint.)

rileyjmurray:
Addressed @PTNobel's comments; merging!

aszekMosek:
@rileyjmurray I see this was merged. Is there still the Mosek issue that needs to be checked?

aszekMosek:
@rileyjmurray Just to finish off the Mosek issue: the problem is caused by the part of the test responsible for checking ``xexp(0)``. If we look at the conic model of that in direct primal form

```
minimize t
st       t >= x*exp(u/x)
         u >= x^2
         x = 0
```

and whether the solver says optimal or ill-posed (which happens in the test) is a little uh-oh depending on how it presolves, what solution it finds, interaction with rest of problem etc. Nothing will be done about it in Mosek at the moment.

rileyjmurray:
So the situation should be fixed if we remove xexp(0) and instead have
xexp(1), right?

aszekMosek:
Anything other than 0 and it's fine, yes

Issue #1983, opened 2022-12-24
account for (xexp, mosek) failure in test_constant_atoms.py


rileyjmurray:
In a recent PR of mine, I noticed that testing the composite atom ``xexp(pos(x))`` failed with MOSEK in ``test_constant_atoms.py``. I tried to add this (atom, solver) pair to a list of known exclusions, via the KNOWN_SOLVER_ERRORS list already in ``test_constant_atoms.py``. I now see that the way I did this was insufficient, since it didn't work with composite atoms. I'm preparing a PR that will change ``test_constant_atoms.py`` to be more robust to this kind of thing.

PR #1984, opened 2022-12-24
correct handling of KNOWN_SOLVER_ERRORS in test_constant_atoms.py

rileyjmurray:
The build failures in https://github.com/cvxpy/cvxpy/actions/runs/3772470280/jobs/6413421816 happened even though ``(cp.xexp, cp.MOSEK)`` were added to the list of known solver errors in ``test_constant_atoms.py``. Upon further inspection, that's because the ``KNOWN_SOLVER_ERRORS`` list wasn't being checked in the appropriate way. This PR addresses that; I've confirmed that it correctly skips the case of ``('xexp', cp.MOSEK)`` (note the problematic atom has to be specified as a string now).

### DIFF
--- a/cvxpy/tests/test_constant_atoms.py
+++ b/cvxpy/tests/test_constant_atoms.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import collections
 import itertools
 import math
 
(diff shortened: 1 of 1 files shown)


rileyjmurray:
This should resolve #1983.

cvxpy/tests/test_constant_atoms.py
```diff
@@ -305,6 +305,10 @@ def log_sum_exp_axis_1(x): return cp.log_sum_exp(x, axis=1)  # noqa E371
 def check_solver(prob, solver_name) -> bool:
     """Can the solver solve the problem?
     """
+    atom_str = str(prob.objective.args[0])
```
SteveDiamond, line 308: I think ``isinstance(prob.objective.args[0], atom)`` is better than string comparison.

rileyjmurray: > I think ``isinstance(prob.objective.args[0], atom)`` is better than string comparison.

The problem is that some of the "atoms" are really lambda function handles. The xexp atom is one such example, since here it's tested as xexp(pos(x)).

SteveDiamond: Ok what class is ``isinstance(prob.objective.args[0]`` then?

SteveDiamond: Sorry, the comparison I meant is not between prob.objective.args[0] and the atom in the list of constant atoms. I meant comparison between prob.objective.args[0] and atom in the list of known solver errors. I just don't like the string comparison.

SteveDiamond: Nvm, this is not important.

Issue #1762, opened 2022-04-28
w.T@V@w is rejected but quad_from is not


BrannonKing:
I attempted to use w.T@V@w for a quadratic objective. This gets rejected as violating DCP rules. After looking online, I replaced that with quad_form(w, V). The quad_form method runs fast and fine. Why the difference? Can we make the matrix multiplication approach work?

 - OS: Linux
 - CVXPY Version: 1.2.0

phschiele:
Hi @BrannonKing, 

thanks for taking the time to create the issue even though you found the currently supported way of doing it using `quad_form`. 

Better handling of such quadratic forms was recently suggested by @SteveDiamond on Discord ([here](https://discord.com/channels/845323972962549790/845327242456203295/916038778085343282)). 

Would you be interested in contributing this feature?

rileyjmurray:
Resolved in PR #1986 and available in CVXPY 1.3.

PR #1986, opened 2022-12-26
Support for Native Quadforms (x.T @ A @ x) 

jlchen0:
Hi -- small QOL upgrade for my personal cvxpy usage to automatically turn native matmul into quadratic forms where applicable. See issue #1762. When expressions are being constructed, then some MulExpression are automatically turned into a QuadForm, when certain criteria hold. 


Notes: 
1) Does not handle associativity. If we have a constant vector c, then `1/2 * c.T @ c @ x.T @ A @ x` is technically a quadratic form `(1/2 * c.T @ c) * QuadForm(x, A)`, but will not be converted in the below change. 
Reasoning here is that the expression tree leading up to the quadratic form can be arbitrarily tall and would require a search for the right-most leaf (to avoid QuadForms in cases like `1/2 * (c.T @ (c @ (d.T @ (d @ x.T)))) @ A @ y`).  However, as long as parantheses are included around `x.T @ A @ x`, the Expr will have a correct QuadForm leaf.  

2) Does not handle asymmetric/non-hermitian matrices. I noticed that the default QuadForm behavior currently is to throw an error if QuadForm(x, A) is called with A being asymmetric/nonhermitian. Could simply replace A with (A + A.H) / 2 but saw that was removed a while back. 

Let me know if these are not the desired behavior -- thanks! 


Best,
Jeffrey


## Type of change
- [ ] Bug fix

- [ ] Add our license to new files. (N/A) 
- [x] Write unittests.

<details>
  <summary>Benchmark Results</summary>

Without Changes: 

```
============================================================================================================================== test session starts ==============================================================================================================================
platform darwin -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/jchen/Desktop/cvxpy, configfile: pyproject.toml
plugins: dash-2.1.0, hypothesis-6.24.1, cov-3.0.0
collected 12 items                                                                                                                                                                                                                                                              

cvxpy/tests/test_benchmarks.py cone_matrix_stuffing_with_many_constraints: avg=1.159e+00 s , std=0.000e+00 s (1 iterations)
.diffcp_sdp: avg=6.678e+00 s , std=0.000e+00 s (1 iterations)
.-------------------------------------------------------------------------------
                                  Compilation                                  
(CVXPY) Dec 25 01:40:42 PM: Compiling problem (target solver=ECOS).
(CVXPY) Dec 25 01:40:42 PM: Reduction chain: Dcp2Cone -> CvxAttr2Constr -> ConeMatrixStuffing -> ECOS
(CVXPY) Dec 25 01:40:42 PM: Applying reduction Dcp2Cone
(CVXPY) Dec 25 01:40:42 PM: Applying reduction CvxAttr2Constr
(CVXPY) Dec 25 01:40:42 PM: Applying reduction ConeMatrixStuffing
(CVXPY) Dec 25 01:40:43 PM: Applying reduction ECOS
(CVXPY) Dec 25 01:40:43 PM: Finished problem compilation (took 6.253e-01 seconds).
Issue #1668 regression test
Compilation time:  0.6262228488922119
.least_squares: avg=4.767e-03 s , std=0.000e+00 s (1 iterations)
.sConic canonicalization
(ECOS) solver time:  0.059706222
cvxpy time:  0.8578147779084473
Quadratic canonicalization
(OSQP) solver time:  0.02199917
cvxpy time:  0.932844112699585
.qp: avg=8.303e-03 s , std=0.000e+00 s (1 iterations)
.small_cone_matrix_stuffing: avg=4.381e-02 s , std=8.711e-03 s (10 iterations)
.small_lp: avg=1.545e-02 s , std=0.000e+00 s (1 iterations)
small_lp_second_time: avg=1.258e-03 s , std=0.000e+00 s (1 iterations)
.sstv_inpainting: avg=4.951e+00 s , std=0.000e+00 s (1 iterations)
.

=============================================================================================================================== warnings summary ================================================================================================================================
cvxpy/tests/test_benchmarks.py::TestBenchmarks::test_parameterized_qp
  /Users/jchen/Desktop/cvxpy/cvxpy/reductions/solvers/solving_chain.py:209: UserWarning: Your problem has too many parameters for efficient DPP compilation. We suggest setting 'ignore_dpp = True'.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```
  
With Changes:
```

============================================================================================================================== test session starts ==============================================================================================================================
platform darwin -- Python 3.9.7, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/jchen/Desktop/cvxpy, configfile: pyproject.toml
plugins: dash-2.1.0, hypothesis-6.24.1, cov-3.0.0
collected 12 items                                                                                                                                                                                                                                                              

cvxpy/tests/test_benchmarks.py cone_matrix_stuffing_with_many_constraints: avg=1.166e+00 s , std=0.000e+00 s (1 iterations)
.diffcp_sdp: avg=6.146e+00 s , std=0.000e+00 s (1 iterations)
.-------------------------------------------------------------------------------
                                  Compilation                                  
(CVXPY) Dec 25 01:40:15 PM: Compiling problem (target solver=ECOS).
(CVXPY) Dec 25 01:40:15 PM: Reduction chain: Dcp2Cone -> CvxAttr2Constr -> ConeMatrixStuffing -> ECOS
(CVXPY) Dec 25 01:40:15 PM: Applying reduction Dcp2Cone
(CVXPY) Dec 25 01:40:15 PM: Applying reduction CvxAttr2Constr
(CVXPY) Dec 25 01:40:15 PM: Applying reduction ConeMatrixStuffing
(CVXPY) Dec 25 01:40:15 PM: Applying reduction ECOS
(CVXPY) Dec 25 01:40:16 PM: Finished problem compilation (took 5.758e-01 seconds).
Issue #1668 regression test
Compilation time:  0.5766229629516602
.least_squares: avg=4.683e-03 s , std=0.000e+00 s (1 iterations)
.sConic canonicalization
(ECOS) solver time:  0.0435112
cvxpy time:  0.8134698390472412
Quadratic canonicalization
(OSQP) solver time:  0.019287906
cvxpy time:  0.873811163595337
.qp: avg=7.598e-03 s , std=0.000e+00 s (1 iterations)
.small_cone_matrix_stuffing: avg=3.792e-02 s , std=8.338e-03 s (10 iterations)
.small_lp: avg=1.440e-02 s , std=0.000e+00 s (1 iterations)
small_lp_second_time: avg=1.113e-03 s , std=0.000e+00 s (1 iterations)
.sstv_inpainting: avg=3.103e+00 s , std=0.000e+00 s (1 iterations)
.

=============================================================================================================================== warnings summary ================================================================================================================================
cvxpy/tests/test_benchmarks.py::TestBenchmarks::test_parameterized_qp
  /Users/jchen/Desktop/cvxpy/cvxpy/reductions/solvers/solving_chain.py:209: UserWarning: Your problem has too many parameters for efficient DPP compilation. We suggest setting 'ignore_dpp = True'.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

</details>

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -624,10 +624,8 @@ class Expression(u.Canonical):
         if isinstance(self, cvxtypes.matmul_expr()):
             # LHS is matrix multiplication expr, so candidate for QuadForm:
             # Specifically, iff the matrix multiplication is of the form x.T @ A @ y
-            # such that x == y, A is hermitian/symmetric constant matrix and x is a variable,
-            # then it is a QuadForm.
-            if self.args[0] is other and not other.is_constant() \
-                    and self.args[1].is_hermitian() and self.args[1].is_constant():
+            # such that x == y, A is constant matrix and x is a variable, then it is a QuadForm.
+            if self.args[0] is other and not other.is_constant() and self.args[1].is_constant():
                 from cvxpy.expressions.cvxtypes import quad_form
                 return quad_form()(other, self.args[1])
 
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -620,6 +620,17 @@ class Expression(u.Canonical):
         """
         if self.shape == () or other.shape == ():
             raise ValueError("Scalar operands are not allowed, use '*' instead")
+
+        if isinstance(self, cvxtypes.matmul_expr()):
+            # LHS is matrix multiplication expr, so candidate for QuadForm:
+            # Specifically, iff the matrix multiplication is of the form x.T @ A @ y
+            # such that x == y, A is hermitian/symmetric constant matrix and x is a variable,
+            # then it is a QuadForm.
+            if self.args[0] is other and not other.is_constant() \
+                    and self.args[1].is_hermitian() and self.args[1].is_constant():
+                from cvxpy.expressions.cvxtypes import quad_form
+                return quad_form()(other, self.args[1])
+
         return cvxtypes.matmul_expr()(self, other)
 
     @_cast_other
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -1401,11 +1401,16 @@ class TestExpressions(BaseTest):
         expr = x.T.__matmul__(A).__matmul__(x)
         assert isinstance(expr, cp.QuadForm)
 
-        # Not a quad_form because matrix is not symmetric
+        # Expect error for asymmetric/nonhermitian matrices
         x = Variable(shape=(2,))
-        A = Constant([[1, 0], [-1, -1]])
-        expr = x.T.__matmul__(A).__matmul__(x)
-        assert not isinstance(expr, cp.QuadForm)
+        A = Constant([[1, 0], [1, 1]])
+        with self.assertRaises(ValueError) as _:
+            x.T.__matmul__(A).__matmul__(x)
+
+        x = Variable(shape=(2,))
+        A = Constant([[1, 1j], [1j, 1]])
+        with self.assertRaises(ValueError) as _:
+            x.T.__matmul__(A).__matmul__(x)
 
         # Not a quad_form because x.T @ A @ y where x, y not necessarily equal
         x = Variable(shape=(2,))
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -24,6 +24,7 @@ import cvxpy.interface.matrix_utilities as intf
 import cvxpy.settings as s
 from cvxpy import Minimize, Problem
 from cvxpy.atoms.affine.add_expr import AddExpression
+from cvxpy.atoms.affine.wraps import psd_wrap
 from cvxpy.expressions.constants import Constant, Parameter
 from cvxpy.expressions.variable import Variable
 from cvxpy.tests.base_test import BaseTest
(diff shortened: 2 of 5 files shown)


SteveDiamond:
This is a great contribution and well implemented! We are discussing internally what sort of additional error messages might be appropriate related to this feature. We will get back to you.

SteveDiamond:
I think our biggest question is whether the different treatment of symmetric and non-symmetric matrices would be confusing. It might be better to map quadratic forms to quad_form regardless of whether the matrix is symmetric or not, and modify the error message for non-symmetric inputs to quad_form. @PTNobel what do you think?

PTNobel:
Very excited about this feature!

First question, is there any situation in CVXPY where we would expect `x.T @ A @ x` with a non-symmetric `A` to not raise an error? If so, I think we should just raise an error about the symmetry instead of raising a DCP error later.

Second question, which I can answer myself tonight when I get back to my computer what do the following errors look like:

 * `x.T @ A @ x` for asymmetric `A`
 * `x.T @ A @ x` for symmetric but not PSD `A`

SteveDiamond:
> Very excited about this feature!
> 
> First question, is there any situation in CVXPY where we would expect `x.T @ A @ x` with a non-symmetric `A` to not raise an error? If so, I think we should just raise an error about the symmetry instead of raising a DCP error later.

In the long run we might want to allow ``x.T @ A @ x`` with non-symmetric ``A`` and symmetrize ``A``. But for now I agree it should raise an error right away. Currently in this PR ``x.T @ A @ x`` doesn't raise an error but is non-DCP. ``quad_form(x, A)`` raises an error immediately for non-symmetric ``A``.

> 
> Second question, which I can answer myself tonight when I get back to my computer what do the following errors look like:
> 
> * `x.T @ A @ x` for asymmetric `A`

DCP error
> * `x.T @ A @ x` for symmetric but not PSD `A`

DCP error

rileyjmurray:
Hi @jlchen0, thanks so much for this contribution! As food for thought when it comes to error messages, here is some undesirable behavior that CVXPY currently implements:

```
import numpy as np
import cvxpy as cp

C = np.array([[1,0],[1,1]])
X = cp.Variable(shape=(2,2))
objective = cp.Minimize(cp.norm(C - X))

cp.Problem(objective, [X >> 0]).solve()
print(X.value.round(decimals=3))
# array([[ 1., -0.],
#        [ 1.,  1.]])

Problem(objective, [cp.lambda_min(X) >= 0]).solve()
print(X.value.round(decimals=3))
# array([[1. , 0.5],
#        [0.5, 1. ]])
```

This example shows that assuming symmetry (or working with the symmetric part of an input) results in inconsistent behavior with ``X >> 0`` and ``lambda_min(X) >= 0``.

So my preference would be to enforce symmetry of A, as with the current implementation of ``quad_form``.

rileyjmurray:
I suggest that we use a ``ValueError`` for ``x.T @ A @ x`` when ``A`` is symmetric, with a warning that quadratic forms should only be defined with symmetric matrices. I think a DCP error should be raised later on (during compilation) if ``A`` is indefinite. My reasoning here is that someone might want to define a library that extends CVXPY to support nonconvex quadratic forms, which solvers like GUROBI already support. We shouldn't get in the way of that kind of valid extension if someone wants to implement it outside of CVXPY.

cvxpy/expressions/expression.py
```diff
@@ -620,6 +620,17 @@ def __matmul__(self, other: "Expression") -> "Expression":
...
+
+        if isinstance(self, cvxtypes.matmul_expr()):
+            # LHS is matrix multiplication expr, so candidate for QuadForm:
+            # Specifically, iff the matrix multiplication is of the form x.T @ A @ y
+            # such that x == y, A is hermitian/symmetric constant matrix and x is a variable,
+            # then it is a QuadForm.
+            if self.args[0] is other and not other.is_constant() \
+                    and self.args[1].is_hermitian() and self.args[1].is_constant():
+                from cvxpy.expressions.cvxtypes import quad_form
+                return quad_form()(other, self.args[1])
```
SteveDiamond, line 632: Based on discussion, what we want IMO is to dispatch to quad_form if self.args[1].is_hermitian() and all the other conditions are satisfied, and raise a ValueError if all the other conditions are satisfied but self.args[1].is_hermitian() is false.

jlchen0: Chose to dispatch to the ValueError raised by QuadForm itself, as to avoid reduplicating error messages for asymmetric matrices. Hope that's okay!

jlchen0:
Thanks everyone for the feedback! Most recent commit now causes `x.T @ A @ x` to throw a `ValueError` if all conditions are met except symmetry/hermitian-ness. This is done by constructing a QuadForm regardless of symmetry, and then deferring to the `ValueError` thrown by QuadForm for asymmetric matrices. 

Also updated the error message to be a bit more clear and an associated test for it. 

Thanks!

SteveDiamond (approved):

cvxpy/tests/test_expressions.py
```diff
@@ -1338,3 +1339,93 @@ def test_log_log_curvature(self) -> None:
...
         assert llcv.curvature == s.LOG_LOG_CONCAVE
+
+    def test_quad_form_matmul(self) -> None:
+        """Test conversion of native x.T @ A @ x into QuadForms.
+        """
+
+        # Trivial quad form
+        x = Variable(shape=(2,))
+        A = Constant([[1, 0], [0, -1]])
+        expr = x.T.__matmul__(A).__matmul__(x)
```
phschiele, line 1350: @jlchen0 just curious, is there any reason why you exclusively use `__matmul__` in the tests?

jlchen0: Not in particular, I mostly modeled my tests after those in `test_matmul_expression`, which uses `__matmul__` exclusively. Will change if needed though

phschiele: Thanks for the quick response, no need to change it from my side. This is a great convenience improvement!

PR #1955, opened 2022-11-22
Cherrypick 1.2.x

SteveDiamond:
## Description
Cherry picks to prepare for new patch

#1931 
#1951 
#1866 
#1962 
Elements of #1978 
#1984
#1987

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [x] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #1987, opened 2022-12-26
Bugfixes for dual variables in Complex2Real, check for real input to certain Constraint objects

rileyjmurray:
### Background on this PR

PR #1978 (targeting CVXPY 1.3) included bugfixes for dual variable recovery in problems with complex expressions. The fixes involved adding tuples called ``UNIMPLEMENTED_REAL_DUALS`` and ``UNIMPLEMENTED_COMPLEX_DUALS`` to the ``Complex2Real`` class. These tuples consisted of certain Constraint classes. If the type of a Constraint object belonged to one of these tuples then dual variable recovery for that Constraint would be skipped.

@SteveDiamond started the process of porting this bugfix to CVXPY 1.1 and 1.2 ([link](https://github.com/cvxpy/cvxpy/pull/1955/commits/5be035253739bf6b0a0a76a14a264e7f3831ac62#diff-06ac9467275a4f02950ff75e8cc0f196fae9c2475e68ea88beb0b80606966a5aR38-R39)). It seemed to Steven that _additional_ constraint classes were needed in these new tuples.  At Steven's suggestion I ran some tests and found that the code introduced in PR #1978 could still break. 

This led me to look into this in more detail. The result of my investigation is that changes are needed on both master (for CVXPY 1.3) as well as CVXPY 1.1 and 1.2. This PR is targeting master.

### Changes in this PR

The main change in this PR is to go through all the Constraint classes and make sure that the list returned by ``constr.get_data()`` satisfies ``constr.get_data()[-1] == constr.id``. This has long been the default behavior for the Constraint class, but it was "lost" in subclasses that overrode the ``get_data`` function. Such classes included ExpCone, PowCone3D, PowConeND, SOC, and the quadrature-based approximate cones (to appear in 1.3). After making this change it was possible for me to get rid of the ``UNIMPLEMENTED_REAL_DUALS`` tuple introduced in PR #1978.

Note: the change to the SOC class wasn't strictly necessary, since SOC has a dedicated complex2real canonicalizer that copies the constraint ID ([see here](https://github.com/cvxpy/cvxpy/blob/1c94aad38d961c754c7c097f66d60f7a62fe10ea/cvxpy/reductions/complex2real/canonicalizers/soc_canon.py#L22)). No change was needed to PSD because the PSD class did not override ``Constraint.get_data()``.

I also added input checks for real data to several Constraint classes. This is necessary for CVXPY 1.3 since all Constraint classes are now being exposed in CVXPY's top-level namespace.

### Are these changes really necessary?

This PR includes two new unittests in ``test_complex.py``. These tests will fail if run on master or if run on Steven's pending PR for 1.1 and 1.2 (including if you modify to account for different import structures in these branches). The tests will also fail before either my fix or Steven's version, when the repo was at [this state](https://github.com/cvxpy/cvxpy/tree/f216089ef53a8496b18643211997e00a83cef9d3).

### DIFF
--- a/cvxpy/tests/test_complex.py
+++ b/cvxpy/tests/test_complex.py
@@ -769,3 +769,18 @@ class TestComplex(BaseTest):
         self.assertItemsAlmostEqual(actual_dual_c[0], expect_dual_c[0], places=2)
         self.assertItemsAlmostEqual(actual_dual_c[1], expect_dual_c[1], places=2)
         pass
+
+    def test_illegal_complex_args(self) -> None:
+        x = cp.Variable(shape=(3,), complex=True)
+        with self.assertRaises(ValueError):
+            cp.ExpCone(x[0], x[1], x[2])
+        with self.assertRaises(ValueError):
+            cp.PowCone3D(x[0], x[1], x[2], [0.5])
+        with self.assertRaises(ValueError):
+            cp.PowConeND(x[0], x[:2], [0.5, 0.5])
+        with self.assertRaises(ValueError):
+            cp.RelEntrConeQuad(x[0], x[1], x[2], 1, 1)
+        with self.assertRaises(ValueError):
+            cp.NonNeg(x)
+        with self.assertRaises(ValueError):
+            cp.NonPos(x)
--- a/cvxpy/tests/test_complex.py
+++ b/cvxpy/tests/test_complex.py
@@ -745,7 +745,6 @@ class TestComplex(BaseTest):
         #   Run tests
         #
         ####################################
-
         x = cp.Variable(shape=(3,), complex=True)
         actual_objective = cp.Minimize(cp.norm(x - u))
         coupling_con = cp.real(x) == y
@@ -753,7 +752,7 @@ class TestComplex(BaseTest):
         con_a_test = con_a.copy()
         prob_a = cp.Problem(actual_objective, [coupling_con, con_a_test])
         prob_a.solve()
-        actual_dual_a = con_a.dual_value
+        actual_dual_a = con_a_test.dual_value
         self.assertItemsAlmostEqual(actual_dual_a, expect_dual_a, places=2)
 
         con_b_test = con_b.copy()
@@ -768,7 +767,6 @@ class TestComplex(BaseTest):
         actual_dual_c = con_c_test.dual_value
         self.assertItemsAlmostEqual(actual_dual_c[0], expect_dual_c[0], places=2)
         self.assertItemsAlmostEqual(actual_dual_c[1], expect_dual_c[1], places=2)
-        pass
 
     def test_illegal_complex_args(self) -> None:
         x = cp.Variable(shape=(3,), complex=True)
--- a/cvxpy/tests/test_complex.py
+++ b/cvxpy/tests/test_complex.py
@@ -710,3 +710,62 @@ class TestComplex(BaseTest):
 
         print(rho_ABC_val)
         assert np.allclose(rho_ABC.value, rho_ABC_val)
+
+    def test_duals(self) -> None:
+        np.random.seed(0)
+        u_real = np.random.rand(3)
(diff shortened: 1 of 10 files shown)


cvxpy/utilities/canonical.py
```diff
@@ -92,6 +92,8 @@ def copy(self, args=None, id_objects=None):
             return id_objects[id(self)]
         if args is None:
             args = self.args
+        else:
```
rileyjmurray, line 95: This case is to guard against invalid ``args``. The tests for ``Constraint.copy()`` in ``cvxpy/tests/test_constraints.py`` provided invalid inputs for ``args``.

cvxpy/tests/test_constraints.py
```diff
@@ -78,7 +78,7 @@ def test_equality(self) -> None:
         self.assertEqual(copy.args, constr.args)
         self.assertFalse(copy.args is constr.args)
         # Test copy with new args
-        copy = constr.copy(args=[self.A])
+        copy = constr.copy(args=[self.A, self.B])
```
rileyjmurray, line 81: The default behavior of ``Constraint.get_data()`` is just to return the list ``[self.id]``. I experimented with removing this behavior as an alternative to the changes proposed in this PR. When I did that, I found that this unittest failed because of an invalid number of arguments. Closer inspection traced the issue to the ``Canonical`` class, which ``Constraint`` subclasses. Specifically, ``Canonical.copy(args)`` had an implicit requirement that the length of ``args`` matched the number of required arguments to the class constructor. I went ahead and made that requirement *explicit*, and modified these tests accordingly.

cvxpy/utilities/canonical.py
```diff
@@ -92,6 +92,8 @@ def copy(self, args=None, id_objects=None):
...
+            assert len(args) == len(self.args)
```
rileyjmurray, line 96: See my comment in changes to ``cvxpy/tests/test_constraints.py`` for an explanation of why I added these two lines of code.

cvxpy/tests/test_complex.py
```diff
@@ -710,3 +710,77 @@ def test_partial_transpose(self) -> None:
...
+        actual_dual_b = con_b_test.dual_value
+        self.assertItemsAlmostEqual(actual_dual_b, expect_dual_b, places=2)
+
+        con_c_test = con_c.copy()
+        prob_c = cp.Problem(actual_objective, [coupling_con, con_c_test])
+        prob_c.solve()
+        actual_dual_c = con_c_test.dual_value
+        self.assertItemsAlmostEqual(actual_dual_c[0], expect_dual_c[0], places=2)
+        self.assertItemsAlmostEqual(actual_dual_c[1], expect_dual_c[1], places=2)
+        pass
```
SteveDiamond, line 771: Remove pass.

cvxpy/tests/test_complex.py
```diff
@@ -710,3 +710,77 @@ def test_partial_transpose(self) -> None:
...
+        ####################################
+
+        x = cp.Variable(shape=(3,), complex=True)
+        actual_objective = cp.Minimize(cp.norm(x - u))
+        coupling_con = cp.real(x) == y
+
+        con_a_test = con_a.copy()
+        prob_a = cp.Problem(actual_objective, [coupling_con, con_a_test])
+        prob_a.solve()
+        actual_dual_a = con_a.dual_value
```
SteveDiamond, line 756: Should this be con_a_test.dual_value? I'm not sure con_a.dual_value is being updated.

SteveDiamond:
Looks great to me! There might be a slight bug in the tests, but I could be wrong about that as well. Otherwise happy to merge.

cvxpy/tests/test_complex.py
rileyjmurray: It should definitely be ``con_a_test.dual_value``. Nice catch!

PR #1998, opened 2023-01-04
Fixes for 1.3.0

h-vetinari:
Same fix for PyPy as in #1208; also remove a relative import that makes the test suite fail to run against an out-of-tree `cvxpy`.

From https://github.com/conda-forge/cvxpy-feedstock/pull/75

### DIFF
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import builtins
 from functools import wraps
 from typing import List, Optional, Tuple
 
@@ -109,6 +110,6 @@ def sum(expr, axis: Optional[int] = None, keepdims: bool = False):
     """Wrapper for Sum class.
     """
     if isinstance(expr, list):
-        return __builtins__['sum'](expr)
+        return builtins.sum(expr)
     else:
         return Sum(expr, axis, keepdims)
--- a/cvxpy/tests/test_perspective.py
+++ b/cvxpy/tests/test_perspective.py
@@ -18,9 +18,8 @@ import numpy as np
 import pytest
 
 import cvxpy as cp
+from cvxpy.atoms.perspective import perspective
 from cvxpy.constraints.exponential import ExpCone
-
-from ..atoms.perspective import perspective
 
 
 def test_monotonicity():


rileyjmurray:
@h-vetinari we're getting some unit test failures that seem to suggest ``builtins`` can be a dict: https://github.com/cvxpy/cvxpy/actions/runs/3835348564/jobs/6528514818.

h-vetinari:
> @h-vetinari we're getting some unit test failures that seem to suggest `builtins` can be a dict: https://github.com/cvxpy/cvxpy/actions/runs/3835348564/jobs/6528514818.

Should be all good now. I had to wait to push to https://github.com/conda-forge/cvxpy-feedstock/pull/75 so as to leave the results of the CI run inspectable (and not deleted by azure), but ran into an excruciatingly slow agent. Normally I would have tested there first.

In any case, #1208 already held the answer of how to fix this. I've pushed it both here and on the feedstock side.

h-vetinari:
Thanks!

Since master contains https://github.com/cvxpy/cvxpy/commit/3d3426f6b20205edb3f8f5307dc9a390ac7099f3 already, I'm guessing the most consistent thing here would be to create a branch `1.3` and backport this commit there?

rileyjmurray:
This fix should be applied to the `release/1.3.x`` branch: https://github.com/cvxpy/cvxpy/tree/release/1.3.x.

h-vetinari:
> This fix should be applied to the `release/1.3.x` branch: https://github.com/cvxpy/cvxpy/tree/release/1.3.x.

Pity that the release was just "announced" on GH <20min ago without these fixes.

rileyjmurray:
Ah crap. Alas, CVXPY 1.3.0 will not be the first version of CVXPY to last
only 24 hours, nor will it be the last.

phschiele:
@h-vetinari yes, unfortunately, it was already released to PyPI at this point. 
We already had planned to release a patch soon after, as there were some build issues with windows/Python 3.8.

We are also reviewing the release process to catch these build failures earlier. If you have any suggestions to catch some of the bugs that you have reported, please let us know.

h-vetinari:
No worries, just thought it might have been avoidable. 🙃 

> If you have any suggestions to catch some of the bugs that you have reported, please let us know.

I can offer to build a release candidate against our infrastructure (on a best effort basis, though anyone can open a PR on the feedstock, and I'd be happy to show you how to do it; besides, several cvxpy people have the commit bit on the feedstock).

This would probably catch a couple things already. I don't think it would be necessary for patch releases, but might be useful for feature releases?

PR #2008, opened 2023-01-13
[CI] Build wheels on push

phschiele:
## Description
Our current build process only builds wheels when we release.
Often, failures of these builds are unrelated to the code changes. When wheels are partially released to PyPI, we have to push a fix in a subsequent release. 

This change builds whenever we merge to master or a release branch (but not on every PR commit).

Regarding changes to the procedures, @h-vetinari you [mentioned](https://github.com/cvxpy/cvxpy/pull/1998#issuecomment-1370662440) that its possible to test pre-releases against your infrastructure. Could you create an example PR on the cvxpy feedstock repo that shows how this is done, then I could add a reference to that in the PROCEDURES document.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,17 +5,17 @@ on:
     push:
         branches:
             - master
+            - 'release/**'
         tags:
           - '*'
 
+# https://docs.github.com/en/enterprise-cloud@latest/actions/using-jobs/using-concurrency#example-using-a-fallback-value
+# Only cancels-in-progress on PRs (head_ref only defined in PR, fallback run_id always unique)
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
-  cleanup-runs:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: rokroskar/workflow-run-cleanup-action@master
-      env:
-        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
 
   linters:
     runs-on: ubuntu-latest
@@ -125,7 +125,7 @@ jobs:
           echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
 
       - name: Build wheels
-        if: ${{env.DEPLOY == 'True' && env.USE_OPENMP != 'True'}}
+        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
         env:
           CIBW_BUILD: "cp3${{env.PYTHON_SUBVERSION}}-*"
           CIBW_SKIP: "*-win32 *-manylinux_i686 *-musllinux*"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -5,17 +5,17 @@ on:
...
+  cancel-in-progress: true
+
 jobs:
-  cleanup-runs:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: rokroskar/workflow-run-cleanup-action@master
-      env:
-        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
```
phschiele, line 18: This action is deprecated and replaced by the builtin `concurrency` keyword now.

.github/workflows/build.yml
```diff
@@ -5,17 +5,17 @@ on:
     push:
         branches:
             - master
+            - 'release/**'
         tags:
           - '*'
 
+# https://docs.github.com/en/enterprise-cloud@latest/actions/using-jobs/using-concurrency#example-using-a-fallback-value
```
phschiele, line 12:
```suggestion
# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
```

h-vetinari:
> Regarding changes to the procedures, @h-vetinari you [mentioned](https://github.com/cvxpy/cvxpy/pull/1998#issuecomment-1370662440) that its possible to test pre-releases against your infrastructure. Could you create an example PR on the cvxpy feedstock repo that shows how this is done, then I could add a reference to that in the PROCEDURES document.

While it's possible to point to an arbitrary commit, the easiest is to have a tagged (pre-)release, and then testing is as simple as pointing the conda-forge infrastructure to that tag - as a recent example from another feedstock, see [here](https://github.com/conda-forge/openssl-feedstock/pull/116/commits/995e15fa86201fe50f631153b04902e4f9e91158).

.github/workflows/build.yml
```diff
@@ -125,28 +125,33 @@ jobs:
...
+
+      - name:  Release to pypi
+        if: ${{env.DEPLOY == 'True' &&  env.USE_OPENMP != 'True'}}
+        shell: bash
+        run: |
           twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_PASSWORD
 
       - name: Upload artifacts to github
-        if: ${{env.DEPLOY == 'True' && env.USE_OPENMP != 'True'}}
+        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
```
SteveDiamond, line 154: Shouldn't this have the same conditions as release to pypi?

.github/workflows/build.yml
```diff
@@ -209,28 +214,33 @@ jobs:
...
+        if: ${{github.event_name == 'push'}}
         run: |
           python -m pip install --upgrade twine
           twine check wheelhouse/*
+
+      - name: Release to pypi
+        shell: bash
+        if: ${{env.DEPLOY == 'True'}}
+        run: |
           twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_PASSWORD
```
SteveDiamond, line 240: Change $PYPI_PASSWORD to $PYPI_BASE_PASSWORD

.github/workflows/build.yml
phschiele: This is just uploading to the artifacts section of the run on GitHub, so we can try out the wheels generated by the run if required.
It does **not** generate a release or similar.

.github/workflows/build.yml
phschiele: Done in [fe338b9](https://github.com/cvxpy/cvxpy/pull/2008/commits/fe338b9ffa9a1eaf61cd954e8a750c0be78c50a2)

SteveDiamond (approved):

Issue #1995, opened 2023-01-03
DCP check fails on a positive definite quadratic form (ARPACK convergence failure)


jomoussafir:
**Describe the bug**

I did solve two equivalent optimisation problems using cvxpy with default solver, cvxopt or mosek and I observed each time 
the same pattern. The two problems are equivalent in the following sense, the second one corresponds to the first one with a simple change of variable. The second problem is quickly solved, but the first one either fails or takes a long time to run.
When both succeed, solutions match.

**To Reproduce**
The following python code reproduces what I got :

```
import cvxpy as cp
import numpy as np
np.set_printoptions(precision=2)
np.set_printoptions(linewidth=200)

import matplotlib.pyplot as plt

print(f"cvxpy: {cp.__version__}")
print(f"numpy: {np.__version__}")

nb_ineq = 10
nb_var = 200

np.random.seed(1)

P = np.random.randn(nb_var, nb_var)
P = P.T @ P

U = np.linalg.cholesky(P).T
U_inv = np.linalg.inv(U)

a = np.random.randn(nb_var)
G = np.random.uniform(low=-10, high=10, size=nb_ineq*nb_var).reshape((nb_ineq, nb_var))
h = np.random.uniform(low=1, high=10, size=nb_ineq)

## initial problem
x = cp.Variable(nb_var)
prob = cp.Problem(cp.Maximize(a.T@x - (1/2)*cp.quad_form(x, P)), [G @ x <= h])
prob.solve()
x_sol = x.value

## transformed problem
y = cp.Variable(nb_var)
prob_tr = cp.Problem(cp.Maximize((a.T @ U_inv)@y - (1/2)*cp.quad_form(y, np.eye(nb_var))), 
                  [(G @ U_inv) @ y <= h])
prob_tr.solve()
x_tr_sol = U_inv @ y.value
```
The run time for the first problem can be several orders of magnitude larger than the run time for the second one.
Most of the time with large values for nb_var, the first problem will fail with error ending like this
`ArpackNoConvergence: ARPACK error -1: No convergence (2001 iterations, 0/1 eigenvectors converged)

**Expected behavior**
I would expect both problem to converge and use approximately the same time. 
The problem is feasible here since $h\geq0$.

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

**Version**
[...]

PR #2009, opened 2023-01-13
Improved psd check message

phschiele:
## Description
Adding a custom error message to `ArpackNoConvergence` during PSD check.

Issue link (if applicable):  #1995 

cc: @PTNobel

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_constant.py
+++ b/cvxpy/tests/test_constant.py
@@ -1,0 +1,31 @@
+import numpy as np
+import pytest
+import scipy.sparse.linalg as sparla
+
+import cvxpy as cp
+from cvxpy import psd_wrap
+
+
+def test_is_psd() -> None:
+
+    n = 50
+
+    # trivial cases
+    psd = np.eye(n)
+    nsd = -np.eye(n)
+
+    assert cp.Constant(psd).is_psd()
+    assert not cp.Constant(psd).is_nsd()
+
+    assert cp.Constant(nsd).is_nsd()
+    assert not cp.Constant(nsd).is_psd()
+
+    np.random.seed(97)
+
+    P = np.random.randn(n, n)
+    P = P.T @ P
+
+    with pytest.raises(sparla.ArpackNoConvergence, match="CVXPY note"):
+        cp.Constant(P).is_psd()
+
+    assert psd_wrap(cp.Constant(P)).is_psd()
--- a/cvxpy/tests/test_constant.py
+++ b/cvxpy/tests/test_constant.py
@@ -1,5 +1,4 @@
 import numpy as np
-import pytest
 import scipy.sparse.linalg as sparla
 
 import cvxpy as cp
@@ -20,12 +19,22 @@ def test_is_psd() -> None:
     assert cp.Constant(nsd).is_nsd()
     assert not cp.Constant(nsd).is_psd()
 
-    np.random.seed(97)
+    # We simulate a scenario where a matrix is PSD but a ArpackNoConvergence is raised.
+    # With the current numpy random number generator, this happens with seed 97.
+    # We test a range of seeds to make sure that this scenario is not always triggered.
 
-    P = np.random.randn(n, n)
-    P = P.T @ P
+    failures = set()
+    for seed in range(95, 100):
+        np.random.seed(seed)
 
-    with pytest.raises(sparla.ArpackNoConvergence, match="CVXPY note"):
-        cp.Constant(P).is_psd()
+        P = np.random.randn(n, n)
+        P = P.T @ P
+
+        try:
+            cp.Constant(P).is_psd()
+        except sparla.ArpackNoConvergence as e:
+            assert "CVXPY note" in str(e)
+            failures.add(seed)
+    assert failures == {97}
 
     assert psd_wrap(cp.Constant(P)).is_psd()
--- a/cvxpy/tests/test_constant.py
+++ b/cvxpy/tests/test_constant.py
@@ -1,5 +1,6 @@
 import numpy as np
 import scipy.sparse.linalg as sparla
+from numpy.random import PCG64, Generator
 
 import cvxpy as cp
 from cvxpy import psd_wrap
@@ -20,14 +21,14 @@ def test_is_psd() -> None:
     assert not cp.Constant(nsd).is_psd()
 
     # We simulate a scenario where a matrix is PSD but a ArpackNoConvergence is raised.
-    # With the current numpy random number generator, this happens with seed 97.
+    # With the numpy PCG64-based random number generator, this happens with seed 119.
     # We test a range of seeds to make sure that this scenario is not always triggered.
 
     failures = set()
-    for seed in range(95, 100):
-        np.random.seed(seed)
+    for seed in range(115, 120):
+        rng = Generator(PCG64(seed))
 
-        P = np.random.randn(n, n)
+        P = rng.standard_normal((n, n))
         P = P.T @ P
 
         try:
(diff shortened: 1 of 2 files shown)


cvxpy/utilities/linalg.py
```diff
@@ -79,12 +79,26 @@ def SA_eigsh(sigma):
     ev = np.NaN
     try:
         ev = SA_eigsh(-tol)  # might return np.NaN, or raise exception
-    finally:
-        if np.isnan(ev).all():
-            # will be NaN if A has an eigenvalue which is exactly -tol
-            # (We might also hit this code block for other reasons.)
-            temp = tol - np.finfo(A.dtype).eps
-            ev = SA_eigsh(-temp)
```
phschiele, line 87: @rileyjmurray According to the [Python docs](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) a `try ... finally` statement reraises any unhandled exceptions if it does not contain a `break`, `return`, or `continue` statement. So if I am reading this correctly, this means that if we did encounter an exception in the `try` block, we would reraise, and if we didn't, we don't need the `finally`.

I changed this slightly to only catch the `ArpackNoConvergence` specifically, let me know what you think of that.

cvxpy/utilities/linalg.py
```diff
@@ -79,12 +79,26 @@ def SA_eigsh(sigma):
...
+
+        message = """
+        CVXPY note: This failure was encountered while trying to certify
+        that a matrix is positive semidefinite. In rare cases, this method
+        fails for numerical reasons even when the matrix is positive semidefinite.
+        If you know that you're in that situation, you can replace the matrix A by
+        cvxpy.psd_wrap(A).
+        """
+
+        error_with_note = f"{str(e)}\n\n{message}"
```
phschiele, line 93: The better exception handling in Python 3.11 would have allowed a cleaner syntax here:
```py
e.add_note(message)
```

cvxpy/utilities/linalg.py
rileyjmurray: It looks like I misunderstood the semantics of try-finally. Good catch on this.

cvxpy/utilities/linalg.py
rileyjmurray: This message looks good to me.

rileyjmurray (approved):

cvxpy/tests/test_constant.py
```diff
@@ -0,0 +1,31 @@
...
+    psd = np.eye(n)
+    nsd = -np.eye(n)
+
+    assert cp.Constant(psd).is_psd()
+    assert not cp.Constant(psd).is_nsd()
+
+    assert cp.Constant(nsd).is_nsd()
+    assert not cp.Constant(nsd).is_psd()
+
+    np.random.seed(97)
```
PTNobel, line 23: I'm curious does this always turn to give a convergence issue, or did you have to try around for a seed?

cvxpy/utilities/linalg.py
PTNobel: Could we consider a link or reference to what a PSD matrix is in the message? I worry it is too math-speak.

cvxpy/utilities/linalg.py
```diff
@@ -79,12 +79,26 @@ def SA_eigsh(sigma):
...
+
+        raise sparla.ArpackNoConvergence(error_with_note, e.eigenvalues, e.eigenvectors)
+
+    if np.isnan(ev).all():
```
PTNobel, line 97: Shouldn't we be worried about any nan, not just all?

cvxpy/tests/test_constant.py
phschiele: I had to play around to find this. I tested seeds starting from 0, and 97 was the first one that failed, indicating that the `ArpackNoConvergence` only is happening quite rarely (at least for this example).

PTNobel: Hmm so officially, NumPy reserves the right to change the generator algorithm, so I wonder if being dependent on an exact seed is wise.
Should we add a comment? Maybe write a test that tries a bunch of seeds (if it is fast enough) and then also ensures that it is still a rare event?

cvxpy/utilities/linalg.py
rileyjmurray: Ah, good catch. Yes I think it should by "any".

rileyjmurray:
Replying to https://github.com/cvxpy/cvxpy/pull/2009#discussion_r1071652053:

I agree that providing a link to the Wikipedia page on PSD matrices is a good idea.

cvxpy/tests/test_constant.py
rileyjmurray: Alternative idea: could we find a smaller example and hard-code the test matrix?

cvxpy/utilities/linalg.py
phschiele: Added a reference in de61f26

cvxpy/utilities/linalg.py
phschiele: Changed in https://github.com/cvxpy/cvxpy/commit/de61f260e4d520327b8c237313defebd77a32967

cvxpy/tests/test_constant.py
phschiele: @PTNobel I followed your suggestion and try 5 seeds ensuring that only the expected one fails. Testing 5 seeds takes about 0.2 seconds. I did not find a tiny example that we could easily hardcode and I wanted to avoid adding an external file for the test.

PTNobel: That sounds good! I don't have a sense of how long we are willing to let things run for in tests, but I wonder if doing 25 or so it is a little more useful for testing the statistical properties?

SteveDiamond: I'm not very concerned about numpy changing the algorithm. If this is really an issue surely you can specify the algorithm?

SteveDiamond: Also this approach has the flavor of adding randomness to tests, which seems much worse than relying on a seed.

PTNobel: Julia recently changed their random number generation scheme and we've been remaking a bunch of the convex book additional exercises accordingly, but NumPy is probably stabler than Julia so I'm not too worried.

I think that for a given NumPy version the tests should test the same thing every time.

I think that we should test this warning stays rare because changes in how we handle things could change that. I think the easiest way is to run it on many psuesorandom matrices and make sure we get the same count of warnings every time.

phschiele: I fixed the algorithm in [0b26946](https://github.com/cvxpy/cvxpy/pull/2009/commits/0b2694637270813878d45f5b80274ffe604c2b7c). 
Testing a small range of seeds seems like a good balance between coverage and test time.

phschiele: Had to revert as this was not available on the earlier numpy versions we support.
However, according to the numpy docs, it seems rather unlikely that there would be a breaking change, and in the worst case it's only a single test that needs to be updated.

_Compatibility Guarantee_ A fixed seed and a fixed series of calls to ‘RandomState’ methods using the same parameters will always produce the same results up to roundoff error except when the values were incorrect. Incorrect values will be fixed and the NumPy version in which the fix was made will be noted in the relevant docstring. Extension of existing parameter ranges and the addition of new parameters is allowed as long the previous behavior remains unchanged.

SteveDiamond (approved):

Issue #2012, opened 2023-01-16
deepcopy exception is a breaking change


michael-123123:
**Describe the bug**
PR #1852 adds an override to `__deepcopy__` in cvxpy objects, 
which now throws an exception instead of using the default implementation provided by `object`.
Our code base contains multiple classes which have cvxpy objects as attributes.
These classes contain data as well as cvxpy models, and we `deepcopy` them.

For example, we might have a list which is partially numbers (or cvxpy constants) and partially cvxpy variables.
And they may be deepcopied (usually after the model is solved) as part of a serialization process.
This worked fine until the changes in PR #1852

It is not realistic for us to customize the entire class hierarchy to call `.copy` for cvxpy objects whenever `deepcopy` is used.
I would suggest that either:
1. `__deepcopy__` will not throw an exception and use the default implementation or
2. `__deepcopy__` will print a warning and call `.copy` instead

**To Reproduce**
```
expr = cp.Variable()
deepcopy(expr)
```

**Expected behavior**
A deep copy of the expression, whatever that means semantically.

**Version**
 - OS: Windows/Linux 
 - CVXPY Version: 1.3.0

adishavit:
Indeed, this is breaking our code.  
Even after reviewing the PR #1852 I don't understand:

1. The original semantics of `copy` and `deepcopy`;
2. The indented semantics for `copy` and `deepcopy` ;
3. How the PR fixes the original semantics - and if the change is worth the deep breaking changes it entails.

phschiele:
@michael-123123 @adishavit Thanks for reporting this.

The intended semantics are that if you create a copy of a CVXPY expression (as originally intended via the `.copy()` method) the copied expression still refers to the same instance of the variables.

The PR was initially meant to fix a [subtle bug](https://github.com/cvxpy/cvxpy/issues/1827#issuecomment-1184870476), but after some discussion, I was questioning the reasoning for when a deepcopy of a cvxpy expression even makes sense. It turned out that that consideration was too narrow, as, for example, deep copying a mutable object that contains CVXPY expressions as well as other objects, can make sense. This seems to be the use case reported above. Indeed, found out as much myself here, where I had to resort to an inelegant workaround: https://github.com/robertmartin8/PyPortfolioOpt/pull/493/files). 

I would support restoring the deepcopy functionality with the soon-to-be-released 1.3.1 patch, if there are no objections by @cvxpy/maintainers.

SteveDiamond:
[...]

PR #2014, opened 2023-01-18
SPEC: deepcopy behaviour 

phschiele:
## Description
Based on a recent discussion about the behaviour of `deepcopy`, this PR collects examples of how CVXPY expressions might be copied, in an effort to define the **intended** behaviour and **document** it once defined.

Note that as written, this PR **would** restore the `deepcopy` capability, but **would not** treat deepcopied variables as two distinct objects.

Issue link (if applicable): #2012

@michael-123123 @adishavit @rileyjmurray @SteveDiamond 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_copy.py
+++ b/cvxpy/tests/test_copy.py
@@ -1,9 +1,10 @@
-from copy import deepcopy, copy
+from copy import copy, deepcopy
 from dataclasses import dataclass
 
-import pytest
-import cvxpy as cp
 import numpy as np
+import pytest
+
+import cvxpy as cp
 
 
 @dataclass
@@ -15,8 +16,8 @@ class MockMutableObject:
     P: list[list[int]]  # requires deepcopy
 
 
-
-def test_valid_deepcopy_example():
+@pytest.mark.parametrize("copy_func", [copy, deepcopy])
+def test_valid_deepcopy_example(copy_func):
     """
     Example where deepcopy is required.
     Even though for the CVXPY expression, deepcopy is not required, it is required for the
@@ -31,8 +32,8 @@ def test_valid_deepcopy_example():
     assert problem1.status == cp.OPTIMAL
     assert np.isclose(problem1.value, 2)
 
-    # Deepcopy should work.
-    obj_copy = deepcopy(obj)
+    # Deepcopy should work
+    obj_copy = copy_func(obj)
     obj_copy.P[0][0] = 2
 
     problem2 = cp.Problem(cp.Minimize(cp.quad_form(obj_copy.x, obj_copy.P)), [obj_copy.x == 1])
@@ -40,10 +41,15 @@ def test_valid_deepcopy_example():
     assert problem2.status == cp.OPTIMAL
     assert np.isclose(problem2.value, 3)
 
-    # Original problem should not be affected.
+    problem1 = cp.Problem(cp.Minimize(cp.quad_form(obj.x, obj.P)), [obj.x == 1])
     problem1.solve()
     assert problem1.status == cp.OPTIMAL
-    assert np.isclose(problem1.value, 2)
+    if copy_func == deepcopy:
+        # Original problem not affected by deepcopy
+        assert np.isclose(problem1.value, 2)
+    else:
+        # Original problem affected by shallow copy
+        assert np.isclose(problem1.value, 3)
 
 
 def test_deepcopy_same_identity():
@@ -58,7 +64,7 @@ def test_deepcopy_same_identity():
     constraints = [x >= y + 1]
 
     copied_constraint = deepcopy(constraints[0])
-    # Other expressions change their identity (id()), but their .id stays the same
+    # Constraints change their identity (id()), but their .id stays the same
     assert copied_constraint.id == constraints[0].id
     assert id(copied_constraint) != id(constraints[0])
 
--- a/cvxpy/tests/test_copy.py
+++ b/cvxpy/tests/test_copy.py
@@ -1,3 +1,19 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_copy.py
```diff
@@ -0,0 +1,112 @@
...
+        # Original problem affected by shallow copy
+        assert np.isclose(problem1.value, 3)
+
+
+def test_deepcopy_same_identity():
+    x = cp.Variable(nonneg=True, name="x")
+    y = deepcopy(x)
+
+    # Leafs keep their identity (id()), as well as their name and .id
+    assert y.name() == "x"
```
phschiele, line 76: If `deepcopy` would create a new independent variable, the name would still remain the same, which could be confusing. 
In general, I think there could be alternative implementations where `.id` is simply `id()` (both with pros and cons), but given that this is a central design decision, I don't think that `deepcopy(x)` should create an independent variable.

SteveDiamond:
Making a note that deepcopy is disable on the latest 1.1, 1.2, and 1.3.

SteveDiamond:
@michael-123123 can you repost your desired behavior? I understand why abstractly it would make sense for deepcopy to change the variable id, but is there a concrete use case where it would matter, other than mixing in deepcopied variables with the original variables and getting unexpected results?

michael-123123:
Regardless of the semantics chosen, I will say that, imo, most behaviors are preferable to throwing an exception.
Even a `return self` in `__deepcopy__`. So, either way, this fix is very welcome -- thanks all.

SteveDiamond:
Thanks @michael-123123 for such a thoughtful response! After I read it, I remembered there is a place in cvxpy where we deepcopy in the sense you're describing. It's for the partial_optimize transform:
https://github.com/cvxpy/cvxpy/blob/726a4b6d6e824b441d8e7017a561a19954e816ad/cvxpy/transforms/partial_optimize.py#L95

rileyjmurray:
I like the semantics proposed by @michael-123123. I can even think of a use-case for deep copy where variable ids are changed but variable names are not: solving CVXPY problems in parallel. Right now we don't make promises about being thread-safe, but people regularly use CVXPY with Dask. If we know that canonicalization of one problem is completely independent of canonicalization of others, then that takes a big step toward being thread-safe.

@phschiele my point above is related to your [comment here](https://github.com/cvxpy/cvxpy/pull/2014#discussion_r1073872159). I'm suggesting that it would be okay to create a new Variable object. A name clash like ``x.name() == y.name()`` even though ``x.id != y.id`` is already possible without getting into copy semantics:
```
In [1]: import cvxpy as cp
In [2]: x = cp.Variable(name='x')
In [3]: y = cp.Variable(name='x')

In [4]: x.id
Out[4]: 1
In [5]: y.id
Out[5]: 2
```

adishavit:
My .02$: I agree.
`deepcopy` has well defined semantics: 

```
from copy import deepcopy

a = 'hello'
d = {'a': a}

d2 = deepcopy(d)
d2['a'] += ' world'

print(a)
print(d)
print(d2)
```
returns:
```
hello
{'a': 'hello'}
{'a': 'hello world'}
```

After a `deepcopy` we get a totally new and independent object.  

CVXPY introduces an "external" notion of identiy (via `id`). 
When `deepcopy`ing, the usual semantics should be maintained - i.e. ensuring the clones get fresh `id`s.

This is the element of least surprise - even if implementing it might be difficult.

SteveDiamond:
@adishavit that sounds fine to me. We are working on a new implementation of deepcopy that will give fresh ids.

michael-123123:
Hi, there's a relatively quick way to accomplish this task if all you want is a default deepcopy with change of id. 
I'm including the implementation here - it can be made into a more properly programmed PR (if anyone is interested).

```
from copy import deepcopy
from cvxpy.utilities import Canonical

_MISSING = object()

class DefaultDeepCopyContextManager:
    """ override custom __deepcopy__ implementation and call copy.deepcopy's implementation instead"""
    
    def __init__(self, item):
        self.item = item
        self.deepcopy = _MISSING

    def __enter__(self):
        self.deepcopy = getattr(self.item, '__deepcopy__', _MISSING)
        if self.deepcopy is not _MISSING:
            self.item.__deepcopy__ = None

    def __exit__(self, exc_type, exc_val, exc_tb):
        if self.deepcopy is not _MISSING:
            self.item.__deepcopy__ = self.deepcopy
            self.deepcopy = _MISSING


delattr(Canonical, '__deepcopy__')


ID = 110

def fresh_id():
    """ this should use CVXPY's internal id system - fresh_id is here for the example """
    global ID
    ID += 1
    return ID


def _canon_deep_copy(canon: Canonical, memo):
    with DefaultDeepCopyContextManager(canon):
        canon = deepcopy(canon, memo)
    canon.id = fresh_id()
    return canon


setattr(Canonical, '__deepcopy__', _canon_deep_copy)
```


Usage example:
```
import cvxpy as cp

a = cp.Variable()
b = deepcopy(a)
print(a)
print(b)
print(a.id)
print(b.id)
print(id(a))
print(id(b))
```

output:
```
var1
var1
1
111
140346094408208
140346089449808

```

phschiele:
Suggested course of action:
1. Review and merge this PR to enable `deepcopy` to be backported to previous versions.
2. Follow up with the semantics change, which will not be backported.

I just started working on an implementation for the new semantics similar to the proposal of @michael-123123 above.

SteveDiamond:
I know it's a violation of semantic versioning, but I'd rather implement the new semantics and backport that. It seems like lining us up for very difficult bug reports if we have such a subtle difference in the deepcopy behavior between versions. @rileyjmurray what do you think? It's a change in interface regardless.

rileyjmurray:
@SteveDiamond I agree that backporting in this situation would be appropriate. I'm definitely in favor of backporting to 1.3.x. I'm not so sure about 1.2.x; I'd like to hear from others on that one. I would not be in favor of backporting 1.1.x.

PR #2017, opened 2023-01-20
Attach Mosek log handler before inputting data

aszekMosek:
Changes the order of operations when initializing the Mosek task: first process options and, in particular, attach the log handler (if verbose), then input data, then optimize.

Attaching the log handler before inputting allows the user to see if Mosek produces any warnings about the magnitude of the data which could be a potential hint in case of numerical issues later in the solver.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -225,6 +225,7 @@ class MOSEK(ConicSolver):
             else:
                 env = mosek.Env()
                 task = env.Task(0, 0)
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Looks good, other than flake8 failure.

aszekMosek:
Uh oh. I edited online in github.  Let me see how I can fix it

pt., 20 sty 2023, 22:07 użytkownik Steven Diamond ***@***.***>
napisał:

SteveDiamond:
@aszekMosek no worries, I fixed it. Just trailing whitespace.

SteveDiamond (approved):

PR #2020, opened 2023-01-23
conic solver, unknown is approximate optimal

Mojusko:
## Description
I added the UNKNOWN into approximate solution.

### DIFF
--- a/cvxpy/reductions/solvers/solver.py
+++ b/cvxpy/reductions/solvers/solver.py
@@ -68,14 +68,14 @@ class Solver(Reduction):
             return False
 
     @abc.abstractmethod
-    def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
+    def solve_via_data(self, data, warm_start: bool, verbose: bool,certified_soln: bool, solver_opts, solver_cache=None):
         """Solve a problem represented by data returned from apply.
         """
         raise NotImplementedError()
 
-    def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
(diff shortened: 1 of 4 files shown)


cvxpy/reductions/solvers/solver.py
```diff
@@ -68,14 +68,14 @@ def is_installed(self) -> bool:
...
 
     @abc.abstractmethod
-    def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
+    def solve_via_data(self, data, warm_start: bool, verbose: bool,certified_soln: bool, solver_opts, solver_cache=None):
         """Solve a problem represented by data returned from apply.
         """
         raise NotImplementedError()
 
-    def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
+    def solve(self, problem, warm_start: bool, verbose: bool, certified_soln: bool, solver_opts):
```
SteveDiamond, line 76: Same as above, this should be in solver_opts not a new arg.

cvxpy/settings.py
```diff
@@ -48,6 +48,7 @@
 # Solver Constants
 OPTIMAL = "optimal"
 OPTIMAL_INACCURATE = "optimal_inaccurate"
+NOTCERTIFIED = 'unknown'
```
SteveDiamond, line 51: I don't think there should be a new status. Map unknown to optimal_inaccurate instead.

Mojusko:
Will do indeed, cheers for feedback.

the-moliver:
@Mojusko Curious if you plan to finish this up at it would be most useful to me :-)

rileyjmurray:
This PR has been subsumed by #2117, which should be merged soon.

Issue #2021, opened 2023-01-24
Setuptools is pinned to a vulnerable version


jgopel:
**Describe the bug**
Pinning `setuptools` is preventing projects that consume cvxpy from taking on critical bugfixes. [CVE-2022-40897](https://nvd.nist.gov/vuln/detail/CVE-2022-40897) affects all versions of `setuptools` before `65.5.1`. https://github.com/cvxpy/cvxpy/pull/1951/ pinned `setuptools` to the vulnerable version `64.0.2`.

SteveDiamond:
Fixed on master.

SteveDiamond:
Thanks for pointing this out!

jgopel:
Thanks for the quick turnaround on the fix!

arsuryan:
When can we expect a cvxpy release that contains this vulnerability fix? Thanks

panteparak:
@SteveDiamond  Do you have a release schedule for this fix? We are awaiting for this fix too :D

phschiele:
@panteparak the next patch release is scheduled for the next few days.

PR #2022, opened 2023-01-24
Unpin setuptools

SteveDiamond:
## Description
Unpins setuptools so the latest version can be installed.
Issue link (if applicable): #2021

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -10,7 +10,10 @@ conda config --set remote_max_retries 10
 conda config --set remote_backoff_factor 2
 conda config --set remote_read_timeout_secs 120.0
 
-if [[ "$PYTHON_VERSION" == "3.7" ]] || [[ "$PYTHON_VERSION" == "3.8" ]]; then
+# Issue with installing setuptools > 65.5.1 through conda on mac with Python 3.7.
+if [[ "$PYTHON_VERSION" == "3.7" ]] && [[ "$RUNNER_OS" == "macos-11" ]]; then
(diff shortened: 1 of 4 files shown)


phschiele:
Some context: the pinning was added in #1950.
The root cause seems to be resolved: 
https://github.com/pypa/setuptools/issues/3693#issuecomment-1325363410

phschiele (approved):
LGTM. Perhaps we could also pin a lower version to make sure it installs on all systems, as anyone requiring more recent versions (e.g. because of vulnerabilities) would not be blocked by having a loose lower bound. I looked at the logs of the Mac/3.7 build but it did not say which version was installed, presumably it used an already installed version.

The pin was backported, so I'm adding a label to backport the unpin as well.

SteveDiamond:
> LGTM. Perhaps we could also pin a lower version to make sure it installs on all systems, as anyone requiring more recent versions (e.g. because of vulnerabilities) would not be blocked by having a loose lower bound. I looked at the logs of the Mac/3.7 build but it did not say which version was installed, presumably it used an already installed version.

I removed the upperbound so this shouldn't be an issue right?

phschiele:
> > LGTM. Perhaps we could also pin a lower version to make sure it installs on all systems, as anyone requiring more recent versions (e.g. because of vulnerabilities) would not be blocked by having a loose lower bound. I looked at the logs of the Mac/3.7 build but it did not say which version was installed, presumably it used an already installed version.
> 
> 
> 
> I removed the upperbound so this shouldn't be an issue right?

@SteveDiamond yes, exactly. Just saying we could also put the lower bound that also works with MacOS/3.7 for example. But fine for me as is 👍

PR #2028, opened 2023-01-30
P-norm issue with DGP problems

jlchen0:
## Description
Ran into the following bug with p-norms in DGP problems.

In particular, the below assertion fails: 

```
x = cp.Variable(pos=True)
norm_expr = cp.norm(cp.Constant([3,4]), p=2)
prob = cp.Problem(cp.Minimize(norm_expr * x ** 2), [x >= 1])
solution = prob.solve(gp=True) # Correctly returns 5 
print(prob.solution.opt_val) # Incorrectly prints 5.8
assert np.isclose(prob.solution.opt_val, solution)
```

Looks like it's due to two issues:
1. In the DGP canon functions dict, the key for pnorms is the pnorm function wrapper, not the Pnorm class. 
2. Should be multiplication instead of exponentiation during the canonicalization process. 

$\ln\(||x||_p\) = \ln\(\(\sum |x_i|^p \)^\frac{1}{p} \) = \frac{1}{p} \ln \(\sum |x_i|^p\)$ 

Since we take the logs of all the leaves first and pass that in as args, we are given $z = \ln (x)$. Then, our expr becomes 
$\frac{1}{p} \ln \(\sum e^{zp} \) = \frac{1}{p} \text{log-sum-exp} (z * p)$. 

Added some tests in `test_dgp2dcp.py` for $p = 2$ and $p = 3$, as well as a matrix fro norm.  

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_dgp2dcp.py
+++ b/cvxpy/tests/test_dgp2dcp.py
@@ -624,7 +624,8 @@ class TestDgp2Dcp(BaseTest):
 
     def test_pnorm(self) -> None:
         x = cvxpy.Variable(pos=True)
-        p = cvxpy.Problem(cvxpy.Minimize(cvxpy.norm(cvxpy.Constant([3, 4]), p=2) * x ** 2), [x >= 1])
+        p = cvxpy.Problem(cvxpy.Minimize(cvxpy.norm(cvxpy.Constant([3, 4]), p=2) * x ** 2),
+                          [x >= 1])
         self.assertAlmostEqual(p.solve(gp=True), 5)
         self.assertAlmostEqual(p.solution.opt_val, 5)
         self.assertAlmostEqual(x.value, 1.0)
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_dgp2dcp.py
```diff
@@ -621,3 +621,31 @@ def test_xexp(self) -> None:
                              [x >= b])
         prob.solve(solver=SOLVER, gp=True)
         self.assertItemsAlmostEqual(expr.value, [np.e, 0.5 * np.e**.5])
+
+    def test_pnorm(self) -> None:
+        x = cvxpy.Variable(pos=True)
+        p = cvxpy.Problem(cvxpy.Minimize(cvxpy.norm(cvxpy.Constant([3, 4]), p=2) * x ** 2),
```
SteveDiamond, line 627: Could you add a test where the input to pnorm is a variable? You could replicate this test and set new_var == constant.

jlchen0: Sure, done! thanks for the quick feedback

SteveDiamond (approved):
Thanks for fixing!

Issue #2023, opened 2023-01-26
Same hashes for different mosek parameters


EmJay276:
**Describe the bug**
I am using cvxpy with the mosek solver.
When I try to pass `mosek.dparam.intpnt_tol_dfeas` and `mosek.iparam.intpnt_solve_form` as keys to a python dict it results in a dict with just one entry. The hashes of `mosek.dparam.intpnt_tol_dfeas` and `mosek.iparam.intpnt_solve_form` are the same!

```
mosek_params={mosek.dparam.intpnt_tol_dfeas: (1.0e-8 / length.max()), 
              mosek.iparam.intpnt_solve_form: mosek.solveform.dual}
```

Result

```
>>> mosek_params
{intpnt_tol_dfeas: dual}
```

Hashes
```
>>> hash(mosek.dparam.intpnt_tol_dfeas)
28
>>> hash(mosek.iparam.intpnt_solve_form)
28
```

**To Reproduce**
```
mosek_params={mosek.dparam.intpnt_tol_dfeas: 1.0e-8), 
              mosek.iparam.intpnt_solve_form: mosek.solveform.dual}
```
Result
```
>>> mosek_params
{intpnt_tol_dfeas: dual}
```

**Expected behavior**
```
mosek_params={mosek.dparam.intpnt_tol_dfeas: 1.0e-8), 
              mosek.iparam.intpnt_solve_form: mosek.solveform.dual}
```
Result
```
>>> mosek_params
{intpnt_tol_dfeas: 1.0e-8, intpnt_solve_form: dual}
```

**Version**
 - OS: Windows 10
 - CVXPY Version: 1.3.0

**Additional context**
The hash values are probably unique (I haven't checked all) within `mosek.iparam` and `mosek.dparam` but not across them.

I personally think this is a mosek issue and should be fixed there, I already reported it https://github.com/MOSEK/Tutorials/issues/5
A workaround would be to separate `iparam` and `dparam` in different dicts for cvxpy
```
mosek_params={'dparam': {mosek.dparam.intpnt_tol_dfeas: 1.0e-8)}, 
              'iparam': {mosek.iparam.intpnt_solve_form: mosek.solveform.dual}}
```
Or to use the string version
```
mosek_params = {'MSK_DPAR_INTPNT_TOL_DFEAS': 1.0e-8,
                'MSK_IPAR_INNTPNT_SOLVE_FORM': mosek.solveform.dual}
```
Currently this can lead to unexpected behaviour. 
[...]

PR #2029, opened 2023-02-03
Cleanup in MOSEK parameters setting

aszekMosek:
## Description
Tries to make Mosek parameters systematic. See https://github.com/cvxpy/cvxpy/issues/2023

Adds proper handling of the case where both parameter name and value are strings. Deprecates the use of constants from the Python API. Adapts the documentation and test accordingly.

Documentation mentions https://docs.mosek.com/latest/faq/faq.html#cvxpy  That link will be separately updated in the same style next time new Mosek docs are uploaded.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -14,9 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import warnings
 from collections import defaultdict
 
-import warnings
 import numpy as np
 import scipy as sp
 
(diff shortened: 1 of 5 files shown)


SteveDiamond (approved):
LGTM. I fixed the flake8 and isort issues.

SteveDiamond:
@phschiele @rileyjmurray does this change look good to you? We need two approvers per our policy.

phschiele (approved):
LGTM, two minor remarks:
- is it intentional to raise both a deprecation and a user warning?
- one test case still uses the deprecated parameter. We could capture the warning using `with pytest.warns()` to ensure it is raised correctly.

aszekMosek:
@phschiele Regarging the warning I followed the only existing case in https://github.com/cvxpy/cvxpy/blob/master/cvxpy/expressions/expression.py#L612 I can do it in a different way

SteveDiamond:
> @phschiele Regarging the warning I followed the only existing case in https://github.com/cvxpy/cvxpy/blob/master/cvxpy/expressions/expression.py#L612 I can do it in a different way

Sounds good. I added ``with pytest.warns()`` to the mosek test. I'll merge after the CI passes.

PR #2041, opened 2023-02-20
MI DGP support

SteveDiamond:
## Description
Adds support for mixed-integer geometric programs, via the FiniteSet constraint.
Issue link (if applicable): #1590

@adishavit

The canonicalization is straightforward. I am not allowing the values of the finite set to be parameterized right now, because ``is_pos`` is only supported for leafs but it's needed here for more complex expressions, since theoretically the set values could be something like param1 + param2. I could hack around this though if the functionality is desired.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/constraints/finite_set.py
+++ b/cvxpy/constraints/finite_set.py
@@ -60,9 +60,9 @@ class FiniteSet(Constraint):
         if isinstance(vec, set):
             vec = list(vec)
         vec = Expression.cast_to_const(vec).flatten()
-        if not expre.is_affine():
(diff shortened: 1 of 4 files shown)


adishavit:
Thanks for this 👍 
And since GP does not support zeros, one cannot use Boolean variables - right?

SteveDiamond:
> Thanks for this 👍 And since GP does not support zeros, one cannot use Boolean variables - right?

That's right. The elements of the finite set must be positive.

phschiele (approved):
Looks good, nice usage of `FiniteSet`!

PR #2046, opened 2023-02-21
Add compilation time property

alxhslm:
## Description
Exposes the `compilation_time` attribute as a property of the `Problem`. I wanted to be able to access this programmatically in a project I was working on.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -409,7 +409,7 @@ class Problem(u.Canonical):
     @property
     def compilation_time(self) -> float | None:
         """float : The number of seconds it took to compile the problem the
-                   last time it was solved.
+                   last time it was compiled.
         """
         return self._compilation_time
 
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -405,14 +405,13 @@ class Problem(u.Canonical):
         """:class:`~cvxpy.problems.problem.SolverStats` : Information returned by the solver.
         """
         return self._solver_stats
-    
+
     @property
     def compilation_time(self) -> float:
         """float : The number of seconds it took to compile the problem the
                    last time it was solved
         """
         return self._compilation_time
-
 
     def solve(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


SteveDiamond:
This sort of thing is supposed to go in ``problem.solver_stats``. It might already be there.

alxhslm:
> This sort of thing is supposed to go in `problem.solver_stats`. It might already be there.

I don't believe this is currently available from `problem.solver_stats`. 

I did originally consider adding the compilation time to `problem.solver_stats`, but it is not strictly a property of the solver itself, so I thought it didn't really "belong" there. However, I'm happy to put it there instead.

SteveDiamond:
> > This sort of thing is supposed to go in `problem.solver_stats`. It might already be there.
> 
> I don't believe this is currently available from `problem.solver_stats`.
> 
> I did originally consider adding the compilation time to `problem.solver_stats`, but it is not strictly a property of the solver itself, so I thought it didn't really "belong" there. However, I'm happy to put it there instead.

That's fair, but compilation time does depend on which solver is targeted, so I'd still put it in solver_stats.

SteveDiamond:
I didn't realize that _compilation_time is something we already record and you're just exposing it. No need to move it to solver_stats.

SteveDiamond (approved):

cvxpy/problems/problem.py
```diff
@@ -406,6 +406,13 @@ def solver_stats(self) -> "SolverStats":
         """
         return self._solver_stats
 
+    @property
+    def compilation_time(self) -> float:
+        """float : The number of seconds it took to compile the problem the
+                   last time it was solved
```
phschiele, line 412:
```suggestion
    def compilation_time(self) -> float | None:
        """float : The number of seconds it took to compile the problem the
                   last time it was compiled.
```

phschiele (approved):
Thanks @alxhslm, this looks good! Had a tiny annotation suggestion above.

SteveDiamond:
Thanks @alxhslm !

PR #2047, opened 2023-02-22
Fix conv and add convolve

SteveDiamond:
## Description
The ``conv`` operator was extremely broken, and gave incorrect shape information. This PR fixes bugs in ``conv``, deprecates ``conv``, and adds a ``convolve`` atom that matches np.convolve.

@PTNobel 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/conv.py
+++ b/cvxpy/atoms/affine/conv.py
@@ -61,8 +61,8 @@ class conv(AffAtom):
     def validate_arguments(self) -> None:
         """Checks that both arguments are vectors, and the first is constant.
         """
-        if not self.args[0].is_vector() or not self.args[1].is_vector():
-            raise ValueError("The arguments to conv must resolve to vectors.")
+        if not self.args[0].ndim == 1 or not self.args[1].ndim == 1:
+            raise ValueError("The arguments to conv must be 1D.")
         if not self.args[0].is_constant():
             raise ValueError("The first argument to conv must be constant.")
 
@@ -71,7 +71,7 @@ class conv(AffAtom):
         """
         lh_length = self.args[0].shape[0]
         rh_length = self.args[1].shape[0]
-        return (lh_length + rh_length - 1, 1)
+        return (lh_length + rh_length - 1,)
 
     def sign_from_args(self) -> Tuple[bool, bool]:
         """Same as times.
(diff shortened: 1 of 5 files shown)


cvxpy/atoms/affine/conv.py
```diff
@@ -69,9 +72,94 @@ def validate_arguments(self) -> None:
...
+
+    def sign_from_args(self) -> Tuple[bool, bool]:
+        """Same as times.
+        """
+        return u.sign.mul_sign(self.args[0], self.args[1])
+
+    def is_incr(self, idx) -> bool:
+        """Is the composition non-decreasing in argument idx?
+        """
+        return self.args[0].is_nonneg()
```
PTNobel, line 91: Why do you not need idx? Shouldn't it be `self.args[0][idx].is_nonneg()`?

cvxpy/atoms/affine/conv.py
```diff
@@ -69,9 +72,94 @@ def validate_arguments(self) -> None:
...
+
+    def is_decr(self, idx) -> bool:
+        """Is the composition non-increasing in argument idx?
+        """
+        return self.args[0].is_nonpos()
```
PTNobel, line 96: Ditto above.

cvxpy/atoms/affine/conv.py
```diff
@@ -69,9 +72,94 @@ def validate_arguments(self) -> None:
...
+            raise ValueError("The first argument to conv must be constant.")
+
+    def shape_from_args(self) -> Tuple[int, int]:
+        """The sum of the argument dimensions - 1.
+        """
+        lh_length = self.args[0].size
+        rh_length = self.args[1].size
+        return (lh_length + rh_length - 1,)
 
     def sign_from_args(self) -> Tuple[bool, bool]:
```
PTNobel, line 164: I think for the copied methods above, my questions from above might need to be copied below too.

PTNobel (commented):
I didn't look too closely at the tests. They made sense to me, but I can check them more carefully if we're worried they may be missing edge cases.

cvxpy/atoms/affine/conv.py
SteveDiamond: If idx were used, it would be ``self.args[idx].is_nonneg()``. But this atom requires that the first argument be constant, so it's always the first argument that is checked.

PTNobel: Ah okay; I had misunderstood the function. Looks good to me then!

SteveDiamond:
Thanks @PTNobel for the quick turn around!

PR #2049, opened 2023-02-22
Introduce ruff linter

phschiele:
## Description
This PR adds configuration for [ruff](https://github.com/charliermarsh/ruff), an extremely fast linter for Python projects and is used by SciPy, pandas, and many others. It includes the relevant rules from flake8 and isort.
Currently, the `isort` pre-commit hook needs an upgrade and I used this as an opportunity to unify all of our linting under one umbrella.
You can run
```
pip install ruff
ruff check cvxpy
```
to try it out (yes, it actually did execute). 

To update you're pre-commit hooks in this branch, run
```
pre-commit clean
pre-commit install
```
Mostly put this together out of curiosity and so other people can try it out easily to see if they like it. 

Seems to work nicely with the GitHub inline errors:
image

Tasks:
- [x] Configuration
- [x] Documentation
- [x] Linter action

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,8 @@ exclude = [
     "cvxpy/cvxcore/*",
     "*__init__.py"
 ]
-# Assume Python 3.8.
-target-version = "py38"
+# The minimum Python version that should be supported
+target-version = "py37"
 
 
 [tool.pytest.ini_options]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,5 @@
 [tool.ruff]
+select = ["E", "F", "I"]
 line-length = 100
 exclude = [
     "build",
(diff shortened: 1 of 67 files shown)


phschiele:
Updated the docs:
image

pyproject.toml
```diff
@@ -1,11 +1,16 @@
...
+line-length = 100
+exclude = [
+    "build",
+    "examples",
+    "doc",
+    "cvxpy/cvxcore/*",
+    "*__init__.py"
 ]
+# Assume Python 3.8.
+target-version = "py38"
```
SteveDiamond, line 12: What is this about?

phschiele: It is the "minimum Python version that should be supported". Was copied from another code base, changed it to 3.7.

phschiele, line 12:
```suggestion
# The minimum Python version that should be supported
target-version = "py37"
```

phschiele: I think this is to keep track of which Python features should be supported (like f-strings only with Python 3.6 etc).

PR #2053, opened 2023-02-24
New deepcopy semantics

phschiele:
## Description
Resumes #2014. This PR can be used to work on the new semantics, and we can still decide which version to backport to 1.1, 1.2, 1.3, respectively. 

Implementation based on a suggestion by @michael-123123.

TODO: Documentation for new behaviour. 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_copy.py
+++ b/cvxpy/tests/test_copy.py
@@ -27,10 +27,6 @@ def test_leaf():
     b = copy.copy(a)
     c = copy.deepcopy(a)
 
-    assert a is b
-    assert a is not c
-    assert b is not c
-
     assert a.id == b.id
     assert a.id != c.id
 
@@ -44,10 +40,6 @@ def test_constraint():
     a = Equality(x, 0)
     b = copy.copy(a)
     c = copy.deepcopy(a)
-
-    assert a is not b
-    assert a is not c
-    assert b is not c
 
     assert a.id == b.id
     assert a.id != c.id
@@ -63,10 +55,6 @@ def test_expression():
     a = x + 1
     b = copy.copy(a)
     c = copy.deepcopy(a)
-
-    assert a is not b
-    assert a is not c
-    assert b is not c
 
     assert a.id != b.id
     assert a.id != c.id
--- a/cvxpy/tests/test_copy.py
+++ b/cvxpy/tests/test_copy.py
@@ -1,0 +1,134 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


cvxpy/tests/test_copy.py
```diff
@@ -0,0 +1,122 @@
...
+    x = cp.Variable()
+
+    a = Equality(x, 0)
+    b = copy.copy(a)
+    c = copy.deepcopy(a)
+
+    assert a.id == b.id
+    assert a.id != c.id
+
+    assert id(a) != id(b)
```
phschiele, line 47: Constraints have special handling in that after a copy `.id` matches but `id()` doesn't.

SteveDiamond: I think this is ok. Everything matches after deepcopy as expected.

SteveDiamond:
This looks fine to me, once there's some documentation. The memo field in deepcopy reminds of how convex.jl would memoize during canonicalization, so if the same expression appears multiple times it will only be canonicalized once. That might actually be quite easy to add.

cvxpy/tests/test_copy.py
SteveDiamond: It's ok IMO. I guess the idea here is a constraint is not a leaf, ``copy`` does need to rebuild it.

phschiele: Makes sense, so I guess my question would have been why `.id` remains the same. Not suggesting that it should be changed, though.

SteveDiamond: Constraint.id is used to propagate dual variables. So it's part of constraint.get_data, which is preserved by copy.

SteveDiamond (approved):
LGTM

cvxpy/tests/test_copy.py
```diff
@@ -0,0 +1,122 @@
...
+    assert id(b) != id(c)
+
+
+def test_problem():
+    x = cp.Variable()
+    y = cp.Variable()
+    obj = cp.Minimize((x + y) ** 2)
+    constraints = [x + y == 1]
+    prob = cp.Problem(obj, constraints)
+    prob_copy = prob.copy()
```
PTNobel, line 73: Documentation mentions `copy.copy` but we're only testing `Problem.copy`. I realize `__copy__` just calls `Problem.copy` but I am not sure if we're worried about that code being changed (I think "No we're not" is perfectly valid).

PTNobel (approved):
Overall, LGTM.

I don't think we should backport semantics changes like this to old versions, so I think should be in 1.3.1 and later only.

cvxpy/tests/test_copy.py
SteveDiamond: Changed to copy.copy(prob)

PR #2056, opened 2023-02-25
add a cvxpy_shape type

### LINKED ISSUE #2026: reshape(...) hint type for shape should accept also int and Tuple[int]
Just to avoid pycharm highlighting a type error when calling cvxpy.reshape(A, (10,))
image

In https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/affine/reshape.py#L46, 
`shape: Tuple[int, int]`
should be replaced by
`shape: Tuple[int, int] | Tuple[int] | int`

sdementen:
## Description
Add a new type cvxpy_shape ~ Tuple[int, ...] and use it acress the code base + fix #2026 by changing the signature of reshape.

## Type of change
- [X] Other

### DIFF
--- a/cvxpy/atoms/affine/add_expr.py
+++ b/cvxpy/atoms/affine/add_expr.py
@@ -22,6 +22,7 @@ import cvxpy.lin_ops.lin_utils as lu
 import cvxpy.utilities as u
 from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.constraints.constraint import Constraint
+from cvxpy.utilities.shape import cvxpy_shape
 
 
 class AddExpression(AffAtom):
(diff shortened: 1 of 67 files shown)


phschiele:
@sdementen Thanks a lot for going through all the files and finding the places where we specify some shape information!
Some places (like variables or reshape) will either take a single int, or a tuple consisting of one or two ints. 
So its definitely a welcome improvement to change the hardcoded `Tuple[int, int]` to `int | Tuple[int, ...]` (or, technically correct it would be `int | Tuple[int] | Tuple[int,int]`, since as of now, only 2D shapes are possible. I realize that my suggestion of `cvxpy_shape` is very uninformative. It probably should be something like `upto2D`, but at this point, I'm afraid this would rather decrease readability for anyone unfamiliar with the codebase. 

Maybe it would be best for the scope of this PR to fix the ones that are wrong in the current state.

phschiele (approved):
@sdementen I have reverted the change to you'r original issue you've encountered. Thanks again for bringing it up and stepping up to make a PR!

Issue #808, opened 2019-09-12
Offset for the diag function 


cvanwynsberghe:
In matlab, cvx enables one to set a matrix with an offset of k from the diagonal, ie by `diag(v, k)`.

If I write, for example, the following code :

```
import cvxpy as cvx

v = cvx.Variable(10)
cvx.diag(v, 2)
```
then it returns 
> TypeError: diag() takes 1 positional argument but 2 were given

Same remark about extracting the k-th diagonal from a matrix X by `diag(X, k)`.

Is it possible to add this feature ?

Transurgeon:
done with PR above :)

PR #2057, opened 2023-03-05
Added the diag_offset parameter to the diag atom

pragmaticTNT:
## Description
Added an offset parameter to the diag atom. Now it is possible to extract an off-diagonal vector from a square matrix and make a square matrix with a vector on the off-diagonal. 

I had some questions regarding the `graph_implementation` method of the `diag_vec` and `diag_mat` classes. It is yet unclear to me what the graph_implementation is doing and I are not sure where I should add the offset parameter. Currently the code will fail if `diag_offset != 0`.

I was also wondering if I should add assert statements to ensure that the value of `diag_offset` is no more than the dimension of the input matrix or vector and would like to know where the best place to add it is.   

Issue link (if applicable): #808 

## Type of change

- [x] Make modifications to C++ backend
- [x] Add tests comparing the behaviour of cvxpy and python
- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/lin_op.py
+++ b/cvxpy/lin_ops/lin_op.py
@@ -72,13 +72,13 @@ TRACE = "trace"
 # Data: None
 RESHAPE = "reshape"
 # Converts a vector to a diagonal matrix.
-# Data: None
+# Data: int, diagonal offset
 DIAG_VEC = "diag_vec"
 # Converts the diagonal of a matrix to a vector.
-# Data: float
+# Data: int, diagonal offset
 DIAG_MAT = "diag_mat"
 # Vectorized upper triangular portion of a matrix.
-# Data: float
+# Data: None
 UPPER_TRI = "upper_tri"
 # The 1D discrete convolution of two vectors.
 # Data: LinOp evaluating to the left hand term.
--- a/cvxpy/lin_ops/lin_op.py
+++ b/cvxpy/lin_ops/lin_op.py
@@ -75,10 +75,10 @@ RESHAPE = "reshape"
 # Data: None
 DIAG_VEC = "diag_vec"
 # Converts the diagonal of a matrix to a vector.
-# Data: None
+# Data: float
 DIAG_MAT = "diag_mat"
 # Vectorized upper triangular portion of a matrix.
-# Data: None
+# Data: float
 UPPER_TRI = "upper_tri"
 # The 1D discrete convolution of two vectors.
 # Data: LinOp evaluating to the left hand term.
--- a/cvxpy/atoms/affine/diag.py
+++ b/cvxpy/atoms/affine/diag.py
@@ -24,13 +24,25 @@ from cvxpy.atoms.affine.vec import vec
 from cvxpy.constraints.constraint import Constraint
 
 
-def diag(expr) -> Union["diag_mat", "diag_vec"]:
+def diag(expr, diag_offset=0) -> Union["diag_mat", "diag_vec"]:
     """Extracts the diagonal from a matrix or makes a vector a diagonal matrix.
 
     Parameters
     ----------
     expr : Expression or numeric constant
         A vector or square matrix.
+
+    diag_offset : int
+        An integer offset for the diagonal to extract into a
+        vector or diagonal to add the vector to. Zero is the
+        default (as if there was no offset). Positive integers
+        on matrices extracts vectors above the diagonal.
+        Negative integers on matrices extracts vectors below
+        the diagonal. Similarly, positive integers on vectors
+        creates a square matrix with the vector above the
+        diagonal. Negative integers on vectors creates a square
+        matrix with the vector below the diagonal.
+        TODO: assert that the offset is not out-of-bounds.
 
     Returns
     -------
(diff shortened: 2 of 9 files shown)


SteveDiamond:
This is a great idea! It's a bit hard to implement new functionality in core linear operators, for somewhat complicated reasons. @phschiele would it be easy to propagate this to the new Python backend?

phschiele:
@SteveDiamond we could use the data field of the LinOps class to pass the offset to the backend. I can take a closer look at this.

phschiele:
@SteveDiamond @pragmaticTNT I've implemented the change for the Python backend in 4dee5e8
I've also aligned the API with `np.diag` to consistently use `k`.

pragmaticTNT:
Ok, great thanks a lot! Is there anything else I need to do before the changes get merged?

rileyjmurray:
@SteveDiamond @phschiele I hesitate to merge this without a plan for resolving a discrepancy between functionality offered by the C++ and Python backends. What happens right now if someone tries to use this atom with the standard C++ backend?

pragmaticTNT:
Hum... that's a good point. I wasn't very clear on how the C++ backend worked with the Python so I did not make any changes to it. However I would be happy to try making changes there and submitting another pull request if someone could briefly point out where the changes needs to be made.

phschiele:
@pragmaticTNT You would need to implement the changes I've made to `cvxpy/lin_ops/canon_backend.py` and implement them in the [C++ equivalent](https://github.com/cvxpy/cvxpy/blob/efc70e9b44e81f6d5296a6112dfc697571bd97de/cvxpy/cvxcore/src/LinOpOperations.cpp#L583C8-L630).

See the cvxcore [README](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/cvxcore/README.md) for instructions on how to make changes there.

To see if it worked, you should get the tests passing that I've added to `test_problem.py` after changing the backend to `canon_backend=cp.CPP_CANON_BACKEND`.

pragmaticTNT:
Ok, will do. Expect the change to be ready by next week.

pragmaticTNT:
@phschiele, it seems like your code is in column major format when doing the conversions. Is this the standard in cvxpy?

phschiele:
> @phschiele, it seems like your code is in column major format when doing the conversions. Is this the standard in cvxpy? 

@pragmaticTNT That's right, the backend uses column-major order (compare the example in the test I've added).

pragmaticTNT:
Hey @phschiele, so I am trying to test my changes in cvxcore but I am running into some issues. I saw the issue #1697 and added `undef_macros = ['NDEBUG']` to `Extension` as suggested in the README, but the tests no longer pass. Instead I get:

```cvxpy/tests/test_problem.py python: cvxpy/cvxcore/src/ProblemData.hpp:37: void ProblemData::init_id(int, int): Assertion `TensorV.count(new_param_id) == 0' failed.
Fatal Python error: Aborted```

To be clear, I rolled back all of my changes and only added the line to `setup.py` (and also removed the extra parenthesis in `LinOpOperations.cpp`) and I still get the above error when I run `pytest` on `test_problem.py`.

I mostly want to turn on the debugger so I can see what errors I am getting on my changes. How should I proceed?

**Comment by phschiele** [comment]
@pragmaticTNT I could replicate the error on master, but did not see the error message `assert(TensorV.count(new_param_id) == 0);`. Which command did you run to see this?
It is this line:

https://github.com/cvxpy/cvxpy/blob/66b58cfb8af91e134ec4b512f5670bc234e4739d/cvxpy/cvxcore/src/ProblemData.hpp#L37

@cvxpy/maintainers does that look familiar to anyone?

**Comment by pragmaticTNT** [comment]
I was running `pytest` on `test_problem.py` with the scipy backend after rebuilding with the `undef_macros = ['NDEBUG']` change. However when I did the same thing again today, it just gave me `cvxpy/tests/test_problem.py Fatal Python error: Aborted` so maybe the error message doesn't mean much.

**Review by SteveDiamond — APPROVED** [review]
Really nice work!

File: cvxpy/lin_ops/lin_op.py [hunk 1]
```diff
@@ -75,10 +75,10 @@ def __repr__(self) -> str:
 # Data: None
 DIAG_VEC = "diag_vec"
 # Converts the diagonal of a matrix to a vector.
-# Data: None
+# Data: float
```
phschiele, line 78: Shouldn't these be `int`?

pragmaticTNT: Yeah, but when I was stepping through the code it seems to be passed to the CPP backend as a float.

pragmaticTNT:
> Really nice work!

Thanks!

cvxpy/lin_ops/lin_op.py
phschiele: I changed both to `int`. I debugged and at the point I checked it was an int. You also declared int in the C++ code, so that should match.

phschiele (approved):
Thanks a lot @pragmaticTNT !
Sidenote: At some point we should look into the debug problems for CVXCORE.

cvxpy/lin_ops/lin_op.py
pragmaticTNT: Ok, great

pragmaticTNT:
Yes, for sure! (I'm a bit curious as to why the debugging doesn't work as well.)

I would also like to submit a pull request to update the list of problems first-time contributors could work on (it seems like another problem has also been solved already). Are there any open issues that you think would be good to add?

Issue #2059, opened 2023-03-09
Quad obj + conic constraints bug


SteveDiamond:
**Describe the bug**
Solvers that support quadratic objectives and conic constraints can crash when certain atoms are present in the objective, such as power. 

**To Reproduce**
```
import cvxpy as cp
x = cp.Variable()

prob = cp.Problem(cp.Minimize(x**1.6 + x**2), [x >= 1])
prob.solve(solver=cp.SCS)
```

**Expected behavior**
The power atom and other similar atoms should be canonicalized into conic form rather than trying and failing to be canonicalized as quadratic terms.

**Output**
```
Traceback (most recent call last):
  File "/Users/stevend2/PythonProjects/cvxpy/test.py", line 5, in <module>
    prob.solve(solver=cp.SCS)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/problems/problem.py", line 500, in solve
    return solve_func(self, *args, **kwargs)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/problems/problem.py", line 1061, in _solve
    data, solving_chain, inverse_data = self.get_problem_data(
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/problems/problem.py", line 688, in get_problem_data
    data, inverse_data = solving_chain.apply(self, verbose)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/reductions/chain.py", line 76, in apply
    problem, inv = r.apply(problem)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/reductions/dcp2cone/dcp2cone.py", line 57, in apply
    canon_objective, canon_constraints = self.canonicalize_tree(
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/reductions/dcp2cone/dcp2cone.py", line 98, in canonicalize_tree
    canon_arg, c = self.canonicalize_tree(arg, affine_atom and affine_above)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/reductions/dcp2cone/dcp2cone.py", line 98, in canonicalize_tree
    canon_arg, c = self.canonicalize_tree(arg, affine_atom and affine_above)
  File "/Users/stevend2/PythonProjects/cvxpy/cvxpy/reductions/dcp2cone/dcp2cone.py", line 101, in canonicalize_tree
    canon_expr, c = self.canonicalize_expr(expr, canon_args, affine_above)
[...]

PR #2060, opened 2023-03-09
Fix bug with quadratic objective and conic constraints

SteveDiamond:
## Description
Adds special case handling for power in the objective for the quadratic objective + conic constraints canonicalization path.
Issue link (if applicable): #2059

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/power.py
+++ b/cvxpy/atoms/elementwise/power.py
@@ -338,6 +338,11 @@ class power(Elementwise):
         else:
             return self.args[0].is_constant()
 
+    def _quadratic_power(self) -> bool:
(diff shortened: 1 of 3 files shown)


SteveDiamond:
Tests are passing. Red mark is from SonarCloud.

phschiele (approved):
LGTM, makes sense that `power` can only go into the quadratic objective in special cases.

phschiele:
@SteveDiamond The low coverage is only because the numpy/scipy version pinned in the config we use to run the coverage (macOS 3.8) is not compatible with SCS 3. The test is run successfully in other configs.

Issue #1723, opened 2022-03-23
Allow -1 (negative one) in cp.reshape


rileyjmurray:
In NumPy, calling ``b = np.reshape(a, shape=s)`` for a tuple ``s`` permits one occurrence of -1 in ``s``. The index -1 is automatically replaced by whatever value is needed for ``s`` to specify valid axis dimensions in reshaping ``a``.

For example, ``np.reshape(a, (-1, 1))`` will reshape ``a`` into a column vector, and ``np.reshape(a, (1, -1))`` will reshape ``a`` into a row vector. More generally, ``np.reshape(a, (-1, k))`` will reshape ``a`` into a matrix of shape ``(a.size / k, k)`` if ``a.size / k`` is an integer and raise an error if ``a.size / k`` is not an integer.

CVXPY's reshape function should have the same behavior as NumPy's (in this regard).

This is minor syntactic sugar as CVXPY exists today, since we only support Expressions with at most two dimensions. This feature would become more important once CVXPY supports n-dimensional Expressions.

rileyjmurray:
See also #1695

phschiele:
[Here](https://github.com/scipy/scipy/blob/0dfb5d600299af16b1482731def7c91f7662ec28/scipy/sparse/_sputils.py#L279-L326) is the shape handling of scipy sparse, which might be a relevant starting point for the logic and error checks. It only works for two dimensions, though.

Another quite compact point of reference is the pydata/sparse implementation found [here](https://github.com/pydata/sparse/blob/07eeab2fe9de57904dc590da001bc2f83ba803b8/sparse/_coo/core.py#L1010-L1019).

rileyjmurray:
This missed v1.3. It would be great to have in 1.4!

PR #2061, opened 2023-03-09
Allow -1 in reshape

phschiele:
## Description
Allows replacing one dimension by -1 in reshape.
Issue link (if applicable): #1723 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -605,7 +605,7 @@ class TestAtoms(BaseTest):
         expr_reshaped = cp.reshape(expr, -1)
         self.assertEqual(expr_reshaped.shape, (6,))
 
-        expr_reshaped = cp.reshape(expr, -1)
+        expr_reshaped = cp.reshape(expr, (-1,))
         self.assertEqual(expr_reshaped.shape, (6,))
 
         with pytest.raises(ValueError, match="Cannot reshape expression"):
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_atoms.py
```diff
@@ -588,6 +588,44 @@ def test_reshape(self) -> None:
...
+        self.assertEqual(expr_reshaped.shape, (6,))
+
+        with pytest.raises(ValueError, match="Cannot reshape expression"):
+            cp.reshape(expr, (8, -1))
+
+        with pytest.raises(AssertionError, match="Only one"):
+            cp.reshape(expr, (-1, -1))
+
+        with pytest.raises(ValueError, match="Invalid reshape dimensions"):
+            cp.reshape(expr, (-1, 0))
```
phschiele, line 618: This would be valid in numpy, but we generally do not allow 0 shapes.

cvxpy/tests/test_atoms.py
```diff
@@ -588,6 +588,44 @@ def test_reshape(self) -> None:
         self.assertItemsAlmostEqual(b_reshaped, X_reshaped.value)
         self.assertItemsAlmostEqual(b, X.value)
 
+    def test_reshape_negative_one(self) -> None:
```
SteveDiamond, line 591: For these tests can you add a comparison with NumPy? I'm sure you already did that on your own but it's better to be explicit IMO

phschiele: Done in 2079dfc

SteveDiamond (approved):
LGTM

Issue #2051, opened 2023-02-22
SDP Problem Reduction High Memory Use for MOSEK


cohnt:
**Describe the bug**
Solving a large SDP requires a very large amount of memory to set up in MOSEK, in comparison to other methods such as SCS. For a 150x150 semidefinite matrix variable, MOSEK requires several GB of memory during problem compilation, whereas SCS uses virtually none.

**To Reproduce**
```
import numpy as np
import cvxpy

n = 150
np.random.seed(43)
Q = np.random.normal(loc=0, scale=1, size=(n,n))

X = cvxpy.Variable((n,n), symmetric=True)
constraints = [X[i,i] == 1 for i in range(n)] + [X >> 0]
objective = cvxpy.Minimize(cvxpy.trace(Q @ X))
problem = cvxpy.Problem(objective, constraints)
problem.solve(verbose=True, solver=cvxpy.SCS)
```
Replacing `solver=cvxpy.SCS` with `solver=cvxpy.MOSEK` will demonstrate the greater memory use.

**Expected behavior**
I would expect the problem compilation to not require that much memory.

**Output**
N/A

**Version**
 - OS: Ubuntu 22.04
 - CVXPY Version: 1.3.0

**Additional context**
Possibly related to #1011?

rileyjmurray:
I'm pretty confident that using the MOSEK 10 interface will fix this.

phschiele:
This is caused by the function [vectorized_lower_tri_to_triples](https://github.com/cvxpy/cvxpy/blob/2a23c109e49577d8da4b97bbf6c866400da105c4/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py#L52). As [this comment](https://github.com/cvxpy/cvxpy/blob/2a23c109e49577d8da4b97bbf6c866400da105c4/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py#L179-L181) suggests, it's a known inefficiency.  Since this is called when specifying PSD constraints, I'd agree that this will likely be fixed by the new interface.

The following gives a substantial speedup over the current implementation for the given example:
```py
    A = A_row.tocoo()
    vals = A.data
    flattened_cols = A.col
    cum_cols = np.cumsum(np.arange(dim, 0, -1))
    rows, cols = [], []
    current_col = 0
    for v in flattened_cols:
        for c in range(current_col, dim):
            if v <= cum_cols[c]:
                cols.append(c)
[...]

PR #2063, opened 2023-03-10
Mosek Interface: Keep sparsity for PSD constraints

phschiele:
## Description
The current implementation casts PSD constraints to a dense array before passing them to mosek. This PR keeps the sparsity.
Issue link (if applicable):  #2051 

For reference, here is the script from the to reproduce the issue:

```py
import numpy as np
import cvxpy

n = 130
np.random.seed(43)
Q = np.random.normal(loc=0, scale=1, size=(n, n))

X = cvxpy.Variable((n, n), symmetric=True)
constraints = [X[i, i] == 1 for i in range(n)] + [X >> 0]
objective = cvxpy.Minimize(cvxpy.trace(Q @ X))
problem = cvxpy.Problem(objective, constraints)
problem.get_problem_data(verbose=True, solver=cvxpy.MOSEK)
```

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -43,19 +43,19 @@ def vectorized_lower_tri_to_mat(v, dim):
     :return: Return the symmetric 2D array defined by taking "v" to
       specify its lower triangular entries.
     """
-    rows, cols, vals = vectorized_lower_tri_to_triples(sp.sparse.coo_matrix(v), dim)
+    rows, cols, vals = vectorized_lower_tri_to_triples(v, dim)
     A = sp.sparse.coo_matrix((vals, (rows, cols)), shape=(dim, dim)).toarray()
     d = np.diag(np.diag(A))
     A = A + A.T - d
     return A
 
 
-def vectorized_lower_tri_to_triples(A: sp.sparse.coo_matrix, dim: int) \
+def vectorized_lower_tri_to_triples(A: sp.sparse.coo_matrix | list[float] | np.ndarray, dim: int) \
         -> tuple[list[int], list[int], list[float]]:
     """
     Attributes
     ----------
-    A : scipy.sparse.coo_matrix
+    A : scipy.sparse.coo_matrix | list[float] | np.ndarray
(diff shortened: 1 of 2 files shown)


cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -168,19 +204,18 @@ def bar_data(A_psd, c_psd, K):
...
-                # A_row defines a symmetric matrix by where the first "order" entries
-                #   gives the matrix's first column, the second "order-1" entries gives
-                #   the matrix's second column (diagonal and blow), and so on.
-                rows, cols, vals = vectorized_lower_tri_to_triples(A_row, dim)
-                # TODO: replace the above function with something that only reads the nonzero
-                #  entries. I.e. return *actual* sparse matrix data, rather than data for a dense
-                #  matrix stated in a sparse format.
+
+                A_row_coo = A_row.tocoo()
+                rows, cols, vals = vectorized_lower_tri_to_triples(A_row_coo, dim)
```
phschiele, line 215: I did add a debug snippet to ensure that `rows, cols, vals` is the same as the old implementation (after sparsifying) in all our tests.

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -42,21 +43,56 @@ def vectorized_lower_tri_to_mat(v, dim):
     :return: Return the symmetric 2D array defined by taking "v" to
       specify its lower triangular entries.
     """
-    rows, cols, vals = vectorized_lower_tri_to_triples(v, dim)
+    rows, cols, vals = vectorized_lower_tri_to_triples(sp.sparse.coo_matrix(v), dim)
```
SteveDiamond, line 46: What is the reason for casting ``v`` to a coo_matrix?

SteveDiamond: Ok sorry, I thought ``v`` was always a list.

SteveDiamond (approved):

SteveDiamond:
Should we include this in 1.3? I think so.

phschiele:
> Should we include this in 1.3.1? I think so.

Yes, I'd also add it to the patch release.

PR #2067, opened 2023-03-16
Fix issue with matrix multiplication involving quad

SteveDiamond:
## Description
Right now the dimensions of quad_form are usually (1, 1). This is not right at all, and leads to errors such as #2065. However changing the dimension of quad_form probably needs to wait for the 1.4 release. This is a hotfix that resolves #2065 more narrowly.

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [x] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2058, opened 2023-03-08
Initial cut at adding error on raw float keys and tuples of floats

PTNobel:
## Description

Per the chat, we decided that `cp.Variable(4)[2.5]` should be an error. This is my first cut at making it one.

I wasn't exactly sure where to put the source of the error or how to structure it. My preference would be to whitelist certain types ala NumPy and validate them that way, but I figured I'd start by banning floats since that occurs the most (in code like `x[5/2]`).

It is worth noting we already have this error for arrays of floats:
```python
ipdb> cp.Variable(5)[2:][np.array([1.5])]
*** IndexError: arrays used as indices must be of integer (or boolean) type

```

Let me know if we should move it and how many unit tests this needs!

## Type of change
- [ ] New feature (backwards compatible)
- [x] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2054, opened 2023-02-25
[CI] aarch64 wheels

phschiele:
## Description
Adding wheel build for `aarch64` to the CI.
Issue link (if applicable): #2048 

Tested the build successfully. Results can be found [here](https://github.com/cvxpy/cvxpy/actions/runs/4270415474) (bottom of page).

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2044, opened 2023-02-21
fixing typo in documentation

Transurgeon:
## Description
I found a small typo in the documentation. 
It is in the "derivative fundamental" sub-section. 
"sensitvity" -> "sensitivity"

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2068, opened 2023-03-16
Prep for 1.3.1

SteveDiamond:
## Description
This MR is prep for patch 1.3.1. It includes the following MRs:

#2053 
#2044 
#2043 
#2029 
#2028 
#2022 
#2017 
#2016 
#2008 
#2054
#2063 
#2067
#2058 
#2009
#2060 
#2001

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,13 +124,20 @@ jobs:
           echo "PYTHON_SUBVERSION=$(echo $PYTHON_VERSION | cut -c 3-)" >> $GITHUB_ENV
           echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
 
+      - name: Set up QEMU  # For aarch64, see https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
       - name: Build wheels
         if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
         env:
           CIBW_BUILD: "cp3${{env.PYTHON_SUBVERSION}}-*"
           CIBW_SKIP: "*-win32 *-manylinux_i686 *-musllinux*"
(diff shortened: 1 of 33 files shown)


SteveDiamond:
Unsure whether to include #2054

rileyjmurray:
@SteveDiamond I think that backporting #2054 is fine.

Porting #2009 is a little delicate because I think the updated code might raise a new type of exception that it wasn't raising earlier. @phschiele can say whether or not that's the case. Assuming that the type of exceptions __wouldn't be changed__ by #2009, I think it should be backported.

I'm not in favor of porting #2007. The renaming of those folders is just there to help developers. Moving forward, developers only need to concern themselves with the code that exists on ``master``. So there isn't much value-add in porting #2007. By contrast, porting #2007 might break some peoples' code. (Note: I don't think that would violate semantic versioning by breaking such code. It would be permissible since our declaration of a public API wouldn't be affected. I'm coming at this from a practical perspective of "if it ain't broke, don't fix it.")

PTNobel:
@rileyjmurray re #2007, do you think there will be easier to backport future fixes to ther 1.3 series if we backport the refactor?

rileyjmurray:
@parthb83 hmm ... that's a good question. Hopefully, it wouldn't be too bad. I suggest that we release without #2007 for 1.3.1. If we have trouble preparing 1.3.2 due to missing the changes in #2007, then we can merge #2007 into the 1.3.x branch before any other change needed for 1.3.2. How's that sound?

PTNobel:
SGTM.

phschiele:
@rileyjmurray #2009 reraises the same exception type `ArpackNoConvergence`, only with an extended error message. So it should be fine to backport it.

SteveDiamond:
Ok I think this is done. I don't see any remaining MRs that could go in.

phschiele (approved):
Looks good, double checked all master merges since the 1.3. release. 
The only question I had was #2001, but I saw you removed the backport label. What was the reason for that?

SteveDiamond:
> Looks good, double checked all master merges since the 1.3. release. The only question I had was #2001, but I saw you removed the backport label. What was the reason for that?

Thanks! I thought #2001 was already integrated, but I checked again and it wasn't. Merged in now.

SteveDiamond:
@h-vetinari we are going to release this as a patch soon.

Issue #1724, opened 2022-03-23
Atom for the "outer product"


rileyjmurray:
CVXPY should have an atom that computes outer products, akin to https://numpy.org/doc/stable/reference/generated/numpy.outer.html.

Here's a rough implementation of how that could work:
```
import cvxpy as cp

def outer(a, b):
    a = cp.Expression.cast_to_const(a)  # if a is an Expression, return it unchanged.
    assert a.ndim == 1
    b = cp.Expression.cast_to_const(b)
    assert b.ndim == 1
    a = cp.reshape(a, (a.size, 1))
    b = cp.reshape(b, (1, b.size))
    expr = a @ b
    return expr
```

rileyjmurray:
This missed v1.3. It would be a shame to not get it into 1.4! Help very much appreciated on this.

SteveDiamond:
In master.

PR #2069, opened 2023-03-17
New atom: outer product

Paulnkk:
## Description
I added a function and unit test for the outer product of two vectors mentioned in [#1724](https://github.com/cvxpy/cvxpy/issues/1724).

## Type of change
- [ ] Bug fix

- [x] Write unittests.

bench_test_with
bench_test_without


cvxpy/atoms/affine/binary_operators.py
```diff
@@ -449,3 +451,27 @@ def scalar_product(x, y):
...
+    y : Expression, int, float, NumPy ndarray, or nested list thereof.
+        The transposed-linear argument to the outer product.
+
+    Returns
+    -------
+    expr : Expression
+        The outer product of (x,y), linear in x and transposed-linear in y.
+    """
+    x = Expression.cast_to_const(x)
+    assert x.ndim == 1
```
SteveDiamond, line 472: It might be better to throw errors here rather than just having asserts.

SteveDiamond:
You should install the precommit: https://www.cvxpy.org/contributing/index.html#code-style

SteveDiamond:
Thanks for adding this! It looks generally fine to me. The main missing piece is updating the web documentation.

cvxpy/atoms/affine/binary_operators.py
Paulnkk: @SteveDiamond sure I can add this !

Paulnkk:
I can update the web docu, is the convention to start a new PR for this update ?

SteveDiamond:
> I can update the web docu, is the convention to start a new PR for this update ?

Usually we update the docs within the PR that adds the function. Relevant page is doc/source/tutorial/functions/index.rst
There's also another change needed to have outer_product show up in the API reference but I'm not sure how to do that. @rileyjmurray can you help?

cvxpy/atoms/affine/binary_operators.py
Paulnkk: @SteveDiamond I added the error messages here: [Add error message](https://github.com/cvxpy/cvxpy/pull/2069/commits/96a032d1516e133f32a3b2637f362a03cedbbb92). Just the web docu left from my side

Paulnkk:
@phschiele added outer in docu and api here: [Add outer to docu and api](https://github.com/cvxpy/cvxpy/pull/2069/commits/a5e9afcdfd2a9c20d3f60fce38b66b004676f58c)

phschiele:
@Paulnkk @SteveDiamond In numpy the function is just called `outer`, not `outer_product`.
https://numpy.org/doc/stable/reference/generated/numpy.outer.html
I think we should follow numpy syntax here.

Paulnkk:
@phschiele made adjustments for outer convention in code here: [Change outer_product to outer](https://github.com/cvxpy/cvxpy/pull/2069/commits/06d3aa765cf153d7b98af550a3e6ce02e630d729) and in documentation here: [Adjust docu outer_product to outer](https://github.com/cvxpy/cvxpy/pull/2069/commits/d971849629ae3d543d7faf1ed9ecc967602a0237)

Paulnkk:
Also one question, I found a spot in the code (please find screenshot attached in line 40) where the outer product is calculated manually, is it desirable to make a change here ? 
Screenshot 2023-03-25 at 16 33 22

phschiele:
> Also one question, I found a spot in the code (please find screenshot attached in line 40) where the outer product is calculated manually, is it desirable to make a change here ? Screenshot 2023-03-25 at 16 33 22

@Paulnkk yes, that would be great. I have also searched the codebase using this regex `.*reshape.*@.*reshape`. There are no more occurrences.

Paulnkk:
@phschiele I have made this change and have encountered a problem: The part of the code that needs to be changed also accepts scalars and matrices for x and w_unique which causes problems because the outer product (as it is currently implemented) is only defined for vectors. This is related to the tests for the 'dotsort' atom, which also accepts matrices and scalars (see failing tests in the screenshot). Accepting scalars for the current outer product makes sense, but applying the outer product to matrices I think rather not (e.g. AB^T would be a special form of matrix multiplication but not the outer product defined for matrices). How should I proceed with the proposed change?
Screenshot 2023-03-26 at 18 52 56

doc/source/api_reference/cvxpy.atoms.affine.rst
```diff
@@ -117,6 +117,13 @@ multiply
 .. autoclass:: cvxpy.multiply
     :show-inheritance:
 
+.. _outer:
```
Paulnkk, line 120: Here I just moved the outer product api reference to the position that it fits in alphabetical order and replaced vstack, everything is complied as it should be, if wanted I can provide a screenshot with the compiled version

phschiele:
> @phschiele I have made this change and have encountered a problem: The part of the code that needs to be changed also accepts scalars and matrices for x and w_unique which causes problems because the outer product (as it is currently implemented) is only defined for vectors. This is related to the tests for the 'dotsort' atom, which also accepts matrices and scalars (see failing tests in the screenshot). Accepting scalars for the current outer product makes sense, but applying the outer product to matrices I think rather not (e.g. AB^T would be a special form of matrix multiplication but not the outer product defined for matrices). How should I proceed with the proposed change? Screenshot 2023-03-26 at 18 52 56

@Paulnkk numpy does accept scalars, vectors, and matrices:
```    
a : (M,) array_like
First input vector.  Input is flattened if
not already 1-dimensional.
```
We could adopt a similar convention, i.e., we say:
```py
if x.ndim != 1:
    x = cp.vec(c)
```

Paulnkk:
@phschiele thanks for the suggestion ! Made changes here: [Flatten outer input](https://github.com/cvxpy/cvxpy/pull/2069/commits/0c421cc41864f52bfb3cf67391c9162897bfed90). I also added a comment that input is flattened. Additionally, I made changes related to the dotsort atom that we discussed earlier here: [Adjust dotsort with outer](https://github.com/cvxpy/cvxpy/pull/2069/commits/45d0de57152f1eec43c4a99536b35025c30f517d)

cvxpy/tests/test_atoms.py
```diff
@@ -1248,6 +1248,24 @@ def test_scalar_product(self) -> None:
...
+
+        # Test with parameter
+        c = Parameter((2,))
+        expr = cp.outer(c, a)
+        self.assertEqual(expr.shape, (2, 3))
+
+        d = np.ones((4,))
+        expr = cp.outer(a, d)
+        true_val = np.outer(a, d)
+        assert np.allclose(expr.value, true_val, atol=1e-1)
```
SteveDiamond, line 1267: Can you add some more tests with matrix inputs and verify that the behavior is the same as numpy? We already know the answer but it's good to put it in the tests as a kind of documentation.

phschiele: @SteveDiamond Added scalar and matrix tests below. Note that there is a difference for matrix inputs because numpy flattens in row-major and cvxpy in column-major order. I have added a note to the docstring.

Also removed the `vec()` because it was redundant with the immediately following reshape.

SteveDiamond: Hmm do you think the discrepancy is ok? It's a bit worse than for reshape because you can't pass an argument to outer that controls how it's flattened.

phschiele: I think it is consistent with the remainder of the code. Numpy also does not allow you to specify the order of the flattening in `np.outer`, so if you want to do a row-major flattened outer product in cvxpy, you'd have to do:
`cp.outer(cp.vec(A,order='C'),cp.vec(A,order='C'))`.
I agree that this can lead to very subtle bugs, though. So we could also just not allow matrix inputs, even though they are allowed in numpy.

rileyjmurray:
I think we should avoid diverging from NumPy’s behavior in our API. The
divergence with how we handle broadcasting (or rather, how we don’t handle
it) is a big pain point. This is a much smaller-scale point, but the
philosophy still applies.

cvxpy/tests/test_atoms.py
SteveDiamond: I agree with Riley. I think the best move here for long term compatibility with NumPy is to not allow matrix inputs for now and then enable them with column-major flattening when we switch to full NumPy matching in 2.0.

SteveDiamond: Edit: column-major -> row-major

phschiele: Disallowed 2D inputs in the latest commit. Also added a small reminder assertion that alerts us to change this once we switched to row-major default.

SteveDiamond (approved):
LGTM

phschiele (approved):

PR #2072, opened 2023-03-20
Fix typo in semidefinite program example

domenicocinque:
## Description
The minimization should be with respect to the trace of Sigma.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
I'm not sure this is a typo. You could have an objective for the problem but you don't need one.

domenicocinque:
Yes you are right, that makes sense. I will close this PR.

Issue #2070, opened 2023-03-17
Change quad_form shape


phschiele:
**Describe the bug**
`cp.quad_form` currently has shape `(1,1)`, but should be scalar, i.e., `()`. 
#2067 Fixes a special case for subsequent matrix multiplication, but the proper fix is to correct the shape with the next major release. 

**To Reproduce**
```py
import cvxpy as cp
import numpy as np

x = cp.Variable(2)
P = np.eye(2)
print(cp.quad_form(x, P).shape)
>>> (1,1)

y = np.arange(2)
print((y.T @ P @ y).shape)
>>> ()
```

Note: PR should also remove the comment about the wrong shape of `quad_form` added in #2067.

SteveDiamond:
Fixed

PR #2073, opened 2023-03-20
Change shape quadratic form 

Paulnkk:
## Description
I adjusted the returned shape for 1-dimensional quadratic form and adjusted test_matmul_scalars. 
Issue link (if applicable):
Found issue on this page: https://github.com/cvxpy/cvxpy/issues/2070

## Type of change
- [ ] Bug fix

- [x] Write unittests.
bench_without
Screenshot 2023-03-20 at 20 05 53

### DIFF
--- a/cvxpy/atoms/quad_form.py
+++ b/cvxpy/atoms/quad_form.py
@@ -122,7 +122,7 @@ class QuadForm(Atom):
         return [sp.csc_matrix(D.ravel(order="F")).T]
 
     def shape_from_args(self) -> Tuple[int, ...]:
-        return tuple() if self.args[0].ndim == 0 else ()
+        return tuple()
 
 
 class SymbolicQuadForm(Atom):
--- a/cvxpy/atoms/quad_form.py
+++ b/cvxpy/atoms/quad_form.py
@@ -122,7 +122,7 @@ class QuadForm(Atom):
         return [sp.csc_matrix(D.ravel(order="F")).T]
 
     def shape_from_args(self) -> Tuple[int, ...]:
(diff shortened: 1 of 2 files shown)


cvxpy/atoms/quad_form.py
```diff
@@ -122,7 +122,7 @@ def _grad(self, values):
         return [sp.csc_matrix(D.ravel(order="F")).T]
 
     def shape_from_args(self) -> Tuple[int, ...]:
-        return tuple() if self.args[0].ndim == 0 else (1, 1)
+        return tuple() if self.args[0].ndim == 0 else ()
```
phschiele, line 125:
```suggestion
        return tuple()
```
`tuple()` is the same as `()`. 
Also because of the second part of this validation on the inputs above
https://github.com/cvxpy/cvxpy/blob/bf4634cc6c9a35ec31b70c8df81e8aa9759aef55/cvxpy/atoms/quad_form.py#L54-L55
we can never have `self.args[0].ndim == 0`, so this `if` case appears to be redundant.

Paulnkk: @phschiele thanks a lot for the suggestion! Adjusted here: [Remove redundant if condition](https://github.com/cvxpy/cvxpy/pull/2073/commits/9c119a5bf1461aa1405910f529c6f17495e59cce) and added assertion statement in tests: [Add assert statement in tests](https://github.com/cvxpy/cvxpy/pull/2073/commits/b0c28015cdd11da47fab399aed261518b9c44b1a)

phschiele (approved):
LGTM, thanks @Paulnkk!

SteveDiamond (approved):

PR #2079, opened 2023-03-27
Mosek IIS

SteveDiamond:
## Description
This adds basic support for extracting the IIS from Mosek. The IIS is returned in ``problem.solver_stats.extra_stats["IIS"]``. The IIS is represented as a dict mapping constraint id to a flattened output from Mosek for that constraint. An obvious improvement would be to reshape the vector to match the constraint dimension.

This is WIP because I'm not sure this is the best approach for exposing the IIS.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


aszekMosek:
Just to say that the Farkas certificate from Mosek is not necessarily irreducible. It will be in practice in most "easy" cases esp. when detected in presolve.

Issue #2050, opened 2023-02-22
SCIP reports OPTIMAL_INACCURATE solution, eventhough no solution has been found


Old-Shatterhand:
Hi,

**Describe the bug**
I'm trying to solve some DQCPs with SCIP but I encounter some "unintuitive" behavior. SCIP reports having found an optimal_inaccurate solution, even though nothing has been found.

**To Reproduce**
This [file](https://github.com/cvxpy/cvxpy/files/10803299/cluster_cold_single_issue.txt) contains the python code to produce the error. I had to rename it to `.txt` as `.py` is prohibited by github. To execute it, rename it to `.py` and run it in a python environment with `cvxpy` and `numpy` installed.

**Expected behavior**
In the following, I'm talking about the `_solve` method in `scip_conif.py` ([link](https://github.com/cvxpy/cvxpy/blob/687f45c7fe9493db480fbbbc234a4bf89f96f6ca/cvxpy/reductions/solvers/conic_solvers/scip_conif.py#L347)) and especially about lines 347 to 350.
The program is terminated due to reaching the maximal time, i.e., `model.getStatus()` is `'timelimit'`. At the same time `model.getNSols()` and `model.getNCountedSols()` are both 0, i.e. no solution has been found at all. I expect the cvxpy to report accordingly and set `solution["status"]` to be some constant indicating that no solution has been found. I suggest setting `solution["status"] = s.SOLVER_ERROR` as the solver hasn't found a solution and we cannot know if the program is even feasible.

**Output**
This [file](https://github.com/cvxpy/cvxpy/files/10803332/issue_stacktrace.txt) contains the program output produced by the code linked above.

**Version**
 - OS: Windows 11 and Rocky Linux (error was produced on both systems)
 - CVXPY Version: 1.2.2
 - NumPy Version: 1.22.3

**Additional context**
Nothing.

Thank you in advance for your help.

Best, Roman

Old-Shatterhand:
A minor update: If I run the program with MOSEK instead of SCIP, I don't get the error. But because MOSEK is proprietary and SCIP is not, I really would like to stick to SCIP.

phschiele:
@Old-Shatterhand I agree, for `timelimit`, the status should be mapped to `s.OPTIMAL_INACCURATE` only if at least one solution was found.

Transurgeon:
Hey @Old-Shatterhand, I was wondering if you help me reproduce the bug? 
[...]

PR #2080, opened 2023-03-27
fix: scipy time limit raises SolverError

allenlawrence94:
## Description
When using SciPy, if the solver reaches it's set time limit, a `SolverError` is raised. This change makes it so these instances are instead reported as `OPTIMAL_INACCURATE`. I think this behavior is more consistent with other solvers - this is at least how GLPK works currently.

I haven't added a unit test for this. Let me know if this seems necessary.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
@@ -42,7 +42,7 @@ class SCIPY(ConicSolver):
 
     # Map of SciPy linprog status
     STATUS_MAP = {0: s.OPTIMAL,  # Optimal
-                  1: s.SOLVER_ERROR,  # Iteration limit reached
+                  1: s.OPTIMAL_INACCURATE,  # Iteration limit reached
                   2: s.INFEASIBLE,  # Infeasible
                   3: s.UNBOUNDED,  # Unbounded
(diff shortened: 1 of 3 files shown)


SteveDiamond:
Thanks for making this PR! You should make sure a solution is actually present when the iteration limit/time limit is reached. We made a similar change to the SCIP interface, which caused a different bug (see #2050). This is a bit hard to unit test but if you could try I'd appreciate it (in test_conic_solvers.py).

allenlawrence94:
Will do! Apologies for not searching issues first

allenlawrence94:
@SteveDiamond You were right to be concerned! It seems that when a time limit is reached, scipy will only sometimes return a solution. I was only able to get "Time limit reached" with a solution present when solving a MI problem with presolve=True (which is the default). 

The unit test I added may be flaky since it uses a time limit, but, unfortunately, SciPy/HiGHS has no iteration limit for MI problems. If you're concerned about this, we could drop the "run without enough time to find optimum" part.

SteveDiamond:
Since the  "run without enough time to find optimum" test passes on all CI platforms I think it's probably ok. @rileyjmurray @phschiele what do you think?

SteveDiamond (approved):

PR #2084, opened 2023-03-29
Fix scip timelimit no solution

Paulnkk:
## Description
I added conditions to check if timelimit for SCIP solver is reached and no solutions (feasible and/or optimal) are generated. If those conditions are met, solution status is set to SOLVER_ERROR. I also added the case if timelimit is reached and a non-optimal and feasible solution is generated to set OPTIMAL_INACCURATE. Additionally, I added a unit test to check for the SOLVER_ERROR case.
Issue link (if applicable):
I found the issue here: https://github.com/cvxpy/cvxpy/issues/2050

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
@@ -351,6 +351,16 @@ class SCIP(ConicSolver):
         if solution["status"] == s.SOLVER_ERROR and model.getNCountedSols() > 0:
             solution["status"] = s.OPTIMAL_INACCURATE
 
+        if model.getStatus() == 'timelimit' \
+                and model.getNCountedSols() == 0 \
+                and model.getNSols() == 0:
+            solution["status"] = s.SOLVER_ERROR
+
+        if model.getStatus() == 'timelimit' \
+                and model.getNSols() > 0 \
+                and model.getNCountedSols() == 0:
+            solution["status"] = s.OPTIMAL_INACCURATE
+
         return solution
 
     def add_model_lin_constr(
--- a/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
@@ -17,7 +17,7 @@ limitations under the License.
 import logging
 from typing import Any, Dict, Generic, Iterator, List, Optional, Tuple, Union
 
-from numpy import array, ndarray
+import numpy as np
 from scipy.sparse import dok_matrix
 
 import cvxpy.settings as s
@@ -190,7 +190,7 @@ class SCIP(ConicSolver):
         dims = dims_to_solver_dict(data[s.DIMS])
         return A, b, c, dims
 
-    def _create_variables(self, model: ScipModel, data: Dict[str, Any], c: ndarray) -> List:
+    def _create_variables(self, model: ScipModel, data: Dict[str, Any], c: np.ndarray) -> List:
         """Create a list of variables."""
         variables = []
         for n, obj in enumerate(c):
@@ -211,7 +211,7 @@ class SCIP(ConicSolver):
             model: ScipModel,
             variables: List,
             A: dok_matrix,
-            b: ndarray,
+            b: np.ndarray,
             dims: Dict[str, Union[int, List]],
     ) -> List:
         """Create a list of constraints."""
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_conic_solvers.py
```diff
@@ -1813,6 +1814,15 @@ def test_scip_test_params__invalid_scip_params(self) -> None:
             exc = "One or more scip params in ['a'] are not valid: 'Not a valid parameter name'"
             assert ke.exception == exc
 
+    def test_scip_time_limit_no_solution(self) -> None:
+        sth = sths.mi_lp_6()
+
+        with pytest.raises(error.SolverError) as se:
+            sth.solve(solver="SCIP", scip_params={"limits/time": 0.01})
```
SteveDiamond, line 1821: You can probably set the timelimit to 0.

Paulnkk:
> Thanks for adding this! Could you model your tests off of #2080 ? It might be better to merge #2080 first and then you can use exactly the same test.

Yes sure, I can use the same milp but I think the unit tests might be different (due to different solvers). I currently think about adding an additional unit test or extend the existing one since I added another case (timelimit reached and feasible solution generated which is not optimal) but this case is more difficult to achieve for a test problem.

SteveDiamond:
Thanks @Paulnkk! I merged #2080 and reworked the tests to be similar to that. Let's iterate on this.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -1817,7 +1817,7 @@ def test_scip_time_limit_reached(self) -> None:
         sth = sths.mi_lp_7()
 
         # run without enough time to find optimum
-        sth.solve(solver="SCIP", scip_params={"limits/time": 0.01})
```
Paulnkk: I tried to increase the timelimit in order to get the sweet spot between getting a Solver error (timelimit too low) and generating an optimal solution (timelimit too high). For the tests on my Machine everything is doing well but it seems that the test on Windows is failing (timelimit seems too high that an optimal solution is generated). How can I make adjustments on the tests so that the other ones are passing ?

cvxpy/tests/test_conic_solvers.py
```diff
@@ -1817,7 +1817,7 @@ def test_scip_time_limit_reached(self) -> None:
...
+        sth.solve(solver="SCIP", scip_params={"limits/time": 0.03})
```
Paulnkk, line 1820: I tried to increase the timelimit in order to get the sweet spot between getting a Solver error (timelimit too low) and generating an optimal solution (timelimit too high). For the tests on my Machine everything is doing well but it seems that the test on Windows is failing (timelimit seems too high that an optimal solution is generated). How can I make adjustments on the tests so that the other ones are passing ?

SteveDiamond: Yeah I don't understand what's happening on windows. There's no way it's generating an optimal solution. I think there must be some difference with how SCIP is working on windows and how it's working on linux. @phschiele can we use codespaces to debug the windows tests?

phschiele: @SteveDiamond I'm not aware of using codespaces to run on a windows machine.
The Linux machine I tested on this problem only took 0.07s. It would be curious if windows were faster here, but it is still possible.
@Paulnkk I assume if you add less time, other tests would already fail with solver error?
We might add a range of time limits (e.g., `0.01, 0.02, 0.03`) and check that we have an inaccurate solution at least once to make it a bit more stable. However, this generally indicates that this test is indeed a bit flaky.

SteveDiamond: It's ok if we remove the test.

SteveDiamond: Also Windows still passes with 0.01 seconds.

SteveDiamond: I may know what the issue is. Trying a fix.

Paulnkk: @phschiele Yes, tests will fail with a solver error if I reduce the timelimit

SteveDiamond: My fix didn't work. I commented out the test.

SteveDiamond:
Ok I think this is good to merge. I found one more bug in determining whether a solution was present but non-optimal.

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -365,8 +365,8 @@ def _solve(
             solution["status"] = s.SOLVER_ERROR
 
         if model.getStatus() == 'timelimit' \
-                and model.getNSols() > 0 \
-                and model.getNCountedSols() == 0:
```
Paulnkk: One question; for this changing conditions, I am not sure to map the solution status to OPTIMAL_INACCURATE since if the condition model.getNCountedSols() > 0 implies that we have an optimal solution (considering the or)?

cvxpy/reductions/solvers/conic_solvers/scip_conif.py
```diff
@@ -365,8 +365,8 @@ def _solve(
...
+                and (model.getNSols() > 0 \
+                or model.getNCountedSols() > 0):
```
Paulnkk, line 369: One question; for this changing conditions, I am not sure to map the solution status to OPTIMAL_INACCURATE since if the condition model.getNCountedSols() > 0 implies that we have an optimal solution (considering the or)?

SteveDiamond: I was confused about this too, but getNCountedSols actually is the number of feasible solutions found. So getNCoutnedSols > 0 doesn't mean there's an optimal solution.

phschiele (approved):

Issue #2083, opened 2023-03-29
.curvature randomly being unknown


timothy-nunn:
**Describe the bug**
Curvature of the attached example is not deterministic. It appears to be coming from the `quad_form` routine which will sometimes be convex, and sometimes be unknown. 

**To Reproduce**
Please run attached python file (as a file because the matrix used is huge and I cannot replicate this issue for random numbers, for this I can only apologise). The gist is we create what should be a convex atom over and over again and print its curvature each time.

**Expected behavior**
The eigenvalues of `B` are all positive, so I would expect the `quad_form_part.curvature` line to print as "Convex" everytime. Regardless, I would expect consensus on whether the problem has unknown or convex curvature.

**Output**
```
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'UNKNOWN'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'UNKNOWN'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'UNKNOWN'
quad_form_part.curvature = 'CONVEX'
quad_form_part.curvature = 'UNKNOWN'
quad_form_part.curvature = 'CONVEX'
...
```

**Version**
 - OS: macOS Ventura 13.0.1 (Apple M1 MacBook Air)
 - CVXPY Version: 1.2.1 and 1.3.1
 - Python Version: 3.8.13
 - Scipy Version: 1.9.1
 - Numpy Version: 1.23.2
[non_deterministic_curvature.py.txt](https://github.com/cvxpy/cvxpy/files/11099460/non_deterministic_curvature.py.txt)

rileyjmurray:
Thanks for reporting, @timothy-nunn. The problem is that we use the scipy method ``eigvalsh`` to try and compute the (algebraically) smallest eigenvalue of a given matrix. That function uses a random starting vector in an iterative method. We can fix this specific issue by generating the random starting vector for ``eigvalsh`` ourselves, with a fixed random seed. We'd welcome a contribution from you if you would be up for implementing that.

Paulnkk:
[...]

PR #2086, opened 2023-03-31
Fix rare bugs in curvature calculation: identity matrices and nondeterministic starting vectors in eigsh

Paulnkk:
## Description
I have included the suggestions from the issue link below. I used a seed and generated a random vector for Lanczos method to calculate eigenvalues. The curvature from the issue link below is now set to UNKNOWN because one eigenvalue which is close to 0.
Issue link (if applicable):
The issue was found with the following link: https://github.com/cvxpy/cvxpy/issues/2083
## Type of change

- [x] Bug fix

- [ ] Write unittests.


Transurgeon:
@Paulnkk I looked at your PR and it resolves the two issues I was talking about...
This was my implementation:
        np.random.seed(0)
        return sparla.eigsh(A, k=1, sigma=sigma, v0=A[np.random.randint(A.shape[0])],
                            which='SA', return_eigenvectors=False)
I used randint to get an index within the size of the array.. not sure why it causes problems.
Anyways, what defined seed produces this new bug you mention? (where the curvature is always unknown)

Also, we probably need to add some tests. Is it feasible to just add the 19 by 19 example we have been testing with into test_quad_form.py?

Paulnkk:
> @Paulnkk I looked at your PR and it resolves the two issues I was talking about... This was my implementation: np.random.seed(0) return sparla.eigsh(A, k=1, sigma=sigma, v0=A[np.random.randint(A.shape[0])], which='SA', return_eigenvectors=False) I used randint to get an index within the size of the array.. not sure why it causes problems. Anyways, what defined seed produces this new bug you mention? (where the curvature is always unknown)
> 
> Also, we probably need to add some tests. Is it feasible to just add the 19 by 19 example we have been testing with into test_quad_form.py?

@Transurgeon, the example mentioned in the issue link has a positive semi-definite matrix which should imply convexity for its quad_form from my point of view (not additionally "unknown" as shown: https://github.com/cvxpy/cvxpy/issues/2083), here it seems to me that the output of the curvature is dependent on some randomness which should not be the case (?). We now try to avoid this "randomness" by setting the seed and a random starting vector for Lanczos method to determine the max eigenvalue. 

Now I see one problem as the following (what I also mentioned in the PR): By setting the seed and having this "switching" property from the example of 'unknown' and 'convex', I can determine the 'unknown' curvature instead of the 'convex' curvature and get in the example (I added in the PR) all the times 'Unknown' instead of 'Convex'. 

Another problem that is not explicitly related to the problem are the tests (please see the one failing macos test below); there seems to be a technical issue with a psd check for a specific mac version. Also another thing is the scipy check (I also ran into an issue on my machine with this) which was added a couple PRs before. But actually, I do not know if these problems were created by my 'master merge' (or some local changes) or for another reason.

Maybe you can help me (or another person who knows how to do it) to run all the relevant tests for a specific branch without pushing changes so that I can check for version dependent fails (as in this PR).

Paulnkk:
@rileyjmurray I am currently facing the issue that the suggested lines I added are leading to the desired output concerning the problem mentioned in the issue link but I do not pass a particular test. Since I can not reproduce this error on my machine and I assume its inconvenient to push commits every time I make a change to see if this error is fixed, is there a way to run all the tests on my machine to potentially reproduce this error ?

SteveDiamond:
> @rileyjmurray I am currently facing the issue that the suggested lines I added are leading to the desired output concerning the problem mentioned in the issue link but I do not pass a particular test. Since I can not reproduce this error on my machine and I assume its inconvenient to push commits every time I make a change to see if this error is fixed, is there a way to run all the tests on my machine to potentially reproduce this error ?

@Paulnkk @phschiele would github codespaces help here?

phschiele:
I tried it on codespaces and the tests are passing there as well.
The failing test is this one:

```py
        n = 50
    
        # trivial cases
        psd = np.eye(n)
        nsd = -np.eye(n)
    
        assert cp.Constant(psd).is_psd()
>       assert not cp.Constant(psd).is_nsd()
```
It is a bit odd that fixing the seed vector would break the test that was specifically declared as a "trivial case" .

rileyjmurray:
1. The initial vector should have iid Gaussian entries.

2. I’m pretty sure our code special-cases for diagonal matrices. Does that
path only get hit for sparse matrices? If so we should definitely update it
to apply to dense matrices as well.

rileyjmurray:
Actually in retrospect the identity matrix is known to create problems for
iterative eigenvalue solvers. The only way we can catch this case reliably
is by special-casing diagonal matrices.

Paulnkk:
There are several things I have questions about: 1.) I picked Gaussian entries with mu = 0 and sigma = 1 but since we hand over this randomly generated vector for Arnoldi, I am not sure how convergency of Arnoldi is affected but this initially randomly generated starting vector, could potentially affect speed for large scale dimensions. 2.) I now understand the error I got from your explanation and I tried to make the trivial test more versatile in terms of eigenvalues and eigenvectors but I seem to get additional failing tests now. 3.) Could you please explain by what do you mean with "The only way we can catch this case reliably is by special-casing diagonal matrices." ? Is it related to differently generate curvature info for diag matrices ? For instance Check for diag matrix property and get the minimum value from the trace instead of using numerical methods here ?

Paulnkk:
I had to revert the trivial psd checks from this commit [Adjust trivial psd check](https://github.com/cvxpy/cvxpy/pull/2086/commits/18a0222569669bd0a89e17d7bbd83eea67c870fc) back because almost all tests were failing.

rileyjmurray:
> There are several things I have questions about: 1.) I picked Gaussian entries with mu = 0 and sigma = 1 but since we hand over this randomly generated vector for Arnoldi, I am not sure how convergency of Arnoldi is affected but this initially randomly generated starting vector, could potentially affect speed for large scale dimensions. 2.) I now understand the error I got from your explanation and I tried to make the trivial test more versatile in terms of eigenvalues and eigenvectors but I seem to get additional failing tests now. 3.) Could you please explain by what do you mean with "The only way we can catch this case reliably is by special-casing diagonal matrices." ? Is it related to differently generate curvature info for diag matrices ? For instance Check for diag matrix property and get the minimum value from the trace instead of using numerical methods here ?

@Paulnkk, for (1), Arnoldi will only converge to the correct eigenspace if the starting vector has a nontrivial projection onto that eigenspace. Sampling a vector whose components are iid Gaussian ensures that this happens with probability one, no matter the matrix or eigenspace that you're considering. The complexity of forming the Gaussian random vector is negligible.

For (3), I mean that we should check if the input matrix is diagonal. If it is, then we should determine whether or not it's PSD simply by checking that its minimum diagonal entry is nonnegative. That is, we would simply check that ``min([A[i,i] for i in range(n)]) >= 0``, if ``A`` is an array of order ``n``. (I'm not sure how you would consider using the trace. That would not suffice for a PSD check even if the matrix was diagonal.)

Paulnkk:
@rileyjmurray Thanks! Concerning 3.) sorry translation mistake from my side; I meant using the min element from the diagonal (I mentioned as trace) of matrix as you mentioned it

Paulnkk:
I used this branch on codespace, but I am still not sure how to run specific tests from the build-in codespace to check why those tests are failing, very grateful for a small hint how to do that :) Also, I do not know why but I am using a mac m1 version 12 and tests are passing without a fail.

phschiele:
@Paulnkk Make sure to have pytest and the Python extension installed. You then need to configure tests by typing:
`Ctrl + P`, `Python: Configure Tests`, `Pytest`, `cvxpy`.

You should then see a beaker icon in the left sidebar. Clicking on this reveals the individual test cases that can be run one at a time (see below).

Since #2088 this is done automatically, so if it doesn't work out you can create a new codespace after making sure that the current master is merged into this branch.

image

Paulnkk:
@phschiele, thanks for the support! It works itself but I am currently facing the issue that the tests which are failing and I need to check for fixing the errorors are skipped (please find attached a screenshot for one example test). I see from the test report that I am using the test on linux(?). Maybe I can change the system on which the test runs and therefore those tests are not skipped (or maybe I can avoid skipping the test). 
Screenshot 2023-04-26 at 17 46 26

cvxpy/utilities/linalg.py
```diff
@@ -61,8 +74,23 @@ def is_psd_within_tol(A, tol):
     if gershgorin_psd_check(A, tol):
         return True
 
+    if is_diagonal(A):
+        if isinstance(A, csc_matrix):
+            A = A.toarray()
+            min_diag_entry = np.min(np.diag(A))
```
rileyjmurray, line 80: This conversion to a dense array can be avoided. Since we've checked that the matrix is diagonal, we know that ``A.data`` contains the diagonal of ``A`` (possibly with explicit zeros elsewhere). So it suffices to check ``np.all(A.data >= -tol)`` without converting to an array.

Or you could set ``min_diag_entry = np.min(A.data)``, with the understanding that this might equal zero in an unusual case where off-diagonal zero entries are stored explicitly.

rileyjmurray (changes requested):
Mostly looks good! I have one small comment in ``is_psd_within_tol``.

cvxpy/utilities/linalg.py
Paulnkk: @rileyjmurray thanks for the suggestion! Made the change here: [Remove dense array conversion](https://github.com/cvxpy/cvxpy/pull/2086/commits/81b8778b192d3f122b1a12ed5b10d58cacdc3a1a)

rileyjmurray:
@Paulnkk, it occurred to me that we shouldn't be setting the global seed for NumPy's random number generator. 

We can define our own random stream with this API: https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng. For example
```
g = np.random.default_rng(123)
v0 = g.normal(loc=0.0, scale=1.0, size=n)
```
That won't quite cut it though, since we need to support the legacy API as well: https://numpy.org/doc/stable/reference/random/legacy.html#legacy
```
g = np.random.random_state(123)
v0 = g.normal(loc=0.0, scale=1.0, size=n)
```
So we should really get the random vectors using a code snippet like this:
```
if hasattr(np.random, 'default_rng'):
    g = np.random.default_rng(123)
else:
    g = np.random.RandomState(123)
v0 = g.normal(loc=0.0, scale=1.0, size=n)
```
I think we should be good to merge once you make this change in the appropriate places.

rileyjmurray:
@Paulnkk I'll go ahead and make the changes I requested in my last comment.

Paulnkk:
@rileyjmurray I can also make the changes, sorry for the late response. I visited my graduation. Planned to make changes on my way back and commit changes when I have access again.

rileyjmurray:
@Paulnkk ah! Well, in that case, I'll let you finish. I'll be sure to review shortly after your changes show up :)

Paulnkk:
@rileyjmurray thanks a lot ! :) 
I will tag you as soon as I made the changes and wifi connection (will have a bigger travel now and unfortunately no wifi)

rileyjmurray (approved):
Looks good to me!

rileyjmurray (approved):
Looks good to me!

rileyjmurray (approved):
Looks good to me! (For reasons that I don't understand, GitHub posted my approval comment three times.)

Issue #2089, opened 2023-04-01
Improve string representations


phschiele:
Going through some of the major classes and improving their string representations.
Two examples:
```py
cp.Variable(2, name='x').__str__()
>>> 'x'
cp.Variable(2, name='x').__repr__()
>>> 'Variable((2,))' 
```
Maybe the latter could still show the variable name.

```py
cp.Constant(np.ones((3,3))).__str__()
>>> '[[1. 1. 1.]\n [1. 1. 1.]\n [1. 1. 1.]]'
cp.Constant(np.ones((3,3))).__repr__()
>>> 'Constant(CONSTANT, NONNEGATIVE, (3, 3))'
```
Maybe the former could print only a few elements, as it gets hard to read when the arrays contain many elements.

Transurgeon:
Hey @phschiele , I tried some fixes for this issue and just wanted to make sure they were in-line with what you wanted.
Here is what I get for the two changes you mentioned.
```
cp.Variable(2, name='x').__repr__()
'Variable((2,), x)'
cp.Constant(np.ones((3,3))).__str__()
'[[1. 1. 1.]\n [1. 1. 1.]\n  ...]'
```

phschiele:
@Transurgeon That's already looking good! What about the number of columns when the data matrix gets larger? Will they also be partially hidden?

Transurgeon:
@phschiele No, but I can try to work on that. I think for more than 5 columns we should add some '...' to hide the rest of the data. 
```
cp.Constant(np.ones((5,5))).__str__()
'[[1. 1. 1. 1. 1.]\n [1. 1. 1. 1. 1.]\n  ...]'
```
But for more it would look like this (I think adding a comma helps to visually separate the "1." from the three dots)
```
cp.Constant(np.ones((1,7))).__str__()
'[[1. 1. 1. 1. 1. , ... ]]'
```

phschiele:
Are you using the numpy functionality for this?
```py
np.set_printoptions(edgeitems=3)
print(np.arange(60).reshape((6, 10)))
# [[ 0  1  2 ...  7  8  9]
#  [10 11 12 ... 17 18 19]
#  [20 21 22 ... 27 28 29]
#  [30 31 32 ... 37 38 39]
#  [40 41 42 ... 47 48 49]
#  [50 51 52 ... 57 58 59]]

np.set_printoptions(edgeitems=2)
print(np.arange(60).reshape((6, 10)))
[...]

PR #2091, opened 2023-04-01
improve string representation, especially shortening expression.tostring

Transurgeon:
## Description
I made changes to three files:

<b>expression.py</b>: here I changed the expression toString method, by using [numpy.array2string](https://numpy.org/doc/stable/reference/generated/numpy.array2string.html) we are able to take advantage of formatting (floating point numbers with 2 decimals), and hiding additional elements (using edgeitems=2).

The last part of the if statement was added to bypass some tests that weren't passing for test_transpose.
```
self.value is not None and "\n" in name
```
This ensures that we only use numpy.array2string on expressions whose name are arrays. (For example, x.T.T wouldn't be affected as it's name doesn't contain a newline).

<b>variable.py</b>: expanded variable representation by including the variable name in addition to the variable shape.

<b>test_expression.py</b>: made changes to two asserts that verified the representation of the expressions.

Here is an example of the changes made in this PR
```
print(cp.Constant(np.ones((5,5))).__str__())
[[1.00 1.00 ... 1.00 1.00]
 [1.00 1.00 ... 1.00 1.00]
 ...
 [1.00 1.00 ... 1.00 1.00]
 [1.00 1.00 ... 1.00 1.00]]
```

Issue link (if applicable): #2089 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/expressions/constants/constant.py
+++ b/cvxpy/expressions/constants/constant.py
@@ -61,6 +61,9 @@ class Constant(Leaf):
     def name(self) -> str:
         """The value as a string.
         """
+        if len(self.shape) == 2 and "\n" in str(self.value):
+            return np.array2string(self.value, edgeitems=2, threshold=5,
+                                    formatter={'float': lambda x: f'{x:.2f}'})
         return str(self.value)
 
     def constants(self) -> List["Constant"]:
--- a/cvxpy/expressions/constants/constant.py
+++ b/cvxpy/expressions/constants/constant.py
@@ -21,9 +21,9 @@ import scipy.sparse as sp
 
 import cvxpy.interface as intf
 import cvxpy.lin_ops.lin_utils as lu
+import cvxpy.settings as s
 import cvxpy.utilities.linalg as eig_util
 from cvxpy.expressions.leaf import Leaf
-from cvxpy.settings import EIGVAL_TOL
 from cvxpy.utilities import performance_utils as perf
 
 
@@ -62,8 +62,10 @@ class Constant(Leaf):
         """The value as a string.
         """
         if len(self.shape) == 2 and "\n" in str(self.value):
-            return np.array2string(self.value, edgeitems=2, threshold=5,
-                                    formatter={'float': lambda x: f'{x:.2f}'})
+            return np.array2string(self.value,
+                                   edgeitems=s.PRINT_EDGEITEMS,
+                                   threshold=s.PRINT_THRESHOLD,
+                                   formatter={'float': lambda x: f'{x:.2f}'})
         return str(self.value)
 
     def constants(self) -> List["Constant"]:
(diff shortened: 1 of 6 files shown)


cvxpy/expressions/expression.py
```diff
@@ -133,7 +133,11 @@ def domain(self):
     def __str__(self) -> str:
         """Returns a string showing the mathematical expression.
         """
-        return self.name()
+        name = self.name()
+        if len(self.shape) == 2 and self.value is not None and "\n" in name:
```
SteveDiamond, line 137: Would it be equivalent in terms of functionality to put this change in the name function of constant.py? That would be easier to understand in the long run.

cvxpy/expressions/variable.py
```diff
@@ -140,4 +140,4 @@ def __repr__(self) -> str:
         if len(attr_str) > 0:
             return "Variable(%s%s)" % (self.shape, attr_str)
         else:
-            return "Variable(%s)" % (self.shape,)
+            return "Variable(%s, %s)" % (self.shape, self.__str__())
```
phschiele, line 143:
```suggestion
            return f"Variable({self.shape}, {self.__str__()})"
```

cvxpy/expressions/expression.py
Transurgeon: I just tried, it most certainly would be the same functionality

cvxpy/expressions/constants/constant.py
```diff
@@ -61,6 +61,9 @@ def __init__(self, value) -> None:
     def name(self) -> str:
         """The value as a string.
         """
+        if len(self.shape) == 2 and "\n" in str(self.value):
+            return np.array2string(self.value, edgeitems=2, threshold=5,
```
Transurgeon, line 65: @phschiele Using edgeitems=2 here makes it so that your example of a 3 by 3 matrix doesn't change. Hopefully that is ok with you? 
However, anything that has more than 4 rows or columns will be truncated, and I think it does look quite appealing visually.

phschiele: That would be fine with me. I've added a few tests and made it so users can overwrite these values if necessary in 
1084f7e.

cvxpy/tests/test_expressions.py
```diff
@@ -61,8 +61,10 @@ def test_variable(self) -> None:
         # self.assertEqual(x.canonical_form[0].shape, (2, 1))
         # self.assertEqual(x.canonical_form[1], [])
 
-        self.assertEqual(repr(self.x), "Variable((2,))")
-        self.assertEqual(repr(self.A), "Variable((2, 2))")
+        self.assertEqual(repr(self.x), "Variable((2,), x)")
+        self.assertEqual(repr(self.A), "Variable((2, 2), A)")
+        self.assertEqual(repr(cp.Variable(name='x', nonneg=True)), "Variable((), x, nonneg=True)")
```
SteveDiamond, line 66: Would we want the output here to be ``Variable((), 'x', nonneg=True)`` or ``Variable((), x, nonneg=True)`` as you have now? @phschiele

SteveDiamond: Nvm this is fine.

SteveDiamond (approved):

cvxpy/tests/test_expressions.py
phschiele: Just saw this now, I think the current form of not having to escape the quotes within the string representation is cleaner 👍

PR #2094, opened 2023-04-03
Fixes Python 2 print statements

phschiele:
## Description
Replaces some Python 2 print statements with functions.
There are a few more places in larger examples, but these would need a larger overhaul.

## Type of change
- [ ] Bug fix

### DIFF
--- a/doc/source/faq/index.rst
+++ b/doc/source/faq/index.rst
@@ -27,7 +27,7 @@ run the following code snippet in the Python prompt:
 .. code:: python
 
   import cvxpy
-  print cvxpy.__version__
+  print(cvxpy.__version__)
 
 What do I do if I get a ``DCPError`` exception?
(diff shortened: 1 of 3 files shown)


PTNobel:
Consider fixing the same issue in `doc/source/examples/applications/nba_ranking.rst` and `doc/source/faq/index.rst`

phschiele:
> Consider fixing the same issue in `doc/source/examples/applications/nba_ranking.rst` and `doc/source/faq/index.rst`

Fixed the FAQ page, but the NBA example has more Python 2 syntax so just fixing the print would not be sufficient. I suggest we coordinate an overhaul of some of the prominent examples via a new (good first)issue(s).

PTNobel:
We could also adapt the ILP NFL one, that I sent to Riley and he forwarded to the rest of us

PR #2097, opened 2023-04-05
Multiprecision and Python 3.11 support for SDPA

usamamuneeb:
## Description

SDPA for Python now supports [SDPA Multiprecision](https://github.com/sdpa-python/sdpa-multiprecision), a fork of SDPA-GMP. This change will test the multiprecision backend and Python 3.11 wheels in the CI pipeline.

SDPA with OpenBLAS backend does not pass LP5 test on the Windows runners, so I changed `install_dependencies.sh` to use the multiprecision backend on Windows and the regular (OpenBLAS) backend on Linux and macOS.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


rileyjmurray:
This is super cool!!

rileyjmurray:
@usamamuneeb tests are failing because of unrelated issues with COPT.

Can you please update our web documentation to mention these changes? You'll want to update the solver install guide and the material about choosing solver settings.

It would be nice to add a dedicated test that showcases the accuracy guarantees of SDPA-multiprecision. You can have a test under ``TestSDPA`` that only runs if multiprecision support is available. I don't know exactly what test problem you'd use here. It would be okay to define a new problem specifically for this purpose.

One possibility is to have a basic LP/SDP where some variable has optimal value sqrt(2). Using double-precision arithmetic, an IPM should only be able to compute a solution to accuracy 1e-8 or so. With multiprecision support, you can ask for a solution accurate to 1e-16. I don't think you can ask for a solution with accuracy better than 1e-16 since CVXPY will probably convert to double-precision somewhere in its rewriting process. (It would be good to check if that happens and to document it if so.)

usamamuneeb:
Hi @rileyjmurray, I will go over the [SDPA-GMP paper](https://doi.org/10.1109/CACSD.2010.5612693) to select some problems which show the benefit of using the multiprecision variant. If I recall correctly, besides a more accurate solution, multiprecision version also has a secondary benefit in sense that it can solve some ill posed problems in [SDPLIB](https://github.com/vsdp/SDPLIB) that the regular SDPA package cannot.

As test cases, I will try to select some problems to demonstrate each of these benefits.

rileyjmurray:
Hi @usamamuneeb. Thanks for adding the new test problem!

I think we should wrap ``test_lp_7`` with a check that we're using SDPA-GMP, not just SDPA.

I suggest you add an enumeration like SDPA_Variants  with values ``Standard`` and ``GMP``. Then have ``SDPA.variant`` be a class property that returns ``Standard`` by default or ``GMP`` if appropriate to do so. If you make other versions of SDPA-Python in the future, then you can extend the ``SDPA_Variants`` enumeration and update the implementation of ``SDPA.variant`` accordingly.

usamamuneeb:
Hi @rileyjmurray, just made the change. LP7 will now run only if GMP support is available.

In the next release of `sdpa-python` and `sdpa-multiprecision`, I can create an alias for `sdpap.sdpacall.sdpacall.get_backend_info()` to simplify it to `sdpap.get_backend_info()`. That might act as a suitable replacement for the enumeration.

rileyjmurray (approved):
Approved! @phschiele @SteveDiamond can one of you review? It's a little ambiguous as to whether this is changing the public API, but regardless it would be good to get more eyes on it.

SteveDiamond (approved):
LGTM

phschiele (approved):
@usamamuneeb Great feature, thanks a lot. LGTM

Issue #2098, opened 2023-04-05
Construct a jupyter book (during ci/cd)


tschm:

tschm:
Checkout https://www.cvxgrp.org/cvx_short_course

Transurgeon:
closing this issue as we might not be needing this anymore.

PR #2099, opened 2023-04-06
JupyterBook for cvxpy

tschm:
## Description
The idea is to create a jupyterbook. Such books can contain several notebooks and markdown files.
see https://jupyterbook.org/en/stable/intro.html

Each notebook contains a link to a binder server.  It is possible to add further steps to prefill the cache for binder.
I have also updated some of the ci/cd for the benchmark step.

Issue link (if applicable):
#2098 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


PTNobel:
Hey Thomas,

This looks like a real cool sketch of a new workflow for the examples! Do you have an example of the output of this workflow I could look over? Would you be able to port over one or two of the examples to give us a sense of what it would look like to maintain this way?

tschm:
Hi @PTNobel, I am currently in 212 of the Packard Building. I can show you examples live. The basic idea is simple. You provide Notebooks and Markdown files as you do already and using the _toc.yml file you merge them into the book. I have a trivial example here: https://github.com/tschm/babynames

There are more impressive examples here: https://python.quantecon.org/intro.html

One can go quite wild and you can for example include your sphinx documentation in the book, etc.

tschm:
Ok, I have added advanced.ipynb which is almost a copy of the (outdated) version in examples/notebook. There are two important points here:
1) Each notebook has to start with a markdown cell with a # header
2) The notebook doesn't execute properly as you are using cvxopt which has been removed(?). However, if you check the log of the build book step you get a warning and can take it from there. All notebooks are executed when you build the book. This can be controlled, notebooks can be cached, execution can be switched off etc. It's imho much more reliable to execute notebooks and verify they actually run.

You will find that binder is somewhat slow, as a docker image has to be compiled. However, please check the caching workflow in my babynames repo. I would move this into my next pull request.

PTNobel:
Awesome! I'm heading into the office now, I'll come by 212

SteveDiamond:
Is this PR dropped? It's an interesting idea.

tschm:
I think we give it another go once the cvx_short_course has finished. We are learning there...

Issue #2095, opened 2023-04-04
Some examples still use python 2


PTNobel:
Some of the examples still use python 2; it would be nice to replace them with up to date code!

As an example, https://www.cvxpy.org/examples/applications/nba_ranking.html uses `print` as a statement.

It would be nice to have someone go through examples, test if they run and fix them if they don't. The source code for the examples is mostly in https://github.com/cvxpy/cvxpy/tree/master/doc/source/examples

CC: @phschiele

phschiele:
To make the most out of this effort, we should give some thought to how we want to manage the examples going forward.
At the very least, we should mention something like "last updated using CVXPY 1.3.1".
Ideally, we would further include some automated testing.

PTNobel:
Now that we have backwards compatibility guarantees, we could just make the examples integration tests we have to pass or justify failing before we can make a release.

rileyjmurray:
That sounds reasonable, provided the examples don’t run too long.

tschm:
I would recommend to use only notebooks as examples and integrate them in a jupyterbook. They can then be executed during the compilation of the notebook. Problems can be observed...

Transurgeon:
this issue could be closed I think since there was a PR that addressed it above.

PR #2096, opened 2023-04-04
Fix-some-python-2-prints-in-nba-and-others

Paulnkk:
## Description
Please include a short summary of the change.
I fixed some python 2 prints in the nba example and the docscrape file.
Issue link (if applicable):
Found issue here: https://github.com/cvxpy/cvxpy/issues/2095

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2101, opened 2023-04-12
Replacing nba_ranking.rst from python2 to python3

kohliaryan:
## Description

This commit updates the nba_ranking.py example file to fix compatibility issues with Python 2. The changes include updating the print statements to use the print() function and fixing the syntax of the division operator. This will allow users running Python 2 to run the example without encountering errors.

Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/2095

## Type of change
-  Bug fix

-  Check that your code adheres to our coding style.
-  Run the unittests and check that they’re passing.

### DIFF
--- a/doc/source/examples/applications/nba_ranking.rst
+++ b/doc/source/examples/applications/nba_ranking.rst
@@ -212,12 +212,12 @@ notebook.
             pid2idx = []
     
         for i in range(N):
-            print '\nPartition {}'.format(i),
+            print('\nPartition {}'.format(i), end='')
             games_t, games_h = split_games(games, holdout)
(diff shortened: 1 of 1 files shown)


PTNobel:
The diff looks great; did you get a chance to run it and see if it works?

phschiele:
I was going through the example fixing some additional issues, but realized that the data is currently not available for this example:
https://www.cvxpy.org/examples/applications/nba_ranking.html#todo

Do we have the data?

phschiele:
As discussed in #2096, we will remove this example, as we cannot run it due to the data being missing.

PR #2103, opened 2023-04-14
feat: retrieve mip gap from SCIPY solution

allenlawrence94:
## Description
When using scipy to solve a MIP, this change makes the solution's relative mip gap available in `Problem.solver_stats.extra_stats["mip_gap"]`

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
@@ -245,7 +245,9 @@ class SCIPY(ConicSolver):
                     inverse_data[self.NEQ_CONSTR])
                 eq_dual.update(leq_dual)
                 dual_vars = eq_dual
-
-            return Solution(status, opt_val, primal_vars, dual_vars, {})
+            attr = {}
+            if inverse_data['is_mip']:
+                attr[s.EXTRA_STATS] = {"mip_gap": solution['mip_gap']}
+            return Solution(status, opt_val, primal_vars, dual_vars, attr)
         else:
             return failure_solution(status)
--- a/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
@@ -246,7 +246,7 @@ class SCIPY(ConicSolver):
                 eq_dual.update(leq_dual)
                 dual_vars = eq_dual
             attr = {}
-            if inverse_data['is_mip']:
+            if "mip_gap" in solution:
                 attr[s.EXTRA_STATS] = {"mip_gap": solution['mip_gap']}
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
LGTM. CI failures are unrelated.

phschiele (approved):

phschiele:
@allenlawrence94 There appears to be a small issue with this PR.
It looks like the way we call SciPy for integer problems (`linprog(..., integrality=integrality, ...`) does not return the MIP gap: [linprog docs](https://docs.scipy.org/doc/scipy-1.9.1/reference/optimize.linprog-highs.html?highlight=linprog)
We would need to call the solver via the milp function, right? [milp docs](https://docs.scipy.org/doc/scipy-1.10.1/reference/generated/scipy.optimize.milp.html)

allenlawrence94:
@phschiele - I see that `linprog`'s docs don't mention returning a mip gap, but when using linprog I see a "mip_gap" in the returned solution:
```
solution
        message: Optimization terminated successfully. (HiGHS Status 7: Optimal)
        success: True
         status: 0
            fun: -8373.0
              x: [ 0.000e+00  0.000e+00 ...  1.000e+00  0.000e+00]
            nit: -1
          lower:  residual: [ 0.000e+00  0.000e+00 ...  1.000e+00
                              0.000e+00]
                 marginals: [ 0.000e+00  0.000e+00 ...  0.000e+00
                              0.000e+00]
          upper:  residual: [ 1.000e+00  1.000e+00 ...  0.000e+00
                              1.000e+00]
                 marginals: [ 0.000e+00  0.000e+00 ...  0.000e+00
                              0.000e+00]
          eqlin:  residual: []
                 marginals: []
        ineqlin:  residual: [ 2.400e+01]
                 marginals: [ 0.000e+00]
 mip_node_count: 1
 mip_dual_bound: -8373.0
        mip_gap: 0.0
```

Are you seeing situations where it isn't returned?

allenlawrence94:
I got the above by adding a `print(solution)` to the end of `SCIPY.solve_via_data` and running
`pytest -s cvxpy/tests/test_conic_solvers.py::TestSCIPY::test_scipy_mi_lp_2`

phschiele:
@allenlawrence94 I looked into it a bit more. When using SciPy 1.10, I get the mip gap for all test cases.
When running SciPy 1.9 (the earliest version that supported MIP), we only get the mip gap for the problems with optimal status, but not for the time-limited test case. When using `milp` instead of `linprog` we get the mip gap for all cases even under SciPy 1.9.  So the only test case that checked for the mip gap was exactly the one that did not have it in the 1.9/`linprog` setup.
I have a first setup working, I will create a PR, but it needs a bit more refinement.

PR #2104, opened 2023-04-14
add __all__ to __init__.py

akshayka:
Add public symbols to `__all__`.

## Description
Some language servers, like pyright (used by vscode), use `__all__` to determine which symbols to show in autocomplete. So with this change, vscode will stop showing internal APIs in its completions.

See
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

## Type of change
- [ ] Bug fix


cvxpy/__init__.py
```diff
@@ -13,6 +13,78 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+
+__all__ = [
+    "__version__",
```
SteveDiamond, line 18: All the atoms are supposed to be at the top level too.

SteveDiamond: @rileyjmurray @phschiele are reductions part of the API too? I don't remember why they're imported at the top level.

rileyjmurray:
I think reductions shouldn’t be (/aren’t) part of the public API. There can
be exceptions to this rule on a case by case basis. Right now I don’t think
we have any reductions that a user should need to access.

cvxpy/__init__.py
akshayka: > All the atoms are in the API too.

Oops, good catch.

akshayka:
@SteveDiamond, regarding this change: I misinterpreted the pyright docs. Pyright doesn't limit completions to `__all__`, instead you can use `__all__` to export additional members that pyright wouldn't export by default. That's not what we want.

I think we should abandon this PR.

> I think reductions shouldn’t be (/aren’t) part of the public API. There can
be exceptions to this rule on a case by case basis. Right now I don’t think
we have any reductions that a user should need to access.

I agree. I'd like to remove the reductions import. But removing them might be a breaking change?

rileyjmurray:
@akshayka it's been our stated policy that reductions are not part of the public API:

https://www.cvxpy.org/api_reference/cvxpy.reductions.html#disclaimer

This has conflicted with our more recently-adopted policy that anything in the top-level cvxpy namespace is public.

While removing reductions from the top-level namespace might be considered a breaking change, I think it is okay in this case. It can be understood as a "bug fix" that brings conflicting policies in line with one another, where we're deferring to the more established policy.

akshayka:
> While removing reductions from the top-level namespace might be considered a breaking change, I think it is okay in this case. It can be understood as a "bug fix" that brings conflicting policies in line with one another, where we're deferring to the more established policy.

Sounds good. I'll put up a PR.

PR #2105, opened 2023-04-15
fix: remove reductions import in top-level __init__.py

akshayka:
## Description

Reductions are stated to not be in the public API
(https://www.cvxpy.org/api_reference/cvxpy.reductions.html#disclaimer, https://github.com/cvxpy/cvxpy/pull/2104#issuecomment-1509886345), but they are imported in the top-level `__init__.py`.

This change removes that import.

## Type of change
- [x] Bug fix

### DIFF
--- a/cvxpy/__init__.py
+++ b/cvxpy/__init__.py
@@ -28,7 +28,6 @@ from cvxpy.expressions.variable import Variable
 from cvxpy.problems.objective import Maximize, Minimize
 from cvxpy.problems.problem import Problem
 from cvxpy.transforms import linearize, partial_optimize, suppfunc
-from cvxpy.reductions import *
 from cvxpy.reductions.solvers.defines import installed_solvers
 from cvxpy.settings import (CBC, CLARABEL, COPT, CPLEX, CPP_CANON_BACKEND,
                             CVXOPT, DIFFCP, ECOS, ECOS_BB, GLOP, GLPK, GLPK_MI,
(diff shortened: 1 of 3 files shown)


SteveDiamond (approved):
LGTM

SteveDiamond:
Is it possible we put the reductions import there because it helped build the sphinx API docs? Not a reason to keep it of course.

rileyjmurray:
Honestly I have no idea how long reductions have been in there or for what
reasons. Your guess is at least as good as mine.

SteveDiamond:
I checked with git blame and it looks like it was related to the API docs:
https://github.com/cvxpy/cvxpy/commit/19c5177c3e2224ccd861bb222b52dfeef518229b
Anyway we can make the docs work easily enough.

rileyjmurray:
When you said it was added in the API docs I started to sweat since I was
the last person to revise those. I was relieved when I saw it was way back
in the 0.4.x release series, lol.

akshayka:
> I checked with git blame and it looks like it was related to the API docs: [19c5177](https://github.com/cvxpy/cvxpy/commit/19c5177c3e2224ccd861bb222b52dfeef518229b) Anyway we can make the docs work easily enough.

We might as well remove them from the API docs, since they're not public.

rileyjmurray:
@akshayka I think they should be documented on the website. Sometimes advanced users benefit from knowing what goes on under the hood. We can put them under a "private API" heading if that feels more appropriate.

akshayka:
> @akshayka I think they should be documented on the website. Sometimes advanced users benefit from knowing what goes on under the hood. We can put them under a "private API" heading if that feels more appropriate.

I think users who are that advanced should just read the source code. The risk in documenting private symbols on our website is that regular users will find these symbols and become dependent on them; people will just treat the private API as public. Similar to `black.format_str`: https://github.com/search?q=black.format_str&type=code, https://github.com/psf/black/issues/779

That said, I don't feel strongly. I just think removing these symbols will make maintainers' lives easier.

Issue #1222, opened 2021-01-24
Exotic2Common: avoid stack atoms in PowConeND to PowCone3D reduction 


rileyjmurray:
Following PR #1211, cvxpy supports 3D and ND power cone constraints. The main way for doing this is the ([Exotic2Common](https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/cone2cone/exotic2common.py)) that maps an ND power cone to a collection of (N-2) 3D power cones. As-implemented, this reduction has two downsides:
1. The canonicalization involves use of hstack atoms and other high-level affine transformations. This use of high-level atoms will likely be more expensive to canonicalize in cvxcore compared to if we built suitable sparse matrices directly.
2. ~~The ``PowConeND`` class currently does not have dual variable support.~~ (Resolved in 2023 by PR #2241.)

~~We'd welcome contributions handling either of these issues.~~

We'd welcome contributions for handling the remaining issue.

rileyjmurray:
For (2), see [this writeup](https://groups.google.com/g/mosek/c/fmQFe0C4Ups) on how dual variables for the 3D power cone lift of the ND power cone can be propogated back as dual variables for the original ND power cone.

pat749:
hi @rileyjmurray currently i am working on it and i am also  reading arxiv research paper which you have given ,but i have a simple doubt should hamza fawzi paper also have to implemented using cvxpy   and also any suggestion for this task? thanks

pat749:
and also any discourse group where i can join

rileyjmurray:
Hi @pat749, nice to hear you're taking this on! Hamza Fawzi's papers are not needed for this task. We can have discussions here, on Discord, or on GitHub Discussions.

pat749:
ok thanks

rileyjmurray:
@pat749 have you worked on this? If so, can you please push your code to a public repository so I can see where you're at?

goulart-paul:
We have added native support for ND power cones in Clarabel v0.6.0, including updates to the python wrapper.

Very basic example [here](https://github.com/oxfordcontrol/ClarabelDocs/blob/main/examples/py/example_genpowcone.py).     Better documentation than that still to come.

Transurgeon:
closing this as it was taken care of in #2971 , thanks for the awesome work @Ni2002ka !

PR #2108, opened 2023-04-25
Support SDPs for Clarabel ver >= 0.5.0

goulart-paul:
## Description

Adds support for SDPs for Clarabel versions >= v0.5.0.   Wrapper is backwards compatible for older versions.   

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/tests/test_conic_solvers.py
```diff
@@ -483,6 +483,12 @@ def test_clarabel_pcp_1(self) -> None:
     def test_clarabel_pcp_2(self) -> None:
         StandardTestSOCPs.test_socp_2(solver='CLARABEL')
 
+    def test_clarabel_sdp_1min(self) -> None:
+        StandardTestSDPs.test_sdp_1min(solver='CLARABEL')
+
+    def test_clarabel_sdp_2(self) -> None:
+        StandardTestSDPs.test_sdp_2(solver='CLARABEL')
```
SteveDiamond, line 490: For the moment you can set the arg ``places=2``. Paging our expert @rileyjmurray in case this suggests any changes to the test case.

rileyjmurray:
I don't know if this problem has non-unique optimal solutions. My
recommended course of action is to build a function like test_sdp_2()
that skips the check for specific values of primal variables. The current
code is

https://github.com/cvxpy/cvxpy/blob/2200e903a25912e35b8f8cac3bb28ecc600e7f5e/cvxpy/tests/solver_test_helpers.py#L1131-L1142
.

You'd want to adapt that code by removing "sth.verify_primal_values(places)".
The resulting function will still check primal feasibility, complementary
slackness, dual domains (i.e. dual variables being PSD), and the reported
objective.

For an example of a test on test_conic_solvers.py that similarly needs to
customize the test code, see

https://github.com/cvxpy/cvxpy/blob/2200e903a25912e35b8f8cac3bb28ecc600e7f5e/cvxpy/tests/test_conic_solvers.py#L911-L920
.

cvxpy/tests/test_conic_solvers.py
rileyjmurray: See my response here: https://github.com/cvxpy/cvxpy/pull/2108#issuecomment-1522268306.

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
```diff
@@ -53,15 +56,88 @@ def dims_to_solver_cones(cone_dims):
...
+        mats.append(mat)
+    V = sum(mats)
+    return V
+
+
 class CLARABEL(ConicSolver):
     """An interface for the Clarabel solver.
     """
 
+    import clarabel
```
SteveDiamond, line 133: We can't import clarabel here because then cvxpy will break if clarabel is not installed. Maybe the version check can be removed and instead bump the minimum version for clarabel to 0.5.0? I'm open to alternatives as well if you prefer.

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
```diff
@@ -53,15 +56,88 @@ def dims_to_solver_cones(cone_dims):
...
+
     # Solver capabilities.
     MIP_CAPABLE = False
     SUPPORTED_CONSTRAINTS = ConicSolver.SUPPORTED_CONSTRAINTS \
-        + [SOC, ExpCone, PowCone3D]
-    REQUIRES_CONSTR = True
```
SteveDiamond, line 64: REQUIRES_CONSTR = False or not present means that the solver will work if called with a linear objective only. We of course should handle this in a preprocessor but for the moment this is how it works. I just want to verify that Clarabel will work without constraints.

phschiele:
Here is the link to the doc table that shows the available cones for each solver, with Clarabel now being on the (short) list of solvers that support all the listed cones: https://github.com/cvxpy/cvxpy/blob/a6e88643570c3b4852d62d5932ba3db7c2b10bd9/doc/source/tutorial/advanced/index.rst?plain=1#L435

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
goulart-paul: The solver works on unconstrained problems, even if unbounded in cost / linear objective.   it seems that with `REQUIRES_CONSTR = False` the solver is still being passed a `0 x N` sparse matrix for the constraints, instead of NULL or something, which is what we want.   I will remove this b/c it looks to default to False.

goulart-paul:
> Here is the link to the doc table that shows the available cones for each solver, with Clarabel now being on the (short) list of solvers that support all the listed cones:
> 
> https://github.com/cvxpy/cvxpy/blob/a6e88643570c3b4852d62d5932ba3db7c2b10bd9/doc/source/tutorial/advanced/index.rst?plain=1#L435

Thanks - I forgot to update that.   We also have native ND power cone support prototyped and expect to have that for 0.6.0 (see https://github.com/cvxpy/cvxpy/issues/1222).   Table will need more columns!

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
```diff
@@ -53,15 +56,92 @@ def dims_to_solver_cones(cone_dims):
...
 class CLARABEL(ConicSolver):
     """An interface for the Clarabel solver.
     """
 
     # Solver capabilities.
     MIP_CAPABLE = False
     SUPPORTED_CONSTRAINTS = ConicSolver.SUPPORTED_CONSTRAINTS \
-        + [SOC, ExpCone, PowCone3D]
-    REQUIRES_CONSTR = True
+        + [SOC, ExpCone, PowCone3D, PSD]
```
SteveDiamond, line 136: Looks like PSD is accidentally included for clarabel < 0.5.0 here.

SteveDiamond (approved):

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
```diff
@@ -61,7 +134,14 @@ class CLARABEL(ConicSolver):
     MIP_CAPABLE = False
     SUPPORTED_CONSTRAINTS = ConicSolver.SUPPORTED_CONSTRAINTS \
         + [SOC, ExpCone, PowCone3D]
-    REQUIRES_CONSTR = True
+    
+    try:
```
PTNobel, line 138: Should we consider adding a TODO comment for whenever we bump up the minimum Clarabel version to be > 0.5.0 to make this unconditional?

PTNobel (commented):
Have we thought about what our policy is for changing default solvers? I imagine that people are generally supportive of eventually making Clarabel default over ECOS for SOCPs and/or Clarabel default over SCS for SDPs. Presumably, at about that time we would require that Clarabel be installed by default (and maybe remove ECOS if it makes people's lawyers happier). 

Is there a certain amount of time we want support to bake? Is changing default solvers a breaking change we have to wait for CVXPY 2.0 to do?

I recognize this might not be the perfect venue for this, so happy to move this question to either Discord or a dedicated issue.

rileyjmurray:
I think that changing the solver order isn’t a breaking change. Removing a
solver is another story. As for ECOS and lawyers, there’s still the matter
of how CVXOPT is licensed.

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
SteveDiamond: Could you approve the MR? I added the TODO as you suggested.

phschiele (approved):

Issue #2109, opened 2023-04-25
Make `Objective` available in top-level namespace


jonathanberthias:
Most (all?) constraints were added to the top-level namespace in #1909, as well as the `Constraint` class itself, making type hints for code that uses cvxpy much easier to write.

Could the same be done for the `cvxpy.problems.objective.Objective` base class? I use it quite a bit when an argument can be either `Minimize` or `Maximize`.

Thanks!

SteveDiamond:
Done in master.

PR #2110, opened 2023-04-25
Add objective as top level import

SteveDiamond:
## Description
This adds Objective as a top level import, to facilitate type hinting.
Issue link (if applicable): #2109 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
Unrelated failures on Python 3.11. @phschiele this is due to a change in the HiGHS interface right?

phschiele (commented):
Agree re adding `Objective` to the public API.
We already import `Maximize` and `Minimize` in l28, we might want to add Objective there (unless I'm missing some import issues that I'm missing).

I verified that the 3.11 failures come from the MIP gap PR.

phschiele (approved):
Agree re adding the objective

Issue #2112, opened 2023-04-25
Retrieve MIP gap from SciPy


phschiele:
#2103 added a feature to retrieve the MIP gap from SciPy which was incompatible with the current solver interface. 

Specifically, the way we call SciPy for integer problems (`linprog(..., integrality=integrality, ...`) does not return the MIP gap: [linprog docs](https://docs.scipy.org/doc/scipy-1.9.1/reference/optimize.linprog-highs.html?highlight=linprog)
We would need to call the solver via the milp function [milp docs](https://docs.scipy.org/doc/scipy-1.10.1/reference/generated/scipy.optimize.milp.html)

This function is also available for SciPy 1.9.0, and since we branch on whether or not we solve a MIP, the solution could be as simple as changing linprog to milp.

Cc: @allenlawrence94

phschiele:
Closed via #2113

PR #2113, opened 2023-04-26
SciPy MIP gap

phschiele:
## Description
Slight adaption of #2103 using `milp` instead of `linprog` which also provides the mip gap for problems with time limits.
Issue link (if applicable): #2112 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
@@ -245,7 +245,9 @@ class SCIPY(ConicSolver):
                     inverse_data[self.NEQ_CONSTR])
                 eq_dual.update(leq_dual)
                 dual_vars = eq_dual
-
-            return Solution(status, opt_val, primal_vars, dual_vars, {})
+            attr = {}
+            if "mip_gap" in solution:
+                attr[s.EXTRA_STATS] = {"mip_gap": solution['mip_gap']}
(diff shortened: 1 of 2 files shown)


cvxpy/reductions/solvers/conic_solvers/scipy_conif.py
```diff
@@ -187,10 +188,23 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
 
         if problem_is_a_mip:
-            solution = opt.linprog(data[s.C], A_ub=data[s.G], b_ub=data[s.H],
-                                   A_eq=data[s.A], b_eq=data[s.B], method=meth,
-                                   options=solver_opts['scipy_options'],
-                                   integrality=integrality, bounds=bounds)
+            constraints = []
+            G = data[s.G]
+            if G is not None:
+                ineq = scipy.optimize.LinearConstraint(G, ub=data[s.H])
```
phschiele, line 194: Using `LinearConstraint` and `Bounds` makes the interface a bit nicer, but I could not unify it with the non-milp case because they are not available for lower versions of SciPy.

SteveDiamond (approved):
LGTM

allenlawrence94:
:tada: Thanks @phschiele!

phschiele:
@SteveDiamond The code analysis fails because the MIP test only runs on Python 3.11, but we extract coverage information from the 3.8 run.

Merging here.

PR #2116, opened 2023-04-30
Add SOC2PSD Reduction

usamamuneeb:
## Description

Added a reduction `SOC2PSD` that converts the SOC constraints to PSD constraints, for solvers that support the PSD cone but not the second order cone.

This enables the SDPA solver to be used for solving QPs and SOCPs.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -296,8 +297,10 @@ def construct_solving_chain(problem, candidates,
             supported_constraints = solver_instance.MI_SUPPORTED_CONSTRAINTS
         else:
             supported_constraints = solver_instance.SUPPORTED_CONSTRAINTS
-        if (all(c in supported_constraints for c in cones)
-                and (has_constr or not solver_instance.REQUIRES_CONSTR)):
+            unsupported_constraints = [
```
SteveDiamond, line 300: Move this out of the else clause so it's calculated in both cases.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -313,10 +316,20 @@ def construct_solving_chain(problem, candidates,
...
-                solver_instance
             ]
-            return SolvingChain(reductions=reductions)
+            if all(c in supported_constraints for c in cones):
+                reductions += [
+                    ConeMatrixStuffing(quad_obj=quad_obj, canon_backend=canon_backend),
+                    solver_instance
+                ]
+                return SolvingChain(reductions=reductions)
+            elif all(c==SOC for c in unsupported_constraints):
```
SteveDiamond, line 326: This feels a bit fragile and SDPA specific. We probably want something more general. For example doesn't this code assume the solver supports PSD constraints?

SteveDiamond:
This is quite a nice contribution, thanks! We definitely should have an SOC2PSD reduction. The solving_chain code is very critical and already pretty messy so may require more organization.

cvxpy/reductions/solvers/solving_chain.py
usamamuneeb: Indeed. Just fixed.

rileyjmurray (changes requested):
@usamamuneeb in your ``soc2psd.py`` file, please replace the ``np.identity(k)`` calls to ``scipy.sparse.eye(k)``. This will result in far lower memory consumption. (You might need to change the format of the SciPy sparse matrix to ``csc_matrix``, not sure.)

usamamuneeb:
`scipy.sparse.eye` uses the `DIAgonal format` by default and `cvxpy.bmat` accepts it.

I believe conversion from diagonal format to either `CSR` or `CSC` should have the same complexity, so we should perhaps let CVXPY decide whichever is more appropriate when it needs to performs matrix operations internally.

rileyjmurray:
Thanks for making the change!

> `scipy.sparse.eye` uses the `DIAgonal format` by default and `cvxpy.bmat` accepts it.
> 
> I believe conversion from diagonal format to either `CSR` or `CSC` should have the same complexity, so we should perhaps let CVXPY decide whichever is more appropriate when it needs to performs matrix operations internally.

I agree that we should let cvxpy decide on any conversions if needed. I just didn't know if we accepted diagonal format SciPy sparse matrices.

cvxpy/reductions/cone2cone/soc2psd.py
```diff
@@ -0,0 +1,172 @@
...
+                """
+                Constrain M to the PSD cone. The `PSD` class constructor has the input parameter
+                `expr` and constrains `expr + expr.T` to the PSD cone.
+
+                Since M is symmetric, (M/2 + (M/2).T) is simply M which we want to constrain to
+                the PSD cone.
+
+                To this end, we set the input `expr` to M/2.
+                """
+                new_psd_constraint = PSD(M / 2)
```
rileyjmurray, line 105: Since the PSD cone is a cone, scaling ``M`` doesn't change whether or not the constraint holds. Is there a reason why you use this scaled version instead of ``new_psd_constraint = PSD(M)``?

The only thing I can think of is that you might need it to get dual variables right. If that's the case, then I think you can replace this long comment with a single-line comment saying ``# scale M by 1/2 so the PSD cone dual variables get correctly mapped to SOC dual variables``.

If this scaling isn't needed for dual variable purposes then I suggest it be removed, along with this comment.

rileyjmurray:
Note: the test failures are unrelated. (They're due to a test where we expect an ARPack convergence failure in ``is_psd`` but ARPack actually converges.)

cvxpy/reductions/cone2cone/soc2psd.py
usamamuneeb: Yup, it's the dual variables :)

If I don't divide the matrix `M` by 2, I have to multiply the dual variables by 2. The reason I chose to divide `M` by 2 was that this way it's easier to explain to the reader why this scaling is needed in the first place.

If we don't need to explain to the reader, then we can simply multiply the dual variables by 2 instead. This is because the number of elements requiring division (i.e. the entire PSD constraint matrix) is a square of the number of the elements that would otherwise require multiplication (i.e. just a row of the PSD dual variable matrix).

usamamuneeb: In addition, multiplication of the dual variables by 2 can also be further vectorized if performed after we `np.hstack` all the dual variables of a packed SOC constraint.

usamamuneeb: I've simplified the comment and moved the scaling to the dual variables for the reasons mentioned above.

cvxpy/reductions/cone2cone/soc2psd.py
```diff
@@ -0,0 +1,174 @@
...
+                Constrain M to the PSD cone. Since the `PSD` class has the input parameter
+                `expr` and constrains `expr + expr.T` to the PSD cone, this will cause a scaling
+                in the dual variables corresponding to these constraints.
+
+                To this end, we can either constrain `M/2` to the PSD cone (since M is symmetric),
+                or we can multiply the corresponding dual variables by 2.
+
+                For efficiency, instead of dividing M by 2, we will multiply the coresponding
+                dual variables by 2 instead.
+                """
```
rileyjmurray, line 106: I think this comment isn't necessary. There's no way this scaling could have a measurable impact on compile time or solve time. You can just have a small comment in the invert function saying that the PSD representation requires scaling the dual variables by 2 to get the SOC dual variable. These sorts of things are not uncommon and don't need a lengthy explanation.

usamamuneeb: Just fixed.

rileyjmurray (approved):
Looks good to me!

rileyjmurray:
@SteveDiamond can you take another look and see if it's good from your perspective?

SteveDiamond (approved):
This LGTM as well.

Issue #1787, opened 2022-05-27
Handling of Mosek's `Unknown` Problem Status


pqrz:
**Describe the bug**
We run large scale optimisation problems using Cvxpy + Mosek.

One of our optimisation problems yielded `problem status` and `solution status` as [Unknown](https://docs.mosek.com/latest/capi/accessing-solution.html#:~:text=unknown%20%E2%80%94%20the%20solver%20was%20unable%20to%20reach%20a%20conclusion%2C%20most%20likely%20due%20to%20numerical%20issues.), and Cvxpy returned solver error for same (instead of returning `unknown` status).

In our case, we found - Mosek was able to correctly find optimal objective and optimal decision variables (since it matched with Cvxpy+Gurobi results - which returned `optimal status`), we are not sure if this will always be the case.

Ideally, if Mosek is able to return a objective and decision variables - we would expect the same with Cvxpy (rather than solver error).

This can be resolved if Cvxpy adds `unknown` status in its [STATUS_MAP Dictionary](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py#L411) and [SOLUTION_PRESENT list](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/settings.py#L59).

SteveDiamond:
I ran into the same issue and monkey patched cvxpy with exactly the fix you proposed. If we wanted to add this change to cvxpy, I think we should introduce an UNKNOWN status in ``cvxpy.settings``. I'm not comfortable labeling it ``optimal_inaccurate``, because sometimes UNKNOWN in mosek does correspond to quite wrong values.

erling-d-andersen:
A solution with a status of unknown can be anything from arbitrary junk to a for-all practical purpose good enough solution.

For instance, if Mosek runs out of space during setup it will say unknown.

rileyjmurray:
@SteveDiamond @erling-d-andersen what do you think about not changing the status code and just always attempting to propagate MOSEK's solution back through the solving chain? We could introduce a new type of ``SolverError`` where the exception isn't raised until all variables are populated.

aszekMosek:
@rileyjmurray I would just comment that it must be some approach that makes the user painfully aware that what they are getting is in the category "this could be arbitrarily bad". Otherwise users will be fetching that solution (or, even worse, "solution") and using it without checking anything.

[...]

PR #2117, opened 2023-05-01
Mosek: Accept Unknown

Paulnkk:
## Description
I added a parameter for the mosek solver option to map solver status 'unknown' to an inaccurate solution if applicable.
Issue link (if applicable):
I found the issue here: https://github.com/cvxpy/cvxpy/issues/1787

## Type of change
- [x] Bug fix

- [ ] Write unittests.


cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -469,6 +469,18 @@ def invert(self, solver_output, inverse_data):
                       mosek.solsta.prim_feas: s.OPTIMAL_INACCURATE,    # for integer problems
                       mosek.solsta.prim_infeas_cer: s.INFEASIBLE,
                       mosek.solsta.dual_infeas_cer: s.UNBOUNDED}
+
+        solver_opts = solver_output['solver_options']
+        # man muss hier über
```
phschiele, line 474: :de: comment :)

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -469,6 +469,18 @@ def invert(self, solver_output, inverse_data):
...
+        if 'accept_unknown' in solver_opts:
+            if solver_opts['accept_unknown']:
```
phschiele, line 476: Can be combined as `if solver_opts.get('accept_unknown', False)`

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -469,6 +469,18 @@ def invert(self, solver_output, inverse_data):
...
+                STATUS_MAP[mosek.solsta.unknown] = s.OPTIMAL_INACCURATE
+                STATUS_MAP[mosek.solsta.optimal] = s.OPTIMAL
+                STATUS_MAP[mosek.solsta.integer_optimal] = s.OPTIMAL
+                STATUS_MAP[mosek.solsta.prim_feas] = s.OPTIMAL_INACCURATE
+                STATUS_MAP[mosek.solsta.prim_infeas_cer] = s.INFEASIBLE
+                STATUS_MAP[mosek.solsta.dual_infeas_cer] = s.UNBOUNDED
```
phschiele, line 482: We would only need to update the entries that changed from above.

phschiele:
@Paulnkk Thanks for adding this! 
In addition to a few small comments above, we should also document the setting here:
https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options

It would be even better if we could add a test case. Paging @pqrz, @aszekMosek, @erling-d-andersen, @rileyjmurray , and @SteveDiamond from the discussion in #1787 for ideas about a minimal example that can trigger the unknown status (with solution).

rileyjmurray:
@phschiele: a few months ago (between 1.2 and 1.3 releases) we found an example in ``test_constant_atoms`` that resulted in an ill-posed exponential cone program. I think the exponential cone is the only one that MOSEK supports which is defined as a closure of some inequality-representable set. So I would try making such an example where the only feasible solutions lay on the boundary that needs the closure.

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
Paulnkk: Merged conditions together here: [Adjust accept_unknown condition](https://github.com/cvxpy/cvxpy/pull/2117/commits/abf94f70f559daef24aacb5f1efa1257e319d318)

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
Paulnkk: Removed conditions here: [Adjust accept_unknown condition](https://github.com/cvxpy/cvxpy/pull/2117/commits/abf94f70f559daef24aacb5f1efa1257e319d318)

aszekMosek:
To reliably get solution status unknown take any problem that is nontrivial enough not to get solved by presolve and set time limit or max number of iterations to 0.

Paulnkk:
@phschiele @rileyjmurray @aszekMosek I added a problem from a simple cone program which already exists in cvxpy with setting timelimit to 0 but I need to double check if the problem is actually non-trivial enough to to trigger the unknown status and that the assertion I build actually leads to the desired check we want here.

rileyjmurray:
I think it suffices to check that the status is "optimal - inaccurate".
With this PR it's clear that we send that status back only when MOSEK
returns an "unknown" status code. If MOSEK's presolve gets better in the
future we may need to update the test problem. I think we can cross that
bridge if / when we come to it.

Paulnkk:
@rileyjmurray thanks, I think I implemented it with the desired output but termination with OPTIMAL_INACCURATE is actually trivial since we have no timelimit.

rileyjmurray (changes requested):
@Paulnkk can you please update the web documentation to mention this parameter?

The relevant section that needs to change is here: https://github.com/cvxpy/cvxpy/blob/master/doc/source/tutorial/advanced/index.rst#setting-solver-options.

rileyjmurray (approved):

SteveDiamond (approved):

phschiele (approved):

PR #2118, opened 2023-05-01
`setup.py`: Remove `setup_requires`, remove setuptools from `install_requires'

mkoeppe:
## Description
Issue link (if applicable):

`setup_requires` is outdated; the build dependencies are declared in `pyproject.toml`.
Also `setuptools` should not appear in `install_requires`; these are the runtime dependencies.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -231,7 +231,5 @@ setup(
         "scs >= 1.1.6",
         "numpy >= 1.15",
         "scipy >= 1.1.0",
-        "setuptools > 65.5.1",
     ],
-    setup_requires=["numpy >= 1.15"],
 )


setup.py
```diff
@@ -231,7 +231,5 @@ def not_on_windows(s: str) -> str:
         "scs >= 1.1.6",
         "numpy >= 1.15",
         "scipy >= 1.1.0",
-        "setuptools > 65.5.1",
```
SteveDiamond, line 234: This is here because of security vulnerability https://github.com/cvxpy/cvxpy/issues/2021.

SteveDiamond:
Could you elaborate on how setup_requires is outdated and pyproject.toml is read by ``python setup.py install``? Since what Python version?

setup.py
SteveDiamond: I see, you're saying this should be in pyproject.toml.

mkoeppe:
It's true that `python setup.py install` would not look at `pyproject.toml`.
But `python setup.py install` -- like all direct invocations of `setup.py` by users -- has long been deprecated by setuptools.

setup.py
mkoeppe: Yes (and it's already there).

SteveDiamond (approved):

PTNobel:
Hey @mkoeppe, I had a few questions that I wasn't understanding about the longer term directions for our setup.py and pyproject.toml; I'm being a bit presumptuous in assuming you're familiar with the python packaging ecosystem, so no worries if you don't have a good answer on what the new best practices are!

1. Do you think we should eliminate our `setup.py` eventually? You mentioned it is deprecated (but perhaps I'm misreading you and only `./setup.py install` is deprecated). Is removing it compatible with our need to compile a C++ extension on install?
2. Should we move our dependencies list to `pyproject.toml` at some point? Is there a benefit in machine readability or some other means?

PTNobel (approved):

mkoeppe:
Hi @PTNobel , what is deprecated is calling `setup.py` directly, with whatever commands (`setup.py install`, `setup.py build`) etc. Such commands should no longer be advertised to users.

In modern Python packaging, `setuptools` is invoked in a different way, namely as a build backend, following PEP 517. (`pip` and [`build`](https://pypi.org/project/build/) are frontends.)
Basic distribution packages that just consist of some static files can indeed be built with setuptools without a `setup.py`, declaring everything in `setup.cfg` (or, at some point in the future, in `pyproject.toml`). But this does not apply to distributions that build extension modules etc.

The build backend is declared in `pyproject.toml`; and there are many options available as alternatives to setuptools. An interesting option is [`meson-python`](https://meson-python.readthedocs.io/en/latest/); the SciPy folks are offering it already as an alternative to their traditional distutils-based build system.

Declaring metadata such as runtime dependencies (install-requires) in `pyproject.toml` will have the advantage that this will be uniform across all build systems; so it's a first easy step to enable experimentation with other build systems. (That's our plan for SageMath - https://github.com/sagemath/sage/issues/33577)

PTNobel:
Thanks for the explanation! (Also I'm a huge SageMath fan! Thanks for all your work there! I've been using it since I was 15!) 

I'll have to experiment with the alternative build systems and see if we can find something better for our setup.

Issue #1206, opened 2021-01-08
MOSEK: one keyword argument for all continuous accuracy parameters


rileyjmurray:
This is a follow-up to the comments on #890. For one reason or another, certain parameters passed in ``mosek_params`` produce unexpected behavior. Maybe the behavior in question is not a bug, and I'm just misunderstanding these parameters. Maybe there is a bug in CVXPY or MOSEK's python interface. Paging @aszekMosek, so you can investigate when you have the time. Also paging  @iarbel84 since you have an interest in specifying termination criteria.

Based on the following small example I know that we are correctly passing some parameters to MOSEK (specifically, ``"MSK_IPAR_INTPNT_MAX_ITERATIONS"`` is being respected). However, the precision parameters aren't having the expected effect on MOSEK's solution status.
```
n = 10
m = 4
np.random.seed(0)
A = np.random.randn(m, n)
x = np.random.randn(n)
y = A.dot(x)

# Solve a simple basis pursuit problem for testing purposes.
z = cp.Variable(n)
objective = cp.Minimize(cp.norm1(z))
constraints = [A @ z == y]
problem = cp.Problem(objective, constraints)

mosek_params = {
    #mosek.dparam.basis_tol_x: 1e-8,
    "MSK_DPAR_INTPNT_CO_TOL_DFEAS": 1e-5,
    "MSK_DPAR_INTPNT_CO_TOL_PFEAS": 1e-5,
    "MSK_DPAR_INTPNT_CO_TOL_REL_GAP": 1e-5,
    "MSK_IPAR_INTPNT_MAX_ITERATIONS": 5
}
problem.solve(solver=cp.MOSEK, mosek_params=mosek_params, verbose=True)
```
With the problem data below on CVXPY 1.1.7 and MOSEK 9.2.18, I get the following output. Note that at iteration 5, all residuals are within the ``1e-5`` tolerance specified in ``mosek_params``. Despite this, MOSEK returns an "unknown" solution status.
```
Problem
  Name                   :                 
  Objective sense        : max             
  Type                   : LO (linear optimization problem)
  Constraints            : 20              
  Cones                  : 0               
  Scalar variables       : 24              
  Matrix variables       : 0               
  Integer variables      : 0               

Optimizer started.
Problem
  Name                   :                 
  Objective sense        : max             
  Type                   : LO (linear optimization problem)
  Constraints            : 20              
  Cones                  : 0               
  Scalar variables       : 24              
  Matrix variables       : 0               
[...]

PR #2119, opened 2023-05-04
Mosek: Add eps for optimality tolerance parameters

Paulnkk:
## Description
I added a parameter 'eps' which automatically maps a user-defined tolerance to Mosek optimality tolerance parameters. 
Issue link (if applicable):
I found the issue here: https://github.com/cvxpy/cvxpy/issues/1206

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -481,6 +481,22 @@ class MOSEK(ConicSolver):
         task = solver_output['task']
         solver_opts = solver_output['solver_options']
 
+        if solver_opts.get('eps') is not None:
+            # Fixed list for optimality tolerance parameters from
+            # https://docs.mosek.com/7.0/toolbox/Parameters.html
+            tol_params = {"MSK_DPAR_INTPNT_CO_TOL_MU_RED", "MSK_DPAR_INTPNT_CO_TOL_NEAR_REL"}
+            if solver_opts.get('mosek_params') is not None:
+                for key in tol_params:
+                    # If tolerance parameter is already provided by the user, user choice is used
+                    if key not in solver_opts['mosek_params']:
+                        solver_opts['mosek_params'][key] = solver_opts.get('eps')
+            else:
+                mosek_params_tol_dict = {}
+                for key in tol_params:
+                    mosek_params_tol_dict[key] = solver_opts.get('eps')
+                mosek_params_dict = {'mosek_params': mosek_params_tol_dict}
+                solver_opts.update(mosek_params_dict)
+
         if task.getnumintvar() > 0:
             sol_type = mosek.soltype.itg
         elif 'bfs' in solver_opts and solver_opts['bfs'] and task.getnumcone() == 0:
@@ -651,6 +667,8 @@ class MOSEK(ConicSolver):
         if 'bfs' in kwargs:
             bfs = solver_opts['bfs']
             kwargs.remove('bfs')
+        if 'eps' in kwargs:
+            kwargs.remove('eps')
         if kwargs:
             raise ValueError("Invalid keyword-argument '%s'" % kwargs[0])
 


Paulnkk:
There are two things I am curious about:

1. For now, I hardcoded the optimality tolerance parameters from https://docs.mosek.com/7.0/toolbox/Parameters.html but just for conic solvers. Should I also add the other optimality tolerance parameters (some might be irrelevant depending on the solver used).
2. Potentially, it may be also useful to implement a test for the 'eps' parameter.

rileyjmurray:
I'd like for us to reference a more recent version of the MOSEK docs, as well as the Python-specific docs:

https://docs.mosek.com/9.3/pythonapi/param-groups.html

Conic interior-point method

[dparam.intpnt_co_tol_dfeas](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.intpnt_co_tol_dfeas)
[dparam.intpnt_co_tol_infeas](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.intpnt_co_tol_infeas)
[dparam.intpnt_co_tol_mu_red](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.intpnt_co_tol_mu_red)
[dparam.intpnt_co_tol_pfeas](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.intpnt_co_tol_pfeas)
[dparam.intpnt_co_tol_rel_gap](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.intpnt_co_tol_rel_gap)

Simplex optimizer

[dparam.basis_rel_tol_s](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.basis_rel_tol_s)
[dparam.basis_tol_s](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.basis_tol_s)
[dparam.basis_tol_x](https://docs.mosek.com/9.3/pythonapi/parameters.html#mosek.dparam.basis_tol_x)

I think we should set all of these parameters when the user provides ``eps``. @aszekMosek, do you agree? If this seems like too blunt of a solution, perhaps the parameter name could be ``eps_all``, to indicate how broadly the setting applies.

Paulnkk:
@aszekMosek thanks very much for the suggestion!

Paulnkk:
I added the parameters from our discussion. I have the following thoughts:

1. The tolerances refer to different applications (admissibility of the solution, termination criteria, etc.), is it ok if 'eps' references all these tolerances? It may also make sense to set MIO tolerances separately (eps_mio?), as it makes no sense to set the parameters for non-mip.
2. Does a unit test for the eps feature make sense? If so, I would like to know what this looks like.

phschiele:
@Paulnkk You added the parameters to the `invert` function, which is called after the problem is solved. Instead, these tolerances should be applied before the solver is called.

Paulnkk:
@phschiele, thanks a lot for your support here! I merged changes into the branch. Should we use an existing problem to test for the eps parameter instead of the new one? I can also write a test that a specific parameter defined by the user is not overwritten by eps if this could increase coverage. Are there any additional changes I can make ?

phschiele:
@Paulnkk A few things before we can merge:
1. Adding the new keyword to the web docs
2. Make it work with enum parameters
3. Add the test you mentioned

Since we are testing a few different cases, let's keep this as a separate test case.

Paulnkk:
I need to check why this particular test is failing, not sure why since I just added the documentation. 

Also concerning the test I proposed, I am currently working on it but I am not sure how to access the parameters during and after the optimization since we do not return them to the user (pls correct me if I am wrong from my current understanding). I can make it work that the `solver_opts` are returned but that would make changes to productive code (actually pretty big changes if I think about it) and I am not sure if this is desired with the purpose of this PR.

phschiele:
@Paulnkk I would not make larger changes to the output in this PR. A simple (indirect) check is saving the error message from the test I've added, then including this parameter explicitly and checking that the error message changes.

Paulnkk:
@phschiele thanks for adjustments, was two minutes late to push haha, this was the test I wrote:


        first_error_message = ''
        first_error_param = ''

        try:
            prob.solve(solver=cp.MOSEK,
                       eps=1e-1,
                       mosek_params={'MSK_DPAR_INTPNT_CO_TOL_DFEAS': 1e-6})
        except mosek.Error as e:
            # Save the error message and mosek parameter
            first_error_message = str(e)
            first_error_param = str(e).split("for parameter '")[1].split("'")[0]
        try:
            prob.solve(solver=cp.MOSEK,
                       eps=1e-1,
                       mosek_params={'MSK_DPAR_INTPNT_CO_TOL_DFEAS': 1e-6,
                                     first_error_param: 1e-1})
        except mosek.Error as e:
            # Check that the error message is the same as before
            assert str(e) == first_error_message

phschiele:
@Paulnkk @rileyjmurray @SteveDiamond I pushed a few small changes to make this PR ready to merge from my side.

Since this is a mosek PR from a fork, we should get in the habit of verifying all tests passing before merging, which I did for this PR.

Verified that the docs render as well:
image

phschiele:
> @phschiele thanks for adjustments, was two minutes late to push haha, this was the test I wrote (please let me know if I understood it correctly):

Yes, that was about what I had in mind. I think the new test I've added achieves the same coverage and was a bit shorter to write.
I removed the enum you introduced mainly to avoid confusion with the existing mosek enums, and changed from a set to a tuple to have the same ordering each time.

phschiele (approved):

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -472,7 +475,7 @@ def invert(self, solver_output, inverse_data):
 
         solver_opts = solver_output['solver_options']
 
-        if solver_opts.get('accept_unknown', False):
+        if solver_opts['accept_unknown']:
```
SteveDiamond, line 478: Is accept_unknown always present?

SteveDiamond (approved):
LGTM

PR #2120, opened 2023-05-05
[CI] Use build instead of setup.py

phschiele:
## Description
Follow-up to #2118

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -258,7 +258,7 @@ jobs:
           twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_BASE_PASSWORD
 
       - name: Upload artifacts to github
-        if: ${{github.event_name == 'push'}}      
+        if: ${{github.event_name == 'push'}}
         uses: actions/upload-artifact@v1
         with:
           name: wheels-base
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,7 +142,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.12.0
 
       - name: Build source
-        if: ${{env.SINGLE_ACTION_CONFIG == 'True'}}
+        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
         run: |
           pip install build
           python -m build --sdist -o wheelhouse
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -142,9 +142,10 @@ jobs:
         uses: pypa/cibuildwheel@v2.12.0
 
       - name: Build source
-        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
+        if: ${{env.SINGLE_ACTION_CONFIG == 'True'}}
```
PTNobel, line 145: What is this diff doing?

.github/workflows/build.yml
```diff
@@ -142,9 +142,10 @@ jobs:
         uses: pypa/cibuildwheel@v2.12.0
```
PTNobel, line 142: I'm surprised this image doesn't already have `build` installed; do we really need the `pip` call? Could we upgrade to a newer builder to avoid it?

Just trying to save a little CI time.

PTNobel: Their docs don't seem to have a good answer, and actually points to our config (in a big list) as an example of using the image...

phschiele: Good point. We use `cibuildwheel` to build the wheels (parallelized across the build matrix), but we still need the source distribution.

This seems to be expected:
> Make your SDist with the [build](https://github.com/pypa/build) tool, and your wheels with cibuildwheel.

In the overall workflow `build` does not seem to be preinstalled, as I found this in the logs:

> Successfully installed build-0.10.0 packaging-23.1 pyproject_hooks-1.0.0 tomli-2.0.1

.github/workflows/build.yml
phschiele: Usually, the build (and upload of the artifact to github) only happens when we merge to master (i.e., push).
To test this change before merging, I manually enabled this by removing the condition. 
Was about to undo this before asking for reviews.

phschiele:
Fixed the failing step: https://github.com/cvxpy/cvxpy/actions/runs/4898400517/jobs/8747579373

.github/workflows/build.yml
PTNobel: Ah of course, sorry I jumped the gun!

.github/workflows/build.yml
PTNobel: Ah okay; sounds good then

PTNobel (approved):

SteveDiamond (approved):

PR #2125, opened 2023-05-14
Add pybind11 extension for sparse Cholesky decompositions

rileyjmurray:
_Preface: many of the comments below concerned an earlier SWIG implementation. The discussion with the current Pybind11 implementation starts [here](https://github.com/cvxpy/cvxpy/pull/2125#issuecomment-1575261171)._

### Context
We often need to decompose a positive definite matrix into a Gram matrix format. Right now, we do this by eigendecomposition followed by scaling the eigenvectors by the square roots of eigenvalues. This is a potentially huge expense for sparse positive definite matrices. In the past, we discussed the possibility of exposing Eigen's sparse Cholesky decomposition functionality through cvxcore; see [here](https://github.com/cvxpy/cvxpy/issues/1995#issuecomment-1373006848) and [here](https://github.com/cvxpy/cvxpy/discussions/1651#discussion-3873350).

### Changes
This (work-in-progress) PR exposes Eigen's sparse Cholesky functionality with a Pybind11 C++ extension.

Here are the primary changes:
* I added a folder called "cpp" underneath "cvxpy/utilities". I suggest we put all pybind11 C++ extension code there moving forward.
* I added a "sparsecholesky" folder underneath "cvxpy/utilities/cpp" that includes the source files for the new extension. Remarkably, we only need two files! That's a big improvement over swig.
* The new functionality is exposed with a single new function, ``sparse_cholesky``, which can be found in cvxpy/utilities/linalg.py.
* The new sparse_cholesky function is tested in cvxpy/tests/test_linalg_utils.py. Right now, that file _only_ tests sparse Cholesky. Longer term, I'd like all of our all tests for code in ``cvxpy/utilities.linalg.py`` to be in this file.

Here are some secondary changes:
* I found that our ``setup.py`` file was cluttered and hard to understand, so I factored out much of its code into two separate modules: ``setup/versioning.py`` and ``setup/extensions.py``. If anyone objects to this, please speak up.
* I added pybind11 to our list of dependencies in several places. I may have added it in places where we don't need it (i.e., runtime requirements vs build requirements). It would be great if someone could check my work.
* I replaced setup.py's use of the deprecated ``distutils.sysconfig`` with plain ``sysconfig``. 

### TODOs
I still need to use this in at least one place within cvxpy's existing code.


SteveDiamond:
This is great! Are we sure we want to double down on Swig though? @akshayka has used pybind11 for a C++ interface and it seemed much nicer than Swig.

phschiele:
Awesome contribution, @rileyjmurray !

Before checking this PR in detail, do you think it could make sense to have this in its own repo (e.g. `cvxpy/spchol`)? I can imagine this functionality alone would be interesting to some, who could then use it directly.
That separation also would make it modular if some users want to use e.g. Mosek's sparse Cholesky functionality or at some point other implementations become available.

PTNobel:
I'm inclined to agree with Philipp that maintaining this as a separate repo is worth considering...

Do we expect to ever have an API change here? Would supporting multiple versions of this library if it were installed separately be difficult?

Also worth considering, could/should we try to get this merged into SCIPY?

rileyjmurray:
Thanks, for the comments, everyone!

@SteveDiamond, SWIG wasn't actually so bad. The interface file I had to write was tiny. If we keep the core C++ implementation as simple as it is now, then it would be easy to port to PyBind 11 if we ever want to do that.

I don't know how I feel about making a separate repo or package. On the one hand, Eigen is header-only, so we're less likely to run into build issues. On the other hand, I don't want people to come in asking for new features (e.g., templating different numerical precisions, using different pivoting rules, incomplete Cholesky, etc...). I'm not sure about the modularity mentioned by @phschiele; that's opening the door for complexity that I think I'd like to avoid.

I agree with @parthb83 that it would be nice if SciPy had this functionality. Whether or not we try to make that happen is a function of a variety of factors. The biggest factor is how SciPy devs feel about using (some of)Eigen as a dependency. Luckily, Eigen is MIT-licensed as of 2012, so that might not be a deal breaker. The main problem is that I don't have the bandwidth to lead the charge on getting this in SciPy.

PTNobel:
Another thing to consider: there was a discussion today in the OSQP Slack about modifying the python interface of qdldl to expose the underlying factors. Could it be a smaller maintenance burden to use that library instead of maintaining our own bespoke wrapper for sparse cholesky factoring?

To be clear, I'm fully supportive of this work (after I get a chance to do a more thorough code review), just want to enumerate a few of the alternatives for consideration.

(aside: riley you tagged the wrong parth)

bstellato:
I think it should not be hard to extend qdldl. It should a matter of:
1. adding functions to the C++ wrapper [here](https://github.com/osqp/qdldl-python/blob/master/cpp/qdldl.hpp) to retrieve the factors
2. Wrap them in the [pybind11 object](https://github.com/osqp/qdldl-python/blob/master/cpp/wrapper.cpp)

PTNobel:
I opened a PR against qdldl following Bart's outline that exposes its factors: osqp/qdldl-python#32

rileyjmurray:
@PTNobel I support accessing QDLDL's factors when QDLDL is installed. That should be the case generically but it won't be the case when people install cvxpy-base. I'll revise this PR so that the sparse Cholesky functionality is accessible through a more generic interface.

rileyjmurray:
We just had a close call on this one, friends.

I looked at Eigen's implementation of sparse Cholesky and saw that it contains GPL'd code from SuiteSparse. This was allowed under Eigen's old LGPL license, but it isn't compatible with Eigen's current MPL license in general. Luckily for us Google has negotiated a license for that functionality under the MPL, specifically in the context of Eigen and Eigen-derivative works. Just pointing this out in case questions arise as to whether or not this functionality is properly licensed. (And to raise the issue of maybe updating our Eigen source.)

rileyjmurray:
Making a note of a weird experience I had:

My first push to this branch today resulted in many failing CI tests. The tests were failing for low-level reasons (like, accessing out-of-bounds memory) and before any of my modified code should have run.

The code was generated by a newer version of SWIG that I had just installed (4.1.1). Because this newer version of SWIG said our option of ``-py3`` was deprecated and ignored, I thought maybe SWIG was the problem. This led me to switch to the Python environment on which I started this PR, which used SWIG 4.0.2. Things started to work again once I used that version of SWIG. I tried to google the possible causes for the error, but I gave up once I found a fix.

PTNobel:
In re licensing, I almost wonder if making this required by `cvxpy-base` kind of kills the "we keep licensing simple with this subpackage" promise. Is there a reason to not make `qdldl-python` a `cvxpy-base` dependency?

Second, do you have reading materials on the licensing scheme you mentioned? Also, if this code is LGPL, we have to expose it as a separate library to keep cvxcore from being LGPL.

rileyjmurray:
Eigen has an option to disable all functionality that doesn’t fall under
MPL. Cvxcore compiles and runs correctly with that functionality disabled.
This sparse Cholesky functionality requires a more up to date version of
Eigen (from 2019 IIRC) in order to compile in MPL-only mode. If we update
the Eigen source code that we ship with cvxcore then this is a non-issue. I
can include those changes in this PR of people would like. In either case —
we should at least remove any Eigen code that’s under LGPL (which we
currently ship as text files that do nothing programmatically).

In the past we’ve had qdldl wheels be a blocking issue in supporting the
latest version of Python. I don’t want to run into similar problems in the
future with cvxpy-base.

rileyjmurray:
In case anyone's wondering why so many files changed, that's because I screwed up some merges and/or squashes and/or reverts.

rileyjmurray:
Tests pass on my laptop but fail on the majority of CI builds. The CI that ran on my latest commit said that a segfault occurs at this line: https://github.com/cvxpy/cvxpy/blob/a9e14b2d037c91c5e2854c4effaf7b68dfd0a115/cvxpy/utilities/linalg.py#L193.

That line is when SWIG's wrapper around C++ std::vector is converted into a list.

rileyjmurray:
Last week @SteveDiamond mentioned that with these poorly understood segfaults stemming from SWIG, it might make sense to consider PyBind 11 seriously. After giving it some thought, I've decided that I'd like to try that route. Ping @akshayka -- is this something you could do? If so, feel free to make commits directly to this branch or to make a new branch with commit [afaa92b](https://github.com/cvxpy/cvxpy/pull/2125/commits/afaa92bafe2365645d23b44b9c4035c37c363cfa) as the root.

PTNobel:
I'll quickly chime in, I just interacted with pybind11 for the qdldl-python changes and it is a joy to work with compared to my experience with swig.

rileyjmurray:
I'm starting work on switching from swig to PyBind11.

rileyjmurray:
I've got the bulk of the PyBind11 approach working! It's definitely _much_ nicer than swig. I'll push my changes once I've cleaned up data types and implemented exception handling.

PTNobel:
Feel free to request a code review from me once it is ready.

rileyjmurray:
ping @PTNobel. There are some failing tests due to unsupported syntax on Python 3.7, but that'll be trivial to fix.

cvxpy/atoms/quad_form.py
```diff
@@ -197,7 +198,14 @@ def decomp_quad(P, cond=None, rcond=None, lower=True, check_finite: bool = True)
 
     """
     if is_sparse(P):
-        P = np.array(P.todense())  # make dense (needs to happen for eigh).
+        try:
+            sign, Lp = sparse_cholesky(P)
+            if sign > 0:
+                return 1.0, Lp, None
```
PTNobel, line 204: Docs for this function do not describe how to interpret a None value, tho I do think it is obvious...

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,81 @@ def gershgorin_psd_check(A, tol):
...
+
+    If Cholesky succeeds and permute_L=False, then we return a lower-triangular
+    matrix L in CSR-format and a permutation vector p so that
+    (L[p, :]) @ (L[p, :]).T == A within numerical precision.
+
+    We raise a ValueError if Eigen's Cholesky fails or if we certify indefiniteness
+    before calling Eigen.
+    """
+    if not isinstance(A, spar.spmatrix):
+        raise ValueError('Input must be a SciPy sparse matrix.')
```
PTNobel, line 206: Why is this not an entry in `SparseCholeskyMessages`

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,81 @@ def gershgorin_psd_check(A, tol):
...
+            raise ValueError(SparseCholeskyMessages.ASYMMETRIC)
+        # check a necessary condition for positive/negative definiteness; call this
+        # function on -A if there's evidence for negative definiteness.
+        d = A.diagonal()
+        maybe_posdef = np.all(d > 0)
+        maybe_negdef = np.all(d < 0)
+        if not (maybe_posdef or maybe_negdef):
+            raise ValueError(SparseCholeskyMessages.INDEFINITE)
+        if maybe_negdef:
+            res = sparse_cholesky(-A, sym_tol, permute_L)
```
PTNobel, line 221: I'd rather see 
```python
_, M1, M2 = sparse_cholesky(...)
return -1, L
```
I think it better captures the intent here.

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,81 @@ def gershgorin_psd_check(A, tol):
...
+            return -1.0, *res[1:]
+
+    A_coo = spar.coo_matrix(A)
+    n = A.shape[0]
+
+    # Call our C++ extension
+    inrows = spchol.IntVector(A_coo.row.tolist())
```
PTNobel, line 228: Correct me if I'm wrong but this is two copies of all of `A_coo`.

One in `.tolist` and one in `spchol.IntVector`.

cvxpy/utilities/linalg.py
PTNobel: Wait our return type changes depending on the arguments (which I don't like).

I would still do
```python
_, *rets = sparse_cholesky(...)
return -1, *rets
```

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,81 @@ def gershgorin_psd_check(A, tol):
...
+            raise ValueError(e.args)
+        else:
+            raise RuntimeError(e.args)
+
+    # error checking and return values
+    outvals = list(outvals)
+    outrows = list(outrows)
+    outcols = list(outcols)
+    outpivs = list(outpivs)
+    L = spar.csr_matrix((outvals, (outrows, outcols)), shape=(n, n))
```
PTNobel, line 251: Two copies again.

cvxpy/utilities/linalg.py
PTNobel, line 1: A lot of copies in this file, I don't know a way to avoid it (tho I would hope there is a way to go straight from np.array to std::vector that would let us only have one copy, but it does make me sad.

If we were using hamr and eigen exposed APIs with T *, we could do this all as zerocopy, fun fact.

setup.py
PTNobel, line 1: It feels like there is a lot of stuff going on here in terms of switching to new sysconfig and setuptools features.

setup/__init__.py
PTNobel, line 1: All other new files have contained an empty line, why does this one not?

setup/versioning.py
PTNobel, line 1: Is this just copy and pasted from old code? I didn't read it closely, will if it is new.

setup.py
PTNobel: Ah I misunderstood I think, sysconfig is a just a small helper lib you made?

PTNobel (changes requested):
Overall, looks fine. I have a lot of comments, some of which are actionable; some of which just need responses.

I'm marking it as requested changes, but certainly most of the comments probably don't need much in the way of changes...

cvxpy/atoms/quad_form.py
rileyjmurray: I guess I could return ``1.0, Lp, np.empty(0)``, or something to that effect.

cvxpy/atoms/quad_form.py
```diff
@@ -197,7 +198,14 @@ def decomp_quad(P, cond=None, rcond=None, lower=True, check_finite: bool = True)
 
     """
     if is_sparse(P):
-        P = np.array(P.todense())  # make dense (needs to happen for eigh).
+        try:
```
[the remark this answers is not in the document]
rileyjmurray: Totally!

cvxpy/utilities/linalg.py
rileyjmurray: No deliberate reason; I'll move it there.

cvxpy/utilities/linalg.py
rileyjmurray: That might very well be the case; I'll try removing the ``tolist`` and see if Pybind11 automatically handles iteration.

cvxpy/utilities/linalg.py
rileyjmurray: I'll try to fix.

setup.py
rileyjmurray: I don't think we're using any new setuptools features. The only setuptools thing I changed was moving where it was imported to reflect a comment that was left a while ago. 

``sysconfig`` is a built-in module added in Python 3.2. I used it in the recommended way to replace a deprecated use of ``distutils``.

setup/__init__.py
rileyjmurray: Will add

setup/versioning.py
rileyjmurray: Just copy-pasted

cvxpy/utilities/linalg.py
rileyjmurray: I've simplified this function's return types; we no longer permute rows of L inside the function.

cvxpy/utilities/linalg.py
rileyjmurray: I tried removing this on my machine, and it worked!

rileyjmurray:
@PTNobel, I believe I addressed all of your concerns. Can you take another look?

rebuild_cvxcore.sh
PTNobel, line 1: Why not do `cd ../../` as you did in the last revision? Is this also swig revision?

PTNobel (approved):
One trivial comment, but otherwise the diff looks great.

rebuild_cvxcore.sh
rileyjmurray: No, that was a manual change. I noticed there were several files that I could leave unchanged in this PR if I just reverted style changes.

PTNobel: Sounds good. Maybe we should have a small style pr with some of them later...

rileyjmurray:
Ping @phschiele. Can you take a look at benchmarks? I think they're failing for some pybind11-related issue.

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,73 @@ def gershgorin_psd_check(A, tol):
...
+
+class SparseCholeskyMessages:
+
+    ASYMMETRIC = 'Input matrix is not symmetric to within provided tolerance.'
+    INDEFINITE = 'Input matrix is neither positive nor negative definite.'
+    EIGEN_FAIL = 'Cholesky decomposition failed.'
+    NOT_SPARSE = 'Input must be a SciPy sparse matrix.'
+
+
+def sparse_cholesky(A, sym_tol=1e-12, assume_posdef=False):
```
phschiele, line 190: We have a number of tolerances defined in `settings.py`. We could add `sym_tol` there to make it easier to reuse/keep track of the tolerances we are using.

cvxpy/utilities/linalg.py
```diff
@@ -176,3 +177,73 @@ def gershgorin_psd_check(A, tol):
...
+    try:
+        spchol.sparse_chol_from_vecs(
+            n, inrows, incols, invals,
+            outpivs, outrows, outcols, outvals
+        )
+    except RuntimeError as e:
+        if e.args[0] == SparseCholeskyMessages.EIGEN_FAIL:
+            raise ValueError(e.args)
+        else:
+            raise RuntimeError(e.args)
```
phschiele, line 242: Just to clarify, this would not be handled in the `try/catch` in `decomp_quad` because it is not expected/unrelated to the Cholesky decomposition.

phschiele (approved):
@rileyjmurray I think this is because we always run the benchmarks twice: once with the HEAD of the PR and once with the latest master. This is to reduce the variance from getting assigned different machines over time.
It looks like reusing the environment between these runs does not work when the requirements change within the PR.
I suggest merging this PR (assuming we're happy except for the failing benchmark). Then, we update another PR to see if the benchmarks work again. If not, the problem must be in the benchmark repo and I can take a closer look there.

I did verify that the code in this PR installs without issues on a fresh codespace, passing all tests. Just had two small additional comments above. Thanks again, this is awesome!

cvxpy/utilities/linalg.py
rileyjmurray: Good idea!

cvxpy/utilities/linalg.py
rileyjmurray: If we raise a ValueError then it should be caught in decomp_quad, but if we raise a RuntimeError then it won't be caught. That was intentional since I only wrote one type of exception for the C++ code.

SteveDiamond (approved):
Looks great!

PR #2127, opened 2023-05-15
Follow-up Mosek accept unknown

phschiele:
## Description
Trying to fix the CI after merging #2117.
Under windows, the test gives `cp.optimal` instead of `cp.optimal_inaccuracte`.

Looks like @rileyjmurray called it?
https://github.com/cvxpy/cvxpy/pull/2117#issuecomment-1530464403

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -677,7 +677,7 @@ class TestMosek(unittest.TestCase):
         }
         sth = sths.lp_5()
         sth.solve(solver=cp.MOSEK, accept_unknown=True, mosek_params=mosek_param)
-        assert sth.prob.status == cp.OPTIMAL_INACCURATE
+        assert sth.prob.status in {cp.OPTIMAL_INACCURATE, cp.OPTIMAL}
 
         with pytest.raises(cp.error.SolverError, match="Solver 'MOSEK' failed"):
             sth.solve(solver=cp.MOSEK, mosek_params=mosek_param)
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -681,16 +681,13 @@ class TestMosek(unittest.TestCase):
                        cp.sum(x) >= 0.1,
                        x >= 0,
                        cone_con]
-        obj = cp.Minimize(3 * x[0] + 2 * x[1] + x[2])
+        obj = cp.Minimize(3 * x[0] + 2 * x[1] + x[2] + cp.exp(x[0]))
         prob = cp.Problem(obj, constraints)
         prob.solve(solver=cp.MOSEK, accept_unknown=True, mosek_params=mosek_param)
-        assert prob.status is cp.OPTIMAL_INACCURATE
-        with pytest.raises(cp.error.SolverError) as se:
+        assert prob.status == cp.OPTIMAL_INACCURATE
+
+        with pytest.raises(cp.error.SolverError, match="Solver 'MOSEK' failed"):
             prob.solve(solver=cp.MOSEK, mosek_params=mosek_param)
-            exc = " Solver 'MOSEK' failed. " \
-                  "Try another solver, or solve with verbose=True " \
-                  "for more information."
-            assert str(se.value) == exc
 
 
 @unittest.skipUnless('CVXOPT' in INSTALLED_SOLVERS, 'CVXOPT is not installed.')
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -673,7 +673,7 @@ class TestMosek(unittest.TestCase):
 
     def test_mosek_accept_unknown(self) -> None:
         mosek_param = {
-            "MSK_DPAR_OPTIMIZER_MAX_TIME": 0
+            "MSK_IPAR_INTPNT_MAX_ITERATIONS": 0
         }
         sth = sths.lp_5()
         sth.solve(solver=cp.MOSEK, accept_unknown=True, mosek_params=mosek_param)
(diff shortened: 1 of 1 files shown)


phschiele:
I think there are two issues here:
1. In the initial PR, mosek was not even tested, see the list of installed solvers from the successful run below. This is because secrets are not shared by default with PRs, and the mosek license is one such secret.
```
['CLARABEL', 'COPT', 'CPLEX', 'CVXOPT', 'DIFFCP', 'ECOS', 'ECOS_BB', 'GLOP', 'GLPK', 'GLPK_MI', 'GUROBI', 'OSQP', 'PDLP', 'PROXQP', 'SCIP', 'SCIPY', 'SCS', 'SDPA', 'XPRESS']
```
2. Either the peresolve is much better under Windows, leading to `optimal` results in all three problems I've tested, or the time constraint is not taken into account under in those cases.

SteveDiamond:
> I think there are two issues here:
> 
> 1. In the initial PR, mosek was not even tested, see the list of installed solvers from the successful run below. This is because secrets are not shared by default with PRs, and the mosek license is one such secret.
> 
> ```
> ['CLARABEL', 'COPT', 'CPLEX', 'CVXOPT', 'DIFFCP', 'ECOS', 'ECOS_BB', 'GLOP', 'GLPK', 'GLPK_MI', 'GUROBI', 'OSQP', 'PDLP', 'PROXQP', 'SCIP', 'SCIPY', 'SCS', 'SDPA', 'XPRESS']
> ```
> 
> 2. Either the peresolve is much better under Windows, leading to `optimal` results in all three problems I've tested, or the time constraint is not taken into account under in those cases.

I see, I think we should change the test to just verify that the accepts_unknown keyword doesn't crash things. Or maybe that the solution is either OPTIMAL or OPTIMAL_INACCURATE

rileyjmurray:
@phschiele, is there a way to make sure MOSEK is installed in PRs by default? Based on the current settings, it seems like CI for MOSEK-based PRs is useless. I'm also reasonably sure that we used to have MOSEK in CI in the recent past.

@SteveDiamond, I agree with your suggestion.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -675,22 +675,12 @@ def test_mosek_accept_unknown(self) -> None:
...
-        assert prob.status is cp.OPTIMAL_INACCURATE
-        with pytest.raises(cp.error.SolverError) as se:
-            prob.solve(solver=cp.MOSEK, mosek_params=mosek_param)
-            exc = " Solver 'MOSEK' failed. " \
-                  "Try another solver, or solve with verbose=True " \
-                  "for more information."
-            assert str(se.value) == exc
+        sth = sths.lp_5()
+        sth.solve(solver=cp.MOSEK, accept_unknown=True, mosek_params=mosek_param)
+        assert sth.prob.status == cp.OPTIMAL_INACCURATE
```
phschiele, line 680:
```suggestion
        assert sth.prob.status in {cp.OPTIMAL_INACCURATE, cp.OPTIMAL}
```

phschiele:
> @phschiele, is there a way to make sure MOSEK is installed in PRs by default? Based on the current settings, it seems like CI for MOSEK-based PRs is useless. I'm also reasonably sure that we used to have MOSEK in CI in the recent past.
> 
> @SteveDiamond, I agree with your suggestion.

I agree, and I also was under the impression new used to have mosek tests running. Those may have been PRs from within the cvxpy repo, as the mosek license is stored as a repository secret, which is not copied to forks.

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. [Source](https://docs.github.com/en/actions/security-guides/encrypted-secrets)

Generally, this is a good thing, as otherwise, a forked PR could access the PYPI token, etc. 

I have seen a workaround where you can manually trigger an action that duplicates the PR from within the repo and copies over the test results, but that seems quite cumbersome. Most of the commercial solvers that we test allow solving small problems without a license, which makes things a bit easier.
I can try to look for other ways to achieve this, but at least at first glance, it does not seem possible to selectively share secrets with forked PRs.

phschiele:
@aszekMosek Do you have an idea how this can solve with `optimal` status with a max time of 0 under windows only? Seems like the time limit is not applied correctly.

```py
mosek_param = {"MSK_DPAR_OPTIMIZER_MAX_TIME": 0}
sth = sths.lp_5()  # I have tried other problems as well
sth.solve(solver=cp.MOSEK, mosek_params=mosek_param)
```

SteveDiamond:
@phschiele for some reason the CI is all passing in MR #2131.

phschiele:
> @phschiele for some reason the CI is all passing in MR #2131.

That's because the license is not available on the fork, so the install is skipped and thus the tests are skipped:

https://github.com/cvxpy/cvxpy/blob/df26389685d487a037426f4c1213cfd94c38624c/continuous_integration/install_dependencies.sh#L52-L55

Paulnkk:
> @aszekMosek Do you have an idea how this can solve with `optimal` status with a max time of 0 under windows only? Seems like the time limit is not applied correctly.
> 
> ```python
> mosek_param = {"MSK_DPAR_OPTIMIZER_MAX_TIME": 0}
> sth = sths.lp_5()  # I have tried other problems as well
> sth.solve(solver=cp.MOSEK, mosek_params=mosek_param)
> ```

Maybe it works by setting max iterations on the solver to 0 (since we have a test LP, we can set MSK_IPAR_SIM_MAX_ITERATIONS = 0), should lead to the same Error, right ?

aszekMosek:
The effect is non-deterministic. In some ``pytest`` runs I tried everything is fine, in others it is as if time was frozen and the solver prints to the log as if it did each step in time ``0.0``.

I have no idea what causes this effect, but it seems not to happen if I isolate the test to a stand-alone python script and run it outside the test system.

It will be more reliable to set ``"MSK_IPAR_INTPNT_MAX_ITERATIONS": 0`` (it is still the interior point optimizer being used for LP).

SteveDiamond:
@aszekMosek limiting max iters seems to work.

SteveDiamond (approved):

PR #2130, opened 2023-05-16
Stop cvxpy from crashing when using an old version of clarabel

SteveDiamond:
## Description
Clarabel would crash in the try/except block when using an old version.

## Type of change
- [x] Bug fix

- [ ] Write unittests.


PTNobel (approved):
If there's an exception on import, I think we should treat that as being not installed.

Issue #2128, opened 2023-05-16
Problem MOSEK with Power Cone and SDP


dcajasn:
**Describe the bug**
When I try to solve a problem that involves an SDP constraint and Power Cone constraint with MOSEK an error message "Error: Tyring to append a too big cone." appears but when I choose SCS I don't have problems.

**To Reproduce**
```python
import numpy as np
import pandas as pd
import cvxpy as cp
import riskfolio as rp
from scipy.linalg import sqrtm

rs = np.random.RandomState(seed=123)

N = 10
T = 1000
rows = int(N*(N+1)/2)

R = rs.randn(T, N)

x = cp.Variable((N,1))
X = cp.Variable((N,N), PSD=True)
psi = cp.Variable(nonneg=True)
gamma = cp.Variable((rows, 1))

D_2 = rp.duplication_matrix(N)
L_2 = rp.duplication_elimination_matrix(N)
S_2 = rp.duplication_summation_matrix(N)

Sigma_4 = rp.cokurtosis_matrix(R)
Sigma_4 = S_2 @ Sigma_4 @ S_2.T
Sqrt_Sigma_4 = sqrtm(Sigma_4)

y = L_2 @ cp.reshape(cp.vec(X), (N*N,1))
z = Sigma_4 @ y

M1 = cp.vstack([X, x.T])
M2 = cp.vstack([x, np.ones((1, 1))])
M3 = cp.hstack([M1, M2])

constraints = [cp.SOC(psi, Sqrt_Sigma_4 @ y),
               cp.sum(x) == 1,
               x >= 0,
               M3 >> 0,
               ]

constraints += [cp.constraints.power.PowCone3D(y, z, gamma, 0.5)]

obj = cp.pnorm(psi - gamma, p=2)

objective = cp.Minimize(obj)
problem = cp.Problem(objective, constraints)
problem.solve(solver='MOSEK')
# problem.solve(solver='SCS') # with this line works
```

**Expected behavior**
That the problem can be solve and return a x.value.

**Output**
<img width="762" alt="image" src="https://github.com/cvxpy/cvxpy/assets/61527316/5464b51b-0ff5-4a61-8b75-a98b54c81bfc">

**Version**
 - OS: macOS Ventura 13.3.1 
 - CVXPY Version: 1.3.1

**Additional context**
I use a M1 computer.

aszekMosek:
[...]

PR #2129, opened 2023-05-16
WIP Fix for Mosek interface with SDP + power cone

SteveDiamond:
## Description
Fix for SDP + power cone.
Issue link (if applicable): #2128 

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [x] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [ ] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2126, opened 2023-05-15
Fix bug with calling Mosek simplex solver

SteveDiamond:
## Description
The Mosek simplex and dual simplex solver cannot be used right now due to a simple interface bug, fixed here. I also made the Clarabel interface a bit more robust because it was failing on my old Clarabel version.

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [x] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2131, opened 2023-05-16
Fixes to PowCone3D and updates to test_conic_solvers.py

dcajasn:
Add test for the fix that solves issue #2128

## Description
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.


cvxpy/tests/test_conic_solvers.py
```diff
@@ -640,6 +640,33 @@ def test_mosek_simplex(self) -> None:
             mosek_params={"MSK_IPAR_OPTIMIZER": "MSK_OPTIMIZER_DUAL_SIMPLEX"}
         )
 
+    def test_mosek_sdp_power(self) -> None:
+        """Test the problem in issue #2128"""
+        T, N = 10, 1000
```
SteveDiamond, line 645: can you make this problem smaller?

SteveDiamond:
Can you include the change in #2129 to mosek_conif.py here and then I will close that PR?

cvxpy/tests/test_conic_solvers.py
dcajasn: When you say smaller, it means less lines or decrease the values of N and T?

dcajasn:
Ok

cvxpy/tests/test_conic_solvers.py
SteveDiamond: Smaller N

dcajasn:
I've just changed the branch from steven-sdp-power to master. It includes all the changes you made on steven-sdp-power.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -622,6 +622,51 @@ def test_mosek_params(self) -> None:
...
+           x >= 0,
+           y >= 0.01,
+           M3 >> 0,
+           cp.PowCone3D(x, np.ones((N,1)), y, 0.9)
+          ]
+
+        obj =  cp.trace(Sigma @ X) + cp.norm(y, p=2)
+        objective = cp.Minimize(obj)
+        problem = cp.Problem(objective, constraints)
+        problem.solve(solver='MOSEK')
```
rileyjmurray, line 668: General comment 1: since you aren't testing MOSEK's output, there's no reason to actually solve the problem. You should set MOSEK's time and iteration limits to 0.

General comment 2: it would be nice to have a mixed powercone / SDP test that can be used in ``solver_test_helpers.py``. If you define a _much_ smaller problem, then you can write down the answer reported by MOSEK explicitly. You can also solve the problem with SCS with very small ``eps`` to check that the solutions are the same. I'm pretty sure that very small problems, say with ``T=2, N=3`` would suffice to trigger the failure case you observed.

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -245,7 +245,7 @@ def apply(self, problem):
                 total_psd = sum([d * (d+1) // 2 for d in K[a2d.PSD]])
                 A_psd = A[:, idx:idx+total_psd]
                 c_psd = c[idx:idx+total_psd]
-                if K[a2d.DUAL_EXP] == 0:
+                if (K[a2d.DUAL_EXP] == 0) and (K[a2d.DUAL_POW3D] == 0):
```
SteveDiamond, line 248: All the changes besides this one line are part of #2126. I included them accidentally in #2129 . Could you remove them and keep just the change at line 248?

cvxpy/tests/test_conic_solvers.py
dcajasn: I modified the test according the first comment. I'm working on the second comment.

dcajasn: I've just finished the test helpers. Let me know if it is ok 😅

SteveDiamond (approved):
LGTM! Thanks so much!

rileyjmurray (changes requested):
@dcajasn, thanks for updating solver_test_helpers! I played around with that new test and found some undiscovered bugs in the process. So I'm going to ask you to do a few things.

1) Please add ``sth.check_primal_feasibility(places)`` right before the ``sth.verify_primal_values(places)`` in all functions in ``StandardTestPCPs``.  This change will reveal a bug in the ``PowCone3D`` class that causes MOSEK and SCS tests to fail.

2) Please add the ``@property`` decorator to the ``residual`` method of ``PowCone3D``. This is the fix that's needed to resolve the failures introduced by my first request.

3) Please change the default value of ``duals`` in ``StandardTestMixedCPs.test_sdp_pcp_1`` to ``False``.

4) Please replace the ``test_mosek_sdp_power`` function with the following
  ```
      def test_mosek_sdp_pcp_1():
          StandardTestMixedCPs.test_sdp_soc_1(solver='MOSEK')
  ```
    
5) Please add the following function to ``TestSCS(BaseTest)``
  ```
    def test_scs_sdp_pcp_1():
        StandardTestMixedCPs.test_sdp_soc_1(solver='SCS')
  ```
  
  That's all from me. Thanks for your patience in this process!

cvxpy/tests/test_conic_solvers.py
rileyjmurray: @dcajasn the new test case ended up being very revealing! I just submitted a review that asks you to resolve some previously-undiscovered bugs.

dcajasn:
Hi @rileyjmurray, I've just made the changes. Thank you for all your hard work developing and improving this terrific software.

SteveDiamond:
I fixed some minor bugs that were causing test failures in the CI.

rileyjmurray (approved):

Issue #1136, opened 2020-09-29
solver_stats.num_iters no longer reported for MOSEK


farrellm:
**Describe the bug**
Previously (in version 0.4.x), `problem.solver_stats.num_iters` was populated after solving with MOSEK. Now, it remains `None` after solving.

**To Reproduce**

1. Take any `problem`
2. Solve using MOSEK
3. Look at `problem.solver_stats.num_iters`, it will be `None`

**Expected behavior**
`problem.solver_stats.num_iters` is some positive integer after solving with MOSEK.

**Version**
 - OS: Linux
 - CVXPY Version: 1.1.5

**Additional context**
Here's where `num_iters` should be set:
https://github.com/cvxgrp/cvxpy/blob/master/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py#L434

Transurgeon:
closing this issue since it was addressed in #2137

PR #2137, opened 2023-05-26
Mosek: Add number of iterations attribute

Paulnkk:
## Description
I have added an attribute to access the number of iterations until termination.
Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/1136

## Type of change
- [x] Bug fix

- [ ] Write unittests.


Paulnkk:
I have the following thoughts concerning this PR:  	

1. I have not yet found a parameter that stores the number of iterations of the method used. Therefore, I have taken variables that store the number of iterations of specific methods (Simplex, Interior-Point, and Interior-Point MIO). Here I have assumed that if a procedure terminates without pre-solve, one of the values is greater than 0 and returns a positive number as the number of iterations. For pre-solve, a 0 is returned. 

2. What would a unit test for this feature look like? Does it make sense to write a test that confirms that the parameter exists (if 'None' is returned via the num_iters parameter, an assertion is activated).

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -508,7 +508,11 @@ def invert(self, solver_output, inverse_data):
                 K = inverse_data['K_dir']
                 prim_vars = MOSEK.recover_primal_variables(task, sol_type, K)
                 dual_vars = MOSEK.recover_dual_variables(task, sol_type)
-        attr = {s.SOLVE_TIME: task.getdouinf(mosek.dinfitem.optimizer_time)}
+        attr = {s.SOLVE_TIME: task.getdouinf(mosek.dinfitem.optimizer_time),
+                s.NUM_ITERS: task.getdouinf(mosek.iinfitem.intpnt_iter) +
+                             task.getdouinf(mosek.liinfitem.simplex_iter) +
+                             task.getdouinf(mosek.liinfitem.mio_simplex_iter)
```
phschiele, line 514:
```suggestion
                s.NUM_ITERS: task.getintinf(mosek.iinfitem.intpnt_iter) +
                             task.getlintinf(mosek.liinfitem.simplex_iter) +
                             task.getlintinf(mosek.liinfitem.mio_intpnt_iter) + 
                             task.getlintinf(mosek.liinfitem.mio_simplex_iter)
```

Haven't tried, but based on the [mosek docs](https://docs.mosek.com/latest/pythonapi/optimizer-task.html#mosek.task.getintinf), I would expect to use `getintinf` for `iinfitem` and `getlintinf` for `liinfitem`.

I have added `mio_intpnt_iter`, I think that's all after searching for `_iter` in the docs.

Paulnkk: @phschiele thanks! Made changes here: [Add mio interior point](https://github.com/cvxpy/cvxpy/pull/2137/commits/3095d223b24e9144eaec69485a514d050e34076a). 

I could not find other parameters that refer to iterations used until termination. Is a unit test also needed for the attribute ?

phschiele: Yes, you can add a check that the key is present. I would not check the specific number, but `solution[s.NUM_ITERS] >= 0` should work.
Maybe you can add this to an existing test with a similar scope.

Paulnkk: @phschiele thanks ! I added a new test here: [Add test for number of iterations](https://github.com/cvxpy/cvxpy/pull/2137/commits/4e22b4bc4277ae0685683e27f39465a9f1d2db13). Instead of the new test I could also put an assertion in `test_mosek_params` (but I wasn't sure because the mentioned test refers to solver specific mosek parameters).

aszekMosek:
One could argue that for MIO https://docs.mosek.com/latest/pythonapi/constants.html#mosek.iinfitem.mio_num_branch is more relevant / meaningful for the user than the number of internal iterations but of course that depends on what you want to measure.

phschiele:
@Paulnkk @aszekMosek How about the following:
We use `mio_num_relax` when computing the single `num_iters` number, but we provide all of the iteration fields separately in the `extra_stats` dictionary.

Paulnkk:
@phschiele works for me, great idea! I have seen that you raised a PR for the solver statistics, should we wait until this branch is merged and after that make changes here ?

phschiele:
> @phschiele works for me, great idea! I have seen that you raised a PR for the solver statistics, should we wait until this branch is merged and after that make changes here ?

#2141 should not interfere with this PR, so you can go ahead and make the changes.

Paulnkk:
I added suggested changes and additional tests, will also add the updated doc

phschiele:
@Paulnkk @rileyjmurray @SteveDiamond:
I've pushed a slight modification. The semantics are now as follows:
For `num_iters`, we return our best guess of what the user might want, i.e., number of iterations for simplex/interior point, and number of relaxations for MIP.
In the latter case, a user might want to know the (total?) number of simplex/interior point iterations, so we provide `mio_intpnt_iter` and `mio_simplex_iter` in the extra stats.

Again, since this is a mosek PR coming from a fork, I confirm that all tests are passing.

rileyjmurray:
The latest changes that I see show that NUM_ITERS is being mapped to a sum
```
s.NUM_ITERS: task.getintinf(mosek.iinfitem.intpnt_iter) +
                task.getlintinf(mosek.liinfitem.simplex_iter) +
                task.getintinf(mosek.iinfitem.mio_num_relax)
```
Summing intpnt_iter and simplex_iter makes sense, as these should be
mutually exclusive. But it seems weird to me to sum these with
mio_num_relax.

I suggest that NUM_ITERS be the sum of intpnt_iter and simplex_iter for
continuous problems. Then discrete problems would have
NUM_ITERS=mio_num_relax, and we'd store the sum of intpnt_iter and
simplex_iter as additional information.

rileyjmurray:
The latest changes that I see show that NUM_ITERS is being mapped to a sum
```
s.NUM_ITERS: task.getintinf(mosek.iinfitem.intpnt_iter) +
                task.getlintinf(mosek.liinfitem.simplex_iter) +
                task.getintinf(mosek.iinfitem.mio_num_relax)
```
Summing intpnt_iter and simplex_iter makes sense, as these should be
mutually exclusive. But it seems weird to me to sum these with
mio_num_relax.

I suggest that NUM_ITERS be the sum of intpnt_iter and simplex_iter for
continuous problems. Then discrete problems would have
NUM_ITERS=mio_num_relax, and we'd store the sum of intpnt_iter and
simplex_iter as additional information.

phschiele:
> The latest changes that I see show that NUM_ITERS is being mapped to a sum
> 
> ```
> s.NUM_ITERS: task.getintinf(mosek.iinfitem.intpnt_iter) +
>                 task.getlintinf(mosek.liinfitem.simplex_iter) +
>                 task.getintinf(mosek.iinfitem.mio_num_relax)
> ```
> 
> Summing intpnt_iter and simplex_iter makes sense, as these should be mutually exclusive. But it seems weird to me to sum these with mio_num_relax.
> 
> I suggest that NUM_ITERS be the sum of intpnt_iter and simplex_iter for continuous problems. Then discrete problems would have NUM_ITERS=mio_num_relax, and we'd store the sum of intpnt_iter and simplex_iter as additional information.

@rileyjmurray As far as I can tell, all three of `intpnt_iter`, `simplex_inter`, and `mio_num_relax` are mutually exclusive.
In the discrete case, the simplex/interior point iterations carry a `mio_` prefix, i.e., `mio_intpnt_iter` and `mio_simplex_iter`. So the current implementation does exactly what you describe. We could think about not including the extra stats for continuous problems, but I'd rather have them available with a value of 0.

Paulnkk:
@phschiele @rileyjmurray I am okay with both approaches, and I can make changes if necessary; how should we proceed here? I will update the web doc if you agree that we leave it as it is and/or make changes.

phschiele:
If my above explanation suffices (@rileyjmurray please jump in if it didn't), we can add the web docs and merge them here.

rileyjmurray:
@phschiele your explanation suffices indeed; thanks for double checking!

Paulnkk:
@phschiele @rileyjmurray Since we changed `attr` and not a solver-specific input parameter (input parameter configs for solvers is mentioned here: [Solver options](https://www.cvxpy.org/tutorial/advanced/index.html#setting-solver-options)), I am not sure where the best place is to make the change in the documentation.

Paulnkk:
@phschiele @rileyjmurray I added documentation here: [Add docu](https://github.com/cvxpy/cvxpy/pull/2137/commits/36f8daa1e6de419a3d34bd9bbbada72d7000ac64).
Since there is no solver-specific section about the `solver_stats`, I left a note in that section.

rileyjmurray (approved):

rileyjmurray:
Thanks, @Paulnkk !

PR #2141, opened 2023-06-02
Human-readable SolverStats

phschiele:
## Description
I wanted to add a human-readable string representation to `SolverStats`. Using a `dataclass`, this comes for free.

The output for `extra_stats` depends on the solver. For example, ECOS and SCS give quite detailed information, other solvers point to the solution object or provide no further information.  If you think it is too unwieldy, we can remove `extra_stats` from the string. 

Before:
image

After:
image

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -1517,10 +1517,10 @@ class SolverStats:
         """
         return cls(
             solver_name,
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
Nice idea!

Issue #1158, opened 2020-10-27
cvxpy throws error when Mosek solver encounters time limit


wkschwartz:
This is #735 but for Mosek: cvxpy crashes when Mosek hits a user-set time limit rather than providing partial solution information or giving an informative error.

# To Reproduce

```python
import cvxpy
x = cvxpy.Variable(1)
cvxpy.Problem(
    cvxpy.Maximize(x),
    [x <= 1]
).solve(
    solver=cvxpy.MOSEK,
    mosek_params={'MSK_DPAR_OPTIMIZER_MAX_TIME': 0.0}
)
```

# Expected behavior
Maybe return `cvxpy.settings.USER_LIMIT`, but it seems there's a lengthy discussion of desired behavior over at #735. I'm not sure if partial solution information is available for MIPs, but it should be available if so. If there's never partial solution information, then raising a new `cvxpy.exceptions.TimeLimitError` exception would be best, especially if that exception object could carry information such as those that would normally be carried by the `SolverStats` object.

# Output
From the session above:

```python-traceback
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../cvxpy/problems/problem.py", line 396, in solve
    return solve_func(self, *args, **kwargs)
  File ".../cvxpy/problems/problem.py", line 754, in _solve
    self.unpack_results(solution, solving_chain, inverse_data)
  File ".../cvxpy/problems/problem.py", line 1070, in unpack_results
    self.unpack(solution)
  File ".../cvxpy/problems/problem.py", line 1030, in unpack
    raise ValueError("Cannot unpack invalid solution: %s" % solution)
ValueError: Cannot unpack invalid solution: Solution(status=UNKNOWN, opt_val=nan, primal_vars={}, dual_vars={}, attr={'solve_time': 0.00013589859008789062})
```

# Version
 - **OS**: macOS 10.15.7
- **Python**: 3.7.9
 - **CVXPY Version**: 1.1.7

SteveDiamond:
We would address this currently by returning the ``USER_LIMIT`` status. That can involve returning a solution.

Paulnkk:
@phschiele @SteveDiamond I am currently taking a look at the issue here and do I understand it correctly if the solver hits a timelimit (timelimit is set by the user) and no solution was generated (presolve optimal solution, inaccurate solution etc.) a timelimit error is returned to the user ?

Transurgeon:
[...]

PR #2144, opened 2023-06-08
Mosek: Draft for handling time limits

Paulnkk:
## Description
I introduced conditions that are triggered if a max time limit is reached and throws a user warning.

Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/1158

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


Paulnkk:
If the time limit triggers and a solver error is raised, the user can at least still see the best solution generated by mosek. 

Currently the question is whether a return as specified above is desired. Furthermore I will add unit tests and fix conflicts with the existing ones.

Paulnkk:
@aszekMosek Hey, I opened this Mosek-related PR and wanted to pick it up again. Does this PR generally satisfy the purpose of the issue mentioned above (https://github.com/cvxpy/cvxpy/issues/1158)?

aszekMosek:
If the solver terminated because of time limit then the return value of ``task.optimize()`` will be ``rescode.trm_max_time ``  (https://docs.mosek.com/latest/pythonapi/optimizer-task.html#mosek.task.optimize). I think that will be a much more simple and reliable way to check for this situation, rather than the more complicated logic involving prosta etc., which can be much more involved and depend on which solver is used, etc.

aszekMosek:
On second thought, I don't really understand what is going on here: if the solution status is ``unknown`` then CVXPY normally returns that the solver failed. Why should it return something else if the solver hit the time limit? How is that different than any other ``unknown`` situation?

If the solver thinks it has something interesting, then it will return some other status, and then CVXPY can make the solution available.

If the solver has nothing, then it has nothing, regardless if it is because of time limit or anything else.

The only thing I think should be done is that if the termination code is ``rescode.trm_max_time`` then set some flag/status informing the user about it, if possible. 

Note that there already was a PR recently which allows the user to say that they want to fetch anything, regardless of solution status being ``unknown``.

Paulnkk:
@aszekMosek, thanks for your suggestions, and sorry for getting back on the PR that late. I inserted a condition to throw a user warning if the time limit is reached within the solver (please let me know if the user warning satisfies what you mentioned as a flag). Benchmarks changes which may be related to the case that `task.optimize()` is called a second time.

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -296,6 +296,11 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
         # Optimize the Mosek Task, and return the result.
         task.optimize()
 
+        if task.optimize() == mosek.rescode.trm_max_time:
+            warnings.warn(
+                "Optimization terminated by time limit; solution may be imprecise or absent.",
+            )
+
```
phschiele, line 303:
```suggestion
        rescode = task.optimize()

        if rescode == mosek.rescode.trm_max_time:
            warnings.warn(
                "Optimization terminated by time limit; solution may be imprecise or absent.",
            )

```

phschiele (commented):
There is an empty `test.py` file floating around. Made a suggestion to only call optimize once.

Paulnkk:
> There is an empty `test.py` file floating around. Made a suggestion to only call optimize once.

@phschiele Removed test file here, leftover from debugging ...

phschiele (approved):
Thanks for the swift update @Paulnkk!

SteveDiamond (approved):

PR #2150, opened 2023-06-19
Makes vec_to_upper_tri top level import

phschiele:
## Description
This PR makes `vec_to_upper_tri` an atom accessible as a top-level import, i.e., part of our public API.

TODO: Add docs (Inline and web)

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/upper_tri.py
+++ b/cvxpy/atoms/affine/upper_tri.py
@@ -137,15 +137,9 @@ def vec_to_upper_tri(expr, strict: bool = False):
     # compute expr3 = reshape(expr2, shape=(n, n)).T
     #   expr3 is the matrix formed by reading length-n blocks of expr2,
     #   and letting each block form a row of expr3.
-    P_rows = []
-    P_row = 0
-    for mat_row in range(n):
-        entries_in_row = n - mat_row
-        if strict:
-            entries_in_row -= 1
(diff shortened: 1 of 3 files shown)


cvxpy/atoms/affine/upper_tri.py
```diff
@@ -111,6 +112,12 @@ def graph_implementation(
 
 def vec_to_upper_tri(expr, strict: bool = False):
     expr = Expression.cast_to_const(expr)
+
+    if not expr.is_vector():
+        raise ValueError("The input must be a vector.")
+    if expr.ndim != 1:
```
PTNobel, line 118: Are we sure we don't want to error here? If I were a user I might expect a 500 x 6 matrix input to become a 500 x 3 x 3 tensor output.

phschiele: That case would already fail `not expr.is_vector()` above. This is basically to handle row vectors.

PTNobel: Ah of course. Thanks.

rileyjmurray:
The current changes look good to me. I think it'd be okay to merge once we have source and web docs.

phschiele:
Subsumed by #2238

PR #2155, opened 2023-06-25
Deprecate NonPos in favor of NonNeg

rileyjmurray:
I recently opened a GitHub Discussion about awkwardness with the NonPos constraint class: https://github.com/cvxpy/cvxpy/discussions/2153. This PR offers the solution where we deprecate NonPos in favor of NonNeg. It includes a deprecation warning, updates to tests, and changes to canonicalization code paths. NonPos constraints are still supported, but we never construct those constraints outside of our testing code.

### DIFF
--- a/cvxpy/constraints/nonpos.py
+++ b/cvxpy/constraints/nonpos.py
@@ -54,7 +54,7 @@ class NonPos(Constraint):
     """
 
     def __init__(self, expr, constr_id=None) -> None:
-        warnings.warn(NonPos.DEPRECATION_MESSAGE)
+        warnings.warn(NonPos.DEPRECATION_MESSAGE, DeprecationWarning)
         super(NonPos, self).__init__([expr], constr_id)
         if not self.args[0].is_real():
             raise ValueError("Input to NonPos must be real.")
--- a/cvxpy/constraints/nonpos.py
+++ b/cvxpy/constraints/nonpos.py
@@ -19,18 +19,22 @@ import numpy as np
 # Only need Variable from expressions, but that would create a circular import.
 from cvxpy.constraints.constraint import Constraint
 from cvxpy.utilities import scopes
+import warnings
 
 
 class NonPos(Constraint):
-    """A constraint of the form :math:`x \\leq 0`.
+    """An inequality constraint of the form :math:`x \\leq 0`.
 
-    The preferred way of creating a ``NonPos`` constraint is through
-    operator overloading. To constrain an expression ``x`` to be non-positive,
-    simply write ``x <= 0``; to constrain ``x`` to be non-negative, write
-    ``x >= 0``. The former creates a ``NonPos`` constraint with ``x``
-    as its argument, while the latter creates one with ``-x`` as its argument.
-    Strict inequalities are not supported, as they do not make sense in a
-    numerical setting.
+    The preferred way of creating an inequality constraint is through
+    operator overloading. To constrain an expression ``x`` to be nonpositive,
+    write ``x <= 0``; to constrain ``x`` to be nonnegative, write ``x >= 0``.
+
+    Dual variables associated with this constraint are nonnegative, rather
+    than nonpositive. As such, dual variables to this constraint belong to the
+    polar cone rather than the dual cone.
+
+    Note: strict inequalities are not supported, as they do not make sense in
+    a numerical setting.
 
     Parameters
     ----------
@@ -39,7 +43,17 @@ class NonPos(Constraint):
     constr_id : int
         A unique id for the constraint.
     """
+
+    DEPRECATION_MESSAGE = """
+    Explicitly invoking NonPos(expr) to a create a constraint is deprecated.
+    Please use operator overloading or NonNeg(-expr) instead.
+    
+    Sign conventions on dual variables associated with these constraints may
+    change in the future.
+    """
+
     def __init__(self, expr, constr_id=None) -> None:
+        warnings.warn(NonPos.DEPRECATION_MESSAGE)
         super(NonPos, self).__init__([expr], constr_id)
         if not self.args[0].is_real():
             raise ValueError("Input to NonPos must be real.")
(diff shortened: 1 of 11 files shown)


cvxpy/reductions/complex2real/canonicalizers/__init__.py
```diff
@@ -91,8 +91,6 @@
     Constant: constant_canon,
     Parameter: param_canon,
     Inequality: inequality_canon,
-    NonPos: inequality_canon,
-    NonNeg: inequality_canon,
```
rileyjmurray, line 95: It made no sense for NonPos and NonNeg to map to inequality_canon. It turns out that our test code never even attempted to use these functions. We do have ``nonpos_canon`` and ``nonneg_canon`` functions in the complex2real code path, but it looks like they are currently sitting unused! I can either remove those functions as well, or add these lines back while having them point to the appropriate (currently unused) functions.

rileyjmurray: It looks like removing these is the appropriate thing to do. We already have a test that requires that NonPos and NonNeg raise errors when provided with complex inputs.

cvxpy/constraints/nonpos.py
```diff
@@ -156,7 +180,6 @@ class Inequality(Constraint):
     """
     def __init__(self, lhs, rhs, constr_id=None) -> None:
         self._expr = lhs - rhs
-        # TODO remove this restriction.
```
rileyjmurray, line 159: I removed this TODO since I don't think adding this "feature" would actually be a good idea.

cvxpy/reductions/complex2real/canonicalizers/inequality_canon.py
```diff
@@ -39,27 +39,3 @@ def inequality_canon(expr, real_args, imag_args, real2imag):
...
-    # Created by Riley; copied nonpos_canon code, and replaced imag_cons'
-    # call to "NonPos" with a call to "NonNeg".
-    if imag_args[0] is None:
-        return [expr.copy(real_args)], None
-
-    imag_cons = [NonNeg(imag_args[0], constr_id=real2imag[expr.id])]
-    if real_args[0] is None:
-        return None, imag_cons
-    else:
-        return [expr.copy(real_args)], imag_cons
```
rileyjmurray, line 65: This was dead code.

cvxpy/reductions/complex2real/complex2real.py
```diff
@@ -123,11 +120,11 @@ def invert(self, solution, inverse_data):
                         imag_id = inverse_data.real2imag[cid]
                         dvars[cid] = 1j*solution.dual_vars[imag_id]
                     # All cases that follow are for complex-valued constraints:
-                    #   1. check inequality / equality constraints.
+                    #   1. check equality constraints.
                     #   2. check PSD constraints.
                     #   3. check if a constraint is known to lack a complex dual implementation
                     #   4. raise an error
-                    elif isinstance(cons, (Equality, Zero, Inequality, NonNeg, NonPos)):
+                    elif isinstance(cons, (Equality, Zero)):
```
rileyjmurray, line 127: The ``Inequality, NonNeg, NonPos`` cases all amounted to dead code.

cvxpy/constraints/nonpos.py
rileyjmurray: To explain my position here, Python doesn't allow inequality-based comparisons between complex numbers. MATLAB allows it but disregards the imaginary part when making the comparison. There's no viable course of action where we can allow complex-valued inequality constraints and not end up confusing the heck out of some people. It's better not to support this and fail loudly than to risk hidden bugs for a tiny bit of syntactic sugar.

cvxpy/constraints/nonpos.py
```diff
@@ -39,7 +44,17 @@ class NonPos(Constraint):
     constr_id : int
         A unique id for the constraint.
     """
+
+    DEPRECATION_MESSAGE = """
+    Explicitly invoking "NonPos(expr)" to a create a constraint is deprecated.
```
SteveDiamond, line 49: Why is NonPos deprecated?

SteveDiamond:
I don't really understand the approach of deprecating NonPos. Isn't the fix to change the duals for NonPos?

cvxpy/constraints/nonpos.py
rileyjmurray: The semantics of dual variables for NonPos are inconsistent with those of other cones we support. I think we should mark it as deprecated for a few releases (e.g., 1.4 and 1.5) before we change the dual variable semantics to be consistent with my proposal here: https://github.com/cvxpy/cvxpy/discussions/2153#discussioncomment-6272802.

Now, we technically never specified the dual variable convention in a formal API, so one could argue that the appropriate fix is just to adopt our preferred convention right away. I'm just trying to follow the spirit of semantic versioning by warning the user before a potentially breaking change. What are your thoughts on this?

rileyjmurray:
> I don't really understand the approach of deprecating NonPos. Isn't the fix to change the duals for NonPos?

@SteveDiamond I replied directly in-thread to your other comment.

cvxpy/constraints/nonpos.py
SteveDiamond: That sounds fine to me, thanks for clarifying. I was also thinking after that having both NonPos and NonNeg is a small but consistent source of bugs, for a very small gain in expressiveness. So it might be worth consolidating on one form. We don't have a negative semidefinite cone constraint class.

SteveDiamond (approved):
LGTM. We maybe should also check the mosek interface locally, since the mosek tests aren't run on PRs.

cvxpy/constraints/nonpos.py
rileyjmurray: I'd definitely be up for removing NonPos eventually. We can revisit whether we want to keep it once we're considering the 1.5 or 1.6 releases.

rileyjmurray:
> LGTM. We maybe should also check the mosek interface locally, since the mosek tests aren't run on PRs.

I can confirm that MOSEK tests pass locally.

rileyjmurray:
Failing benchmarks are unrelated to this PR. Merging!

Issue #2158, opened 2023-06-26
Import error with latest version of scipy installed


ehtec:
**Describe the bug**
When using this together with the latest version 1.11.0 of scipy (from pypi), you get an error when importing cvxpy:

```
File "/usr/local/lib/python3.10/site-packages/cvxpy/__init__.py", line 18, in <module>
    import cvxpy.interface.scipy_wrapper
  File "/usr/local/lib/python3.10/site-packages/cvxpy/interface/scipy_wrapper.py", line 37, in <module>
    method = getattr(spmatrix, method_name)
AttributeError: type object 'spmatrix' has no attribute '__div__'
```

**Version**
 - OS: Debian 11
 - CVXPY Version: 1.3.1

**Additional context**
I encountered this because my CI pipeline is failing since the latest version of scipy was released a few hours ago. It works with scipy 1.10.1 (the second-latest version).

phschiele:
@ehtec Thanks for reporting this! I was able to replicate the issue and found the following:

- As stated in the message, the error originates from [scipy_wrapper.py](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/interface/scipy_wrapper.py)
- Adding a `try ... except` around the loop in the file fixes all tests
- Indeed, so does commenting out all of the contents in the file
- I used the below snippet to understand the intended behavior under SciPy 1.10:

```py
import cvxpy as cp
import numpy as np
import scipy.sparse as sp

a = cp.Variable((1,3))
b = sp.coo_matrix(np.arange(3))
a + b
>>> Expression(AFFINE, UNKNOWN, (1, 3))
b + a
>>> Expression(AFFINE, UNKNOWN, (1, 3))
a.__add__(b)
>>> Expression(AFFINE, UNKNOWN, (1, 3))
b.__add__(a)
>>> NotImplemented
```

My thoughts are:
1. At the very least, we had some missing test coverage on the intended effect of these operations
2. If anything, we should probably `raise NotImplementedError` instead of `return NotImplemented`
3. The operation `b + a` does not fail because it is wrapped again by https://github.com/cvxpy/cvxpy/blob/8f3e2b95051d2fc0977735ebfebdd34b53295a13/cvxpy/expressions/expression.py#L46-L51, so the combination of the wrappers is likely not intended behavior

The files have been there for quite a long time, so maybe @cvxpy/maintainers can give some context.

niklasvm:
I can confirm we are experiencing the same issue since a few hours ago

[...]

PR #2159, opened 2023-06-26
Update scipy_wrapper

SteveDiamond:
## Description
Updates the scipy_wrapper as suggested in https://github.com/scipy/scipy/issues/18749
Issue link (if applicable): #2158 

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/interface/scipy_wrapper.py
+++ b/cvxpy/interface/scipy_wrapper.py
@@ -24,7 +24,8 @@ SPARSE_MATRIX_CLASSES = [
     sparse.coo_matrix, 
     sparse.bsr_matrix, 
     sparse.lil_matrix, 
-    sparse.dia_matrix
+    sparse.dia_matrix,
+    sparse.dok_matrix,
 ]
 BIN_OPS = ["__div__", "__mul__", "__add__", "__sub__",
            "__le__", "__eq__", "__lt__", "__gt__"]
--- a/cvxpy/interface/scipy_wrapper.py
+++ b/cvxpy/interface/scipy_wrapper.py
@@ -14,10 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from scipy.sparse import spmatrix
+from scipy import sparse
 
 from cvxpy.expressions import expression as exp
 
+SPARSE_MATRIX_CLASSES = [
+    sparse.csc_matrix, 
+    sparse.csr_matrix, 
+    sparse.coo_matrix, 
+    sparse.bsr_matrix, 
+    sparse.lil_matrix, 
+    sparse.dia_matrix
+]
 BIN_OPS = ["__div__", "__mul__", "__add__", "__sub__",
            "__le__", "__eq__", "__lt__", "__gt__"]
 
@@ -32,8 +40,8 @@ def wrap_bin_op(method):
             return method(self, other)
     return new_method
 
-
-for method_name in BIN_OPS:
-    method = getattr(spmatrix, method_name)
-    new_method = wrap_bin_op(method)
-    setattr(spmatrix, method_name, new_method)
+for cls in SPARSE_MATRIX_CLASSES:
+    for method_name in BIN_OPS:
+        method = getattr(cls, method_name)
+        new_method = wrap_bin_op(method)
+        setattr(cls, method_name, new_method)


SteveDiamond:
Please include #2160.

cvxpy/interface/scipy_wrapper.py
```diff
@@ -14,10 +14,18 @@
...
 
 from cvxpy.expressions import expression as exp
 
+SPARSE_MATRIX_CLASSES = [
+    sparse.csc_matrix, 
+    sparse.csr_matrix, 
+    sparse.coo_matrix, 
+    sparse.bsr_matrix, 
+    sparse.lil_matrix, 
+    sparse.dia_matrix
```
ivirshup, line 27:
```suggestion
    sparse.dia_matrix,
    sparse.dok_matrix,
```

I forgot `dok` in my comment

phschiele (approved):
LGTM. Let me summarize the understanding that I've gathered as of now (this might be obvious to some readers).

1. Returning `NotImplemented` from a binary operation (instead of raising `NotImplementedError`) is a signal that a certain operation is not implemented, but a fallback may be available (e.g., if `a.__eq__(b)` is not implemented, `b.__eq__(a)` might still be available). Only if all recovery attempts fail, an exception is being raised.
2. `scipy_wrapper` ensures that all binary operations fall back to the cvxpy implementation when `other` is a cvxpy expression.
3. This might not always be necessary, though, as SciPy may already provide this signal in their implementation, e.g., [here](https://github.com/scipy/scipy/blob/8e55aea98ee0a89e1404acec410aeb879aef33a7/scipy/sparse/_base.py#L678).

PTNobel:
> 1. Returning NotImplemented from a binary operation (instead of raising NotImplementedError) is a signal that a certain operation is not implemented, but a fallback may be available (e.g., if a.__eq__(b) is not implemented, b.__eq__(a) might still be available). Only if all recovery attempts fail, an exception is being raised.

Correct.

> 3.    This might not always be necessary, though, as SciPy may already provide this signal in their implementation, e.g., [here](https://github.com/scipy/scipy/blob/8e55aea98ee0a89e1404acec410aeb879aef33a7/scipy/sparse/_base.py#L678).

Do we know how recent that code was? I wonder if we even need this hotfix nowadays...It was added in 2013 per git blame, and it seems this originated as: https://github.com/cvxpy/cvxpy/commit/f8b8e7ed05bfd57b11f3234bbc1ccbf8d87b56bb#diff-5b5c7742b5ff35de17876c5a718a7f6f1aeb774037c2e73d3c39bb7e937607c3
which doesn't seem relevant to this file...

In any case, I wonder if there is still a need. Especially on the `__div__, __add__, __sub__, __mul__` arguments.

Testing with
```python
class C:
    pass
    
getattr(sp.dok_matrix(np.zeros((5, 5))), func)(C())
```

only errored for `__ge__`, `__le__`, `__lt__`, `__gt__` and returned `NotImplemented` for the rest.

phschiele:
> > 1. Returning NotImplemented from a binary operation (instead of raising NotImplementedError) is a signal that a certain operation is not implemented, but a fallback may be available (e.g., if a.**eq**(b) is not implemented, b.**eq**(a) might still be available). Only if all recovery attempts fail, an exception is being raised.
> 
> Correct.
> 
> > 3. This might not always be necessary, though, as SciPy may already provide this signal in their implementation, e.g., [here](https://github.com/scipy/scipy/blob/8e55aea98ee0a89e1404acec410aeb879aef33a7/scipy/sparse/_base.py#L678).
> 
> Do we know how recent that code was? I wonder if we even need this hotfix nowadays...It was added in 2013 per git blame, and it seems this originated as: [f8b8e7e#diff-5b5c7742b5ff35de17876c5a718a7f6f1aeb774037c2e73d3c39bb7e937607c3](https://github.com/cvxpy/cvxpy/commit/f8b8e7ed05bfd57b11f3234bbc1ccbf8d87b56bb#diff-5b5c7742b5ff35de17876c5a718a7f6f1aeb774037c2e73d3c39bb7e937607c3) which doesn't seem relevant to this file...
> 
> In any case, I wonder if there is still a need. Especially on the `__div__, __add__, __sub__, __mul__` arguments.
> 
> Testing with
> 
> ```python
> class C:
>     pass
>     
> getattr(sp.dok_matrix(np.zeros((5, 5))), func)(C())
> ```
> 
> only errored for `__ge__`, `__le__`, `__lt__`, `__gt__` and returned `NotImplemented` for the rest.

Good point. Indeed, I have checked that, based on our tests, we do not need any patch with recent SciPy versions. However, because I neither know that our tests cover all cases, nor checked with older SciPy versions, I was hesitant to suggest removing the patching entirely.

PTNobel:
> However, because I neither know that our tests cover all cases, nor checked with older SciPy versions, I was hesitant to suggest removing the patching entirely.

Could we deprecate the patch in 1.4 and see if it is still needed? We could raise a warning and asks people to contact us if they hit it.

rileyjmurray:
@PTNobel I'd support deprecating in 1.4.

PR #2171, opened 2023-07-05
Adding functionality to cvxpy/tests: verifying stationarity of the Lagrangians as part of KKT conditions

aryamanjeendgar:
## Description
This PR adds a new method, `check_stationarity_lagrangian`, which:
1. Tries to construct the Lagrangian corresponding to a given problem
2. Check if it is stationary by using the existing `.grad` facility.
3. Can also handle the case of the user passing in implicit constraints via flags (such as `diag, PSD, NSD, nonpos` etc. by checking dual cone membership of the computed gradient instead of stationarity)

I have added in tests for this new method which live in `cvxpy/tests/test_KKT.py` --- where all of the standard test classes from `SolverTestHelpers` make an appearance --- the KKT conditions are verified in full for all of the tests (with the exception of MI-programs)  in the `solver_test_helpers.py` file.

I haven't added documentation for this method yet since it just lives within the `SolverTestHelper` class as of now --- we can probably write a full class for verifying the KKT conditions after proper support for verifying dual feasibility has been added (via the introduction of the `dual` method on `Constraint` objects).

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/tests/test_KKT.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def test_lp_5(self):
+        # LP with redundant constraints
+        sth = STH.lp_5()
+        sth.solve(solver='ECOS')
+        sth.verify_objective(places=4)
+        sth.check_primal_feasibility(places=4)
+        sth.check_complementarity(places=4)
+        sth.check_dual_domains(places=4)
+        sth.check_stationary_lagrangian(places=2)
```
rileyjmurray, line 36: You don't need to use ``verify_objective``, or ``verify_primal_values``, or ``verify_dual_values`` here. Just check the KKT conditions.

cvxpy/tests/test_KKT.py
```diff
@@ -0,0 +1,303 @@
...
+        sth.solve(solver='ECOS')
+        sth.verify_primal_values(places=4)
+        sth.verify_objective(places=4)
+        sth.check_complementarity(places=4)
+        sth.verify_dual_values(places=4)
+        sth.check_stationary_lagrangian(places=4)
+        return sth
+
+
+class StandardTestSOCPs(BaseTest):
```
rileyjmurray, line 52: Update these tests to only check the KKT conditions.

cvxpy/tests/test_KKT.py
rileyjmurray, line 1: Naming conventions of classes: test classes start with ``Test...``. The convention you're using here is like for classes in ``solver_test_helpers.py``, which is not actually dispatched by a test runner.

cvxpy/tests/test_KKT.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+class StandardTestFlags(BaseTest):
+    """
+    A class with tests for testing out the incorporation of
+    implicit constraints within `check_stationarity_lagrangian`
+    """
+
+    @staticmethod
+    def tf_1() -> STH.SolverTestHelper:
```
rileyjmurray, line 183: What's the ``tf`` stand for here?

cvxpy/tests/test_KKT.py
```diff
@@ -0,0 +1,303 @@
...
+        sth.solve(solver='SCS')
+        sth.verify_objective(places=2)
+        sth.check_primal_feasibility(places=2)
+        sth.check_complementarity(places=2)
+        sth.check_dual_domains(places=2)
+        sth.check_stationary_lagrangian(places=2)
+        return sth
+
+    def test_sdp_2(self):
+        # places is set to 3 rather than 4, because analytic solution isn't known.
```
rileyjmurray, line 128: This is actually a great example where we don't need to know the analytic solution. Just check the KKT conditions to the usual level of accuracy.

cvxpy/tests/test_KKT.py
rileyjmurray: Separately: there are no tests for exponential cone constraints here.

rileyjmurray (changes requested):
See my individual comments.

aryamanjeendgar:
@rileyjmurray I think I should've addressed your suggested changes with this commit, please take a look whenever you get time!

rileyjmurray:
Failing tests are unrelated to this PR.

cvxpy/tests/test_KKT.py
rileyjmurray: The nature of the flag isn't obvious from the name itself. (Resolved with the latest naming convention I committed to your branch.)

rileyjmurray (approved):

PR #2172, opened 2023-07-07
Deprecates `scipy_wrapper.py`

PTNobel:
## Description

We automatically detect when `scipy_wrapper.py` actually does something and then raises a warning telling people to contact us. The hope is we will hear if there are cases not covered by our test suite that require scipy_wrapper to do something.

This might have a slight performance regression in cases where SciPy does expensive compute that we throw away, but those cases will result in warnings so we should hear about them and can correct them.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/interface/scipy_wrapper.py
+++ b/cvxpy/interface/scipy_wrapper.py
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import warnings
+
 from scipy import sparse
 
 from cvxpy.expressions import expression as exp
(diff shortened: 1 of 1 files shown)


SteveDiamond:
Won't they hit the warning with an old scipy?

PTNobel:
I need to run some experiments and confirm, but in my investigation here: https://github.com/cvxpy/cvxpy/pull/2159#issuecomment-1608032067 it seemed like some 2013 commits in SciPy were correctly returning Not implemented... I couldn't find the original motivation for adding this.

Also do we think people are updating CVXPY but not SciPy?

I'm happy to run some experiments on older SciPy (maybe releases from 5 years ago?) And see if the test suite errors this way.

SteveDiamond (approved):
LGTM

PTNobel:
Quick comment: with the SciPy 1.3 (released around May 2019 so about 4 years ago) version we test in CI, there were no warnings. I think we could consider bumping our min SciPy to 1.3 if we're not able to test anything older.

Issue #2182, opened 2023-07-12
Complex-valued vector expression dropping imaginary part


tikuma-lsuhsc:
**Describe the bug**

Using complex-valued constants with different precision drops the imaginary part from the resulting cvxpy expression.

**To Reproduce**

```python
import cvxpy as cp
import numpy as np

N = 100
M = 5

x = np.ones(N, "complex64")
A = np.ones((N, M), "complex128")

theta = cp.Variable(M, complex=True)
A @ theta - x
```

This produces a warning:

```
...\lib\site-packages\cvxpy\interface\numpy_interface\ndarray_interface.py:53: ComplexWarning: Casting complex values to real discards the imaginary part
  return result.astype(numpy.float64)
```

Using this expression to form a problem (obviously) yields an incorrect solution.

**Expected behavior**

Maintain the complex-valued expression at the highest precision of the constants.

**Version**
 - OS: Win10
 - CVXPY Version: v1.3.2

SteveDiamond:
That's surprising we missed this case in our tests. I pushed a fix in #2183 with expanded test coverage.

tikuma-lsuhsc:
👍 Nice. Thank you, and keep up the great work! Love this package.

PR #2183, opened 2023-07-12
Don't cast complex numpy dtypes to float64

SteveDiamond:
## Description
An oversight in the code and testing led to conversion of np.complex64 and np.complex128 dtypes to float64.

Issue link (if applicable): #2182 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/interface/numpy_interface/ndarray_interface.py
+++ b/cvxpy/interface/numpy_interface/ndarray_interface.py
@@ -22,6 +22,9 @@ import scipy.sparse
 
 from .. import base_matrix_interface as base
 
+# Possible types for complex data.
+COMPLEX_TYPES = [complex, numpy.complex64, numpy.complex128]
+
 
 class NDArrayInterface(base.BaseMatrixInterface):
     """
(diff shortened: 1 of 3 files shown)


cvxpy/interface/numpy_interface/ndarray_interface.py
```diff
@@ -47,7 +50,7 @@ def const_to_matrix(self, value, convert_scalars: bool = False):
             result = numpy.asarray(value).T
         else:
             result = numpy.asarray(value)
-        if result.dtype in [complex, numpy.float64]:
+        if result.dtype in [numpy.float64] + COMPLEX_TYPES:
```
PTNobel, line 53: Should we convert `np.complex64` to `np.complex128`? That's the equivalent of converting float32 to float64.

PTNobel (approved):

phschiele (approved):
Looks good, @PTNobel's comment seems reasonable.

cvxpy/interface/numpy_interface/ndarray_interface.py
SteveDiamond: I'm not sure that casting from float32 to float64 is a good thing. All these interfaces were written years ago when performance was not a particular concern. I'd rather not cast complex64 to complex128 if it's not strictly necessary.

PTNobel: Sounds good. I was just surprised by the asymmetry.

PR #2186, opened 2023-07-16
NumPy backend

Transurgeon:
## Description
This PR introduces a new backend written in pure NumPy following the dict-based approach of the SciPy backend. 
Because of the lack of sparsity, the implementations are slightly easier to follow and can serve as a reference for other implementations. 
Note that the output is still transformed to a sparse matrix format as the downstream code doesn't yet accept dense matrices.
In the future, we can change this to keep the dense/sparse format coming out of the canonicalization. 

This PR has been in the works and discussed with @phschiele. 
Link: https://github.com/Transurgeon/cvxpy/pull/1
## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
Great work! In terms of CI issues, looks like ``csc_array`` is a new scipy feature, so should drop that for backwards compatibility. Also @Transurgeon have you installed the pre-commit?

Transurgeon:
@SteveDiamond no I didn't have pre-commit installed for this PR, I just downloaded it now. 
I'll look into removing csc arrays.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -826,6 +837,267 @@ def get_param_tensor(self, shape: tuple[int, ...], parameter_id: int) \
...
+        indices = np.arange(shape[0]) * shape[0] + np.arange(shape[0])
+
+        lhs = np.zeros(shape=(1, np.prod(shape)))
+        lhs[0, indices] = 1
+        def func(x):
+            return lhs @ x
+
+        return view.accumulate_over_variables(func, is_param_free_function=True)
+
+    def conv(self, lin: LinOp, view: NumpyTensorView) -> NumpyTensorView:
```
SteveDiamond, line 1005: Should these functions have comments?

cvxpy/lin_ops/canon_backend.py
```diff
@@ -927,7 +1199,67 @@ def create_new_tensor_view(self, variable_ids: set[int], tensor: Any, is_paramet
         pass  # noqa
 
 
-class ScipyTensorView(TensorView):
+class DictTensorView(TensorView, ABC):
```
SteveDiamond, line 1202: What is this class for?

SteveDiamond:
Looks generally good. We try to have comments for all functions. I'm sure copilot can help with this :)

cvxpy/lin_ops/canon_backend.py
Transurgeon: the comments are currently located in the unit-tests implementations. It might be good to add them here also, to help differentiate between the scipy and numpy backends.

cvxpy/lin_ops/canon_backend.py
Transurgeon: This class was introduced to remove redundant code. It takes care of the two layers of dictionaries meant to store information about the tensor (ie: variable ID, constant ID, or param ID) which is common between the numpy and scipy backend. I'll add some further comments explaining this in further detail.

Transurgeon:
Ok noted. I just made two PRs on my branch, one for refactoring the unit-tests and the other one for the comments. 
As most comments are already located in ``PythonCanonBackend``, and ``ScipyCanonBackend`` doesn't have many comments. I'll target the methods that have changed a lot (such as ``conv`` and the toeplitz matrix, ``rmul`` and ``kron``)

phschiele:
I mostly only added the docstring to the base class methods, but [PEP 257](https://www.python.org/dev/peps/pep-0257/) is quite clear about it:

> If a class subclasses another class and its behavior is mostly inherited from that class, its docstring should mention this and summarize the differences. Use the verb "override" to indicate that a subclass method replaces a superclass method and does not call the superclass method; use the verb "extend" to indicate that a subclass method calls the superclass method (in addition to its own behavior).

We can go ahead and point to the base implementation, highlighting differences.

phschiele (approved):
Looks good, made a few tiny changes, please have a look @Transurgeon.

regarding `csc_array`: I think @SteveDiamond was referring to the difference between `csc_array` and `csc_matrix`. Right now, the caller of the backend expects a csc matrix, not a csr or coo matrix. The C++ backend also returns csc.

About the docs: I suggest merging this relatively big PR as is. We are working on the docs in a separate branch, which will add the docstrings to both backends in a way that avoids duplication and only highlights the differences.
We discussed whether it makes sense to build an API reference for the backend using sphinx. We would add that this is for internal documentation and does not suggest that this is a public API.

Transurgeon:
LGTM, I am not sure why the tests are failing. It seems to be something unrelated to the changes you've made.

phschiele:
Indeed, CI failures were unrelated! Merging.

Issue #2191, opened 2023-07-20
PowCone3D compilation fails when alpha is a scalar


rileyjmurray:
@aryamanjeendgar shared the following code snippet with me:
```
import cvxpy as cp
x = cp.Variable(shape=(3,))
cons = [cp.PowCone3D(x[0], x[1], x[2], 0.75)]
obj = cp.Minimize(cp.norm(x))
prob = cp.Problem(obj, cons)
prob.solve()
```
This raises an error 
```
  File "/Users/riley/Documents/cvxpydev/cvxpy/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py", line 86, in __init__
    p3d = np.concatenate([c.alpha.value for c in constr_map[PowCone3D]]).tolist()
  File "<__array_function__ internals>", line 200, in concatenate
```
The source of the error is that the fourth argument to ``PowCone3D`` is a pure scalar, rather than something that's 1d-array-like. If we change the code to use
```
cons = [cp.PowCone3D(x[0], x[1], x[2], [0.75])]
```
then everything works fine.

We should update the ``PowCone3D`` constructor to automatically cast scalar inputs for the final argument to 1d arrays of size 1. There's no need to cast other arguments to 1d arrays; just ``alpha`` is fine.

@aryamanjeendgar please make a PR to this effect when you're able.

rileyjmurray:
Resolved by #2192

PR #2192, opened 2023-07-21
`PowCone3D` fix

aryamanjeendgar:
## Description
Fixes small issue with `PowCone3D` when passing in naked floats as `alpha`.
Issue link (if applicable): [#2191](https://github.com/cvxpy/cvxpy/issues/2191)

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/constraints/power.py
+++ b/cvxpy/constraints/power.py
@@ -50,12 +50,21 @@ class PowCone3D(Constraint):
                 raise ValueError('All arguments must be affine and real.')
         alpha = Expression.cast_to_const(alpha)
         if alpha.is_scalar():
-            alpha = cvxtypes.promote()(alpha, self.x.shape)
+            if self.x.shape:
+                alpha = cvxtypes.promote()(alpha, self.x.shape)
+            else:
+                # when `alpha` is a naked float, it has to be cast into a
+                # 1-D array to be compatible with downstream (vectorized)
+                # processing
+                alpha = cvxtypes.promote()(alpha, (1,))
         self.alpha = alpha
         if np.any(self.alpha.value <= 0) or np.any(self.alpha.value >= 1):
             msg = "Argument alpha must have entries in the open interval (0, 1)."
             raise ValueError(msg)
-        arg_shapes = [self.x.shape, self.y.shape, self.z.shape, self.alpha.shape]
+        if alpha.shape == (1,):
+            arg_shapes = [self.x.shape, self.y.shape, self.z.shape, ()]
+        else:
+            arg_shapes = [self.x.shape, self.y.shape, self.z.shape, self.alpha.shape]
         if any(arg_shapes[0] != s for s in arg_shapes[1:]):
             msg = ("All arguments must have the same shapes. Provided arguments have"
                    "shapes %s" % str(arg_shapes))
--- a/cvxpy/tests/test_constraints.py
+++ b/cvxpy/tests/test_constraints.py
@@ -302,22 +302,14 @@ class TestConstraints(BaseTest):
         Simple test case with scalar AND vector `alpha`
         inputs to `PowCone3D`
         """""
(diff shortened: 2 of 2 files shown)


rileyjmurray:
@aryamanjeendgar can you please add a unit test to ``test_constraints.py``? You can add something right before or after [this test](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/tests/test_constraints.py#L275).

cvxpy/constraints/power.py
```diff
@@ -50,12 +50,21 @@ def __init__(self, x, y, z, alpha, constr_id=None) -> None:
...
         if np.any(self.alpha.value <= 0) or np.any(self.alpha.value >= 1):
             msg = "Argument alpha must have entries in the open interval (0, 1)."
             raise ValueError(msg)
-        arg_shapes = [self.x.shape, self.y.shape, self.z.shape, self.alpha.shape]
+        if alpha.shape == (1,):
+            arg_shapes = [self.x.shape, self.y.shape, self.z.shape, ()]
+        else:
+            arg_shapes = [self.x.shape, self.y.shape, self.z.shape, self.alpha.shape]
         if any(arg_shapes[0] != s for s in arg_shapes[1:]):
             msg = ("All arguments must have the same shapes. Provided arguments have"
```
rileyjmurray, line 69: It looks like this will raise an indexing error if ``alpha.shape = (1,)``, since ``s = arg_shapes[-1]`` is an empty tuple.

cvxpy/tests/test_constraints.py
```diff
@@ -297,6 +297,28 @@ def test_pow3d_constraint(self) -> None:
...
+                        x[0] + x[1] + 3 * x[2] >= 1.0,
+                        y_square <= 25]
+        obj = cp.Minimize(3 * x[0] + 2 * x[1] + x[2])
+        prob = cp.Problem(obj, cons)
+        prob.solve()
+        self.assertLessEqual(cons[0].residual, 1e-6)
+        self.assertLessEqual(cons[1].residual, 1e-6)
+        self.assertItemsAlmostEqual(cons[0].dual_value,
+                                    [np.array(1.98202141),
+                                     np.array(0.98202142), np.array(-2.05393575)])
```
rileyjmurray, line 320: This is excessive. The only reason for this test is to verify that canonicalization happens without error. It would suffice to only check the objective function value.

You should also explain where the test came from. I believe this is mostly copy-pasted from ``solver_test_helpers.py``.

cvxpy/constraints/power.py
aryamanjeendgar: Thanks for the catch! I'll make this `(1,1)`

cvxpy/tests/test_constraints.py
aryamanjeendgar: Yep! The test is `pcp_1` from `solver_test_helpers.py` with an additional `PowCone3D` constraint where all the arguments are scalar

cvxpy/constraints/power.py
aryamanjeendgar: Shouldn't `arg_shapes[-1]` just return `()`?

cvxpy/tests/test_constraints.py
rileyjmurray: Okay, but I still think that's an overkill test. You could just do a trivial projection problem instead. As a rule: tests should be as simple as possible while still hitting the codepaths we care about.

Separately, if you were to repurpose pcp_1, then your source code should include comments that clearly state as much.

cvxpy/constraints/power.py
rileyjmurray: Oh shit, I misunderstood, lol.

rileyjmurray:
Failing tests are unrelated to this PR.

PR #2204, opened 2023-08-05
Adding functionality to cvxpy/tests: verifying dual feasibility for `Constraint` classes as part of the KKT conditions

aryamanjeendgar:
This PR introduces the `dual_cone` methods for every `Constraint` class (which support dual variable recovery, which is all of them, with the exception of `PowConeND` and `OpRelEntrConeQuad`). A complementary `dual_violation` method if also implemented for ease in actually performing the feasibility checks.

These methods are used to update the existing `check_dual_domains` method in `solver_test_helpers.py` --- all the existing tests in `test_KKT.py` remain the same as they were when introduced as part of the `check_stationary_lagrangian` PR, and they all pass.

Again, I haven't added documentation for this method yet since it just lives within the `SolverTestHelper` class as of now.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/constraints/cones.py
+++ b/cvxpy/constraints/cones.py
@@ -18,6 +18,7 @@ import abc
 import typing
 
 from cvxpy.constraints.constraint import Constraint
+import cvxpy as cp
 
 
 class Cone(Constraint):
@@ -46,9 +47,27 @@ class Cone(Constraint):
     def __init__(self, args, constr_id=None) -> None:
         super(Cone, self).__init__(args, constr_id)
 
-    def dual_cone(self) -> typing.Type['Cone']:
+    def dual_cone(self, args: list['cp.Expression'] | None = None) -> typing.Type['Cone']:
+    # def dual_cone(self, args) -> typing.Type['Cone']:
+        """Method for modelling problems with the dual cone of `Cone`
+
+        If the user simply calls the method without any arguments, then
+        the dual cone to the current instance of `Cone` will be returned, else,
+        the user can also choose to freely model with the dual cone of `Cone`
+        by constraining aribitrary expressions in the same.
+
+        Will typically not be used stand-alone, and would be
+        indirectly accessed via `dual_residual` which will return
+        the violation of CVXPY's computed dual variables w.r.t. `Cone`'s
+        dual-cone"""
         raise NotImplementedError
 
     @property
     def dual_residual(self) -> float:
-        return self.dual_cone().residual
+        """Computes the residual (see Constraint.violation for a
+        more formal definition) for the dual cone of the current instance
+        of `Cone` w.r.t. the recovered dual variables
+
+        Primarily intended to be used for full-fledged KKT checks
+        """
+        return self.dual_cone(*self.dual_variables).residual
--- a/cvxpy/constraints/cones.py
+++ b/cvxpy/constraints/cones.py
@@ -1,0 +1,54 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import abc
+import typing
+
+from cvxpy.constraints.constraint import Constraint
+
+
+class Cone(Constraint):
+    """A base class for all conic constraints in CVXPY
+
+    These are special constraints imposing set membership in convex cones
+    CVXPY supports modelling using the following cones as of today:
+    - ExpCone
+    - SOC (Second Order Cone)
+    - PowCone3D
+    - PowConeND
+    - RelEntrConeQuad
+    - OpRelEntrConeQuad
+    - PSD/NSD
+
+    Parameters
+    ----------
+    args : list
+        A list of expression trees.
+    constr_id : int
+        A unique id for the constraint.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, args, constr_id=None) -> None:
+        super(Cone, self).__init__(args, constr_id)
+
+    def dual_cone(self) -> typing.Type['Cone']:
+        raise NotImplementedError
+
+    @property
+    def dual_residual(self) -> float:
+        return self.dual_cone().residual
--- a/cvxpy/constraints/cones.py
+++ b/cvxpy/constraints/cones.py
@@ -17,8 +17,8 @@ limitations under the License.
 import abc
 import typing
 
-from cvxpy.constraints.constraint import Constraint
 import cvxpy as cp
+from cvxpy.constraints.constraint import Constraint
 
 
 class Cone(Constraint):
--- a/cvxpy/constraints/cones.py
+++ b/cvxpy/constraints/cones.py
@@ -15,9 +15,7 @@ limitations under the License.
 """
 
 import abc
-import typing
 
-import cvxpy as cp
 from cvxpy.constraints.constraint import Constraint
 
 
@@ -47,7 +45,7 @@ class Cone(Constraint):
     def __init__(self, args, constr_id=None) -> None:
         super(Cone, self).__init__(args, constr_id)
 
-    def _dual_cone(self, args: list['cp.Expression'] | None = None) -> typing.Type['Cone']:
+    def _dual_cone(self, args):
         """Method for modelling problems with the dual cone of `Cone`
 
         If the user simply calls the method without any arguments, then
--- a/cvxpy/constraints/cones.py
+++ b/cvxpy/constraints/cones.py
@@ -47,8 +47,7 @@ class Cone(Constraint):
     def __init__(self, args, constr_id=None) -> None:
         super(Cone, self).__init__(args, constr_id)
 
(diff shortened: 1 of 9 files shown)


aryamanjeendgar:
Tagging @rileyjmurray for review!

rileyjmurray:
@aryamanjeendgar one small comment. Your ``dual_violation`` function accesses the ``residual`` field of a constraint returned by ``CON.dual_cone``. It would make more sense if you named this function ``dual_residual`` instead. So the ``ExpCone`` constraint would look like the following
```
def dual_residual(self):
        return ExpCone.dual_cone(*self.dual_variables).residual
```
On that note, I'd like to suggest that ``dual_residual`` functions have the ``@property`` decorator so that you can access ``con.dual_residual`` just like you'd access ``con.residual``.

rileyjmurray:
My intuition tells me that ``dual_cone`` and ``dual_residual`` should be defined in a base class somewhere. The problem is that ``dual_cone`` makes little sense for algebraic constraints and basically no sense for constraints like ``FiniteSet``. 

Perhaps this is the time to add a ``Cone`` interface class from which all conic constraints can inherit. If we did this, then we could even write a one-size-fits-all implementation for ``dual_residual``:
```
@property
def dual_residual
    return self.dual_cone(*self.args).residual
```

 @SteveDiamond @akshayka @phschiele, what do you three think about adding such a ``Cone`` interface class? (If you need more context please do ask.)

cvxpy/constraints/exponential.py
```diff
@@ -146,6 +146,13 @@ def save_dual_value(self, value) -> None:
         self.dual_variables[1].save_value(dv1)
         self.dual_variables[2].save_value(dv2)
 
+    @staticmethod
```
SteveDiamond, line 149: staticmethod doesn't seem right...

SteveDiamond:
> My intuition tells me that `dual_cone` and `dual_residual` should be defined in a base class somewhere. The problem is that `dual_cone` makes little sense for algebraic constraints and basically no sense for constraints like `FiniteSet`.
> 
> Perhaps this is the time to add a `Cone` interface class from which all conic constraints can inherit. If we did this, then we could even write a one-size-fits-all implementation for `dual_residual`:
> 
> ```
> @property
> def dual_residual
>     return self.dual_cone(*self.args).residual
> ```
> 
> @SteveDiamond @akshayka @phschiele, what do you three think about adding such a `Cone` interface class? (If you need more context please do ask.)

What about ``Cone`` as a subclass of ``Constraint``?

rileyjmurray:
Replying to @SteveDiamond :
> What about Cone as a subclass of Constraint?

I really like that approach. I wanted to get buy-in from others (e.g., you) before suggesting that @aryamanjeendgar go that route.

@aryamanjeendgar please create a new file ``cvxpy/constraints/cones.py`` that defines an abstract class ``Cone(Constraint)``. That class should have a (non-static!) ``dual_cone`` method that raises a ``NotImplementedError`` by default. It should also have a ``dual_residual`` property that uses my implementation suggested above. Once that class is defined, please update all existing conic constraints to formally subclass it.

phschiele:
I like the idea as well. As I've mentioned before, this could immediately use the common need of adding dual constraints, see, e.g., [here](https://github.com/cvxgrp/dsp/blob/f485eeefe089a4f897130929048159d4b4751ca0/dsp/cone_transforms.py#L380).

Would we also change the inheritance structure of existing cone constraints? I believe it could be considered nonbreaking, as `isinstance(c, Constraint)` would still be valid.

For reference, here is the current type hierarchy for `Constraint`:
image

rileyjmurray:
I would consider it a non-breaking change to change the inheritance
structure in this way

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,54 @@
...
+        A unique id for the constraint.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, args, constr_id=None) -> None:
+        super(Cone, self).__init__(args, constr_id)
+
+    def dual_cone(self) -> typing.Type['Cone']:
+        raise NotImplementedError
```
rileyjmurray, line 50: I think ``dual_cone`` should have the same signature as ``__init__`` as far as ``args`` go.

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @property
+    def dual_residual(self) -> float:
+        return self.dual_cone().residual
```
rileyjmurray, line 54: We need to unpack ``return self.dual_cone(*self.dual_variables).residual``.

cvxpy/constraints/power.py
```diff
@@ -139,17 +139,11 @@ def save_dual_value(self, value) -> None:
...
-    @staticmethod
-    def dual_cone(x, y, z, alpha: float):
-        return PowCone3D(x/alpha, y/(1-alpha), z, alpha)
-
-    @property
-    def dual_residual(self):
-        return PowCone3D.dual_cone(*self.dual_variables,
-                                   self.alpha).residual
+    def dual_cone(self):
+        return PowCone3D(self.x/self.alpha, self.y/(1-self.alpha), self.z, self.alpha)
```
rileyjmurray, line 143: This is just ... wrong? What's the semantic or mathematical meaning of the returned constraint?

rileyjmurray (changes requested):
These changes don't really make sense. They're a big departure from what we had before.

cvxpy/constraints/psd.py
```diff
@@ -83,3 +83,18 @@ def residual(self):
             return None
         min_eig = cvxtypes.lambda_min()(self.args[0] + self.args[0].T)/2
         return cvxtypes.neg()(min_eig).value
+
+    def dual_cone(self, *args):
+        """Implements the dual cone of the exponential cone
+        See Pg 85 of the MOSEK modelling cookbook for more information"""
```
rileyjmurray, line 89: This docstring is copy-pasted.

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,73 @@
...
+
+import cvxpy as cp
+from cvxpy.constraints.constraint import Constraint
+
+
+class Cone(Constraint):
+    """A base class for all conic constraints in CVXPY
+
+    These are special constraints imposing set membership in convex cones
+    CVXPY supports modelling using the following cones as of today:
```
rileyjmurray, line 28: modelling --> modeling

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,73 @@
...
+    constr_id : int
+        A unique id for the constraint.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, args, constr_id=None) -> None:
+        super(Cone, self).__init__(args, constr_id)
+
+    def dual_cone(self, args: list['cp.Expression'] | None = None) -> typing.Type['Cone']:
```
rileyjmurray, line 50: @aryamanjeendgar please make this a private method by renaming it to ``_dual_cone`` as we discussed.

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,73 @@
...
+    # def dual_cone(self, args) -> typing.Type['Cone']:
```
rileyjmurray, line 51: Remove this comment.

cvxpy/constraints/cones.py
```diff
@@ -0,0 +1,73 @@
...
+        dual-cone"""
+        raise NotImplementedError
+
+    @property
+    def dual_residual(self) -> float:
+        """Computes the residual (see Constraint.violation for a
+        more formal definition) for the dual cone of the current instance
+        of `Cone` w.r.t. the recovered dual variables
+
+        Primarily intended to be used for full-fledged KKT checks
```
rileyjmurray, line 71: In
``Primarily intended to be used for full-fledged KKT checks``
remove "full-fledged".

cvxpy/constraints/power.py
```diff
@@ -140,8 +139,25 @@ def save_dual_value(self, value) -> None:
         # TODO: figure out why the reshaping had to be done differently,
         #   relative to ExpCone constraints.
 
-
-class PowConeND(Constraint):
+    def dual_cone(self, *args):
```
rileyjmurray, line 142: Make sure that when you change ``dual_cone`` to ``_dual_cone`` in ``cones.py`` you also make that change in the various classes (like this one).

cvxpy/tests/solver_test_helpers.py
```diff
@@ -74,18 +74,14 @@ def check_dual_domains(self, places) -> None:
...
-            elif isinstance(con, cp.constraints.ExpCone):
-                # TODO: implement this (preferably with ExpCone.dual_violation)
-                raise NotImplementedError()
-            elif isinstance(con, cp.constraints.SOC):
-                # TODO: implement this (preferably with SOC.dual_violation)
-                raise NotImplementedError()
+            if isinstance(con, (cp.constraints.ExpCone,
+                                cp.constraints.PowCone3D,
+                                cp.constraints.PSD,
+                                cp.constraints.SOC)):
```
rileyjmurray, line 80: This if-statement can be simplified to ``if isinstance(con, cp.constraints.Cone)``.

rileyjmurray (changes requested):
See my individual comments.

rileyjmurray:
@aryamanjeendgar this generally looks good. Please make changes to the type annotations so that tests pass on Python 3.8. Once that's done we can request approval from another maintainer before merging.

rileyjmurray (approved):
Failing benchmarks are unrelated to this PR. I'm marking this as approved now. This doesn't change our public API so we don't need approval from a second maintainer to merge. But I'll wait until tomorrow morning ET before merging in case @SteveDiamond or @phschiele have comments they'd like to leave.

phschiele (approved):

PR #2207, opened 2023-08-11
Adds a clear error message to address a common student misuse of geo_mean

PTNobel:
## Description
Many of my students wrote `geo_mean(x, y)` when they needed to write `geo_mean(hstack([x, y]))`. They would get a type error about strict inequalities not being allowed, and be stuck on how to fix the issue. This was disappointing because we encouraged them to pay a lot of attention to error messages, so having a bad error message undermined our ability to make this point.

In order, to guide users to the correct use of the atom, I added an error message when the second argument is $p$. Further, I went through the atoms table and did not find any other atoms who are both sensibly used as a multiple argument function (I think norm taking a vector input is obvious enough there's no need to warn on it) and had a second argument which wasn't what someone might naively think. If I missed any please let me know and I'll add it to this PR.

Do we want tests of error messages? If so, I'll happily add one, but I don't recall the rules on atom testing tbh.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/errormsg.py
+++ b/cvxpy/atoms/errormsg.py
@@ -1,0 +1,5 @@
+SECOND_ARG_SHOULD_NOT_BE_EXPRESSION_ERROR_MESSAGE = """
+The second argument has type cvxpy.Expression. However, that is not allowed.
+Most likely, you want to call this atom by using cvxpy.hstack to combine the
+two arguemnts into a vector.
+"""
--- a/cvxpy/atoms/geo_mean.py
+++ b/cvxpy/atoms/geo_mean.py
@@ -20,6 +20,7 @@ import numpy as np
 import scipy.sparse as sp
 
 from cvxpy.atoms.atom import Atom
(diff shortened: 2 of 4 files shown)


SteveDiamond:
We do have tests for error messages. There should be some examples in test_atoms.py Could you take a look and add a test?

phschiele:
FWIW: At the last tutorial a typical path for the $\sqrt{x^2 + 1}$ example was:
1. `cp.norm2(x, 1)`, raises `ValueError("Invalid argument for axis.")`
2. `cp.norm2([x, 1])`, raises `ValueError: The input must be a single CVXPY Expression, not a list. Combine Expressions using atoms such as bmat, hstack, and vstack.`
3. `cp.norm2(cp.hstack([x, 1]))` -> success on 3rd try

So there is at least some evidence that indicates we should add a similar message for norms.

SteveDiamond:
I added a test and fixed the linter errors.

SteveDiamond (approved):

PR #2213, opened 2023-08-17
Updating backend docstrings

Transurgeon:
## Description
This PR attempts to add docstrings to all functions in the canonicalization backend. Notably, this includes the new NumPy backend that has recently been added in #2186 . This branch is WIP but a good chunk has already been completed, so a first review isn't out of the question. I also found a neat way to refactor indexing in the kron functions. 

P.S: Adding non-trivial docstrings is actually kind of challenging, please bare with all the redundancy and reiterations that I may have introduced.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -845,11 +845,15 @@ class ScipyCanonBackend(PythonCanonBackend):
         assert lhs.ndim == 2
 
         assert len({arg.shape for arg in lin.args}) == 1
-        lin.args[0].shape
+        rhs_shape = lin.args[0].shape
+
+        # TODO: add test to verify that reordering is necessary
+        row_idx = self._get_kron_row_indices(lin.data.shape, rhs_shape)
 
         def func(x: np.ndarray) -> np.ndarray:
             assert x.ndim == 2
             kron_res = sp.kron(lhs, x).tocsr()
+            kron_res = kron_res[row_idx, :]
             return kron_res
 
         return view.accumulate_over_variables(func, is_param_free_function=is_param_free_lhs)
@@ -1172,11 +1176,14 @@ class NumpyCanonBackend(PythonCanonBackend):
         assert lhs.ndim == 2
 
         assert len({arg.shape for arg in lin.args}) == 1
-        lin.args[0].shape
+        rhs_shape = lin.args[0].shape
+
+        row_idx = self._get_kron_row_indices(lin.data.shape, rhs_shape)
 
         def func(x: np.ndarray) -> np.ndarray:
             assert x.ndim == 3
             kron_res = np.kron(lhs, x)
+            kron_res = kron_res[:, row_idx, :]
             return kron_res
 
         return view.accumulate_over_variables(func, is_param_free_function=is_param_free_lhs)
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -751,6 +751,10 @@ class ScipyCanonBackend(PythonCanonBackend):
     def rmul(self, lin: LinOp, view: ScipyTensorView) -> ScipyTensorView:
         """
         Multiply view with constant data from the right.
+        When the rhs is parametrized, multiply each slice of the tensor with the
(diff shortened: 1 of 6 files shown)


SteveDiamond:
Great work adding these docstrings! Having docstrings sets a good pattern for future developers.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -509,11 +509,24 @@ def kron_l(self, lin: LinOp, view: TensorView) -> TensorView:
...
+            .flatten(order="F").astype(int)
+        return row_indices
+
     @abstractmethod
     def get_variable_tensor(self, shape: tuple[int, ...], variable_id: int) -> Any:
         """
-        Returns tensor of a variable node, i.e., eye(n) across axes 0 and 1, where n i the number of
-        entries of the variable.
+        Returns tensor of a variable node, i.e., eye(n) across axes 0 and 1, where n is
+        the number of entries of the variable.
```
phschiele, line 529:
```suggestion
        Returns tensor of a variable node, i.e., eye(n) across axes 0 and 1, where n is
        the size of the variable.
```

cvxpy/lin_ops/canon_backend.py
```diff
@@ -527,13 +540,18 @@ def get_data_tensor(self, data: Any) -> Any:
     @abstractmethod
     def get_param_tensor(self, shape: tuple[int, ...], parameter_id: int) -> Any:
         """
-        Returns tensor of a parameter node, i.e., eye(n) across axes 0 and 2, where n i the number
-        of entries of the parameter.
+        Returns tensor of a parameter node, i.e., eye(n) across axes 0 and 2, where n is
+        the number of entries of the parameter.
```
phschiele, line 544:
```suggestion
        Returns tensor of a parameter node, i.e., eye(n) across axes 0 and 2, where n is
        the size of the parameter.
```

cvxpy/lin_ops/canon_backend.py
```diff
@@ -1029,25 +1178,57 @@ def kron_r(self, lin: LinOp, view: NumpyTensorView) -> NumpyTensorView:
...
+        [[[1   0   0   0],
+         [0   1   0   0],
+         [2   0   0   0],
+         [0   2   0   0],
+         [0   0   1   0],
+         [0   0   0   1],
+         [0   0   2   0],
+         [0   0   0   2]]]
+
+         However computing kron(x, a) directly gives us:
```
Transurgeon, line 1219: I added a comparison of normal kron(x, a) to show why reordering is necessary. I am not sure if this is a great idea as it makes the docstring very lengthy.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -764,25 +847,22 @@ def kron_r(self, lin: LinOp, view: ScipyTensorView) -> ScipyTensorView:
...
-        lhs_ones = np.ones(lin.data.shape)
-        rhs_ones = np.ones(rhs_shape)
-
-        lhs_arange = np.arange(np.prod(lin.data.shape)).reshape(lin.data.shape, order="F")
-        rhs_arange = np.arange(np.prod(rhs_shape)).reshape(rhs_shape, order="F")
-
-        row_indices = (np.kron(lhs_ones, rhs_arange) +
-                       np.kron(lhs_arange, rhs_ones * np.prod(rhs_shape))) \
-            .flatten(order="F").astype(int)
+        # TODO: add test to verify that reordering is necessary
```
Transurgeon, line 850: Added TODO for the kron_r unit-test.

cvxpy/lin_ops/canon_backend.py
phschiele: I think we should move this to the tests and add a pointer here (which we should do in general).

Transurgeon: Ok, I'll try to make some additional changes now.

phschiele (approved):
Good from my side.

phschiele:
Benchmarks are failing because of a new asv release. Merging here as it only added docs.

PR #2217, opened 2023-08-25
Remove setuptools version bound for cvxpy-base build

ghackebeil:
The setuptools lower bound seems to be linked to the SCS installation (unless I am mistaken), so I'm hoping it can be dropped for the cvxpy-base build, or at least loosened.


SteveDiamond:
Pretty sure the version lower bound was due to a massive security issue with setuptools. I don't remember the details though.

ghackebeil:
Maybe this one?
https://github.com/advisories/GHSA-r9hx-vwmv-q579

I suppose it’s probably better to keep it then.

PR #2224, opened 2023-09-12
Adding log for canon backend when setting ``verbose=True``

Transurgeon:
## Description
This PR adds a log output message specifying which canonicalization backend is being used. If none has been specified, it will print out the default backend.

Example below:
```
===============================================================================
(CVXPY) Sep 12 09:32:16 AM: Your problem has 250000 variables, 3 constraints, and 0 parameters.
(CVXPY) Sep 12 09:32:16 AM: It is compliant with the following grammars: DCP, DQCP
(CVXPY) Sep 12 09:32:16 AM: (If you need to solve this problem multiple times, but with different data, consider using parameters.)
(CVXPY) Sep 12 09:32:16 AM: CVXPY will first compile your problem; then, it will invoke a numerical solver to obtain a solution.
(CVXPY) Sep 12 09:32:16 AM: Your problem is compiled with the SCIPY canonicalization backend # changes here
```
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/problems/problem.py
```diff
@@ -1015,7 +1015,9 @@ def _solve(self,
             s.LOGGER.info(
                     'CVXPY will first compile your problem; then, it will '
                     'invoke a numerical solver to obtain a solution.')
-
+            s.LOGGER.info(
+                    'Your problem is compiled with the %s canonicalization backend',
```
SteveDiamond, line 1019: End with a period.

phschiele:
We could change to use f-strings here.

Transurgeon:
I tried f strings, but the condition takes too much space and it would require 3 lines...
I think this method works for now.
But generally, I do agree that the formatting in f strings are much simpler to read off.

phschiele (approved):

PR #2226, opened 2023-09-12
Speed up parameterized QPs of a certain form

SteveDiamond:
## Description
I optimized one piece of the matrix stuffing for parameterized QPs, which arose in a problem similar to
```
import cvxpy as cp
import numpy as np

n = 5
w = cp.Parameter(n)
x = cp.Variable(n)
prob = cp.Problem(cp.Minimize(cp.sum_squares(x - w)), [x >= 2])

w.value = np.ones(n)
prob.solve()
```

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/utilities/coeff_extractor.py
+++ b/cvxpy/utilities/coeff_extractor.py
@@ -152,10 +152,17 @@ class CoeffExtractor:
                 # We multiply the columns of P, by c_part
                 # by operating directly on the data.
                 if var_size > 1:
+                    # TODO remove zeros from data.
                     data = P.data[:, None] * c_part[P.col]
+                    param_idxs = np.arange(c_part.shape[1])
                 else:
-                    data = P.data[:, None] * c_part
-                P_tup = (data, (P.row, P.col), P.shape)
+                    # Eliminate zeros from data by tracking
+                    # which indices of the global parameter vector are used.
+                    nonzero_idxs = c_part[0] != 0
+                    data = P.data[:, None] * c_part[0:, nonzero_idxs]
+                    param_idxs = np.arange(c_part.shape[1])[nonzero_idxs]
+                # TODO make P_tup a data structure.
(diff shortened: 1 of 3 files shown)


SteveDiamond:
I tried this on the larger problem and it doesn't fix the main performance issue, but is still an improvement potentially.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -163,26 +185,39 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
                         coeffs[orig_id]['P'] = P_tup
                     else:
                         coeffs[orig_id]['P'] = P_tup
                 else:
                     coeffs[orig_id] = dict()
                     coeffs[orig_id]['P'] = P_tup
-                    coeffs[orig_id]['q'] = np.zeros((P.shape[0], c.shape[1]))
+                    shape = (P.shape[0], c.shape[1])
+                    # Fast path for no parameters.
+                    if num_params <= 1:
```
Transurgeon, line 201: Is there a reason why you use <= here? Im wondering if num_params can even be < 1?

SteveDiamond: I don't think num_param can be < 1, but I'm not 100% sure.

SteveDiamond: Fixed.

PTNobel (approved):
Read through it today. Not sure I follow why this helps things [I'm also not that familiar with this part of the codebase], but I didn't find any behavior changes or anything that looks terribly off.

SteveDiamond:
@PTNobel basically I looked into what was happening in QP canonicalization when there were parameters, and it was creating massive dense matrices full of zeros. The main changes are to avoid creating the dense columns of zeros when possible and to use sparse matrices to store the data in some cases.

PR #2231, opened 2023-09-23
Atoms datatable

Transurgeon:
Implemented the atom data-table for atomic functions.

The steps were to add a static javascript file which renders the RST ``list-table`` into a dataTable.

The next goal will be to try and do a global atoms table and have the respective types only be views (filters) on that global table. Types can include DGP, DCP, DQCP, and also have sub-types such as matrix and scalar.


Transurgeon:
Oops, please ignore.
Meant to do a PR on my own fork again!

Issue #1106, opened 2020-08-12
Setting variable bounds (domain?)


adishavit:
Bounded variables are very common (LB <= var  <= UB). 
Some have internal bounds like `nonneg` And `boolean` and for others, one needs to add the bound constraints. For example an integer variable cannot be specified to be nonneg. 

When linearizing various formulations (eg big-M methods) one needs to pass around big M bounds in addition to the bound constraints. 
This is inconvenient and makes interfaces cumbersome. 

Is there some place that these bounds are stored within a Leaf, Variable or Atom?
I saw something called domain. 
Is that relevant?

rileyjmurray:
We currently don't have such a feature. You just append ``LB <= var`` and ``var <= UB`` to the list of constraints.

Right now CVXPY doesn't treat domain constraints differently than expression constraints. For example, ``x = Variable(nonneg=True)`` is later replaced by ``y = Variable(),  y >= 0`` when CVXPY prepares data for a solver. This sort of conversion isn't ideal because it means solvers can't exploit efficient methods for bound constraints. I'm personally in favor of changing this so that domain information is processed in a specialized way. However, there is currently no timeline for adding such a feature.

Maybe you're asking about something else, though?

adishavit:
Thanks. That was the question.

The API I was hoping for was something conceptually like having
`lower_bound` and `upper_bound` members or properties which can be set and
queried.

When you have these coupled to the variable they be used in eg big-M
constraints and also no need to add them explicitly to the constraints
(they will be added automatically like with booleans).

SteveDiamond:
Clarifying for people who may work on this, step 1 is to add support for a ``bounds`` attribute to Variables and Parameters. After that is implemented we can look into passing bounds information directly to the solvers.

jonathanberthias:
To add my 2 cents, I don't think adding `bounds` to the variable is a great API gain, as `x >= 0` is easier to understand and more general.
What I think could be really useful is if CVXPY were able to infer some variable bounds from the constraints without having to explicitly use the `bounds` attribute. So that e.g. `x <= 3` can be interpreted as a domain constraint and not just like any other constraint. That information can then be passed to the underlying solver.

SteveDiamond:
@jonathanberthias I have worked now with OR people and they consider ``bounds`` an important part of the API. Adding automatic checks for variable bounds is an interesting idea but not something we intend to add because it would increase the compile time, which is a major concern for many users.

[...]

PR #2234, opened 2023-09-25
Add support for bounds 

Paulnkk:
## Description
The first step of adding the attribute effectively is implementing support for lower and upper bounds for variables and parameters. The second step would involve passing the domain constraints directly to the solver so that solvers can exploit the properties of those types of constraints.
Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/1106

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/problems/problem.py
```diff
@@ -164,6 +164,21 @@ def __init__(
                 warnings.warn(f"Constraint #{i} contains too many subexpressions. "
                               "Consider vectorizing your CVXPY code to speed up compilation.")
 
+        for expr in self.variables() + self.parameters():
```
SteveDiamond, line 167: Adding the constraints should happen in cvx_attr2constr

cvxpy/problems/problem.py
```diff
@@ -164,6 +164,21 @@ def __init__(
...
+            lower_bound = expr.bounds[0]
+            upper_bound = expr.bounds[1]
+
+            if lower_bound is not None and upper_bound is not None:
+                if upper_bound < lower_bound:
+                    raise ValueError(f"Invalid bounds for {expr}: [{lower_bound}, {upper_bound}]")
```
SteveDiamond, line 173: Validation should happen in leaf.py

cvxpy/expressions/leaf.py
```diff
@@ -152,6 +152,8 @@ def __init__(
 
         self.args = []
 
+        self.bounds = bounds if bounds is not None else [None, None]
```
SteveDiamond, line 155: I don't love saving this as [None, None]. [-np.inf, np.inf] might better, because that's something the user might input.

SteveDiamond: But it still might be better to just save it as None if bounds aren't specified.

cvxpy/tests/test_constraints.py
```diff
@@ -394,3 +394,19 @@ def test_nonneg_dual(self) -> None:
...
         prob.solve(solver=cp.OSQP)
         self.assertItemsAlmostEqual(prob.constraints[0].dual_value, dual)
+
+    def test_bounds_attr(self) -> None:
+        """Test that the bounds attribute for variables and parameters is set correctly.
+        """
+        import pytest
+        x = cp.Variable(2, bounds=[11, 10])
+        c = [1, 2]
+        with pytest.raises(ValueError) as ke:
```
SteveDiamond, line 404: We usually do this as
```
        with pytest.raises(
                ValueError, 
                match=<string>
            ):
            raise the error
```

SteveDiamond:
Good start!

phschiele:
Thanks for starting on this @Paulnkk!

I think we should either accept scalar, or the shape has to match the shape of the expression. That is, we could support something like `Variable((2,), bounds = [np.array([1,2]), np.array([3,4])])`. We would need some further checking (or leave that to the constraint initialization that should catch shape mismatches). If neither are specified, we can just keep `self.bounds = None`.

Paulnkk:
> Thanks for starting on this @Paulnkk!
> 
> I think we should either accept scalar, or the shape has to match the shape of the expression. That is, we could support something like `Variable((2,), bounds = [np.array([1,2]), np.array([3,4])])`. We would need some further checking (or leave that to the constraint initialization that should catch shape mismatches). If neither are specified, we can just keep `self.bounds = None`.

@phschiele thanks for the suggestion! I thought about this as well since for now, the implementation imposes the same constraints on all the components of the vector. Your suggestion from above is a nice idea to deal with box constraints which should not be the same, I will try it that way

cvxpy/problems/problem.py
Paulnkk: @SteveDiamond, thanks for your suggestions! I moved, adding constraints here: [Move adding constraints](https://github.com/cvxpy/cvxpy/pull/2234/commits/8afae4488e868468326a59a81dc36654d2e8ae52)

cvxpy/problems/problem.py
Paulnkk: I moved it here: [Move validations to leaf](https://github.com/cvxpy/cvxpy/pull/2234/commits/d9248fa23610ef3befdd951b53884afd7e7b6d37)

Still left from my side to add a check for correct shapes and datatypes that are given by the user.

cvxpy/tests/test_constraints.py
Paulnkk: Sure thanks will do, removed the test coverage for now and will implement it when the basic setup for bounds attribute works.

cvxpy/expressions/leaf.py
```diff
@@ -483,3 +485,34 @@ def is_dpp(self, context: str = 'dcp') -> bool:
 
     def atoms(self) -> list[Atom]:
         return []
+
+    @property
+    def bounds(self):
+        # Assuming bounds is stored as a list of tuples
+        return self._bounds
+
+    @bounds.setter
```
SteveDiamond, line 494: This doesn't validate array inputs in the way @phschiele discussed. You need to 

1. Check that bounds is a list of two items.
2. Check that each item in the list is either a scalar or an array with the same dimensions as the leaf.
3. Check that upper_bound < lower_bound as an elementwise comparison. Make sure it works with arrays.

Please add all three of these checks and tests that hit each check.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -99,14 +100,27 @@ def apply(self, problem):
         id2old_var = {}
         constr = []
         for var in problem.variables():
+            bounds = var.bounds
```
SteveDiamond, line 103: This is somewhat convoluted. bounds should either be None or a list of two elements. It should not be possible to have lower_bound is None but upper_bound is not None. You also need to handle the case where the lower bound is -np.inf or the upper bound is np.inf. These do not generate constraints.

Please try again and move it into the main code path for adding attributes (below ``for key in CONVEX_ATTRIBUTES:``)

cvxpy/expressions/leaf.py
SteveDiamond: That's what I meant by condition number 2

SteveDiamond: Sorry, no bounds is always two items

Paulnkk: @SteveDiamond If I define x as a 3-dimensional vector like x = (x1, x2, x3)' and want to enforce domain constraints on each of components x1, x2, x3, it would look from my example above like 1<=x1<=2, 3<=x2<=4 and 5<=x3<=6. And I understood Philipps comment as x = (x1, x2)' and 1<=x1<=2, 3<=x2<=4. To define a constraint for each component of the vector.

SteveDiamond: In your example you would write``x = cp.Variable(3, bounds=[np.array([1,3,5]), np.array([2,4,6])])

cvxpy/reductions/cvx_attr2constr.py
Paulnkk: I moved the adding constraints operations below the for-loop

cvxpy/expressions/leaf.py
```diff
@@ -95,10 +95,10 @@ class Leaf(expression.Expression):
...
 
     def __init__(
-        self, shape: int | Iterable[int, ...], value=None, nonneg: bool = False,
-        nonpos: bool = False, complex: bool = False, imag: bool = False,
-        symmetric: bool = False, diag: bool = False, PSD: bool = False,
-        NSD: bool = False, hermitian: bool = False,
+        self, shape: int | Iterable[int, ...], value=None, bounds=None,
+        nonneg: bool = False, nonpos: bool = False, complex: bool = False,
+        imag: bool = False, symmetric: bool = False, diag: bool = False,
+        PSD: bool = False, NSD: bool = False, hermitian: bool = False,
```
PTNobel, line 101: Reordering these is a breaking change right?

cvxpy/expressions/leaf.py
```diff
@@ -152,6 +152,8 @@ def __init__(
 
         self.args = []
 
+        self.bounds = bounds if bounds is not None else None
```
PTNobel, line 155: Is this different from `self.bounds = bounds` without the ternary operator?

cvxpy/expressions/leaf.py
```diff
@@ -483,3 +485,47 @@ def is_dpp(self, context: str = 'dcp') -> bool:
 
     def atoms(self) -> list[Atom]:
         return []
+
+    @property
+    def bounds(self):
+        # Assuming bounds is stored as a list of tuples
```
PTNobel, line 491: I think this comment is incorrect

cvxpy/expressions/leaf.py
```diff
@@ -483,3 +485,47 @@ def is_dpp(self, context: str = 'dcp') -> bool:
...
+                and upper_bounds.shape == self.shape)):
+            raise ValueError("Bounds should either contain "
+                             "two scalars or two ndarrays with matching shapes.")
+
+        # Check that upper_bound >= lower_bound
+        if np.any(upper_bounds < lower_bounds):
+            raise ValueError("Invalid bounds: some upper bounds are less "
+                             "than corresponding lower bounds.")
+
+        self._bounds = [lower_bounds, upper_bounds]
```
PTNobel, line 531: Is there a reason to ban inf and -inf as bounds? If I want to provide an upper bound wouldn't `[-np.inf, 0]` be a reasonable way to implement this?

Also, is there a reason we only allow scalar bounds if the variable is a 1D vector with one entry? Wouldn't it be reasonable to say all variables are between 0 and 1 as `bounds=[0, 1]` instead of `bounds=[np.zeros(shape), np.ones(shape)]`? And we are banning scalar variables from having bounds?

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -139,6 +144,25 @@ def apply(self, problem):
...
+                            constr.append(obj <= upper_bounds)
+
+                    # If the bounds are arrays, we apply them element-wise.
+                    elif isinstance(lower_bounds, np.ndarray) \
+                            and isinstance(upper_bounds, np.ndarray):
+                        for idx in np.ndindex(obj.shape):
+                            if lower_bounds[idx] is not None:
+                                constr.append(obj[idx] >= lower_bounds[idx])
+                            if upper_bounds[idx] is not None:
+                                constr.append(obj[idx] <= upper_bounds[idx])
```
PTNobel, line 165: You don't allow mixing scalar and vector bounds?

PTNobel (commented):
I mostly have design questions that I think a maintainer would have to answer before Paul should implement them...

cvxpy/expressions/leaf.py
SteveDiamond: Yes, they should not be reordered.

cvxpy/expressions/leaf.py
SteveDiamond: Both +/-np.inf and scalar bounds should be allowed in all cases.

cvxpy/reductions/cvx_attr2constr.py
Paulnkk: @PTNobel, thanks for your comments! I followed the convention from above to use scalar bounds on scalar variables and arrays on > 1-dimensional variables. From your comment above I assume that you suggest allowing something like `bounds=[1, [2,3]]` so that the constraints would look like `1 <= x_1 <= 2` and `1 <= x_2 <= 3`?

cvxpy/expressions/leaf.py
Paulnkk: You are right, I will remove this part

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -148,21 +148,63 @@ def apply(self, problem):
...
+                        elif upper_bounds == np.inf:
+                            constr.append(obj >= lower_bounds)
+                        else:
                             constr.append(obj >= lower_bounds)
-                        if upper_bounds is not None:
                             constr.append(obj <= upper_bounds)
 
+                    # Case 2: lower bounds are scalar, upper bounds are array
+                    elif isinstance(lower_bounds, np.ndarray) and np.isscalar(upper_bounds):
+                        for idx in np.ndindex(obj.shape):
```
PTNobel, line 165: This will be an incredibly slow loop. Is there a reason we can't use broadcasting and masks?

For example if the lower_bounds is an array, you can implement this as:
```
mask = (lower_bounds == -np.inf)
constr.append(obj[mask] >= lower_bounds[mask])
```

Paulnkk: Thanks made changes here: [Remove slow loop](https://github.com/cvxpy/cvxpy/pull/2234/commits/15c0b904b09f054a8902c95762b6d735a865fae5)
I assume this was the reason why benchmarks were failing... Still not sure while linters fails here

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -139,6 +144,28 @@ def apply(self, problem):
...
+                        lower_bounds = np.full(obj.shape, lower_bounds)
+                    if np.isscalar(upper_bounds):
+                        upper_bounds = np.full(obj.shape, upper_bounds)
+
+                    # Create masks if -inf or inf is present in the bounds
+                    lower_bound_mask = (lower_bounds != -np.inf)
+                    upper_bound_mask = (upper_bounds != np.inf)
+
+                    # Apply constraints based on the presence of a valid bound
+                    for idx in np.ndindex(obj.shape):
```
PTNobel, line 162: This is still a very slow loop. It has the same efficiency as the one you replaced.

The entire loop can be replaced with
```
constr.append(obj[lower_bounds_mask] >= lower_bounds[lower_bounds_mask])
```

Paulnkk: Sorry misunderstanding from my side, I replaced the loop by indexing here: [Remove loop for mask](https://github.com/cvxpy/cvxpy/pull/2234/commits/d60e6eed0e96e3ff5a4cda44a89f827fd6e09d73)

cvxpy/expressions/leaf.py
```diff
@@ -483,3 +485,48 @@ def is_dpp(self, context: str = 'dcp') -> bool:
...
+                and upper_bounds.shape == self.shape) or
+               (is_upper_scalar and is_lower_array
+                and lower_bounds.shape == self.shape) or
+               (is_lower_array and is_upper_array
+                and lower_bounds.shape == self.shape
+                and upper_bounds.shape == self.shape)):
+            raise ValueError("Bounds should contain scalars and/or arrays with the same dimensions")
+
+        # Check that upper_bound >= lower_bound
+        if np.any(upper_bounds < lower_bounds):
```
PTNobel, line 528: So if someone provides `[np.inf, np.inf]` as a bound this fails to catch that. We should probably check if there are `np.inf` lower_bounds or `-np.inf` upper bounds and also check for `np.nan` while we're at it...

Paulnkk: Thanks! I added validations for the situation if `np.inf` and `-np.inf` is encountered in lower and upper bounds at the same position/component of the variable. Also added validations if `np.nan` occurs in lower and/or upper bound (I marked those situation as invalid and give back an error). I think the situation for `np.inf` in lower bound (and same logic `-np.inf` in upper) should be handled by now since if `np.inf` in lower and `np.inf` in upper, we get the Value error from the latest commit + if the user passes something else than `np.inf` in upper it must be considered smaller compared to `np.inf` (and the condition `upper_bounds >= lower_bounds` should trigger?)

Paulnkk:
For now, there is no support for something like `Variable((2,), bounds=[np.array([1]), np.array([2])])`. Should this case be considered the same as `Variable((2,), bounds=[1, 2])` ? Since `np.array` support is currently for variables and parameters with dimension >=2

cvxpy/expressions/leaf.py
```diff
@@ -529,4 +529,46 @@ def bounds(self, value):
...
+            if (lower_bounds == -np.inf
+                    and upper_bounds == -np.inf):
+                raise ValueError("-np.inf is not feasible as lower "
+                                 "and upper bound.")
+            if (lower_bounds == np.inf
+                    and upper_bounds == np.inf):
+                raise ValueError("np.inf is not feasible as lower "
+                                 "and upper bound.")
+            if (lower_bounds == np.nan
+                    or upper_bounds == np.nan):
```
PTNobel, line 542: `np.nan == np.nan` always evaluates to `False`. You have to use `np.isnan`

cvxpy/tests/test_constraints.py
```diff
@@ -464,6 +466,29 @@ def test_bounds_attr(self) -> None:
                                             "bounds are less than corresponding lower bounds."):
             cp.Variable((2,), bounds=[np.array([2,3]), np.array([1,4])])
 
+        with pytest.raises(ValueError, match="-np.inf is not feasible as lower and upper bound."):
+            cp.Variable((2,), bounds=[-np.inf, -np.inf])
+        with pytest.raises(ValueError, match="-np.inf is not feasible as lower and upper bound."):
+            cp.Variable((2,), bounds=[-np.inf, np.array([1,-np.inf])])
```
PTNobel, line 472: The error message does not explain what is wrong here (which is that l > u.

cvxpy/expressions/leaf.py
Paulnkk: I made the change and also found another mistake from my side in the if-conditions...

cvxpy/expressions/leaf.py
```diff
@@ -538,22 +538,22 @@ def bounds(self, value):
...
+            if (np.isnan(lower_bounds) == np.nan
+                    or np.isnan(upper_bounds) == np.nan):
                 raise ValueError("np.nan is not feasible as lower "
                                  "or upper bound.")
 
         if is_lower_scalar:
             # Convert scalar lower bounds to array for -inf and inf conflict mask
             lower_bounds = np.full(self.shape, lower_bounds)
-            if is_lower_scalar == np.nan or np.any(np.isnan(upper_bounds)):
+            if np.any(np.isnan(lower_bounds)) == np.nan or np.any(np.isnan(upper_bounds)):
```
PTNobel, line 549: I think you left the `== np.nan` on this line...

cvxpy/expressions/leaf.py
```diff
@@ -538,22 +538,22 @@ def bounds(self, value):
...
                 raise ValueError("np.nan is not feasible as lower "
                                  "or upper bound.")
 
         if is_upper_scalar:
             # Convert scalar upper bounds to array for -inf and inf conflict mask
             upper_bounds = np.full(self.shape, upper_bounds)
-            if is_upper_scalar == np.nan or np.any(np.isnan(lower_bounds)):
+            if np.any(np.isnan(upper_bounds)) == np.nan or np.any(np.isnan(lower_bounds)):
```
PTNobel, line 556: Same issue here with the `== np.nan`...

cvxpy/expressions/leaf.py
Paulnkk: Sorry, I messed up with the latest branch... pushed update

PTNobel (approved):
Looks great to me!

phschiele (approved):
Looks good!

PR #2236, opened 2023-09-27
Fix #2235

### LINKED ISSUE #2235: osqp_qif.OSQP.solve_via_data - l and u and OSQP scaling problem
**Describe the bug**

With a DPP problem, post parameters update, solve_via_data is used and calls solver.update with only new_args i.e. those affected by the parameters change.

(https://github.com/cvxpy/cvxpy/blob/v1.3.2/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L74-L76)

For `l` and `u` in particular, when OSQP uses scaling ([default 10](https://osqp.org/docs/interfaces/solver_settings.html)), values of `l` and `u` are modified in place hence ... 

.. If only one of `l` or `u` makes it into new args, it is possible that a `i` exist such as $l_i > u_i$  and the underlying call to osqp_update_upper_bound or osqp_update_lower_bound fails and cvxpy raises a ValueError.

See https://github.com/osqp/osqp-python/issues/119 for the related issue in in the osqp-python project.

**Expected behavior**

Checking the value of scaling in solver_opts in solve_via_data and if one of `l` or `u` are in new_args include the other would prevent the problem from occuring.

```python
            if solver_opts.get("scaling", 10) != 0:
                if "u" in new_args and "l" not in new_args:
                    new_args["l"] = data["l"]
                elif "l" in new_args and "u" not in new_args:
                    new_args["u"] = data["u"]
```

**Version**
 - OS: Ventura 13.5.2
 - CVXPY Version: 1.3.2

SteveDiamond:
## Description
Fixes issue where OSQP solver data could become erroneous.
Issue link (if applicable): #2235

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -85,6 +85,13 @@ class OSQP(QpSolver):
                 new_args['Ax'] = A.data
                 factorizing = True
 
+            # When scaling is non-zero, l and u are overwritten by OSQP.
+            if solver_opts.get("scaling", 10) != 0:
+                if "u" in new_args and "l" not in new_args:
+                    new_args["l"] = data["l"]
+                elif "l" in new_args and "u" not in new_args:
+                    new_args["u"] = data["u"]
+
             if new_args:
                 solver.update(**new_args)
             # Map OSQP statuses back to CVXPY statuses
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -86,7 +86,7 @@ class OSQP(QpSolver):
                 factorizing = True
 
             # When scaling is non-zero, l and u are overwritten by OSQP.
-            if solver_opts.get("scaling", 10) != 0:
+            if "scaling" in solver_opts and solver_opts["scaling"] != 0:
                 if "u" in new_args and "l" not in new_args:
                     new_args["l"] = data["l"]
                 elif "l" in new_args and "u" not in new_args:
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -85,7 +85,9 @@ class OSQP(QpSolver):
                 new_args['Ax'] = A.data
                 factorizing = True
 
-            # When scaling is non-zero, l and u are overwritten by OSQP.
+            # When scaling is enabled (non-zero), l and u are overwritten by OSQP.
+            # The default value of 10 is specified by OSQP here:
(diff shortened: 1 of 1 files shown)


cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -85,6 +85,13 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts,
                 new_args['Ax'] = A.data
                 factorizing = True
 
+            # When scaling is non-zero, l and u are overwritten by OSQP.
+            if solver_opts.get("scaling", 10) != 0:
```
phschiele, line 89: Can you explain the 10 in `solver_opts.get("scaling", 10)`?

SteveDiamond: I think the idea is the default scaling is 10, so if scaling is not specified it will still be non-zero.

SteveDiamond: I pushed a less confusing if statement.

phschiele: Ok, I now know where the default value comes from after looking at the linked issue. 
It seems a bit odd to just hard-code default value here, as it might change in the future, but we only care about it being non-zero. I also had to look up that a "scaling" of 0 just means disabling scaling completely (I would have assumed that scaling by 1 means no scaling).

Just to be clear though: This should eventually be fixed within OSQP, right?
Simple example:
I have `l = 1` and `u = 2`, which is modified in place to `l = 10` and `u = 20`.
Now we change `u` to be `u = 3`, which then fails the validation since `u = 3 < 10 = l`.
In such cases, the updated parameters should also be scaled.

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
```diff
@@ -85,6 +85,13 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts,
                 new_args['Ax'] = A.data
                 factorizing = True
 
+            # When scaling is non-zero, l and u are overwritten by OSQP.
```
phschiele, line 88:
```suggestion
            # When scaling is enabled (non-zero), l and u are overwritten by OSQP.
            # The default value of 10 is specified by OSQP here:
            # https://osqp.org/docs/interfaces/solver_settings.html
```

cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
SteveDiamond: I'm not sure if it will be fixed in OSQP. We can ask Bart.

phschiele (approved):

phschiele:
@SteveDiamond Actually, let me quickly push a test.

SteveDiamond:
Closing because this is not an issue in the latest OSQP

PR #2238, opened 2023-09-29
New atoms and adds NumPy API methods

### LINKED ISSUE #2237: Add sum / max / min methods to cp.Expression

Suppose I need a function that takes array-like inputs and outputs a complicated expression involving sum / max / min operations. An array-like input could be one of the following:
- a [numpy](https://github.com/numpy/numpy) array
- a [JAX](https://github.com/google/jax) array
- a [cvxpy](https://github.com/cvxpy/cvxpy) expression

To make the function [polymorphic](https://en.wikipedia.org/wiki/Ad_hoc_polymorphism) in its arguments, I must write something like the following:

```python3
def polymorphic_sum(x):
    if isinstance(x, cp.Expression):
        return cp.sum(x)
    else:
        return x.sum()

def polymorphic_max(x):
    if isinstance(x, cp.Expression):
        return cp.max(x)
    else:
        return x.max()

def polymorphic_min(x):
    if isinstance(x, cp.Expression):
        return cp.min(x)
    else:
        return x.min()

def complicated_function(x):
   return polymorphic_sum(x), polymorphic_max(x), polymorphic_min(x)
```

I'd like to be able to call `sum` / `max` / `min` on a [`cp.Expression`](https://www.cvxpy.org/api_reference/cvxpy.expressions.html) as a _method_ (i.e. through dot syntax), so that I can write such polymorphic functions more easily:
```python3
def complicated_function(x):
  return x.sum(), x.max(), x.min()
```

This can be done by adding `sum` / `max` / `min` methods to `cp.Expression` that call [`cp.sum`](https://www.cvxpy.org/api_reference/cvxpy.atoms.affine.html#sum) / [`cp.max`](https://www.cvxpy.org/api_reference/cvxpy.atoms.other_atoms.html#max) / [`cp.min`](https://www.cvxpy.org/api_reference/cvxpy.atoms.other_atoms.html#min), respectively.

I'm also very much open to adding other [atomic functions](https://www.cvxpy.org/tutorial/functions/index.html)—such as [`abs`](https://www.cvxpy.org/api_reference/cvxpy.atoms.elementwise.html#abs), [`log`](https://www.cvxpy.org/api_reference/cvxpy.atoms.elementwise.html#log), [`exp`](https://www.cvxpy.org/api_reference/cvxpy.atoms.elementwise.html#exp), etc.—as methods to `cp.Expression` for convenience. But that is perhaps a separate issue.

PTNobel:
## Description

Closes #2237 by cloning much of the [NumPy ndarray API](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html).

Supersedes and closes #2150 

Notes on specific methods:

 * I skipped `np.diagonal` because the RO API in NumPy made me feel that maybe we don't want to emulate it, and we don't support the optional arguments in diagonal yet
 * `.flatten` uses the opposite ordering as numpy, but is consistent with our default
 * `.max`, `.min`, `.prod` requires that `keepdims` be a keyword argument because NumPy has `out` as its second arg, and so this makes our API compatible with NumPy
 * `.mean`, `.ptp`, `.std`, `.var` do not have a corresponding `cp.mean`, `cp.ptp`, `cp.std`, `cp.var` respectively
 * I skipped .ravel, .repeat, .swapaxes, .take for no particular reason other than they felt a little weird to me with our API
 * I skipped `.transpose` because it is much less useful on 1d and 2d arrays

Once we confirm that this is more or less the API we're after and we want to include all the methods, I'll write tests.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/upper_tri.py
+++ b/cvxpy/atoms/affine/upper_tri.py
@@ -22,6 +22,7 @@ import cvxpy.lin_ops.lin_op as lo
 import cvxpy.lin_ops.lin_utils as lu
 from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.atoms.affine.reshape import reshape
+from cvxpy.atoms.affine.vec import vec
 from cvxpy.constraints.constraint import Constraint
 from cvxpy.expressions.expression import Expression
 
@@ -111,6 +112,12 @@ class upper_tri(AffAtom):
 
 def vec_to_upper_tri(expr, strict: bool = False):
     expr = Expression.cast_to_const(expr)
+
+    if not expr.is_vector():
+        raise ValueError("The input must be a vector.")
+    if expr.ndim != 1:
+        expr = vec(expr)
+
     ell = expr.shape[0]
     if strict:
         # n * (n-1)/2 == ell
@@ -119,6 +126,9 @@ def vec_to_upper_tri(expr, strict: bool = False):
         # n * (n+1)/2 == ell
         n = ((8 * ell + 1) ** 0.5 - 1) // 2
     n = int(n)
+    if not (n * (n + 1) // 2 == ell or n * (n - 1) // 2 == ell):
+        raise ValueError("The size of the vector must be a triangular number.")
+
     # form a matrix P, of shape (n**2, ell).
     #       the i-th block of n rows of P gives the entries of the i-th row
     #       of the upper-triangular matrix associated with expr.
@@ -126,15 +136,9 @@ def vec_to_upper_tri(expr, strict: bool = False):
     # compute expr3 = reshape(expr2, shape=(n, n)).T
     #   expr3 is the matrix formed by reading length-n blocks of expr2,
     #   and letting each block form a row of expr3.
-    P_rows = []
-    P_row = 0
-    for mat_row in range(n):
-        entries_in_row = n - mat_row
-        if strict:
-            entries_in_row -= 1
-        P_row += n - entries_in_row  # these are zeros
-        P_rows.extend(range(P_row, P_row + entries_in_row))
-        P_row += entries_in_row
+    k = 1 if strict else 0
+    row_idx, col_idx = np.triu_indices(n, k=k)
+    P_rows = n*row_idx + col_idx
     P_cols = np.arange(ell)
     P_vals = np.ones(P_cols.size)
     P = csc_matrix((P_vals, (P_rows, P_cols)), shape=(n**2, ell))
--- a/cvxpy/atoms/affine/upper_tri.py
+++ b/cvxpy/atoms/affine/upper_tri.py
@@ -111,7 +111,10 @@ class upper_tri(AffAtom):
 
 
 def vec_to_upper_tri(expr, strict: bool = False):
-    """Reshapes a vector into an upper triangular matrix in XXX order
+    """Reshapes a vector into an upper triangular matrix in
+    row-major order.
+
+    Inverts cp.upper_tri.
     """
     expr = Expression.cast_to_const(expr)
 
--- a/cvxpy/atoms/affine/upper_tri.py
+++ b/cvxpy/atoms/affine/upper_tri.py
@@ -112,8 +112,8 @@ class upper_tri(AffAtom):
 
 def vec_to_upper_tri(expr, strict: bool = False):
     """Reshapes a vector into an upper triangular matrix in
-    row-major order.
-
+    row-major order. The strict argument specifies whether an upper or a strict upper triangular
+    matrix should be returned.
     Inverts cp.upper_tri.
     """
     expr = Expression.cast_to_const(expr)
--- a/cvxpy/atoms/affine/upper_tri.py
+++ b/cvxpy/atoms/affine/upper_tri.py
@@ -111,6 +111,8 @@ class upper_tri(AffAtom):
 
 
 def vec_to_upper_tri(expr, strict: bool = False):
+    """Reshapes a vector into an upper triangular matrix in XXX order
+    """
     expr = Expression.cast_to_const(expr)
 
     if not expr.is_vector():
--- a/cvxpy/atoms/affine/wraps.py
+++ b/cvxpy/atoms/affine/wraps.py
@@ -64,6 +64,20 @@ class Wrap(AffAtom):
         return (arg_objs[0], [])
 
 
+class nonneg_wrap(Wrap):
+    """Asserts that the expression is nonnegative.
+    """
+    def is_nonneg(self) -> bool:
+        return True
+
+
+class nonpos_wrap(Wrap):
+    """Asserts that the expression is nonpositive.
+    """
+    def is_nonpos(self) -> bool:
+        return True
+
+
 class psd_wrap(Wrap):
     """Asserts that a square matrix is PSD.
     """
@@ -81,6 +95,12 @@ class psd_wrap(Wrap):
 
     def is_nsd(self) -> bool:
         return False
+
+    def is_complex(self) -> bool:
+        return self.args[0].is_complex()
+
+    def is_symmetric(self) -> bool:
+        return not self.args[0].is_complex()
 
     def is_hermitian(self) -> bool:
         return True
--- a/cvxpy/atoms/affine/wraps.py
+++ b/cvxpy/atoms/affine/wraps.py
@@ -37,6 +37,9 @@ class Wrap(AffAtom):
         """
         return values[0]
 
+    def is_complex(self) -> bool:
+        return self.args[0].is_complex()
+
     def shape_from_args(self) -> Tuple[int, ...]:
         """Shape of input.
         """
@@ -95,9 +98,6 @@ class psd_wrap(Wrap):
 
     def is_nsd(self) -> bool:
         return False
-
-    def is_complex(self) -> bool:
-        return self.args[0].is_complex()
 
     def is_symmetric(self) -> bool:
         return not self.args[0].is_complex()
--- a/cvxpy/atoms/ptp.py
+++ b/cvxpy/atoms/ptp.py
@@ -1,0 +1,28 @@
+"""
+Copyright 2013 CVXPY Developers
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+from cvxpy.atoms.max import max as cvxpy_max
+from cvxpy.atoms.min import min as cvxpy_min
+
+
+def ptp(x, axis=None, keepdims=False):
+    """
+    Range of values (maximum - minimum) along an axis.
+
+    The name of the function comes from the acronym for ‘peak to peak’.
+    """
+    return cvxpy_max(x, axis, keepdims) - cvxpy_min(x, axis, keepdims)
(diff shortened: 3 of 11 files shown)


cvxpy/expressions/expression.py
```diff
@@ -750,3 +750,89 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
 
     def __abs__(self):
         raise TypeError(__ABS_ERROR__)
+
+    def conj(self):
+        """
+        Equivalent to `cp.conj(self)`
+        """
+        from cvxpy import conj
```
SteveDiamond, line 758: Normally I use cvxtypes to prevent circular imports in situations like this. Though I'm not sure cvxtypes is better than importing directly like you did.

cvxpy/expressions/expression.py
```diff
@@ -750,3 +750,89 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
...
+        return conj(self)
+
+    def cumsum(self):
+        """
+        Equivalent to `cp.cumsum(self)`
+        """
+        from cvxpy import cumsum
+        return cumsum(self)
+
+    def max(self, axis=None, *, keepdims=False):
```
SteveDiamond, line 775: Could you say more about the * here? I looked it up, and this makes it so keepdims must be specified as a kwarg. Is that right?

cvxpy/expressions/expression.py
```diff
@@ -750,3 +750,89 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
...
+
+    def ptp(self, axis=None, *, keepdims=False):
+        """
+        Peak to peak (maximum - minimum) value along a given axis.
+        """
+        from cvxpy import max as max_
+        from cvxpy import min as min_
+        return max_(self, axis, keepdims) - min_(self, axis, keepdims)
+
+    def reshape(self, shape, order='F'):
```
SteveDiamond, line 811: At some point we should switch from Fortran order default to C order default. Maybe in cvxpy 2.0?

cvxpy/expressions/expression.py
SteveDiamond: Nvm, saw your explanation.

cvxpy/expressions/expression.py
phschiele: Yes, it would certainly be a breaking change, but I would recommend making that change in 2.0.

SteveDiamond:
This set of methods seems good to me. Should we add cp.mean, cp.ptp, cp.std, and cp.var as atoms?

SteveDiamond:
Actually we should add sum and possibly abs.

PTNobel:
> This set of methods seems good to me. Should we add cp.mean, cp.ptp, cp.std, and cp.var as atoms?

I think we should. Same pull request?

PTNobel:
> Actually we should add sum and possibly abs.

sum is an oversight on my part. I'll fix it tonight when I add tests.

Hmmm, abs is not in the NumPy API, do we really want to be expanding beyond it? Should we add other atoms like log and exp then?

SteveDiamond:
> > Actually we should add sum and possibly abs.
> 
> sum is an oversight on my part. I'll fix it tonight when I add tests.
> 
> Hmmm, abs is not in the NumPy API, do we really want to be expanding beyond it? Should we add other atoms like log and exp then?

I see, better to stick with the NumPy API.

cvxpy/expressions/expression.py
rileyjmurray: This could be a _bad_ breaking change. I suggest that once we get a timeline for 2.0 we start raising warnings if people don't specify the order explicitly.

cvxpy/expressions/expression.py
PTNobel: Ah okay; I will adjust this to be consistent with the rest of the codebase unless we want to migrate to this technique?

cvxpy/expressions/expression.py
PTNobel: I agree with Riley; though I've been bitten by the difference between `cp.reshape(x, (2, 3).value` and `np.reshape(x.value, (2, 3)` so I do question if we need a timeline for 2.0 and maybe should just raise a warning now?

PTNobel:
I had a few minutes to spare, so I pushed some new code for the new atoms. I need to test the broadcasting is being handled correctly in the std and var code still, and I'll write docs and tests tonight. Do people mind if I include #2150 in this PR as well? Cherry pick Philipp's commits and then when I'm editing the docs and tests do `vec_to_upper_tri` as well?

cvxpy/expressions/expression.py
SteveDiamond: The cvxtypes is so hideous, it might be worth migrating. What do you all think?

SteveDiamond:
> I had a few minutes to spare, so I pushed some new code for the new atoms. I need to test the broadcasting is being handled correctly in the std and var code still, and I'll write docs and tests tonight. Do people mind if I include #2150 in this PR as well? Cherry pick Philipp's commits and then when I'm editing the docs and tests do `vec_to_upper_tri` as well?

Merging in #2150 sounds good.

cvxpy/expressions/expression.py
PTNobel: I don't think there is any benefit to `cvxtypes` and we should get rid of it. It is opaque, requires more Python bytecode to be executed than the alternative.

cvxpy/atoms/affine/upper_tri.py
```diff
@@ -110,7 +111,15 @@ def graph_implementation(
 
 
 def vec_to_upper_tri(expr, strict: bool = False):
+    """Reshapes a vector into an upper triangular matrix in XXX order
```
PTNobel, line 114: @phschiele @SteveDiamond anyone familiar with the code here and no what order is being used to assemble this matrix from the vector?

SteveDiamond: I assume fortran order?

cvxpy/expressions/expression.py
SteveDiamond: Sg. Let's get rid of it in a separate PR though.

cvxpy/atoms/ptp.py
```diff
@@ -0,0 +1,28 @@
...
+from cvxpy.atoms.min import min as cvxpy_min
+
+
+def ptp(x, axis=None, keepdims=False):
+    """
+    Range of values (maximum - minimum) along an axis.
+
+    The name of the function comes from the acronym for ‘peak to peak’.
+    """
+    return cvxpy_max(x, axis, keepdims) - cvxpy_min(x, axis, keepdims)
```
PTNobel, line 28: Is there a "Asssume non-negative" like our "assume PSD" warpper I could apply here?

PTNobel:
I think everything but tests is done? @SteveDiamond If you want to do a code review, I'll come back in a few hours to write tests and hopefully resolve any outstanding todos.

cvxpy/atoms/ptp.py
SteveDiamond: There isn't. Should be fairly easy to add in wraps.py, but would need tests. Definitely can be a follow up PR.

cvxpy/atoms/affine/upper_tri.py
PTNobel: I tested it; it is C order.

doc/source/tutorial/functions/index.rst
```diff
@@ -311,6 +319,14 @@ and returns a scalar.
      - |concave| concave
      - |incr| incr.
 
+   * - :ref:`ptp(X) <ptp>`
+
+     - :math:`\max_{ij} X_{ij} - \min_{ij} X_{ij}`
+     - :math:`X \in \mathbf{R}^{m \times n}`
+
+     - |unknown| unknown
```
SteveDiamond, line 327: We should probably put in the work to add a nonneg_wrap and make the sign nonnegative before we add it to the API.

cvxpy/atoms/affine/upper_tri.py
SteveDiamond: Amazing. That's one thing we don't need to change later on at least :P

doc/source/tutorial/functions/index.rst
PTNobel: Done.

cvxpy/atoms/affine/wraps.py
```diff
@@ -82,6 +96,12 @@ def is_psd(self) -> bool:
     def is_nsd(self) -> bool:
         return False
 
+    def is_complex(self) -> bool:
```
PTNobel, line 99: Should this be in `class Wrap`?

SteveDiamond: Good idea.

SteveDiamond: Done

PTNobel:
Okay I think this is ready to go.

Outstanding TODOs:
 * Update docs for the methods on Expression
 * Eliminate cvxtypes in a different PR
 * Decide if we should start warning about reshape order before we have a 2.0 timeline

doc/source/tutorial/functions/index.rst
```diff
@@ -311,6 +319,14 @@ and returns a scalar.
      - |concave| concave
      - |incr| incr.
 
+   * - :ref:`ptp(X) <ptp>`
+
+     - :math:`\max_{ij} X_{ij} - \min_{ij} X_{ij}`
+     - :math:`X \in \mathbf{R}^{m \times n}`
+
```
SteveDiamond, line 326: Should this newline be here?

doc/source/tutorial/functions/index.rst
```diff
@@ -912,6 +944,15 @@ Expression and a negative Expression) then the returned Expression will have unk
      - |affine| affine
      - |incr| incr.
 
+   * - :ref:`vec_to_upper_tri(X, strict=False) <vec-to-upper-tri>`
+
+     - :math:`x' \in\mathbf{R}^{n(n-1)/2}` for ``strict=True``
+
+       :math:`x' \in\mathbf{R}^{n(n+1)/2}` for ``strict=False``
+     - :math:`X \in\mathbf{R}^{n \times n}`
```
SteveDiamond, line 952: sign should be same as X

SteveDiamond (approved):
LG other than minor docs issues.

doc/source/tutorial/functions/index.rst
PTNobel: This table has no sign information which is why I didn't include it

PTNobel:
I think public API changes needs to Maintainer signoffs, so @rileyjmurray @phschiele feel free to review!

cvxpy/atoms/stats.py
```diff
@@ -0,0 +1,64 @@
...
+def mean(x, axis=None, keepdims=False):
+    """
+    Returns the mean of x
+    """
+    if axis is None:
+        return cvxpy_sum(x, axis, keepdims) / x.size
+    elif axis in (0, 1):
+        return cvxpy_sum(x, axis, keepdims) / x.shape[axis]
+    else:
+        raise RuntimeError("Invalid axis value.")
```
phschiele, line 33: Based on the error message and looking at the other parts of the code base, a `ValueError` might be clearer.

cvxpy/tests/test_expression_methods.py
```diff
@@ -0,0 +1,400 @@
...
+
+        expr = expr.reshape((2, 2))
+        self.assertEqual(expr.shape, (2, 2))
+
+        expr = cp.square(self.x).reshape((1, 2))
+        self.assertEqual(expr.sign, s.NONNEG)
+        self.assertEqual(expr.curvature, s.CONVEX)
+        self.assertEqual(expr.shape, (1, 2))
+
+        with self.assertRaises(Exception) as cm:
```
phschiele, line 150: We should add one test to check that doing a cp reshape without order argument results in a np reshape with Fortran order.

phschiele (approved):
Left a few tiny comments that we might want to include before merging, otherwise looks good.

Overall I think this is a major improvement for the quality of life! Thanks a lot @PTNobel, awesome job!

PTNobel:
@phschiele can you look at the last commit on improving the tests and then merge?

phschiele:
Looks good, merging as soon as the CI completed.

PR #2215, opened 2023-08-22
New Stacked-Slices backend

Transurgeon:
## Description
This PR attempts to bring a new backend to CVXPY that outperforms the current SciPy backend. 
It is a major step forward towards tackling the numerous "slow DPP compilation time" issues brought by many users. 
Benchmarks have shown that some problems now compile up to 100x faster, but generally, the compilation times range between a 15 to 20% decrease. In addition, a few benchmarks are also much slower when compared to the original CVXCORE backend. 
Issue link (if applicable): #1332 

## Type of change
- [x] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2239, opened 2023-10-01
Prep backends for version 1.4

Transurgeon:
## Description
With the release of cvxpy 1.4 coming to a close, this PR attempts to replace the original SciPy backend with our rewritten version from #2215 . We have decided to rename the ``stacked-slices`` backend back to ``scipy`` to avoid any confusion from the user's perspective. As for the numpy backend, we plan to expose this to the user as a new argument for ``canon_backend``, as it could see some use-cases. What are the next steps? I think we could have a sentence or two in the docs to reflect the changes in the scipy backend.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


phschiele (approved):
Looks good, thanks for the updates @Transurgeon !

phschiele:
Phenomenal speed improvements in some cases! The below comparison shows the old SciPy backend compared to the new one based on the stacked slices data structure.
11s vs 4ms in one case! That's more than a 1000x improvement! 

image

PR #2245, opened 2023-10-03
Prepare for 3.12 / Drop 3.7

phschiele:
## Description
To ensure cvxpy is usable with the newly released 3.12, we planned to ship cvxpy with a 3.12 wheel right away. 
I tested it locally and it works. Currently, our solvers do not have prebuilds for 3.12, but I managed to install all of them locally, except for SCS. I think the build would not go through if this was the case on the CI as well.

This PR also drops Python 3.7 support, as it has reached EOL (https://devguide.python.org/versions/).

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11"]
+        python-version: [ 3.9, "3.10", "3.11"]
         include:
           - os: ubuntu-20.04
             python-version: 3.8
@@ -164,7 +164,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11", "3.12" ]
+        python-version: [ 3.9, "3.10", "3.11", "3.12" ]
         include:
           - os: ubuntu-20.04
             python-version: 3.8
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,7 +85,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11" ]
+        python-version: [ 3.9, "3.10", "3.11" ]
         include:
           - os: ubuntu-20.04
             python-version: 3.8
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,8 +104,8 @@ jobs:
       PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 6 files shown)


continuous_integration/install_dependencies.sh
```diff
@@ -49,8 +46,8 @@ else
   fi
 fi
 
-# cylp has wheels for all versions 3.7 - 3.10, except for 3.7 on Windows
-if [[ "$PYTHON_VERSION" != "3.7" ]] && [[ "$PYTHON_VERSION" != "3.11" ]] && [[ "$RUNNER_OS" != "Windows" ]]; then
+# cylp has no wheels for Windows
+if [ "$RUNNER_OS" != "Windows" ]]; then
```
Transurgeon, line 50: I think you are missing a bracket here.
should be ``if [[ "$RUNNER_OS" != "Windows" ]]; then``

continuous_integration/install_dependencies.sh
```diff
@@ -60,7 +57,7 @@ if [[ "$PYTHON_VERSION" == "3.9" ]] || [[ "$RUNNER_OS" == "Windows" ]]; then
 fi
 
 # Only install Mosek if license is available (secret is not copied to forks)
-if [[ -n "$MOSEK_CI_BASE64" ]] && [[ "$PYTHON_VERSION" != "3.11" ]]; then
+if [[ -n "$MOSEK_CI_BASE64" ]] then
```
Transurgeon, line 60: From the failing CI tests, I think here you need to keep the semi-colon.
``if [[ -n "$MOSEK_CI_BASE64" ]]; then``

.github/workflows/build.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11"]
+        python-version: [ 3.9, "3.10", "3.11"]
```
PTNobel, line 32: Why do we not us strings for 3.9?

.github/workflows/build.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11" ]
+        python-version: [ 3.9, "3.10", "3.11" ]
```
PTNobel, line 88: ditto

.github/workflows/build.yml
```diff
@@ -164,7 +164,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-20.04, macos-11, windows-2022 ]
-        python-version: [ 3.7, 3.9, "3.10", "3.11" ]
+        python-version: [ 3.9, "3.10", "3.11", "3.12" ]
```
PTNobel, line 167: Ditto

continuous_integration/install_dependencies.sh
PTNobel: Agreed

continuous_integration/install_dependencies.sh
PTNobel: There does.

PTNobel (approved):

PTNobel:
My approval is conditional on resolving the typoes

.github/workflows/build.yml
phschiele: I think we had to add it for 3.10 for the first time as otherwise it was somehow registered as 3.1.

continuous_integration/install_dependencies.sh
phschiele: Fixed in [17f41a3](https://github.com/cvxpy/cvxpy/pull/2245/commits/17f41a355ea996f34fa3836c392cf0fdd2696939)

continuous_integration/install_dependencies.sh
phschiele: Fixed in [17f41a3](https://github.com/cvxpy/cvxpy/pull/2245/commits/17f41a355ea996f34fa3836c392cf0fdd2696939)

phschiele:
Resolved the typos. Merging here based on the conditional approval and our discussions offline.

PR #2247, opened 2023-10-04
Follow-up backend 1.4 release prep

phschiele:
## Description
Small follow-up to #2239.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -1056,7 +1056,7 @@ class SciPyCanonBackend(PythonCanonBackend):
         if is_param_free_lhs:
             reps = view.rows // lhs.shape[-1]
             if reps > 1:
-                stacked_lhs = (sp.kron(sp.eye(reps, format="csc"), lhs))
+                stacked_lhs = (sp.kron(sp.eye(reps, format="csr"), lhs))
             else:
                 stacked_lhs = lhs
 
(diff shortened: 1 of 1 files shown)


cvxpy/lin_ops/canon_backend.py
```diff
@@ -1056,7 +1056,7 @@ def mul(self, lin: LinOp, view: SciPyTensorView) -> SciPyTensorView:
         if is_param_free_lhs:
             reps = view.rows // lhs.shape[-1]
             if reps > 1:
-                stacked_lhs = (sp.kron(sp.eye(reps, format="csc"), lhs))
+                stacked_lhs = (sp.kron(sp.eye(reps, format="csr"), lhs))
```
phschiele, line 1059: Here we are in the param-free branch, so we should use CSR.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -1209,7 +1209,7 @@ def func(x, p):
             else:
                 new_rows = x_row * (rows + 1) - k
             new_rows = (new_rows + x_slice * total_rows).astype(int)
-            return sp.csr_matrix((x.data, (new_rows, x.col)), shape)
+            return sp.csc_matrix((x.data, (new_rows, x.col)), shape)
```
phschiele, line 1212: This could be a parametrized expression, so we should use CSC.

PTNobel (approved):
This seems like mostly cleaning up sparse matrix formats and should all be correct

phschiele:
> This seems like mostly cleaning up sparse matrix formats and should all be correct

Some intuition:
We prefer CSR for fast row slicing. However, having parameters leads to **many** zero rows, so CSR becomes impractical / prohibitive. 

E.g., a `Variable(3)` looks like 
```
1 0 0
0 1 0
0 0 1
```
in the backend, but a `cp.multiply(Variable(3), Parameter(3))` looks like
```
1 0 0
0 0 0
0 0 0
0 0 0
0 1 0
0 0 0
0 0 0 
0 0 0
0 0 1
```
I.e., the number of rows is roughly $\mathcal{O}(n * p)$ (where n is the number of variables and p the number of parameters), but the nonzero rows is $\mathcal{O}(n)$. We often even have $n \approx p$.

PR #2251, opened 2023-10-07
Support or-tools 9.7; drop 9.4

mlubin:
Updates GLOP and PDLP to support or-tools 9.7. Drops support for or-tools 9.4.
I tested the code locally on 9.6 and 9.7.

Issue #1718, opened 2022-03-19
Support quadratic objectives in PDLP


rileyjmurray:
CVXPY 1.2 can interface to PDLP. Per the initial post suggesting PDLP support (https://github.com/cvxpy/cvxpy/issues/1649), PDLP allows diagonal quadratic objective functions.

CVXPY 1.3 should support quadratic objectives with PDLP. This can actually be done with diagonal and non-diagonal quadratics. However, it requires a bit of a hack. First, consider CVXPY's standard form for QPs:

My suggestion is that we do the following:
 * move ``pdlp_conif.py`` to ``pdlp_qpif.py``, and have ``PDLP`` inherit from ``QPSolver``.
 * In ``PDLP.apply``, check if the quadratic objective ``P`` is diagonal.
     * If it is, then don't make any changes to low-level problem data (at least, don't make changes beyond those that are required for something like OSQP).
      * If ``P`` is NOT diagonal, then factor it into ``P = M.T @ M``. Ideally, this would be a sparse factorization. Worst-case we could maybe afford to use a block-dense factorization. In any case, once ``M`` is in hand we ... https://github.com/cvxpy/cvxpy/blob/98a9ed721d3ffb7cf62de80374807117e4eb9abb/cvxpy/reductions/solvers/qp_solvers/qp_solver.py#L65-L73

SteveDiamond:
I would favor the reformulation over returning an error. My concern over Miles' initial suggestion of raising an error if it's not diagonal is that it would be hard to communicate to the user what sort of model formulations would be ok. It's not so complicated, but it feels like introducing a messy coupling.

rileyjmurray:
We should hold off on this reformulation until we decide on a standard method for sparse Cholesky and/or sparse LDLT within CVXPY. There are two ways we might access that functionality:

 * Use Eigen (see #1459. CVXPY's cvxcore subsystem already has Eigen as a dependency.
 * Use QDLDL. CVXPY has OSQP as a dependency, and OSQP requires QDLDL.

mlubin:
I'm hesitant to recommend the reformulation via Cholesky before seeing a couple numerical examples where it works well and has reasonable performance. It could be the case that other QP solvers accessible in CVXPY generally dominate PDLP in this setting.

PR #2252, opened 2023-10-07
Major update of PDLP interface

mlubin:
Switches the PDLP interface to use `ortools.pdlp.python.pdlp` instead of `ortools.linear_solver.python.model_builder_helper`. This interface should be much faster because it passes the sparse constraint matrix directly via pybind11; no need to loop through the coefficients in python.

This PR bumps the minimum version of or-tools to 9.7. The PDLP interface has changed across versions, and it would be a pain to manage lots of version branches in the interface.
You may consider tagging a cvxpy release with https://github.com/cvxpy/cvxpy/pull/2251 included before merging this PR.

FYI @mizux @lperron @dapplegate

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
@@ -131,7 +131,7 @@ class PDLP(ConicSolver):
         from ortools.pdlp.python import pdlp
 
         parameters = solvers_pb2.PrimalDualHybridGradientParams()
-        parameters.verbosity_level = 2 if verbose else 0
+        parameters.verbosity_level = 3 if verbose else 0
         # CVXPY reductions can leave a messy problem (e.g., no variable
         # bounds), so we turn on presolving by default.
         parameters.presolve_options.use_glop = True
--- a/cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
@@ -37,8 +37,9 @@ class PDLP(ConicSolver):
 
     SUPPORTED_CONSTRAINTS = ConicSolver.SUPPORTED_CONSTRAINTS
 
-    # The key that maps to the MPModelProto in the data returned by apply().
-    MODEL_PROTO = "model_proto"
+    # The key that maps to the pdlp.QuadraticProgram in the data returned by
+    # apply().
+    PDLP_MODEL = "pdlp_model"
 
     def name(self) -> str:
         """The name of the solver."""
@@ -46,11 +47,10 @@ class PDLP(ConicSolver):
 
     def import_solver(self) -> None:
         """Imports the solver."""
-        import google.protobuf  # noqa F401
         import ortools  # noqa F401
-        if Version(ortools.__version__) < Version('9.5.0'):
+        if Version(ortools.__version__) < Version('9.7.0'):
             raise RuntimeError(f'Version of ortools ({ortools.__version__}) '
-                               f'is too old. Expected >= 9.5.0.')
+                               f'is too old. Expected >= 9.7.0.')
         if Version(ortools.__version__) >= Version('9.8.0'):
             raise RuntimeError('Unrecognized new version of ortools '
                                f'({ortools.__version__}). Expected < 9.8.0. '
@@ -59,7 +59,7 @@ class PDLP(ConicSolver):
 
     def apply(self, problem: ParamConeProg) -> Tuple[Dict, Dict]:
         """Returns a new problem and data for inverting the new solution."""
-        from ortools.linear_solver import linear_solver_pb2
+        from ortools.pdlp.python import pdlp
 
         # Create data and inv_data objects
         data = {}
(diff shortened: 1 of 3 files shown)


rileyjmurray:
Very exciting! I'm not sure how we should time merging this. @SteveDiamond @PTNobel @phschiele what do you think?

My thoughts:
* I wouldn't want to release 1.4.0 without this _and then_ release 1.4.1 with this. I think it would be too disruptive of a change. So in my mind, we need to merge before releasing 1.4.0 or we need to delay this to 1.5.0. (I'm open to being convinced to the contrary.)
* We should release an update to 1.3.x that _does not_ have this PR.

mlubin:
If this can make it in for 1.4.0, that sounds like a reasonable point to bump the or-tools requirement.

PTNobel:
> You may consider tagging a cvxpy release with https://github.com/cvxpy/cvxpy/pull/2251 included before merging this PR.

@mlubin Can you explain this comment?

PTNobel:
Re @rileyjmurray:

I think as a matter of policy we should never increase the minimum version of a solver in a transition from 1.X.Y -> 1.X.(Y+1). Which means I agree with you that 1.3.X should NOT include this PR and that if we release 1.4.0 before merging this PR, we should not merge it for 1.4.1.

I'm neutral on whether this PR should be in 1.4.0; we do need to draw a line somewhere...

mlubin:
> I think as a matter of policy we should never increase the minimum version of a solver in a transition from 1.X.Y -> 1.X.

Given this policy, https://github.com/cvxpy/cvxpy/pull/2251 shouldn't go into 1.3.X because it drops support for or-tools 9.4.

> I wouldn't want to release 1.4.0 without this and then release 1.4.1 with this.

Based on these constraints we could either let https://github.com/cvxpy/cvxpy/pull/2251 go into 1.4 and wait for 1.5 to land https://github.com/cvxpy/cvxpy/pull/2252 or just land this PR for 1.4. 

> @mlubin Can you explain this comment?

https://github.com/cvxpy/cvxpy/pull/2251 lets users use or-tools 9.7 with the old/slow PDLP interface. It's a smaller change so it's less likely to have bugs. Not a big deal either way.

rileyjmurray:
Actually, looking at it more, I think it might make sense to delay this to CVXPY 1.5.0. This PR requires or-tools 9.7, while PR #2251 just added support for or-tools 9.7 in the first place.

Responding to Parth's comment:
> I think as a matter of policy we should never increase the minimum version of a solver in a transition from 1.X.Y -> 1.X.

I don't know if I'd go that far. If a solver makes no promises of backward compatibility / semantic versioning (as is the case with or-tools / PDLP) then it feels artificial for us to impose that. But maybe that's precisely why we _should_ impose this requirement. Hard to say.

Thoughts, @SteveDiamond and @PTNobel ?

PTNobel:
> I don't know if I'd go that far. If a solver makes no promises of backward compatibility / semantic versioning (as is the case with or-tools / PDLP) then it feels artificial for us to impose that. But maybe that's precisely why we should impose this requirement. Hard to say.

I think that's a reason for us to implement the policy. I think upgrading a CVXPY 1.X.Y to 1.X.(Y+1) is supposed to be very boring. One of your solvers no longer working is extremely interesting and breaks out very boring promise.

mlubin:
> I think upgrading a CVXPY 1.X.Y to 1.X.(Y+1) is supposed to be very boring. One of your solvers no longer working is extremely interesting and breaks out very boring promise.

FWIW I agree.

rileyjmurray:
Excellent point! Let’s make sure that neither of the PRs make it into the
1.3.X series. Based on this discussion I think the earlier PR is a good
candidate for 1.4.0, and this PR is better suited to 1.5.0.

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
```diff
@@ -162,31 +142,36 @@ def solve_via_data(
...
+        solution["status"] = self._status_map(result.solve_log)
+
+        convergence_info = self._get_convergence_info(
+            result.solve_log.solution_stats,
+            result.solve_log.solution_type
+        )
+        if convergence_info is not None:
+            solution["value"] = convergence_info.primal_objective
+        else:
+            solution["value"] = -np.inf
```
PTNobel, line 162: How do you detect infeasible problems? It seems you are assuming if it did not converge it was unbounded below?

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
```diff
@@ -77,36 +77,26 @@ def apply(self, problem: ParamConeProg) -> Tuple[Dict, Dict]:
...
+        constraint_upper_bounds = np.full_like(b, np.inf)
+        # Ax + b = 0
+        num_eq = problem.cone_dims.zero
+        constraint_lower_bounds[:num_eq] = -b[:num_eq]
+        constraint_upper_bounds[:num_eq] = -b[:num_eq]
+        # Ax + b >= 0
+        constraint_lower_bounds[num_eq:] = -b[num_eq:]
+
+        model.constraint_lower_bounds = constraint_lower_bounds
+        model.constraint_upper_bounds = constraint_upper_bounds
```
PTNobel, line 97: Out of curiousity what does this parameter of the model represent?

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
```diff
@@ -77,36 +77,26 @@ def apply(self, problem: ParamConeProg) -> Tuple[Dict, Dict]:
         A = csr_matrix(A)
         data["num_constraints"], data["num_vars"] = A.shape
 
-        # TODO: Switch to a vectorized model-building interface when one is
-        # available in OR-Tools.
-        model = linear_solver_pb2.MPModelProto()
+        model = pdlp.QuadraticProgram()
```
PTNobel, line 80: Aside, is this a QP solver? I'm pretty sure we're only treating it an LP solver and it could be nice to upgrade this to a QP solver.

PTNobel (commented):
I had a couple questions, but nothing major.

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
mlubin: The constraints (all linear) are encoded as:
```
constraint_lower_bounds <= constraint_matrix*x <= constraint_upper_bounds
```

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
mlubin: As of last time I checked, there's experimental support for QPs with a diagonal objective. The diagonal-only part would make it difficult to expose, and I haven't seen evidence that it's better than other QP solvers already available in cvxpy.

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
mlubin:
```
solution["status"] = self._status_map(result.solve_log)
```
manages the status reporting. It will return `s.INFEASIBLE` if the problem is infeasible. This code is just filling in a dummy value for `solution["value"]` when there's no objective value to report, because I wasn't certain what's required in this case. If it's valid to not set `solution["value"]` then that's a better option.

cvxpy/reductions/solvers/conic_solvers/pdlp_conif.py
rileyjmurray: @mlubin: we recently added sparse Cholesky decomposition functionality to our internal utilities. This could be used to implement the approach mentioned in https://github.com/cvxpy/cvxpy/issues/1718 for eliminating non-diagonal quadratic terms. This wouldn't affect your point about PDLP's performance relative to other solvers. However, it would address the problem of exposing general QP functionality to users.

mlubin: Ah, I forgot about that discussion. I'm still not too excited about exposing the QP functionality without evidence that it performs competitively with respect to other open source solvers accessible through cvxpy.

SteveDiamond (approved):

SteveDiamond:
Are we ready to merge this?

PTNobel:
Good with me @SteveDiamond

PR #2260, opened 2023-10-12
Patch 1.4.1

SteveDiamond:
## Description
Patch for 1.4

Includes #2259 

## Type of change
- [x] Bug fix

- [ ] Write unittests.


h-vetinari:
We still [manage](https://github.com/conda-forge/cvxpy-feedstock/pull/83) to hit the timelimit that was increased in #2249 when running in emulation (not consistently, but it happens on slow agents). I think if we increase the timeout by another factor of 10 (or at least ~3), it should pass consistently.

SteveDiamond:
> We still [manage](https://github.com/conda-forge/cvxpy-feedstock/pull/83) to hit the timelimit that was increased in #2249 when running in emulation (not consistently, but it happens on slow agents). I think if we increase the timeout by another factor of 10 (or at least ~3), it should pass consistently.

Increased by a factor of 100, since we're only testing that using the solver option doesn't throw an error.

SteveDiamond:
Redoing with cherry pick for clean commit history.

PR #2268, opened 2023-10-22
Reformatting the installation page

Transurgeon:
## Description
This PR introduces inline-tabs for cvxpy installation and drop-downs for solver installation instructions.
The updates can be viewed under this [link](https://cvxpy.readthedocs.io/en/inline-tabs/install/index.html).
I believe there was a concern with drop-downs not showing in the search tool, I think this isn't a problem anymore as the information in the dropdown can still be searched.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
This looks nice!

PTNobel:
Is there a reason that running our test suite comes before installing the other solvers? I feel like the latter is much more useful to users. Also, Clarabel is a default dependency, should we still have install instructions for it? We don't for SCS and ECOS.

PTNobel:
> I think we should definitely remove Clarabel from this list.

👍

> Do you think we would need to add ECOS back?

I don't think so, it'll still be an install dependency till CVXPY 1.7 which is a long way away.

> Also, we discussed briefly the idea of separating a solvers page (where we can put the solver options and stats)
as part of the refactoring of the advanced features

I still think keeping install info together makes sense, but I'm open to arguments against

Transurgeon:
Yea I agree that the solver install should be kept here. Once we have a concrete idea for the solver section and we feel that the install instructions become redundant here we could always remove it then. 
Also, since it is possible to install cvxpy without solvers, (cvxpy-base). I think we can briefly mention some install instructions of clarabel elsewhere, but I will remove it for now.
@SteveDiamond was there any preference for putting the testing suite before or after the solver options?

SteveDiamond:
> Yea I agree that the solver install should be kept here. Once we have a concrete idea for the solver section and we feel that the install instructions become redundant here we could always remove it then. Also, since it is possible to install cvxpy without solvers, (cvxpy-base). I think we can briefly mention some install instructions of clarabel elsewhere, but I will remove it for now. @SteveDiamond was there any preference for putting the testing suite before or after the solver install instructions?

I like @PTNobel's suggestion of putting the test suite instructions at the end.

Transurgeon:
done. I also removed the readthedocs yaml file.
The PR looks good on my end.

SteveDiamond (approved):
Nice job!

PTNobel (approved):

PR #2273, opened 2023-10-26
Update swig autogenerated files to swig 4.1.1

yuval-nextsilicon:
## Description
Swig 4.0.2 is very old and, critically for me, contains [this](https://github.com/swig/swig/issues/1985) bug which causes a memory leak (and crashes the program if LeakSan is attached). In this PR I simply regenerated the files using `rebuild_cvxcore.sh` with swig 4.1.1 installed (I also removed the argument `-py3` from `rebuild_cvxcore.sh` since swig claims is now deprecated and has no use).

## Type of change
- [x] Bug fix

- [ ] Write unittests.


SteveDiamond:
This is helpful! We will update assuming no issues are revealed in testing.

SteveDiamond:
Hmm many test failures.

yuval-nextsilicon:
This seems to be a [very recent regression in `cylp`](https://github.com/coin-or/CyLP/issues/189) which we've inherited because the `pyproject.toml` file doesn't have an upper bound for versions. I'll just rerun it which should use the version with the fix.

yuval-nextsilicon:
Turns out I don't have permission to rerun; @SteveDiamond can you trigger it?

yuval-nextsilicon:
There's a bug with `coptpy`, they probably also need to update their SWIG version - but I can't seem to find it on github... do you know who maintains that package?

yuval-nextsilicon:
I worked around it by making `coptpy` be opportunistically imported before `cvxcore`, which seems to solve the blunder inside SWIG. We need to remove this hack after the `coptpy` maintainers update their SWIG version.

yuval-nextsilicon:
> > I worked around it by making `coptpy` be opportunistically imported before `cvxcore`, which seems to solve the blunder inside SWIG. We need to remove this hack after the `coptpy` maintainers update their SWIG version.
> 
> COPT info is here: https://github.com/COPT-Public/COPT-Release

not open source and no real contact info... great... I've opened an issue but I doubt they'll seriously work on this.
do you wanna merge it like this?

SteveDiamond (approved):

SteveDiamond:
> > > I worked around it by making `coptpy` be opportunistically imported before `cvxcore`, which seems to solve the blunder inside SWIG. We need to remove this hack after the `coptpy` maintainers update their SWIG version.
> > 
> > 
> > COPT info is here: https://github.com/COPT-Public/COPT-Release
> 
> not open source and no real contact info... great... I've opened an issue but I doubt they'll seriously work on this. do you wanna merge it like this?

I think it's ok. Will discuss with other devs.

phschiele (approved):

phschiele:
Paging @wujianjack , who contributed the COPT interface, for the question about the same.

Issue #2278, opened 2023-10-30
SCIP solution not returned when total node limit reached


arvoelke:
**Describe the bug**
A solution is not returned when the SCIP status is "solving was interrupted [total node limit reached]", even when it has found a number of suboptimal solutions. This seems like a bug because it *will* return a solution if the solver is interrupted for other reasons, such as hitting a timeout.

**To Reproduce**
Run the below code twice, once with `be_good = False` and again with `be_good = True`:
1. When `be_good = False`, it runs for ~15 seconds and then hits the total node limit and raises an exception. No solution is returned despite it finding 17 suboptimal solutions.
2. When `be_good = True`, it runs for ~10 seconds and hits the timeout. A solution is available despite spending less time and finding the ~same solutions.

```python
import cvxpy as cp
import numpy as np

be_good = False  # set this to True to work-around the bug

if be_good:
    # Solution is accessible in this case.
    scip_params = {"limits/totalnodes": 5, "limits/time": 10}
else:
    # Solution is *not* accessible even though it spends more time and finds the same solutions.
    scip_params = {"limits/totalnodes": 5}

m = 100
n = 5
h = 1000

rng = np.random.RandomState(seed=0)
A = rng.randint(1, 10, size=(h, m))

X = cp.Variable((m, n), boolean=True)
c = cp.Variable(1)

objective = cp.Minimize(c)
constraints = [
    cp.sum(X, axis=1) == 1,
    A @ X <= c
]

prob = cp.Problem(objective, constraints)
prob.solve(solver=cp.SCIP, verbose=True, scip_params=scip_params)
print(prob.status, c.value)
```

**Expected behavior**
When the SCIP status is "solving was interrupted [total node limit reached]" I expect the best solution to be unpacked and returned, like it is in the case of "solving was interrupted [time limit reached]".

**Output**
```
===============================================================================
                                     CVXPY                                     
                                     v1.4.1                                    
===============================================================================
(CVXPY) Oct 30 02:01:37 PM: Your problem has 501 variables, 2 constraints, and 0 parameters.
(CVXPY) Oct 30 02:01:37 PM: It is compliant with the following grammars: DCP, DQCP
(CVXPY) Oct 30 02:01:37 PM: (If you need to solve this problem multiple times, but with different data, consider using parameters.)
(CVXPY) Oct 30 02:01:37 PM: CVXPY will first compile your problem; then, it will invoke a numerical solver to obtain a solution.
[...]

PR #2279, opened 2023-10-31
SCIP return solution if [total]nodelimit hit

arvoelke:
## Description

Makes it so that the SCIP solver will still return a solution if it hits the maximum number of iterations (consistent with timeout behavior).

Issue link: https://github.com/cvxpy/cvxpy/issues/2278

## Type of change
- [x] Bug fix

- [ ] Write unittests.


SteveDiamond (approved):
LGTM

Issue #1668, opened 2022-02-23
Utils.cpp:mat_vec_mul can be 4000% faster without SparseMatrixBase::pruned() call


lochsh:
**Describe the bug**
We have a convex optimisation problem we're solving with cvxpy. Depending on our inputs, we sometimes get unexpectedly long compilation times.

After doing some investigation, I tried applying this diff to `cvxcore`:

```diff
diff --git a/cvxpy/cvxcore/src/Utils.cpp b/cvxpy/cvxcore/src/Utils.cpp
index 17b11e3f9..62d0d697c 100644
--- a/cvxpy/cvxcore/src/Utils.cpp
+++ b/cvxpy/cvxcore/src/Utils.cpp
@@ -27,7 +27,7 @@ std::vector<Matrix> mat_vec_mul(const std::vector<Matrix> &lh_vec,
   for (unsigned i = 0; i < lh_vec.size(); ++i) {
     for (unsigned j = 0; j < rh_vec.size(); ++j) {
       // prune away explicit nonzeros
-      result.push_back((lh_vec[i] * rh_vec[j]).pruned());
+      result.push_back(lh_vec[i] * rh_vec[j]);
     }
   }
   return result;
```

For the slow example I was investigating, compilation time went down from 55 seconds to 1.26 seconds. The solution is identical.

For some other inputs there's no difference in speed.

**Output**

From profiling, it seems like the `assign_sparse_to_sparse` function is where the time is taken. https://github.com/cvxpy/cvxpy/blob/295f5720a2cbf005119acd18b16a943a4bd41ff0/cvxpy/cvxcore/include/Eigen/src/SparseCore/SparseAssign.h#71

Here is an excerpt of my `perf` report:
```
Samples: 244K of event 'cpu-clock', Event count (approx.): 61139500000
Overhead  Command  Shared Object                                     Symbol
  89.72%  python   _cvxcore.cpython-38-x86_64-linux-gnu.so           [.] Eigen::internal::assign_sparse_to_sparse<Eigen::SparseMatrix<double, 0, int>, Eigen::SparseView<Eigen::Product<Eigen::SparseMatrix<double, 0, int>, Eigen::SparseMatrix<double, 0, int>, 2> > >                   ◆
   1.27%  python   [kernel.kallsyms]                                 [k] __schedule                                                                                                                                                                                                     
```

and an excerpt of a backtrace from gdb:

```c++
[...]

PR #2283, opened 2023-10-31
Adding the ability to generate torch expressions

AmitSolomonPrinceton:
## Description
This change allows users to generate torch expressions from expressions and constraints, by using a new method called gen_torch_exp. Also added unittests for the new functionality, and one for each changed atom. This work continues the discussion in #torch_numeric in the Discord channel.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [X] Write unittests.
====================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.11.5, pytest-7.4.0, pluggy-1.0.0
rootdir: /Users/as3993/Amit/cvxpy
configfile: pyproject.toml
plugins: anyio-4.0.0, typeguard-3.0.2
collected 12 items                                                                                                                                                                              

cvxpy/tests/test_benchmarks.py ss-------------------------------------------------------------------------------
                                  Compilation                                  
(CVXPY) Oct 31 02:49:08 PM: Compiling problem (target solver=ECOS).
(CVXPY) Oct 31 02:49:08 PM: Reduction chain: Dcp2Cone -> CvxAttr2Constr -> ConeMatrixStuffing -> ECOS
(CVXPY) Oct 31 02:49:08 PM: Applying reduction Dcp2Cone
(CVXPY) Oct 31 02:49:08 PM: Applying reduction CvxAttr2Constr
(CVXPY) Oct 31 02:49:08 PM: Applying reduction ConeMatrixStuffing
(CVXPY) Oct 31 02:49:09 PM: Applying reduction ECOS
(CVXPY) Oct 31 02:49:09 PM: Finished problem compilation (took 2.831e-01 seconds).
Issue #1668 regression test
Compilation time:  0.2837810516357422
.least_squares: avg=2.010e-03 s , std=0.000e+00 s (1 iterations)
.ssqp: avg=3.008e-03 s , std=0.000e+00 s (1 iterations)
.small_cone_matrix_stuffing: avg=1.126e-02 s , std=1.509e-03 s (10 iterations)
.small_lp: avg=7.981e-03 s , std=0.000e+00 s (1 iterations)
small_lp_second_time: avg=4.008e-04 s , std=0.000e+00 s (1 iterations)
.sss

======================================================================================= warnings summary ========================================================================================
../../anaconda3/envs/lropt_cvxpy/lib/python3.11/site-packages/diffcp/cones.py:7
../../anaconda3/envs/lropt_cvxpy/lib/python3.11/site-packages/diffcp/cones.py:7
  /Users/as3993/anaconda3/envs/lropt_cvxpy/lib/python3.11/site-packages/diffcp/cones.py:7: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if StrictVersion(scs.__version__) >= StrictVersion('3.0.0'):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================================================== 5 passed, 7 skipped, 2 warnings in 1.45s ============================================================================


SteveDiamond:
This is nice work, but it makes torch a dependency of cvxpy, which we don't want.

lropt/pyvenv.cfg
```diff
@@ -0,0 +1,5 @@
+home = /Users/as3993/anaconda3/envs/lropt/bin
```
SteveDiamond, line 1: accidentally included this file.

cvxpy/tests/test_atoms.py
```diff
@@ -1791,3 +1837,147 @@ def test_exceptions(self):
         with self.assertRaises(Exception) as cm:
             cp.Problem(cp.Minimize(cp.dotsort(self.x, p_squared))).solve(enforce_dpp=True)
         assert "You are solving a parameterized problem that is not DPP" in str(cm.exception)
+
+class TestTorchNumeric(BaseTest):
```
SteveDiamond, line 1841: Nice tests! Maybe it should go in a separate file though?

SteveDiamond: I usually do one class per test file.

SteveDiamond:
> Thanks for the input, I updated the PR. Torch will no longer be a dependency of cvxpy, but is rather imported during a call to torch_numeric.

Ok sg. Great work on this MR! It might be close to mergeable.

PTNobel:
Should we add torch as an optional dependency?

cvxpy/atoms/atom.py
```diff
@@ -26,11 +28,40 @@
...
 from cvxpy.expressions.expression import Expression
+from cvxpy.expressions.variable import Variable
 from cvxpy.utilities import performance_utils as perf
 from cvxpy.utilities.deterministic import unique_list
 
+#from cvxpy.atoms.affine.binary_operators import MulExpression
+
+VAR_TYPE = Enum("VAR_TYPE", "VARIABLE_PARAMETER CONSTANT EXPRESSION")
+
+class VariablesDict():
```
SteveDiamond, line 44: What is this about?

cvxpy/atoms/atom.py
```diff
@@ -26,11 +28,40 @@
...
+        """
+        This function returns True if one of the keys of this variable is of a certain type,
+        and False otherwise
+        """
+        for var in self.vars_dict.keys():
+            if isinstance(var, look_type):
+                return True
+        return False
 
 class Atom(Expression):
```
SteveDiamond, line 66: I don't love the changes to this class. Please discuss with us about the overall picture.

cvxpy/atoms/atom.py
AmitSolomonPrinceton: It's an enum to tell apart between variables/parameters, constants, and expressions.

cvxpy/atoms/atom.py
```diff
@@ -471,3 +503,150 @@ def atoms(self) -> List['Atom']:
...
+                elif isinstance(arg, Parameter) or isinstance(arg, Variable):
+                    ind_to_value_type[i] = (arg, VAR_TYPE.VARIABLE_PARAMETER)
+                    vars_dict.add_var(arg)
+                else:
+                    ind_to_value_type[i] = (arg, VAR_TYPE.EXPRESSION)
+                    _gen_consts_vars(arg, vars_dict)
+            return ind_to_value_type
+        
+        def wrapped_func(self, ind_to_value_type, vars_dict, *args):
+            def transpose_if_matmul(self, res, should_transpose):
```
SteveDiamond, line 540: Wrapping with transpose in all cases seems to be more important than I had understood. I am open to changing how transpose works in expression.py.

rileyjmurray (changes requested):
IMO, we need detailed web documentation before merging this kind of functionality. I'd like for some of the docs to be written for developers, not just users.

A detailed overview of changes should probably be given at the top-level comment of this PR.

cvxpy/atoms/atom.py
```diff
@@ -471,3 +503,150 @@ def atoms(self) -> List['Atom']:
...
+        This function returns self.torch_numeric(values) if it exists,
+        and self.numeric(values) otherwise.
+        """
+        if hasattr(self, "torch_numeric"):
+            return self.torch_numeric(values)
+        else:
+            return self.numeric(values)
+
+
+    def gen_torch_exp(self):
```
SteveDiamond, line 518: gen_torch_exp should be a function in the Expression class, not the Atom class.

rileyjmurray:
Just popping in to say that I will want to review this thoroughly before it’s merged. Once it reaches a plausible level of maturity please request me as a reviewer!

phschiele:
@AmitSolomonPrinceton Great to see you updated this PR! It would be nice to have this functionality in 1.6.
Are there any open todos from your side here?

@rileyjmurray asked for a high level overview of changes before his review.
I think it would be important to also add documentation for this feature.
Maybe we can write some sort of a meta test that checks that all classes in the `cvxpy` namespace that implement `numeric` also implement `torch_numeric`?

AmitSolomonPrinceton:
> @AmitSolomonPrinceton Great to see you updated this PR! It would be nice to have this functionality in 1.6. Are there any open todos from your side here?
> 
> @rileyjmurray asked for a high level overview of changes before his review. I think it would be important to also add documentation for this feature. Maybe we can write some sort of a meta test that checks that all classes in the `cvxpy` namespace that implement `numeric` also implement `torch_numeric`?

Thanks @phschiele! I still iron some details (as you can see from commits, mainly working with sparse tensors).
What would you like to see in the documentation?

PTNobel:
>  What would you like to see in the documentation?

Hey Amit, are you on the CVXPY Discord? It might be easier to brainstorm the docs there and/or just a schedule a call. Personally, I think adding a section to the user guide on derivatives would be ideal.

AmitSolomonPrinceton:
@rileyjmurray @PTNobel I added a Torch expressions section to index.rst. Please let me know if this is what you had in mind and if there is anything else you would like me to add to the PR.

cvxpy/atoms/affine/reshape.py
```diff
@@ -93,6 +93,14 @@ def numeric(self, values):
         """Reshape the value.
         """
         return np.reshape(values[0], self.shape, self.order)
+    
+    def torch_numeric(self, values):
+        import torch
+        from cvxpy.utilities.torch_utils import tensor_reshape_fortran
+
+        if self.order == "F":
+            return tensor_reshape_fortran(values[0], self.shape)
```
Transurgeon, line 102: a bit nitpicky, but a more compact way to do this could be something like the following:
``torch.reshape(values[0].T, shape[::-1]).T``. See the stackoverflow [discussion](https://stackoverflow.com/questions/64433896/pytorch-equivalent-of-numpy-reshape-function)

cvxpy/atoms/affine/transpose.py
```diff
@@ -39,6 +39,9 @@ def name(self) -> str:
     @AffAtom.numpy_numeric
     def numeric(self, values):
         return np.transpose(values[0], axes=self.axes)
+    
+    def torch_numeric(self, values):
+        return values[0].T
```
Transurgeon, line 44: For dimensions greater than 2, I believe using ``.T`` behaves differently as transposing with axes. I don't think it's a big deal, but something to keep in mind. Maybe it would be better to keep using ``torch``?

cvxpy/atoms/affine/upper_tri.py
```diff
@@ -59,6 +60,28 @@ def numeric(self, values):
...
+    def torch_numeric(self, values):
+        import torch
+        value = torch.zeros(self.shape[0])
+        count = 0
+        for i in range(values[0].shape[0]):
+            for j in range(values[0].shape[1]):
+                if i < j:
+                    value[count] = values[0][i, j]
+                    count += 1
+        return value
```
Transurgeon, line 84: I think you wrote this function twice. Also, I am not sure using for loops is the best solution for performance. Could you try to improve this using torch natively, if it's even possible?

cvxpy/atoms/atom.py
```diff
@@ -27,11 +27,14 @@
 from cvxpy import interface as intf
 from cvxpy import utilities as u
 from cvxpy.expressions import cvxtypes
-from cvxpy.expressions.constants import Constant
+
+#from cvxpy.expressions.constants import Constant #TODO (Amit): This was in the original code
+from cvxpy.expressions.constants.constant import Constant
```
Transurgeon, line 32: agreed with Steven's comment below. You changed the import but it doesn't seem like it's doing anything. Could you have another look at the file changes here please?

cvxpy/expressions/leaf.py
```diff
@@ -602,3 +661,12 @@ def bounds(self, value):
...
         self._bounds = value
+
+    def gen_torch_exp(self):
+        """
+        This function generates a torch expression for a leaf.
+        Also returns a vars_dict with the leaf.
+        """
+        from cvxpy.atoms.affine.add_expr import AddExpression
+        tmp_aff_exp = AddExpression([self])
+        return tmp_aff_exp.gen_torch_exp()
```
Transurgeon, line 672: this function is also a duplicate of lines ``556-563`` above. Also it is not clear what this function is doing. It seems like you are using it in constraints, but an explanation would be nice.

cvxpy/tests/test_constraints.py
```diff
@@ -517,3 +587,59 @@ def test_bounds_attr(self) -> None:
...
+        test6 = exp6(T1, T2)
+        test7 = exp7(T1, T2)
+
+        self.assertTrue(all(test1==torch.tensor([3, -3])))
+        self.assertTrue(all(test2==np.array([0, 0])))
+        self.assertTrue(all(test3==torch.tensor([1])))
+        self.assertTrue(all(test4==torch.tensor([-1])))
+        self.assertTrue(np.isclose(test5, 25.9258))
+        self.assertTrue((test6==-n*np.ones((m,m))).all())
+        self.assertTrue((test7==n*np.ones((m,m))).all())
```
Transurgeon, line 645: these tests seem very similar than the lines you have added from 400-468. I am not very sure what is going on.
I think generally we prefer breaking the tests into multiple smaller tests. For your case, it could be one test for each torch numeric function that you have implemented.

Transurgeon (commented):
Nice work overall. Had a few concerns about duplicated code and am not really sure what is going in the tests, please have a look when possible.
Thought I would give a quick review before @rileyjmurray does a more thorough one.

P.S. Since the API is so similar to NumPy for the majority of torch_numeric functions you have added, I am wondering if we could somehow make use of the Scientific [Python Array API standard](https://data-apis.org/array-api/latest/purpose_and_scope.html) (which includes Pytorch and others) to generalize this? Would be happy to discuss more.

cvxpy/atoms/atom.py
AmitSolomonPrinceton: Could you please elaborate what you mean by this comment? This import is needed.

cvxpy/expressions/leaf.py
AmitSolomonPrinceton: Thanks for catching this. This function creates a torch expression for a leaf, as described in the docstring.

cvxpy/tests/test_constraints.py
AmitSolomonPrinceton: Thanks for catching the duplicate definition (I don't know what caused all of these duplicate definitions). This test is a general test to check that multiple constraints can generate torch_expressions (as opposed to say expressions). The only way I see to break this test would be into 7 different tests, one for each constraint. Would you like me to do that?

PTNobel (commented):
I didn't do a full review, but I'm a little concerned about how you handled constants. See a comment below.

cvxpy/atoms/affine/reshape.py
rileyjmurray: @Transurgeon, it looks like the implementation of ``tensor_reshape_fortran`` has been revised to use your suggested approach. I prefer to call the explicit helper function in this case since it's easier to read. Can we mark this as resolved?

cvxpy/expressions/expression.py
```diff
@@ -794,6 +795,261 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
...
 
+    def apply_torch_numeric(self, values):
+        """
+        This function returns self.torch_numeric(values) if it exists,
+        and self.numeric(values) otherwise.
+        """
+        if hasattr(self, "torch_numeric"):
+            return self.torch_numeric(values)
+        else:
+            return self.numeric(values)
```
rileyjmurray, line 806: The rationale for this function isn't clear to me. My instinct is to let the user figure out the AttributeError on their own. Alternatively, we could have a default implementation of ``torch_numeric`` for the Expression class that just raises a NotImplementedError.

cvxpy/expressions/expression.py
```diff
@@ -794,6 +795,261 @@ def __array_ufunc__(self, ufunc, method, *args, **kwargs):
...
+
+    def gen_torch_exp(self, provided_vars_list:list = []) -> tuple[callable, VariablesDict]:
+        """ This function generates a torch expression.
+        Also returns vars_dict to help keep track of the order of the variables.
+        If no vars_dict is provided, this function will generate one automatically.
+        The order of the arguments is the same as in self.args (left to right).
+        """
```
rileyjmurray, line 813: We'll definitely want more documentation here.

cvxpy/expressions/expression.py
```diff
@@ -884,3 +1140,167 @@ def var(self, *, ddof=0):
...
+                    res.append(wrapped_func(curr_arg[0], rec_ind_to_value_type, vars_dict, *args))
+                transposable_elements.append(should_transpose(curr_arg))
+            #If this is a matrix multiplicaiton operation between 2 elements, transpose the second.
+            #This helps with overloading this function to be used with matrices.
+            transpose_if_matmul(self, res, transposable_elements)
+            return self.apply_torch_numeric(res)
+        
+        vars_dict = VariablesDict(provided_vars_list=provided_vars_list)
+        ind_to_value_type = _gen_consts_vars(self, vars_dict)
+        return partial(wrapped_func, self, ind_to_value_type, vars_dict), vars_dict
```
rileyjmurray, line 1306: All of this code is duplicated from lines 798 -- 1052, and should be removed.

cvxpy/expressions/leaf.py
```diff
@@ -602,3 +602,62 @@ def bounds(self, value):
...
+                                "or upper bound.")
+
+        # Upper bound cannot be -np.inf.
+        if np.any(value[1] == -np.inf):
+            raise ValueError("-np.inf is not feasible as an upper bound.")
+        # Lower bound cannot be np.inf.
+        if np.any(value[0] == np.inf):
+            raise ValueError("np.inf is not feasible as a lower bound.")
+
+        self._bounds = value
```
rileyjmurray, line 654: This code is duplicated from lines 556 to 604, and should be removed.

cvxpy/expressions/leaf.py
rileyjmurray: Hi, @AmitSolomonPrinceton. I have to agree with @Transurgeon that the purpose of this function is unclear. What is the formal mathematical meaning of "generating a torch expression for a leaf"? I think this would be clearer if ``Expression`` had a detailed explanation of the intended semantics of ``gen_torch_expr``.

cvxpy/interface/matrix_utilities.py
```diff
@@ -18,13 +18,15 @@
 
 import numpy as np
 import scipy.sparse as sp
+import torch
```
rileyjmurray, line 21: Don't have a bare import statement for torch. We can't be certain that it will be installed.

cvxpy/interface/numpy_interface/tensor_interface.py
```diff
@@ -0,0 +1,102 @@
+"""
+Copyright 2013 Steven Diamond
```
rileyjmurray, line 2: Please use an appropriate copyright assertion. E.g,, "Copyright 2024, the CVXPY Authors"

cvxpy/interface/numpy_interface/tensor_interface.py
```diff
@@ -0,0 +1,102 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+from typing import Tuple
+
+import numpy as np
+import scipy.sparse
+import torch
```
rileyjmurray, line 22: A bare import of torch _might_ be okay at this point in the code, but I would prefer if you had a guard against the possibility that torch isn't available.

cvxpy/interface/numpy_interface/tensor_interface.py
rileyjmurray, line 1: A meta-comment for @AmitSolomonPrinceton.

@PTNobel pointed out a few places where we ought to be explicit about whether a tensor is created on the CPU or GPU. This file might be a good place to add the logic that handles that for us in a clean way. I'd also suggest _not_ using torch.float64 as the precision everywhere. There are some hardware accelerators -- like Apple's Metal Performance Shaders -- that only support 32-bit precision. Users who want 64 bit by default can set with with this function: https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html.

cvxpy/tests/test_constraints.py
rileyjmurray, line 1: I can't really assess the appropriateness of these tests, since I don't know the intended semantics of ``gen_torch_exp()``.

cvxpy/tests/test_torch_numeric.py
```diff
@@ -0,0 +1,172 @@
...
+from cvxpy.tests.base_test import BaseTest
+
+
+class TestTorchNumeric(BaseTest):
+    """ Unit tests for the atoms torch_numeric. """
+
+    def _assert_torch_numeric(self, atom, tests_flags=(True, True, True), repetitions=1,
+                              dtype=torch.float64, *args):
+        """
+        This funciton tests if atom.torch_numeric is correct.
```
rileyjmurray, line 33: "funciton" should be "function"

rileyjmurray, line 33: Please elaborate on what it means for an atom's implementation of ``torch_numeric`` to be correct.

cvxpy/tests/test_torch_numeric.py
```diff
@@ -0,0 +1,172 @@
...
+                and the second element is the 1D vector, pass (1,0).
+
+        Comments:
+            You may notice that during the test, an object is initiated with a dummy variable
+            (often None). That is because to test torch_numeric, this value is often ignored,
+            and only the passed values are used.
+            
+        """
+
+        def perform_test(atom, in_np, in_tch):
```
rileyjmurray, line 63: Please use a more descriptive name. E.g., ``check_torch_against_numeric``

cvxpy/utilities/torch_utils.py
```diff
@@ -0,0 +1,71 @@
+"""
+Copyright 2013 Steven Diamond
```
rileyjmurray, line 2: Please use an appropriate copyright assertion.

cvxpy/utilities/torch_utils.py
```diff
@@ -0,0 +1,71 @@
...
+import torch
+from scipy.sparse import coo_matrix, issparse
+
+VAR_TYPE = Enum("VAR_TYPE", "VARIABLE_PARAMETER CONSTANT EXPRESSION")
+
+class VariablesDict():
+    """ Helper class that contains a dictionary from a non-constant leaf to an index in *args,
+    with a safe add method.
+    If a list is given, will create the dictionary in the order of the list."""
+    def __init__(self, provided_vars_list: list=[]):
```
rileyjmurray, line 29: Don't use mutable defaults: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ requires = [
     "scipy >= 1.1.0",
     # 68.1.0 Promoted pyproject.toml's [tool.setuptools] out of beta.
     "setuptools >= 68.1.0",
+    "torch",
```
rileyjmurray, line 32: Don't include torch as a required dependency.

cvxpy/atoms/dotsort.py
```diff
@@ -134,11 +139,11 @@ def get_data(self):
         return None
 
     @staticmethod
-    def _get_args_from_values(values: List[np.ndarray]) \
+    def _get_args_from_values(values: List[np.ndarray], mod=np) \
             -> Tuple[np.ndarray, np.ndarray]:
```
rileyjmurray, line 143: The changes to this function make the type annotation incorrect.

rileyjmurray (changes requested):
This is a monster of a project! Thanks for all the hard work in getting this far, @AmitSolomonPrinceton. I've left many comments that I'd like you to address. 

## Most important comments

I'd like the semantics of``torch_numeric`` and ``gen_torch_exp`` clearly defined in the appropriate base classes (I think ``Atom`` and ``Expression`` are the appropriate base classes, but I could be wrong).

Here is a template for the kind of specification that I'm looking for.
```python
class SomewhatAbstractClass:
    ... 
    def fairly_abstract_operation(self, ...):
        """
        Implementations of this function perform  <operation involving args>, where <args> are <some datatype>
        representations of <some mathematical objects>, and <operation> can be anything that <properties>.

        For example, if <the operation> was <specific example here>, then a conformant
        implementation might look like:

        .. code:: python

            # Some Python code.
            # A little more Python code.
            # Just a teeny bit more code. But not too much!


        """
        raise NotImplementedError()
```

I also have a handful of comments that I didn't attach to a single file or block of code.

## Handling torch imports

I think importing torch in every function that needs it introduces excessive overhead, and it prevents you from using type annotations. I recently need to add torch functionality to a project that includes some amount of optimization modeling, and I found the following techniques helpful:

* An example that handles the type annotation issue. https://github.com/sandialabs/pyGSTi/blob/911118c3965a4714cbdef3e883e942d1e8c3c6eb/pygsti/modelmembers/torchable.py#L15-L18
* An example that handles type annotation and function invocation: https://github.com/sandialabs/pyGSTi/blob/911118c3965a4714cbdef3e883e942d1e8c3c6eb/pygsti/modelmembers/states/tpstate.py#L14-L21 

## Misc

The screenshot below shows deferences to intf.sparse and scipy.sparse that I wouldn't expect inside these torch-related functions.
Pasted Graphic 1

The screenshot below shows us using python's ``reduce`` function. Can you please verify that using reduce is compatible with the automatic differentiation features of pytorch?
Pasted Graphic 2

cvxpy/atoms/affine/reshape.py
Transurgeon: Yes sure go ahead.

jonathanberthias:
Hi everyone,

Sorry if this has already been discussed, but I wanted to suggest moving this code to an external package, e.g. `cvxpy-torch`. The reason I'm suggesting this is that almost all the code in this PR is new and does not rely much on private implementations. It would also reduce the maintenance burden for `cvxpy` and cost of implementing new atoms. Many review comments relate to making sure `torch` is an optional dependency, which is a constraint an external package would not have.

I could imagine an interface like:
```python
exp = cp.Variable() ** 2
torch_exp = cvxpy_torch.gen_exp(exp)
```
where `gen_exp` could be a [dispatching function](https://docs.python.org/3/library/functools.html#functools.singledispatch). Or it could use a visitor pattern.

Sorry to intrude so long after this feature has been started, I realize there has already been a lot of work put into this!

SteveDiamond:
@jonathanberthias we discussed and decided to develop this functionality as a separate package here https://github.com/cvxpy/cvxtorch
We would greatly appreciate your thoughts on how to best approach the problem!

PR #2291, opened 2023-11-20
Use oldest-supported-numpy

enzbus:
For https://github.com/bodono/scs-python/issues/77; @phschiele @SteveDiamond you can try the CI on this PR, hopefully it works.

### DIFF
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -52,7 +52,7 @@ if [[ "$RUNNER_OS" != "Windows" ]]; then
 fi
 
 # SCIP only works with scipy >= 1.5 due to dependency conflicts when installing on Linux/macOS
-if [[ "$PYTHON_VERSION" == "3.9" ]] || [[ "$RUNNER_OS" == "Windows" ]]; then
+if [[ "$PYTHON_VERSION" == "3.9" ]] && [[ "$RUNNER_OS" == "Windows" ]]; then
   conda install pyscipopt
 fi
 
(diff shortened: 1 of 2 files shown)


SteveDiamond:
It didn't work with oldest-supported-numpy, so trying again with the explicit NumPy versions.

enzbus:
Ok, looks good. It fixed the issue on 3.8. It fails on Windows because we're building from source (you may also add `openblas` to the conda installs, for fun), when installing from pip/pypi it doesn't matter. Then some other failures seem to be due to `scip`.

enzbus:
You can rerun with the `oldest-supported-numpy` to see if there are different failures?

SteveDiamond:
In the interest of getting things working I dropped the pyscipopt installation on linux and mac.

enzbus:
I got the needed fixes from both PRs, if this passes all tests I think you can merge. Also consider the CLA signed. Thanks!

SteveDiamond:
There are some other changes needed. The main PR now is https://github.com/cvxpy/cvxpy/pull/2292

SteveDiamond:
Addressed in #2292

enzbus:
OK @SteveDiamond @phschiele but you can still replace all the `numpy` clauses in the build section of your `pyproject.toml` with `oldest-supported-numpy`. As you see it passes all tests. It is the official solution by the Numpy team for that job, makes things cleaner and more maintenable for you. Sorry but I have to act whenever I see unnecessary complications in dependencies, especially in their package management code. Thanks!

SteveDiamond (approved):

Issue #2287, opened 2023-11-08
pypi dependencies list clarabel twice


adfea9c0:
**Describe the bug**
Looking at the pypi dependencies for 1.4, I see:

```bash
 $ curl https://pypi.org/pypi/cvxpy/1.4/json -H "Accept: application/json" | jq '.info.requires_dist'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 25023  100 25023    0     0  2036k      0 --:--:-- --:--:-- --:--:-- 2036k
[
  "osqp >=0.6.2",
  "ecos >=2",
  "clarabel >=0.5.0",
  "scs >=3.0",
  "numpy >=1.15",
  "scipy >=1.1.0",
  "pybind11",
  "cylp >=0.91.5 ; extra == 'cbc'",
  "clarabel ; extra == 'clarabel'",
  "cvxopt ; extra == 'cvxopt'",
  "diffcp ; extra == 'diffcp'",
  "ortools <9.8,>=9.5 ; extra == 'glop'",
  "cvxopt ; extra == 'glpk'",
  "cvxopt ; extra == 'glpk_mi'",
  "gurobipy ; extra == 'gurobi'",
  "scipy >=1.6.1 ; extra == 'highs'",
  "Mosek ; extra == 'mosek'",
  "ortools <9.8,>=9.5 ; extra == 'pdlp'",
  "piqp ; extra == 'piqp'",
  "proxsuite ; extra == 'proxqp'",
  "PySCIPOpt ; extra == 'scip'",
  "scipy ; extra == 'scipy'",
  "setuptools >65.5.1 ; extra == 'scs'",
  "xpress ; extra == 'xpress'"
]
```

Clarabel is listed twice (`"clarabel >=0.5.0"` and `"clarabel ; extra == 'clarabel'"`) with the first inclusion being unconditional. Based on my reading of the website this is probably unintentional? Should clarabel be optional?

SteveDiamond:
Clarabel should not be optional

rileyjmurray:
That line in the docs (and perhaps a few others) definitely needs to be
updated.

Transurgeon:
@adfea9c0 Thanks for pointing this out! I am currently working on a PR that tries to refactor the "advanced features" page. I will make sure to replace Clarabel instead of ECOS in the docs. 
Was there any other issues that you noticed? Any feedback can be helpful :)

[...]

PR #2292, opened 2023-11-21
Test SCS Python 3.8 fix

phschiele:
## Description
Various updates to the setup and build scripts:
- Use "oldest-supported-numpy". 
- Fix cvxpy-base build.
- Remove clarabel as an *optional* dependency. (Still kept as a *required* dependency.)

Issue link (if applicable): #2287 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        if: ${{env.USE_OPENMP != 'True'}}
+        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
         env:
           CIBW_BUILD: "cp3${{env.PYTHON_SUBVERSION}}-*"
           CIBW_SKIP: "*-win32 *-manylinux_i686 *-musllinux*"
@@ -131,13 +131,13 @@ jobs:
         uses: pypa/cibuildwheel@v2.16.2
 
       - name: Build source
-        if: ${{env.SINGLE_ACTION_CONFIG == 'True'}}
+        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
         run: |
           pip install build
           python -m build --sdist -o wheelhouse
 
       - name: Check wheels
-        if: ${{env.USE_OPENMP != 'True'}}
+        if: ${{github.event_name == 'push' &&  env.USE_OPENMP != 'True'}}
         shell: bash
         run: |
           python -m pip install --upgrade twine
@@ -220,6 +220,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
+        if: ${{github.event_name == 'push'}}
         env:
           CIBW_BUILD: "cp3${{env.PYTHON_SUBVERSION}}-*"
           CIBW_SKIP: "*-win32 *-manylinux_i686 *-musllinux*"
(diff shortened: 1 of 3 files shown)


enzbus:
Try also `oldest-supported-numpy` in your `pyproject.toml`, should make it more maintainable.

pyproject.toml
```diff
@@ -64,7 +64,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 # Solver names as in cvxpy.settings = pip-installable distribution providing it
 CBC = ["cylp>=0.91.5"]
-CLARABEL = ["clarabel"]
+CLARABEL = []
```
SteveDiamond, line 67: Fixes #2287

pyproject.toml
```diff
@@ -47,7 +47,7 @@ dependencies = [
     "osqp >= 0.6.2",
     "ecos >= 2",
     "clarabel >= 0.5.0",
-    "scs >= 3.0",
+    "scs >= 3.2.4.post1",
```
PTNobel, line 50: Is this a prerelease version? Do we want to depend on it?

.github/workflows/build.yml
```diff
@@ -196,16 +196,26 @@ jobs:
           echo "PYTHON_SUBVERSION=$(echo $PYTHON_VERSION | cut -c 3-)" >> $GITHUB_ENV
           echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
 
-      - name: Adapt setup.py
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: ${{ matrix.python-version }}
+          channels: conda-forge,anaconda
```
PTNobel, line 203: Overall, I'm happy with this; maybe we want to make CVXPY a virtual package that depends on cvxpy-base though... It would allow us to have CVXPYgen and DSP to depend on cvxpy-base instead...

PTNobel (approved):
Overall looks good. I had two questions.

pyproject.toml
SteveDiamond: It won't be deleted so I think it's fine to depend on it. It is the only version of SCS that we can install in CI right now.

.github/workflows/build.yml
SteveDiamond: Interesting idea. CVXPYgen and DSP could already depend on cvxpy-base though, right?

.github/workflows/build.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           platforms: all
 
       - name: Build wheels
-        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
```
SteveDiamond, line 125: Why are the if statements here changed?

phschiele: Just so we can test the build of the wheels before merging.

phschiele:
Can't approve my own PR, but looks good to me now!

.github/workflows/build.yml
PTNobel: I don't think cvxpy-base is installed with CVXPY right now so you'd have a conflict by installing both

PR #2304, opened 2023-12-14
Add return types affine add expr

sunnygurm:
## Description
While getting familiar with the repository and codebase, I ran `ruff check --select ANN201 ./cvxpy` to see how many issues existed relating to return type annotations for public functions. 

I found 933 such instances (see attached [ann201_errors.txt](https://github.com/cvxpy/cvxpy/files/13667751/ann201_errors.txt)). 

More details for ANN201 errors and why it might be good to fix them can be found [here](https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/). 

This PR adds return type annotations that were missing in `cvxpy/atoms/affine/add_expr.py`. 

*My questions for the team*
- Do you want to see more of these types of PRs? 
- If so, given the large number of similar errors, could we put this into a 'Good first issue' type of issue?

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


cvxpy/atoms/affine/add_expr.py
```diff
@@ -41,7 +41,7 @@ def shape_from_args(self) -> Tuple[int, ...]:
         """
         return u.shape.sum_shapes([arg.shape for arg in self.args])
 
-    def expand_args(self, expr):
+    def expand_args(self, expr: Any) -> List:
```
SteveDiamond, line 44: I believe expr has to be a cvxpy Expression

SteveDiamond:
Sure, this is helpful. We want all the Python code to have types. It could be a good first issue.

cvxpy/atoms/affine/add_expr.py
```diff
@@ -41,7 +42,7 @@ def shape_from_args(self) -> Tuple[int, ...]:
         """
         return u.shape.sum_shapes([arg.shape for arg in self.args])
 
-    def expand_args(self, expr):
+    def expand_args(self, expr: Expression) -> List:
```
SteveDiamond, line 45: List[Expression]

cvxpy/atoms/affine/add_expr.py
sunnygurm: Updated

cvxpy/atoms/affine/add_expr.py
SteveDiamond: Sorry not quite what I meant

sunnygurm: Oh! My fault, yeah that makes sense now that I look at the function return.

SteveDiamond (approved):

Issue #2310, opened 2023-12-21
add DAQP as a solver


ghorn:
This is a feature request to add DAQP (https://github.com/darnstrom/daqp) as a solver. It is available on pypi.

It's a modern dual active set method that is useful for control problems, including embedded optimization problems.

phschiele:
It would be nice to have an active set solver supported.
Note: this solver is also available via [qpsolvers](https://pypi.org/project/qpsolvers/)

enzbus:
According to the paper it's super fast on small to medium programs (it uses dense LA). Would love to have it interfaced through Cvxpy. One of the co-authors is also co-author of OSQP so it's likely to have a very similar interface.

phschiele:
Thanks again @enzbus, closing this issue.

PR #2309, opened 2023-12-21
Advanced features refactor

Transurgeon:
## Description
Please include a short summary of the change.
This PR introduces some updates to the documentation, especially making many changes to the advanced features section of the tutorial. 
- moved SDP, MIP, attributes to a new section titled ``Advanced Constraints``
- moved solver related stuff to a new section titled ``Solver Features``
- moved DPP and sensitivity analysis to a new section titled ``DPP``
- combined license, related projects, citations and short course into a new top level page titled ``other``
- removed content tree in FAQ since the new theme does it automatically
- other minor changes that use the sphinx immaterial theme features (ex: using collapsible instead of dropdown)
The documentation changes can be previewed [here](https://cvxpy--49.org.readthedocs.build/en/49/).

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #2312, opened 2023-12-24
DAQP solver initial interface

enzbus:
## Description
This is a first implementation of an interface to the DAQP active set (dense) QP solver, an open source solver published in 2023. This was requested in issue #2310.

## Notes
This is a straightforward implementation of the Python interface [described in the documentation](https://darnstrom.github.io/daqp/start/python). The main catch is that DAQP with its default parameters can't solve QPs whose cost matrix has zero eigenvalues (that includes LPs). It can do so if we change the ``eps_prox`` option; I implemented the following logic:

- check if the cost matrix is positive definite; if so leave default ``eps_prox`` = 0 which gives best performance (both numerical and execution time)
- if instead the cost matrix has null eigenvalues, set ``eps_prox=1e-5``; this was obtained by benchmarking with a typical use-case -> https://gist.github.com/enzbus/3d0236c7ed93cff5f0ec3f8587bcd67e

## Type of change
- [X ] New feature (backwards compatible)
- [ ] Bug fix

- [X ] Add our license to new files.
- [ X] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [X ] Run the unittests and check that they’re passing.


enzbus:
@SteveDiamond @phschiele @PTNobel @rileyjmurray any comments? Also @bstellato should be happy (OSQP is both faster and more robust). Thanks and happy holidays.

PTNobel:
Hey Enzo, I'm out of the country and won't be able to review this till after I'm back.

enzbus:
For more context here's another little benchmark, on linearly constrained least squares (so there's not proximal step involved). OSQP clearly in the lead.

https://gist.github.com/enzbus/426bd635d3dd7df559855c934495c899

doc/source/tutorial/advanced/index.rst
```diff
@@ -829,6 +831,51 @@ For others see `OSQP documentation <https://osqp.org/docs/interfaces/solver_sett
...
+``'fval_bound'``
+    Maximum allowed objective function value. The solver terminates if the dual
+    objective exceeds this value (since it is a lower bound of the optimal value,
+    default: 1e30).
+
+``'eps_prox'``
+    Regularization parameter used for proximal-point iterations (0 means that
+    no proximal-point iterations are performed). If the
+    cost matrix has a null eigenvalue, setting this to 0 (upstream's default)
+    makes DAQP fail. Note that Cvxpy's canonicalization procedure may add extra
```
SteveDiamond, line 870: Capitalize CVXPY

SteveDiamond (approved):
This looks great to me! Thanks, Enzo! Let's have a another review before we merge, but seems like everything is there.

enzbus:
Thanks! Fixed that, and DAQP's author approves this initial implementation (see https://github.com/darnstrom/daqp/issues/31 ). Thanks and happy new year!

cvxpy/reductions/solvers/defines.py
```diff
@@ -42,9 +42,10 @@
 from cvxpy.reductions.solvers.conic_solvers.xpress_conif import XPRESS as XPRESS_con
 from cvxpy.reductions.solvers.qp_solvers.copt_qpif import COPT as COPT_qp
```
PTNobel, line 43: Shouldn't this and the line below be moved down below the comment `# QP interfaces` too?

cvxpy/reductions/solvers/qp_solvers/daqp_qpif.py
```diff
@@ -0,0 +1,162 @@
...
+        # Map DAQP statuses back to CVXPY statuses
+        status = self.STATUS_MAP.get(exitflag, s.SOLVER_ERROR)
+
+        if status in s.SOLUTION_PRESENT:
+            opt_val = fval + inverse_data[s.OFFSET]
+            primal_vars = {
+                DAQP.VAR_ID:
+                intf.DEFAULT_INTF.const_to_matrix(np.array(xstar))
+            }
+            # need to skip unused variable bounds constraints
```
PTNobel, line 80: What are these unused variable bounds constraints? Why are they being passed to the solver?

PTNobel (commented):
Overall, this seems like a good interface; thank you Enzo!

I am a bit nervous about adding a "sparse -> dense" step in the solver interface. It feels like using this solver is a big performance footgun, and at the least it should be called out in the docs. Thoughts?

cvxpy/reductions/solvers/defines.py
enzbus: I don't remember but I think the import reorder was done by `ruff`, is there a logic I was missing?

cvxpy/reductions/solvers/qp_solvers/daqp_qpif.py
enzbus: To my understanding variable bounds are mandatory, this point also the author was particular about and asked whether Cvxpy can't handle those natively (I told him there are people looking into those but it's not there yet).

enzbus:
> Overall, this seems like a good interface; thank you Enzo!
> 
> 
> 
> I am a bit nervous about adding a "sparse -> dense" step in the solver interface. It feels like using this solver is a big performance footgun, and at the least it should be called out in the docs. Thoughts?

Agreed, I've done only limited execution time benchmarking but it didn't look to me that the sparse-dense conversion was too costly, worth investigating though. I don't see an alternative, though, other than perhaps factoring it out to be re-used by other interfaces, I think only one other at this point, or writing a new canonicalizer with dense output which is much more work I guess.

SteveDiamond:
For what it's worth, there is a dense canonicalization backend now if you use canon_backend=NUMPY_CANON_BACKEND

I don't think it's necessary or desirable to couple the canonicalization backend to the solver, but it might be worth trying this interface with that backend or documenting that the dense backend is recommended.

Also, I will post some thoughts about propagating variable bounds in the discord

enzbus:
> For what it's worth, there is a dense canonicalization backend now if you use canon_backend=NUMPY_CANON_BACKEND
> 
> I don't think it's necessary or desirable to couple the canonicalization backend to the solver, but it might be worth trying this interface with that backend or documenting that the dense backend is recommended.
> 
> Also, I will post some thoughts about propagating variable bounds in the discord

Ok! I should have thought of that, will do some more interesting benchmarks then. Also, what's the policy for `SolverStats.solve_time` and `SolverStats.setup_time` ? We put there the time reported by the solver (but I'm not sure it's accurate, probably ignores the time spent in its own Python interface)? Would it be better to time it in the interface? In any case will get back to it ASAP.

cvxpy/reductions/solvers/defines.py
PTNobel: I think the QP solver interfaces were supposed to be below the comment about `# QP solvers` and the conic programs above...

cvxpy/reductions/solvers/qp_solvers/daqp_qpif.py
PTNobel: Ah okay that makes sense.

PTNobel:
> For what it's worth, there is a dense canonicalization backend now if you use canon_backend=NUMPY_CANON_BACKEND
> 
> I don't think it's necessary or desirable to couple the canonicalization backend to the solver, but it might be worth trying this interface with that backend or documenting that the dense backend is recommended.

@SteveDiamond correct me if I'm wrong, but the NumPy backend will convert to sparse before passing it into the solver interface right? We should probably have a way to skip that conversion if the solver interface wants a dense object.

SteveDiamond:
> > For what it's worth, there is a dense canonicalization backend now if you use canon_backend=NUMPY_CANON_BACKEND
> > I don't think it's necessary or desirable to couple the canonicalization backend to the solver, but it might be worth trying this interface with that backend or documenting that the dense backend is recommended.
> 
> @SteveDiamond correct me if I'm wrong, but the NumPy backend will convert to sparse before passing it into the solver interface right? We should probably have a way to skip that conversion if the solver interface wants a dense object.

That could be right, I'm not sure though. @phschiele @Transurgeon how does the NumPy backend work in this regard?

phschiele:
> > > For what it's worth, there is a dense canonicalization backend now if you use canon_backend=NUMPY_CANON_BACKEND
> > > I don't think it's necessary or desirable to couple the canonicalization backend to the solver, but it might be worth trying this interface with that backend or documenting that the dense backend is recommended.
> > 
> > 
> > @SteveDiamond correct me if I'm wrong, but the NumPy backend will convert to sparse before passing it into the solver interface right? We should probably have a way to skip that conversion if the solver interface wants a dense object.
> 
> That could be right, I'm not sure though. @phschiele @Transurgeon how does the NumPy backend work in this regard?

That's right, the dense backend has the same (sparse) output format as the other backends. We'd need to have a flag for the output and also update a few transformations that happen after the backend (or at least, verify that all overloaded operators work for sparse and dense matrices).

enzbus:
Do you guys want to look into amending the numpy canonicalizer to produce a dense problem matrix? It would be cool but out of scope for this solver interface I guess, can always be added later. I haven't had time to benchmark it better yet but I was under the impression the sparse-dense conversion wasn't too costly. As an aside, the solver author mentioned he was interested in `cvxpygen`, I guess having a dense canonicalizer in CVXPY would be helpful for that side project too...

SteveDiamond:
We can look into that later. This is good to go.

SteveDiamond:
I did some work on propagating variable upper and lower bounds to the solver:
https://github.com/cvxpy/cvxpy/pull/2321

If the general design seems good, it will be easy to integrate with the DAQP interface.

enzbus:
Hello @Transurgeon, looks like #2309 overwrote the changes to `doc/advanced/solvers/index.rst` introduced by this PR (adding DAQP). Should be fixable by a copy-paste but maybe you should have a look. Cc @SteveDiamond

enzbus:
Thank you @Transurgeon @SteveDiamond ! PS I encountered the same issue recently (splitting a git-tracked file into many ones breaks automatic merge) and with some digging I found the solution, if you use [this script](https://github.com/cvxgrp/cvxportfolio/blob/master/git-split.sh) it keeps the correct git history and allows for automatic merge.

PR #2313, opened 2023-12-25
change some O(n) lookups in scip_conif.py to O(1)

onlyblackstars:
## Description
Minor change for SCIP wrapper that reduces some O(n) lookups to O(1). The lookups get called as many times as there are variables by get_variable_type (in scip_conif.py). 

(Anecdotally, this reduced my create_variables() time from 10 minutes to 0.5 seconds.)

## Type of change
- [X] Bug fix

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scip_conif.py
@@ -125,8 +125,8 @@ class SCIP(ConicSolver):
 
         # data, inv_data = super(SCIP, self).apply(problem)
         variables = problem.x
-        data[s.BOOL_IDX] = [int(t[0]) for t in variables.boolean_idx]
-        data[s.INT_IDX] = [int(t[0]) for t in variables.integer_idx]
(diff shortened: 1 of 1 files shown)


PTNobel (approved):
Overall looks good to me; is there a reason not to use set constructor notation?

onlyblackstars:
> Overall looks good to me; is there a reason not to use set constructor notation?

No reason except to convey intent. The variables `data[s.BOOL_IDX]` and `data[s.INT_IDX]` in some other `ConicSolver` subclasses get passed up as lists (though there's no apparent requirement that they actually be lists/Sequences) and someone might one day think that, e.g., the one use of `{...}` in scip_conif.py was unintentional. 🤷‍♂️

phschiele (approved):
Fine with me to use `set()` to make this choice more explicit.

PTNobel:
That's a good point; could you add a comment explaining the benefit of this optimization to make sure it survives future editing?

SteveDiamond (approved):

PR #2315, opened 2023-12-31
avoid sparse identity matrix multiply in conic_solver.py

ghackebeil:
Compiling a "highly" parameterized linear program seems to hit a slow path involving a number of scipy sparse matrix multiplication operations within the ConicSolver reduction. By avoiding multiplication altogether for the trivial cases involving identity and negative identity operations, the initial compilation path for a parameterized linear program can be sped up by orders of magnitude.

As an example, for a relatively sparse problem with ~7k variables, ~20k constraints and ~7k parameters (all on the rhs of constraints), with this change the initial cvxpy compile time goes from ~140s to 0.4s (**350x speedup**).

pr_benchmark comments:
```
Benchmarks that have improved:

       before           after         ratio
     [8d1c56d4]       [987310df]
-         2.65±0s          2.40±0s     0.91  simple_LP_benchmarks.SimpleFullyParametrizedLPBenchmark.time_compile_problem
-         425±0ms          248±0ms     0.58  high_dim_convex_plasticity.ConvexPlasticity.time_compile_problem
```

## Type of change
- [x] performance enhancement

- [ ] ~Write unittests~ (looks like there's already a benchmark that covers this, seems to have improved)
- [x] Run the unittests and checked that they’re passing (missing some solvers locally though)

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
+++ b/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
@@ -44,6 +44,19 @@ class LinearOperator:
     def __call__(self, X):
         return self._matmul(X)
 
+class IdentityOperator(LinearOperator):
+    """A wrapper for the identity operator."""
+    def __init__(self, n):
+        self.shape = (n,n)
+    def __call__(self, X):
+        return X
+
+class NegativeIdentityOperator(LinearOperator):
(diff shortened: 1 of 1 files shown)


SteveDiamond (approved):
Thanks for finding this!

PR #2319, opened 2024-01-10
Use new NumPy API

phschiele:
## Description
Thanks @PTNobel for the ping about https://numpy.org/devdocs/numpy_2_0_migration_guide.html
I ran `ruff check . --preview` locally, as the "NPY201" rule is still in preview. Nevertheless, I added it to the linting rules so it will become active at some point.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel (approved):
Changes are fine, but ruff is failing in CI.

PR #2321, opened 2024-01-12
Propagate variable bounds to the solver

SteveDiamond:
## Description
This MR adds a flag to solvers that indicates whether they have special handling for upper and lower bounds on variables. If they do have special handling, then variable bounds are passed through to the ParamConeProg object where problem data is stored for the solver. Variable bounds are also passed through to ParamQuadProg for the QP pathway.

The DAQP interface is updated to use variable bounds. Other interfaces can be updated in future MRs.

Issue link (if applicable): #2312

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -145,20 +145,20 @@ def _reductions_for_problem_class(problem, candidates, gp: bool = False, solver_
...
-        if not candidates['conic_solvers']:
-            raise SolverError("Problem could not be reduced to a QP, and no "
-                              "conic solvers exist among candidate solvers "
-                              "(%s)." % candidates)
-
+    # Special reduction for finite set constraint, 
+    # used by both QP and conic pathways.
     constr_types = {type(c) for c in problem.constraints}
     if FiniteSet in constr_types:
         reductions += [Valinvec2mixedint()]
```
SteveDiamond, line 152: This reduction now happens earlier in the QP canonicalization pathway, but I think it's fine.

cvxpy/tests/test_constraints.py
```diff
@@ -396,6 +396,15 @@ def test_nonneg_dual(self) -> None:
         prob.solve(solver=cp.OSQP)
         self.assertItemsAlmostEqual(prob.constraints[0].dual_value, dual)
 
+    def test_bound_properties(self) -> None:
```
SteveDiamond, line 399: We forgot to implement domain for bounded variables, so I added that functionality.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -87,13 +99,31 @@ def lower_value(variable, value):
 class CvxAttr2Constr(Reduction):
     """Expand convex variable attributes into constraints."""
 
+    def __init__(self, problem=None, ignore_bounds: bool = False) -> None:
```
phschiele, line 102: Maybe `reduce_bounds` would be easier to understand.

SteveDiamond: Done.

SteveDiamond: I also added a test for when the bounds are not supposed to be passed to the solver.

SteveDiamond:
@enzbus could you help add support to the DAQP interface for variable bounds? I got the bound info going into the interface but I don't understand the solver formatting super well.

enzbus:
Congrats @SteveDiamond , this is great! I had no idea it was this close already. Let me try it with DAQP.

enzbus:
What about the dual variables associated with the variable bounds @SteveDiamond , how to we pass them back?

SteveDiamond:
> What about the dual variables associated with the variable bounds @SteveDiamond , how to we pass them back?

Hmm not sure. I think the most correct thing would be to attach them to the ``variable.domain`` constraints. I'll see how hard that would be.

SteveDiamond:
> > What about the dual variables associated with the variable bounds @SteveDiamond , how to we pass them back?
> 
> Hmm not sure. I think the most correct thing would be to attach them to the `variable.domain` constraints. I'll see how hard that would be.

This is too difficult for this PR. So the functionality in this PR would not include passing back dual variables for variable attributes. Currently cvxpy doesn't support this anyway, so this isn't a regression.

SteveDiamond:
This is enough work for this MR IMO. Future MRs can add variable bound support to more solver interfaces. I looked into adding it to the mosek interface, but the dualization makes it a bit more complicated. Also there is a rewrite of the mosek interface in the works so better not to mess with it for now.

PTNobel:
> the dualization makes it a bit more complicated

If we switch to the conic interface we no longer need the dualization right?

rileyjmurray:
MOSEK’s “affine” conic interface (as opposed to its traditional “direct”
conic interface) is implemented just by adding slack variables. There are
pretty fundamental reasons why this will make problems more computationally
challenging given how CVXPY represents problems. So  I anticipate
dualization being something we want in place for the foreseeable future.
Happy to revisit if others think differently.

SteveDiamond:
I agree with @rileyjmurray, we will always need dualization. The Mosek devs actually suggested that we have logic that decides whether to dualize or not, since for some SDPs dualization makes the problem slower (and Mosek has automatic dualization for SOCPs and LPs). I think these are SDPs in LMI standard form? I would need to dig up the issue.

PTNobel:
Ah okay; sorry I misunderstood

enzbus:
> I agree with @rileyjmurray, we will always need dualization. The Mosek devs actually suggested that we have logic that decides whether to dualize or not, since for some SDPs dualization makes the problem slower (and Mosek has automatic dualization for SOCPs and LPs). I think these are SDPs in LMI standard form? I would need to dig up the issue.

My 2 cents on Mosek intf from the finance community: just make sure any big change on it is approved by the Mosek people. I had more than one discussion with folks claiming that CVXPY+Mosek is slower than native Mosek, they shut up when I tell them that the CVXPY Mosek interface is maintained by Mosek employees, and performance difference is not an issue.

cvxpy/expressions/leaf.py
```diff
@@ -273,18 +274,73 @@ def is_complex(self) -> bool:
...
+    def _has_lower_bounds(self) -> bool:
+        """Does the variable have lower bounds?"""
+        if self.is_nonneg():
+            return True
+        elif self.attributes['bounds'] is not None:
+            lower_bound = self.attributes['bounds'][0]
+            if np.isscalar(lower_bound):
+                return lower_bound != -np.inf
+            else:
+                return np.any(lower_bound != -np.inf)
```
phschiele, line 286: Is there an else case missing here?
```suggestion
            else:
                return np.any(lower_bound != -np.inf)
        else:
            return False
```

cvxpy/expressions/leaf.py
```diff
@@ -273,18 +274,73 @@ def is_complex(self) -> bool:
...
+    def _has_upper_bounds(self) -> bool:
+        """Does the variable have upper bounds?"""
+        if self.is_nonpos():
+            return True
+        elif self.attributes['bounds'] is not None:
+            upper_bound = self.attributes['bounds'][1]
+            if np.isscalar(upper_bound):
+                return upper_bound != np.inf
+            else:
+                return np.any(upper_bound != np.inf)
```
phschiele, line 297: Same here

cvxpy/tests/test_param_quad_prog.py
```diff
@@ -107,3 +110,56 @@ def test_qp_problem(self) -> None:
...
+        x = cp.Variable((3, 2), bounds=[lower_bounds, upper_bounds])
+        problem = cp.Problem(cp.Minimize(cp.sum(x)))
+        data, _, _ = problem.get_problem_data(solver=solver_instance)
+        param_quad_prog = data[cp.settings.PARAM_PROB]
+
+        assert np.all(param_quad_prog.lower_bounds == lower_bounds)
+        param_upper_bound = np.reshape(param_quad_prog.upper_bounds, (3, 2), order="F")
+        assert np.all(param_upper_bound == upper_bounds)
+
+    @unittest.skipUnless(cp.DAQP in INSTALLED_SOLVERS, 'DAQP is not installed.')
```
phschiele, line 147: I changed to use the `skipUnless` syntax instead of an if clause in the test to prevent a false sense of coverage.

cvxpy/expressions/leaf.py
SteveDiamond: Added.

PR #2326, opened 2024-01-15
Better integration with Mosek>=10; remove Mosek<=9 code

aszekMosek:
As discussed on Discord, this PR slightly modernizes the Mosek interface, getting rid of the ``env`` object (Mosek environment), which was only dragged around through the code, and instead only creating a Mosek task ``task``. A default global environment will be created internally by the library, once for the whole process. See https://docs.mosek.com/latest/pythonapi/environment-and-task.html for more info.

In practice this change will only make a difference for users of commercial floating licenses. They will have one license checkout per process, instead of per solve, ie. the user experience should only improve, judging from some communication we had with users in the past. In any case, should this ever come up as a problem, this is a business issue Mosek support takes responsibility for. Otherwise the only difference is code simplicity.

Since this change **only works with Mosek >= 10**, I went ahead and removed all code parts only relevant for Mosek <= 9. Releasing this code makes CVXPY depend on Mosek>=10.


SteveDiamond (approved):
Thanks!

Issue #2282, opened 2023-10-31
Mutating NumPy array inplace with cp.Expression should have a clear error message 


PTNobel:
Currently, when calling a NumPy function on a CVXPY variable such as `np.log(cp.Variable())`, we raise an error telling people to use a CVXPY atom inside. This error incorrectly triggers on the following code:

```
import numpy as np
import cvxpy as cp

x = np.array([0])
x += cp.Variable()
```

This errors because of [this line](https://github.com/cvxpy/cvxpy/blob/4e76e6075c40367076429973e2116cf2c0daa875/cvxpy/expressions/expression.py#L745), since NumPy sets `kwargs = {'out': x}`. 

**Expected behavior**

It isn't clear to me what we want to do here. On the one hand people might expect this to mutate `x` the way it would for a non-NumPy array. It may not possible to both keep the error and make this have the same behaviour as `x = x + cp.Variable()`.

We could add a new error message that says to do `x = x + expression` in this case, I think this is probably better than trying to get this to work because of my concerns about people expecting mutation. It is possible to write a function that works when given a NumPy array but which depends on mutation that doesn't have the same behaviour when given a CVXPY expression and I think we want to make this an error. 

**Version**
 - OS: Linux
 - CVXPY Version: 1.4.1

rileyjmurray:
I think the most we'd want to do is raise an error saying ``x = x + expression``. The idea of an in-place assignment on a numerical array with changing data types doesn't even make sense. Consider the following code as a non-cvxpy example:
```
>>> import numpy as np
>>> a = np.array([0],'d')
>>> a
array([0.])
>>> a += 1j
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('complex128') to dtype('float64') with casting rule 'same_kind'
```

PR #2332, opened 2024-01-20
Implements inplace mutation error

PTNobel:
## Description

Implementing @rileyjmurray's suggestion from #2282 that we should error when someone attempts inplace mutation of a CVXPY expression.

This is a breaking change for the following code:

```python
x = cp.Variable()
x += 1
``` 

It also improves the error message on code we broke a few releases ago:

```python
x = cp.Variable(4)
c = np.array([1, 2, 3, 4])
x += c
```

It is possible to not break the first code while improving the error messages on the other code.

Once we have consensus on what we want this to do, I'll write unit tests.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -68,6 +68,14 @@ errors or code that doesn't behave as expected. Consider using one of the
 functions documented here: https://www.cvxpy.org/tutorial/functions/index.html
 """
 
+__INPLACE_MUTATION_ERROR__ = """
+You're trying to mutate a CVXPY expression inplace. This is prone to errors or
+code that doesn't behave as expected. Consider implementing replacing, for example:
+> x += 1
+with
+> x = x + 1
+"""
+
 __ABS_ERROR__ = """
 You're calling the built-in abs function on a CVXPY expression. This is not
 supported. Consider using the abs function provided by CVXPY.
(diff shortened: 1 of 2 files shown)


SteveDiamond:
I would prefer that we break as little code as possible. It's definitely a good catch that inplace isn't playing well with NumPy, but I don't think we need to ban ``x += 1`` if we can make it work.

phschiele:
From looking at the linked issue, it seems that two cases are being mixed. Intuitively, I would expect the following:
```py
x = cp.Variable(4)
c = np.array([1, 2, 3, 4])

c += x  # fails, for the reason described by Riley
x += c  # results in x = x + c
```
This appears to be the current behavior. Roughly speaking, my expectation is that the in-place assignment works if the lhs can represent the rhs of the operator. A cvxpy expression can represent a constant array, but a numpy array can't represent a CVXPY variable.

PTNobel:
Alright, i need to email the NumPy mailing list, since they think we are having an XY problem. I'll try and do so this weekend?

PTNobel:
@phschiele @SteveDiamond can one of you help me figure out what is going on with pytest? I'm very confused.

phschiele:
> @phschiele @SteveDiamond can one of you help me figure out what is going on with pytest? I'm very confused.

Should be

```py
match=re.escape(__INPLACE_MUTATION_ERROR__)
```

because of special characters.

PTNobel:
Okay, I think that this is ready to merge @phschiele @SteveDiamond

phschiele (approved):
Thanks for this @PTNobel, let's merge once the final CI run passes.

PR #2334, opened 2024-01-22
Removes pybind11 runtime dependency

### LINKED ISSUE #2330: Unnecessary pybind11 dependency?
https://github.com/cvxpy/cvxpy/pull/2125 (released in 1.4.0) added `pybind11` as a dependency, but it seems like it is just a build dependency, not a run time dependency?

One piece of evidence is that the cvxpy conda-forge feedstock did not add pybind11 as a dependency [here](https://github.com/conda-forge/cvxpy-feedstock/pull/83) and has gotten no complaints. I noticed this because I maintain a downstream feedstock that runs `pip check` as a test and that fails for cvxpy 1.4 because pybind11 is not installed.

phschiele:
## Description
Removes pybind11 runtime dependency from `requirements.txt` and `pyproject.toml`.
Issue link (if applicable): resolves #2330 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ dependencies = [
     "scs >= 3.2.4.post1",
     "numpy >= 1.15",
     "scipy >= 1.1.0",
-    "pybind11"
 ]
 requires-python = ">=3.8"
 urls = {Homepage = "https://github.com/cvxpy/cvxpy"}
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,3 @@ ecos
 scs
 numpy
 scipy
-pybind11


SteveDiamond (approved):
LGTM

phschiele:
@rileyjmurray The only potential downside I saw was that my IDE complains about this line now:
https://github.com/cvxpy/cvxpy/blob/0a8881d9f8ea38a54d992fd6f0cca0db41fda137/setup/extensions.py#L18

Confirmed that running `pip install -e .` works without issues.

SteveDiamond:
There is a CI failure but it is transient. Happened to me earlier. Maybe conda is having issues?

phschiele:
Looks like there are no objections, merging here.

PR #2338, opened 2024-01-30
Improving the Gurobi integration by supporting further error codes

jmarecek:
## Description

CVXPY currently supports only a handful of error codes of Gurobi, which results in translation of many runs in an error, while those are really "user limit" terminations (of which there are many kinds with many different error codes). 

## Type of change
- [X] Bug fix

There are no new files.


SteveDiamond:
Nice! Can you remove all the changes besides updating the status maps?

PTNobel (approved):

PR #2350, opened 2024-02-12
Address deprecation warning

usamamuneeb:
## Description

The solver will accept `numpy.ndarray` as well as one of the sparse array formats (e.g. `csr_array` or `csc_array`) in `scipy.sparse` for each of `A`, `b` and `c`.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
This isn't working with SDPA in the CI: https://github.com/cvxpy/cvxpy/actions/runs/7878652690/job/21497355489?pr=2350
Is this only with a new version of SDPA?

usamamuneeb:
For some reason it fetched the old version (0.2) instead of 0.2.1 for that specific job. We should rerun that specific job (all others are passing).

Yes, I created a patch release which is backward compatible with CVXPY before and after this change.

SteveDiamond (approved):
Thanks!

Issue #2353, opened 2024-02-20
A seperate verbose switch for solvers


hailiangliu89:
is it possible to set verbose=False for cvxpy only, and verbose=True for the solvers? 

I am using cplex as my solver and it seems the switch should be added to prob.solve(verbose=False, solver=cp.cplex, cplex_params={}). Just wondering if anyone knows the parameter name that should be passed to the dictionary here.

PR #2354, opened 2024-02-20
add solver-only verbose switch as kwarg

hailiangliu89:
## Description
adds a solver-only verbose switch as kwarg.

the old behaviour is kept if the user does not provide `solver_verbose` as a kwarg.

Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/2353

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/doc/source/tutorial/solvers/index.rst
+++ b/doc/source/tutorial/solvers/index.rst
@@ -175,7 +175,7 @@ Use the ``installed_solvers`` utility function to get a list of the solvers your
 Viewing solver output
 ^^^^^^^^^^^^^^^^^^^^^
 
-All the solvers can print out information about their progress while solving the problem. This information can be useful in debugging a solver error. To see the output from both cvxpy and the solvers, set ``verbose=True`` in the solve method. If you want to see the output from the solver only, set ``solver_verbose=True``.
+All the solvers can print out information about their progress while solving the problem. This information can be useful in debugging a solver error. To see the output from both CVXPY and the solvers, set ``verbose=True`` in the solve method. If you want to see the output from the solver only, set ``solver_verbose=True``.
 
 .. code:: python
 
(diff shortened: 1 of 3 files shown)


phschiele:
Hi @hailiangliu89, thanks for your contribution! This makes for a neat way to give more granular logging control. 

It looks like you've removed the [checklists](https://github.com/cvxpy/cvxpy/blob/master/.github/pull_request_template.md) from the PR template. Those checklists help us select relevant PRs for new releases or patches, so it would be great if you could bring them back. This PR would also need documentation and a test. Here is our current test of the verbose keyword: 
https://github.com/cvxpy/cvxpy/blob/fc6116e85e91ec3da88283d86834e6998a1e6a37/cvxpy/tests/test_problem.py#L288

PTNobel:
Hey Hailiang, can you provide some more justification for why solver_verbose is needed? I think of the CVXPY verbosity as not very long or impactful on runtime.

hailiangliu89:
> Hey Hailiang, can you provide some more justification for why solver_verbose is needed? I think of the CVXPY verbosity as not very long or impactful on runtime.

Inneed the cvxpy verbosity is not that long and very useful when debugging, but our end-users usually only care about time spent on solvers and do not understand the reduction processes carrried out by cvxpy. Also some reported to us that the compilation printouts comfused them with the original problem.

SteveDiamond:
I see, you are using CVXPY under the hood for another library. I think this feature is reasonable enough.

LiaoSW (approved):
this switch is just what we need for our end-user software.

doc/source/tutorial/solvers/index.rst
```diff
@@ -175,7 +175,7 @@ Use the ``installed_solvers`` utility function to get a list of the solvers your
 Viewing solver output
 ^^^^^^^^^^^^^^^^^^^^^
 
-All the solvers can print out information about their progress while solving the problem. This information can be useful in debugging a solver error. To see the output from the solvers, set ``verbose=True`` in the solve method.
+All the solvers can print out information about their progress while solving the problem. This information can be useful in debugging a solver error. To see the output from both cvxpy and the solvers, set ``verbose=True`` in the solve method. If you want to see the output from the solver only, set ``solver_verbose=True``.
```
SteveDiamond, line 178: capitalize CVXPY

SteveDiamond (approved):
Looks good other than one minor comment

PR #2363, opened 2024-02-26
Allow pure Python builds

phschiele:
## Description
We recently discussed the possibility to run CVXPY within [Pyodide](https://pyodide.org). 
Pure Python wheels are directly supported, and since we a backend that only relies on SciPy, which is already supported by Pyodide, this should be fairly easy.

This PR moves all calls to the C++ backend into a separate file and makes the Eigen-based sparse Cholesky optional

Issue link (if applicable): NA

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -22,6 +22,11 @@ import numpy as np
 import scipy.sparse as sp
 
 import cvxpy.settings as s
+try:
+    import cvxpy.cvxcore.python.cvxcore as cvxcore
+except ModuleNotFoundError:
+    # Allow to run without cvxcore, cannot use C++ backend
+    cvxcore = None
 from cvxpy.lin_ops import lin_op as lo
 from cvxpy.lin_ops.canon_backend import CanonBackend
 
@@ -294,11 +299,6 @@ def get_problem_matrix(linOps,
     canon_backend = default_canon_backend if not canon_backend else canon_backend
 
     if canon_backend == s.CPP_CANON_BACKEND:
-        try:
-            import cvxpy.cvxcore.python.cvxcore as cvxcore
-        except ModuleNotFoundError:
-            raise ImportError("cvxcore is not installed. Please install cvxcore "
-                              "to use the C++ backend.")
 
         lin_vec = cvxcore.ConstLinOpVector()
 
@@ -405,7 +405,6 @@ def get_type_map() -> dict:
     """
     Returns a map from string type to cvxcore type
     """
-    import cvxpy.cvxcore.python.cvxcore as cvxcore
     type_map = {
         "VARIABLE": cvxcore.VARIABLE,
         "PARAM": cvxcore.PARAM,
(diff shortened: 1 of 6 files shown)


cvxpy/cvxcore/python/canonInterface.py
```diff
@@ -251,29 +250,6 @@ def get_matrix_from_tensor(problem_data_tensor, param_vec,
     return (A, b)
 
 
-def get_matrix_and_offset_from_unparameterized_tensor(problem_data_tensor,
```
phschiele, line 254: This functions appears to be unused.

SteveDiamond:
> @SteveDiamond Looks like there is in fact a small performance regression. Alternatively, we could replace the top level import by
> 
> ```python
> try:
>     import cvxpy.cvxcore.python.cvxcore as cvxcore
> except ModuleNotFoundError:
>     cvxcore = None
> ```
> 
> The only downside is that if you do indeed try to use the C++ backend without it being installed you'd get a less descriptive `None object has no attribute ....` error. Since right now we don't actively expose the option to build without the C++ backend, that would be fine with me as well. Alternatively, I could try a refactor where I split `canonInterface.py` into the functionality shared between backends and everything that actually calls cvxcore.

This quick fix is fine by me.

phschiele:
@rileyjmurray @SteveDiamond How do we feel about making the Eigen-based sparse Cholesky optional?

Here is an attempt:
https://github.com/cvxpy/cvxpy/pull/2363/commits/f9ea0b03510dd07ebad228193a59e60a0b57d7fe

Catching the `ModuleNotFoundError` at the call site does not seem optimal. However, since we use that function once, adding another layer of abstraction may be overkill. Let me know if you have a strong preference here.

rileyjmurray:
I’m okay with making sparse Cholesky optional if the goal is to make CVXPY
usable as a pure Python package.

(Strictly speaking I hope no one uses numerical solvers written in pure
Python, but it’s fair for us to get out of the way for people who really
really care about minimizing non-Python dependencies.)

PTNobel:
We might have an easier time getting QDLDL into Pyodide. So it could be nice to implement using that for the sparse Cholesky.

Riley, the proposal is to get Clarabel compiling to WASM not to use a pure Python numerical solvers.

PTNobel:
I think Riley replied to my question elsewhere: https://github.com/cvxpy/cvxpy/discussions/2359#discussioncomment-8599577

phschiele:
@SteveDiamond I did have to move to another file as the regression was coming from moving the `TYPE_DICT` inside a function.

I think the structure is cleaner now; functions inside `canonInterface.py` are called throughout the code base, but everything in `cppbackend.py` is only called from `canonInterface.py`.

cvxpy/cvxcore/python/cppbackend.py
```diff
@@ -0,0 +1,261 @@
...
+
+try:
+    import cvxpy.cvxcore.python.cvxcore as cvxcore
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "Tried using the C++ backend, but the cvxcore module was not installed."
+    )
+
+
+def build_matrix(
```
phschiele, line 34: No functional change to any of these functions, just moved them to a separate file.

PTNobel (approved):

PR #2370, opened 2024-03-05
Improved support for Gurobi environments

jmarecek:
## Description

Gurobi is a commercial solver, with an academic license, trial license, etc. Three years ago, @sjschneider added support for instatiating the model with a custom environment, which would have the license details pre-loaded. It is non-obvious from the documentation how to set the environment up, though. 

This pull request fixes gurobi_conif.py and gurobi_qpif.py such that even without specifying the environment, the license is loaded from the correct environment variables. 

## Type of change
- [X] Bug fix

- [ ] Write unittests.


cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
```diff
@@ -46,15 +47,17 @@ class GUROBI(ConicSolver):
...
-                  7: s.SOLVER_ERROR,
-                  8: s.SOLVER_ERROR,
-                  # TODO could be anything.
-                  # means time expired.
-                  9: s.USER_LIMIT,
-                  10: s.SOLVER_ERROR,
-                  11: s.SOLVER_ERROR,
-                  12: s.SOLVER_ERROR,
-                  13: s.SOLVER_ERROR}
+                  7: s.USER_LIMIT, # ITERATION_LIMIT
```
SteveDiamond, line 50: I'm confused about why this is a delta. This change should be in master. Maybe you need to sync with master?

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -143,8 +147,32 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
+            license_company = os.getenv("GUROBI_LICENSE_COMPANY", None)
+            license_project = os.getenv("GUROBI_LICENSE_PROJECT", None)
+            license_expiration_date = os.getenv("GUROBI_LICENSE_EXPIRATION_DATE", None)
+            license_code = os.getenv("GUROBI_LICENSE_CODE", None)
+            model = None
+            if (license_company is None
+                or license_project is None
+                or license_expiration_date is None
+                or license_code is None):
+                print("Note: License parameters can be provided in enviroment "
```
SteveDiamond, line 161: This should not print if verbose=False. Also I don't understand the meaning of the message.

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -143,8 +147,32 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
             del solver_opts['env']
             model = grb.Model(env=default_env)
         else:
-            # Create Gurobi model using default (unspecified) environment
-            model = grb.Model()
+            # Alternatively, reads the license off the environment variables
+            log_file = os.getenv("GUROBI_LOG_FILE", "gurobi.log")
```
SteveDiamond, line 151: Can you factor out/clean up this code somehow?

SteveDiamond:
Can you explain with a more concrete example what this PR is enabling?

sjschneider:
I think it's bad form to have cvxpy read in environment variables like this.  To the extent the mechanism for providing the Gurobi Environment was unclear, I think this code belongs as just an example.

phschiele:
> I think it's bad form to have cvxpy read in environment variables like this. To the extent the mechanism for providing the Gurobi Environment was unclear, I think this code belongs as just an example.
> 
> I should also note that the Gurobi Env API has not been static and so this puts an additional burden on cvxpy for a specific license type that will not be easy to test.

Considering the concerns about testing, do we want to proceed with this PR? 
I'd prefer simply adding a line to the docs that shows the syntax for using the (existing) environment feature.

Transurgeon:
I'll close this PR since it is very outdated.. let's revisit this (and potential also improve the docs) if it becomes of interest again.

Issue #2336, opened 2024-01-27
Introduce "vdot" as an alias (or replacement) for scalar_product


rileyjmurray:
I wrote the ``scalar_product`` atom a while back to make computing things like ``trace(A @ B)`` more efficient. That atom also handles complex conjugation of the first argument when necessary. It turns out that numpy already has its own function for this, called [``vdot``](https://numpy.org/doc/stable/reference/generated/numpy.vdot.html).

For consistency with NumPy I suggest we introduce a ``vdot`` atom that has the same semantics as``scalar_product``. Supposing we do this, there's a question of whether we should keep ``scalar_product`` as a (much more sensibly named ...) alias for ``vdot``, or if we should deprecate ``scalar_product`` for removal in cvxpy 2.0.

Thoughts, @PTNobel, @phschiele, @SteveDiamond?

PTNobel:
1. I don't think scalar_product is a better name. I think the vdot name makes more sense.
2. My soft preference is to mirror the NumPy API.
3. I would also love to just have a performance optimization that makes the trace form behave similarly... Writing a design doc for this has been on my to-do list for a little bit

SteveDiamond:
I agree on introducing vdot. I don't think it's necessary to deprecate scalar_product, but don't feel strongly about this. I like @PTNobel 's suggestion about catching ``trace(A @ B)`` internally and replacing it with vdot.

rileyjmurray:
Re, Parth's point 1: I actually just figured out what's meant by the ``v`` in ``vdot``. That indeed makes sense! I also agree with Parth's other two points.

So it sounds like we're in agreement about adding ``vdot`` and using it to more efficiently handle ``trace(A @ B)`` (like we do with ``quad_form`` to handle ``x.T @ P @ x``). The question remains as to whether we should deprecate ``scalar_product``. I think I have a soft preference for leaving it in place, but removing it from the web documentation.

rileyjmurray:
Resolved by #2371.

PR #2371, opened 2024-03-05
Replace cp.scalar_product with cp.vdot

Transurgeon:
## Description
This PR attempts to introduce cp.vdot as a replacement for cp.scalar_product. 
In the issue below, it was suggested to not deprecate cp.scalar_product, but instead remove it from the documentation.
When running ``grep -ir "scalar_product"`` I saw that there were a lot of instances under ``cvxcore``, I am not sure if the code there also needs to be updated.
Issue link (if applicable): #2336 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -455,10 +455,10 @@ def vdot(x, y):
 
 
 def scalar_product(x, y):
-    x = deep_flatten(x)
-    y = deep_flatten(y)
-    prod = multiply(conj(x), y)
-    return cvxpy_sum(prod)
+    """
+    Alias for vdot.
+    """
+    return vdot(x, y)
 
 
 def outer(x, y):
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -422,7 +422,7 @@ class DivExpression(BinaryOperator):
         return (lu.div_expr(arg_objs[0], arg_objs[1]), [])
 
 
-def scalar_product(x, y):
+def vdot(x, y):
     """
     Return the standard inner product (or "scalar product") of (x,y).
 
@@ -448,6 +448,13 @@ def scalar_product(x, y):
     real scalars), then this function returns an Expression representing
     ``a * c + b * d``.
     """
+    x = deep_flatten(x)
+    y = deep_flatten(y)
+    prod = multiply(conj(x), y)
+    return cvxpy_sum(prod)
+
+
+def scalar_product(x, y):
     x = deep_flatten(x)
     y = deep_flatten(y)
     prod = multiply(conj(x), y)
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -1270,18 +1270,15 @@ class TestAtoms(BaseTest):
         """Test that norm1 and normInf match definition for matrices.
         """
         A = np.array([[1, 2], [3, 4]])
-        print(A)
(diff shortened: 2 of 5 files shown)


cvxpy/tests/test_atoms.py
```diff
@@ -1349,14 +1349,14 @@ def test_log_normcdf(self) -> None:
             -result, 4 * np.log(scipy.stats.norm.cdf(2)), places=None, delta=1e-2
         )
 
-    def test_scalar_product(self) -> None:
+    def test_vdot(self) -> None:
```
SteveDiamond, line 1352: If we're not going to delete scalar_product, we should keep the test for scalar_product and have an additional test for vdot (which can be the same as what you have here)

SteveDiamond:
Thanks for doing this! It looks like there are some unrelated test failures due to an update to the COPT solver. They might be fixed by changing the test tolerance a bit. Could you try that in this MR?

cvxpy/atoms/affine/binary_operators.py
```diff
@@ -454,6 +454,13 @@ def scalar_product(x, y):
     return cvxpy_sum(prod)
 
 
+def scalar_product(x, y):
+    x = deep_flatten(x)
+    y = deep_flatten(y)
+    prod = multiply(conj(x), y)
+    return cvxpy_sum(prod)
```
phschiele, line 461: Maybe this is simpler:

```suggestion
def scalar_product(x, y):
    """
    Alias for vdot.
    """
    return vdot(x, y)
```

cvxpy/tests/test_atoms.py
phschiele: @Transurgeon You can mark the test parametrized with `@pytest.mark.parametrize("func", [cp.scalar_product, cp.vdot]` and then use `func()`in the body of the test.

cvxpy/atoms/affine/binary_operators.py
Transurgeon: Yes I agree, thanks for the suggestion.

Transurgeon:
> Thanks for doing this! It looks like there are some unrelated test failures due to an update to the COPT solver. They might be fixed by changing the test tolerance a bit. Could you try that in this MR?

I got this error for ``test_sdp_2`` for COPT:
```
E   AssertionError: 21.04819569795163 != 21.04711571 within 3 places (0.0010799879516305566 difference)
```
Changing the accuracy from 3 decimals to 2 fixes the error (see below):
```
    def test_sdp_2(solver, places: int = 2, duals: bool = True, **kwargs) -> SolverTestHelper:
```
Is that a good fix for the failing tests?

cvxpy/tests/test_atoms.py
Transurgeon: Hey Philipp, I tried but was unsuccessful. I believe we are not allowed to pass in callables as parameters to ``pytest.mark.parametrize`` (also I am not sure how we would deal with passing in the parameters ``p`` and ``v``. 
To avoid having to create a new test, do you think I can just add another assertion in the same test?

phschiele: You're right, the arguments must be serialisable. So you could pass strings that are then looked up in a dict to get the function handles, but I'm totally fine with additional assertions like:

```py
assert np.allclose(cp.vdot(v, p).value, cp.scalar_product(v,p).value)
```

SteveDiamond (approved):
LGTM

SteveDiamond:
Merging. CI failures unrelated.

PR #2388, opened 2024-03-23
Next step of the ECOS deprecation

PTNobel:
## Description

Adds the new warning for when ECOS is specified as a solver and makes Clarabel preferred, unless it is an SDP, then prefers SCS.

I still need to update the tests...

### DIFF
--- a/cvxpy/reductions/solvers/solving_chain.py
+++ b/cvxpy/reductions/solvers/solving_chain.py
@@ -60,7 +60,7 @@ ECOS_DEP_DEPRECATION_MSG = (
     You specified your problem should be solved by ECOS. Starting in
     CXVPY 1.6.0, ECOS will no longer be installed by default with CVXPY.
     Please either add an explicit dependency on ECOS or switch to our new
-    default solver, Clarabel by either not specifying a solver argument
+    default solver, Clarabel, by either not specifying a solver argument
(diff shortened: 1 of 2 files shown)


PTNobel:
Seems to be some bugs, I'll try and fix them now

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -55,6 +55,16 @@
     "disciplined-parametrized-programming"
 )
 
+ECOS_DEP_DEPRECATION_MSG = (
+    """
+    You specified your problem should be solved by ECOS. Starting in
+    CXVPY 1.6.0, ECOS will no longer be installed by default with CVXPY.
+    Please either add an explicit dependency on ECOS or switch to our new
+    default solver, Clarabel by either not specifying a solver argument
```
phschiele, line 63:
```suggestion
    default solver, Clarabel, by either not specifying a solver argument
```

phschiele (approved):
Looks good:
- if no solver is selected, Clarabel will not be used for SDPs
- slightly implicit change: SCS has now priority over ECOS for users who uninstall Clarabel or install cvxpy-base+ecos+scs (edge case)

SteveDiamond (approved):

PR #2391, opened 2024-03-29
removing some test warnings

Transurgeon:
## Description
I saw that the summary of the tests had many warnings. This PR attempts to clean up a small subset of those warnings.
I was wondering if we should change all occurrences of using ECOS to CLARABEL in the tests, now that we are planning to change default solvers? 
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/solving_chain.py
+++ b/cvxpy/reductions/solvers/solving_chain.py
@@ -50,9 +50,8 @@ DPP_ERROR_MSG = (
     "You are solving a parameterized problem that is not DPP. "
     "Because the problem is not DPP, subsequent solves will not be "
     "faster than the first one. For more information, see the "
-    "documentation on Discplined Parametrized Programming, at\n"
-    "\thttps://www.cvxpy.org/tutorial/advanced/index.html#"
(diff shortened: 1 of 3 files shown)


cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -50,9 +50,8 @@
     "You are solving a parameterized problem that is not DPP. "
     "Because the problem is not DPP, subsequent solves will not be "
     "faster than the first one. For more information, see the "
-    "documentation on Discplined Parametrized Programming, at\n"
-    "\thttps://www.cvxpy.org/tutorial/advanced/index.html#"
-    "disciplined-parametrized-programming"
+    "documentation on Disciplined Parametrized Programming, at "
+    "https://www.cvxpy.org/tutorial/dpp/index.html"
```
Transurgeon, line 54: I updated the link to the DPP docs in this warning.

SteveDiamond (approved):
Looks good! Definitely right move to change ECOS to CLARABEL.

Issue #2389, opened 2024-03-26
Wheels compatible with Numpy 2.0


skoudoro:
Hi `cvxpy` Team,

Do you have any wheels compatible with the future numpy 2.0 ?

Currently,  when I compile my package with Numpy 2.0.0dev and try to run it with `cvxpy`, I encounter the following error:

```
import _cvxcore
RuntimeError: module compiled against ABI version 0x1000009 but this version of numpy is 0x2000000
```

Numpy 2.0 is coming and I would like to make sure everything will be smooth with the transition.

Any advice will be welcomed. 

Thanks !

Transurgeon:
Hello @skoudoro thanks for opening this issue. I believe cvxpy has already added the ruff rule ``NPY201`` mentioned in the NumPy docs. I will try testing against the main branch today. Is [this](https://numpy.org/devdocs/dev/depending_on_numpy.html#testing-against-the-numpy-main-branch-or-pre-releases) the right place to get the updated version of NumPy 2.0?

skoudoro:
> Is [this](https://numpy.org/devdocs/dev/depending_on_numpy.html#testing-against-the-numpy-main-branch-or-pre-releases) the right place to get the updated version of NumPy 2.0?

Yes, we also use it to put our wheels for Numpy 2.0. You will need to do a request to scientific-python community if you would like to upload the one for cvxpy. 

Thank you for your feedback.

Transurgeon:
Some updates on this. I tried testing with Numpy 2.0 but I also get the following error:
```  
  import _cvxcore
RuntimeError: module compiled against ABI version 0x1000009 but this version of numpy is 0x2000000
```
I first got the nightly build from scipy which created some conflicts with OSQP (since it requires scipy <=1.12).
I can try contacting some folks at scientific python to ask for advice.
 
@phschiele @PTNobel any thoughts on this issue? I am not too familiar with the process of creating new wheels for cvxpy.

PTNobel:
Transurgeon did you try deleting artifacts before installing cvxpy locally?

Transurgeon:
> @Transurgeon did you try deleting artifacts before installing cvxpy locally?

No, I am not sure what artefacts are and where to delete them. Also when I'm installing cvxpy locally, the nightly NumPy version I installed gets overridden by the ``oldest-supported-numpy``. Do we need to update the ``pyproject.toml`` setup file to be able to create new wheels?

rileyjmurray:
[...]

PR #2392, opened 2024-03-31
Prep cvxpy for NumPy 2.0 release

Transurgeon:
## Description
This PR updates the cvxpy build to allow compatible wheels with NumPy 2.0. 
The changes come in the following stages:
- updating the ``numpy.i`` file in cvxcore which is used for the swig interface. (As pointed out by Riley in the corresponding issue linked below). The new file is taken from [here](https://github.com/numpy/numpy/blob/main/tools/swig/numpy.i) and was last updated 2 years ago.
- Rebuild cvxcore using the swig command in the corresponding README located [here](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/cvxcore/README.md).
- Update cvxpy's ``pyproject.toml`` build file. I first introduced newer nightly versions of NumPy and SciPy in the requirements. Then, following a [discussion](https://github.com/pytorch/pytorch/issues/122056#issuecomment-2024662467) about a similar issue to ours, I also added the ``pybind11 >= 2.12.0`` requirement. 

I tried to only update the ``pyproject.toml`` build file but I encountered some errors like the one below:
```
error: no matching function for call to 'PyArray_TYPE'
#define array_type(a)          (int)(PyArray_TYPE(a))
```
So I believe the updates to cvxcore are necessary for wheels compatibility with NumPy 2.0.
Any feedback/questions are welcome! 

P.S: I am quite new to the python build ecosystem so I may have missed some important steps. 

Issue link (if applicable): #2389 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/cvxcore/python/cvxcore.py
```diff
@@ -1,10 +1,13 @@
-# This file was automatically generated by SWIG (https://www.swig.org).
-# Version 4.1.1
+# This file was automatically generated by SWIG (http://www.swig.org).
+# Version 4.0.2
```
PTNobel, line 2: I don't think we shoiuld be downgrading our SWIG version

cvxpy/cvxcore/python/cvxcore.py
```diff
@@ -1,10 +1,13 @@
...
 #
-# Do not make changes to this file unless you know what you are doing - modify
+# Do not make changes to this file unless you know what you are doing--modify
 # the SWIG interface file instead.
 
 from sys import version_info as _swig_python_version_info
+if _swig_python_version_info < (2, 7, 0):
+    raise RuntimeError("Python 2.7 or later required")
```
PTNobel, line 9: We don't support Python 2.7 I thought?

cvxpy/cvxcore/python/cvxcore.py
PTNobel: Not a huge deal; but none of the changes to this file are material, we can probably just remove it from the commit and it will be fine

pyproject.toml
```diff
@@ -25,12 +25,12 @@ testpaths = [
 
 [build-system]
 requires = [
-    "oldest-supported-numpy",
-    "scipy >= 1.1.0",
+    "numpy >= 2.0.0rc1",
```
PTNobel, line 28: Do we need to require the prerelease NumPy? This means we're dropping support for a lot of old NumPy versions right?

cvxpy/cvxcore/python/cvxcore.py
Transurgeon: done, thanks for pointing this out.

cvxpy/cvxcore/python/cvxcore.py
Transurgeon: yes you are right, I've reverted all the changes for ``cvxcore.py``

pyproject.toml
Transurgeon: I don't believe this would drop support for older NumPy versions. see [here](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice). 
Also, NumPy >= 1.15 is still part of the ``dependencies`` for cvxpy.

Transurgeon:
I tried the follow the steps outlined in the NumPy migration guide, but could have certainly missed some steps.
I think the next steps for this would be to update the CI to test both on the oldest numpy version and the nightly version of NumPy. 
Another option could just be to wait until NumPy 2.0 is officially released through pip. (see image below)
image

Transurgeon:
@PTNobel @SteveDiamond @rileyjmurray 
I propose that we first merge this PR without the changes to the``pyproject.toml`` setup file, that way the CI will pass and we can figure out later how we want to go about testing the new builds.
I'll also change the title of the PR to something like "updating numpy.I swig interface".

This could also maybe allow the cvxpy-feedstock to rebuild without any problems. 

What are your thoughts?

PTNobel:
That sounds like a great first step. I think I agree with h-vetenari in another thread that we should release 1.5.2 with NumPy 2.0 support before 2.0 comes out. That'll mean some changes to the GitHub action runners

Transurgeon:
Ok, I just reverted the changes to ``pyproject.toml``. CI failure seems to be unrelated to the PR.

PTNobel:
I want to make sure that the cvxpy-base build is working before merging... Rerunning CI, since the error seemed spurious

cvxpy/cvxcore/python/cvxcore_wrap.cxx
PTNobel, line 1: Not a huge fan of downgrading our swig version I thought we had updated to fix a bug; William can you update the swig on your computer and rerun this?

PTNobel (changes requested):

PTNobel:
Upgrading to Swig 1.4.1 was quite intentional. I don't think we should backout https://github.com/cvxpy/cvxpy/pull/2273

cvxpy/cvxcore/python/cvxcore_wrap.cxx
Transurgeon: I tried rerunning the command, but to no avail. I also couldn't find a way to upgrade the swig version (sorry! I am very unfamiliar with swig). 

I decided to just revert the changes to the file since it seems like it was already generated using the desired swig 4.1.1.

Transurgeon:
> Upgrading to Swig 4.1.1 was quite intentional. I don't think we should backout #2273

Sorry, I wasn't aware of that PR. 

I just tried rebuilding cvxpy with numpy 2.0, but it didn't work. So we do need to rerun the given command with a newer version of swig. 

Anaconda only provides swig 4.0.2, see here: https://anaconda.org/anaconda/swig. Same thing for homebrew. (This means we should probably update the instructions in the cvxcore README as well). 

I found the releases on the swig GitHub, see here: https://github.com/swig/swig/tags. Would we potentially want to update to the latest version which is 4.2.1?

cvxpy/cvxcore/python/cvxcore_wrap.cxx
Transurgeon: Ok I just managed to install swig 4.2.1 (the latest). And regenerated the cvxcore wrap file. Everything seems to work now :).

Transurgeon:
@PTNobel could you have another look at this PR please?

I managed to install the newest swig version and regenerated the cvxcore wrap file (it is now version 4.2.1).
I then retried to build cvxpy with numpy 2.0 and it seems to work without any problems.
Finally, I updated the cvxcore/README.md to reflect anaconda being outdated on the swig version.

PTNobel:
Looks good to me

PTNobel (approved):

phschiele (approved):
Let's see if this works with the pre-release CI!

Issue #2402, opened 2024-04-05
Unexpected handling of coefficient that could be zero in version 1.4.x


codeterm:
**Describe the bug**
Practically, when an expression in the objective has a coefficient (be it constant or parameter) that could potentially be zero, the solution for the scenario where the coefficient = 0 should be the same as if the term is removed from the objective. This seems to be the case for version 1.2 and 1.3, but not 1.4.

**To Reproduce**
A quick example in a portfolio optimization problem with ridge regularization:

```
import numpy as np
import cvxpy as cp

r =  np.array([-0.48,  0.11,  0.09, -0.39,  0.03])
Sigma = np.array([
       [2.4e-04, 1.3e-04, 2.0e-04, 1.6e-04, 2.0e-04],
       [1.3e-04, 2.8e-04, 2.1e-04, 1.7e-04, 1.5e-04],
       [2.0e-04, 2.1e-04, 5.8e-04, 3.3e-04, 2.3e-04],
       [1.6e-04, 1.7e-04, 3.3e-04, 6.9e-04, 2.1e-04],
       [2.0e-04, 1.5e-04, 2.3e-04, 2.1e-04, 3.6e-04]])

w = cp.Variable(5)
risk_aversion = cp.Parameter(value=1., nonneg=True)
ridge_coef = cp.Parameter(value=0., nonneg=True)  # <-- this causes the issue

obj_func = r @ w - risk_aversion * cp.quad_form(w, Sigma) - ridge_coef * cp.sum_squares(w)
objective = cp.Maximize(obj_func)
constaints = [cp.sum(w) == 1]
prob = cp.Problem(objective, constaints)
prob.solve(eps_abs=1e-8, eps_rel=1e-8)
print(w.value)
```

**(Un)Expected behavior**
In versions before 1.4, the snippet above will output:
`[-1959.30787844   871.81166733   461.55434711  -409.40111681, 1036.34298081]`
which is consistent with the result if we directly remove the `ridge_coef * cp.sum_squares(w)` term from the objective.

However, starting from 1.4, the output is `None`. If, instead of a Parameter, we directly set `ridge_coef=0`, the error becomes explicit and points to:

```
File ~/miniconda3/envs/dev/lib/python3.12/site-packages/cvxpy/utilities/coeff_extractor.py:189, in CoeffExtractor.extract_quadratic_coeffs(self, affine_expr, quad_forms)
    183 if 'P' in coeffs[orig_id]:
[...]

PR #2406, opened 2024-04-09
[Bug] Parametrized quad form broken when values are zero

phschiele:
## Description
This PR should eventually contain the fix for #2402. 

Current status:
- [x] Identify origin, this behavior exists since #2226 
- [x] Adding MWE and test case. This test currently fails but passes on version <1.4.
- [x] Fix logic
 
Issue link (if applicable): #2402

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_coeff_extractor.py
+++ b/cvxpy/tests/test_coeff_extractor.py
@@ -33,7 +33,7 @@ def coeff_extractor():
     return CoeffExtractor(inverset_data, backend)
 
 
-def test_issue_2402_scalar():
+def test_issue_2402_scalar_parameter():
     """
     This is the problem reported in #2402, failing to solve when two parameters
     are used on quadratic forms with the same variable.
@@ -59,6 +59,35 @@ def test_issue_2402_scalar():
     
     assert prob.value is not None
     assert w.value is not None
+
+
+def test_issue_2402_scalar_constant():
+    """
+    This slight modification uses a constant instead of a parameter,
+    which was a separate issue in the same problem.
+    """
+
+    r =  np.array([-0.48,  0.11,  0.09, -0.39,  0.03])
+    Sigma = np.array([
+        [2.4e-04, 1.3e-04, 2.0e-04, 1.6e-04, 2.0e-04],
+        [1.3e-04, 2.8e-04, 2.1e-04, 1.7e-04, 1.5e-04],
+        [2.0e-04, 2.1e-04, 5.8e-04, 3.3e-04, 2.3e-04],
+        [1.6e-04, 1.7e-04, 3.3e-04, 6.9e-04, 2.1e-04],
+        [2.0e-04, 1.5e-04, 2.3e-04, 2.1e-04, 3.6e-04]])
+
+    w = cp.Variable(5)
+    risk_aversion = cp.Parameter(value=1., nonneg=True)
+    ridge_coef = 0
+    
+    obj_func = r @ w - risk_aversion * cp.quad_form(w, Sigma) -  ridge_coef * cp.sum_squares(w)
+    objective = cp.Maximize(obj_func)
+    constraints = [cp.sum(w) == 1]
+    prob = cp.Problem(objective, constraints)
+    prob.solve()
+    
+    assert prob.value is not None
+    assert w.value is not None
+
 
 def test_issue_2402_vector():
     """
--- a/cvxpy/tests/test_coeff_extractor.py
+++ b/cvxpy/tests/test_coeff_extractor.py
@@ -5,7 +5,7 @@ import pytest
 
 import cvxpy as cp
 from cvxpy.atoms.quad_form import SymbolicQuadForm
-from cvxpy.utilities.coeff_extractor import CoeffExtractor
+from cvxpy.utilities.coeff_extractor import COOData, CoeffExtractor
 
 
 @dataclass
@@ -78,12 +78,12 @@ def test_coeff_extractor(coeff_extractor):
     coeffs, constant = coeff_extractor.extract_quadratic_coeffs(affine_expr, quad_forms)
     
     assert len(coeffs) == 1
-    assert np.allclose(coeffs[1]["q"], np.zeros((2, 3)))
+    assert np.allclose(coeffs[1]["q"].toarray(), np.zeros((2, 3)))
     P = coeffs[1]["P"]
-    assert len(P) == 3
-    assert np.allclose(P[0], np.array([[1, 0, 0], [1, 0, 0], [0, 1, 0], [0, 1, 0]]))
-    assert len(P[1]) == 2
-    assert np.allclose(P[1][0], np.array([0, 1, 0, 1]))
-    assert np.allclose(P[1][1], np.array([0, 1, 0, 1]))
-    assert P[2] == (2, 2)
+    assert isinstance(P, COOData)
+    assert np.allclose(P.data, np.ones((4)))
+    assert np.allclose(P.row, np.array([0, 1, 0, 1]))
+    assert np.allclose(P.col, np.array([0, 1, 0, 1]))
+    assert P.shape == (2, 2)
+    assert np.allclose(P.param_idxs, np.array([0,1]))
     assert np.allclose(constant.toarray(), np.zeros((3)))
--- a/cvxpy/tests/test_coeff_extractor.py
+++ b/cvxpy/tests/test_coeff_extractor.py
@@ -1,0 +1,89 @@
+from dataclasses import dataclass
+
+import numpy as np
+import pytest
+
+import cvxpy as cp
+from cvxpy.atoms.quad_form import SymbolicQuadForm
+from cvxpy.utilities.coeff_extractor import CoeffExtractor
+
+
+@dataclass
+class MockeInverseData:
+    var_offsets: dict
+    x_length: int
+    var_shapes: dict
+    param_shapes: dict
+    param_to_size: dict
+    param_id_map: dict
+
+
+@pytest.fixture
+def coeff_extractor():
+    inverset_data = MockeInverseData(
+        var_offsets={1: 0},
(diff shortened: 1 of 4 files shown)


cvxpy/utilities/coeff_extractor.py
```diff
@@ -189,6 +190,7 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
                         acc_data = np.concatenate([acc_P.data, data], axis=0)
                         acc_row = np.concatenate([acc_P.row, P.row], axis=0)
                         acc_col = np.concatenate([acc_P.col, P.col], axis=0)
+                        param_idxs = np.concatenate([acc_P.param_idxs, param_idxs], axis=0)
```
phschiele, line 192: @SteveDiamond Previously, `param_idxs` would be overwritten above in every iteration.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -150,70 +174,68 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
             #                      argument index of quad form,
             #                      quad form atom)
             if var.id in quad_forms:
+                # This was a dummy variable
                 var_id = var.id
                 orig_id = quad_forms[var_id][2].args[0].id
                 var_offset = affine_id_map[var_id][0]
                 var_size = affine_id_map[var_id][1]
                 c_part = c[var_offset:var_offset+var_size, :]
-                if quad_forms[var_id][2].P.value is not None:
```
phschiele, line 158: We never hit the `None` case and I'm not sure how this would happen, so I assert `not None` below.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -247,14 +269,14 @@ def quad_form(self, expr):
...
-        for var_id, offset in offsets:
+        for var_id, _ in offsets:
             shape = self.var_shapes[var_id]
             size = np.prod(shape, dtype=int)
             if var_id in coeffs and 'P' in coeffs[var_id]:
                 P = coeffs[var_id]['P']
                 P_entries += P.data.size
             else:
-                P = COOData([], [], [], (size, size), np.arange(num_params))
+                P = COOData.empty_with_shape((size, size))
```
phschiele, line 279: Since it was empty, I don't think we need the `np.arange(num_params)`.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -317,30 +339,35 @@ def merge_P_list(
             ```
             but done by constructing a COO matrix.
             """
-            if len(P.data) > 0:
+            if P.data.size > 0:
```
phschiele, line 342: This was an edge case where we had a sparse matrix with shape but no values, that would then give a `ZeroDivisionError` when computing the number of repetitions.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -35,14 +36,37 @@
...
     shape: tuple[int, int]
     param_idxs: np.ndarray
 
+    def __add__(self, other: COOData) -> COOData:
+        # Concatenation becomes addition when constructing
+        # COO matrix because repeated indices are summed.
+
+        if self.shape != other.shape:
+            raise ValueError("Shapes must match for addition.")
+        data = np.concatenate([self.data, other.data], axis=1)
```
phschiele, line 55: I changed this to add on `axis=1`, such that the columns of `data` correspond to `param_idxs`, instead of blocks of rows.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -150,70 +174,68 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                    P = sp.coo_matrix(P.value)
+
                 # We multiply the columns of P, by c_part
                 # by operating directly on the data.
                 if var_size > 1:
+                    # Multiple quad forms in the same expression, sharing P
                     # TODO remove zeros from data.
                     data = P.data[:, None] * c_part[P.col]
-                    param_idxs = np.arange(c_part.shape[1])
+                    param_idxs = np.arange(num_params)
```
SteveDiamond, line 200: Was the core bug using c_part.shape[1] instead of num_params?

phschiele, line 200: I'm not sure if this is correct if `num_params > 1`, which we haven't covered in our tests.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -317,30 +339,35 @@ def merge_P_list(
...
             """
-            if len(P.data) > 0:
+            if P.data.size > 0:
                 vals[entry_offset:entry_offset + P.data.size] = P.data.flatten(
                     order='F'
                 )
                 P_cols_ext = P.col.astype(np.int64) * np.int64(P_height)
                 base_rows = gap_above + acc_height + P.row + P_cols_ext
-                full_rows = np.tile(base_rows, len(P.param_idxs))
+                full_rows = np.tile(base_rows, P.data.size // P.row.size)
```
SteveDiamond, line 348: For future reference could you add a few comments about the details of what's happening in constructing the matrix?

SteveDiamond (approved):
Thanks for doing this!

SteveDiamond:
Could you add a test for this issue when the numeric value is 0 instead of the parameter with value 0? https://github.com/cvxpy/cvxpy/issues/2402
We should make sure both bugs are covered in the test suite.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -291,56 +282,44 @@ def merge_P_list(
...
+            # Translate local to global indices within the block diagonal matrix.
+            P.row += offset
+            P.col += offset
+            P.shape = (P_height, P_height)
+    
+            offset += m
+
+        combined = TensorRepresentation.combine(P_list)
+
+        return combined.flatten_tensor(num_params).tocoo()  # todo: check if csc works
```
phschiele, line 308: I'm quite happy with this refactoring. Reusing the `TesnsoRepresentation` from the backend removes most of the rather complicated logic here.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -291,56 +282,44 @@ def merge_P_list(
...
+            assert m == n
+    
+            # Translate local to global indices within the block diagonal matrix.
+            P.row += offset
+            P.col += offset
+            P.shape = (P_height, P_height)
+    
+            offset += m
+
+        combined = TensorRepresentation.combine(P_list)
```
SteveDiamond, line 306: This is great!!!

cvxpy/tests/test_coeff_extractor.py
```diff
@@ -0,0 +1,147 @@
...
+        [1.3e-04, 2.8e-04, 2.1e-04, 1.7e-04, 1.5e-04],
+        [2.0e-04, 2.1e-04, 5.8e-04, 3.3e-04, 2.3e-04],
+        [1.6e-04, 1.7e-04, 3.3e-04, 6.9e-04, 2.1e-04],
+        [2.0e-04, 1.5e-04, 2.3e-04, 2.1e-04, 3.6e-04]])
+
+    w = cp.Variable(5)
+    risk_aversion = cp.Parameter(value=1., nonneg=True)
+    ridge_coef = cp.Parameter(value=0., nonneg=True)
+    
+    obj_func = r @ w - risk_aversion * cp.quad_form(w, Sigma) -  ridge_coef * cp.sum_squares(w)
```
SteveDiamond, line 54: Could you add this same test but with ``ridge_coef = 0``? That triggered a separate bug.

phschiele: Done here: [0968ef0](https://github.com/cvxpy/cvxpy/pull/2406/commits/0968ef0ec4fa686a9f5502b12a82cf927c950ccc)

cvxpy/utilities/coeff_extractor.py
```diff
@@ -56,21 +44,6 @@ def __init__(self, inverse_data, canon_backend: str | None) -> None:
...
-        elif expr.is_quadratic():
-            return self.quad_form(expr)
-        else:
-            raise Exception("Unknown expression type %s." % type(expr))
-
-    def constant(self, expr):
-        size = expr.size
-        return sp.csr_matrix((size, self.N)), np.reshape(expr.value, (size,),
-                                                         order='F')
-
```
phschiele, line 73: This was unused.

cvxpy/utilities/coeff_extractor.py
```diff
@@ -150,70 +123,84 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                    scaled_c_part = P @ c_part
+                    paramx_idx_row, param_idx_col = np.nonzero(scaled_c_part)
+                    c_vals = c_part[paramx_idx_row, param_idx_col]
+                    P_tup = TensorRepresentation(
+                        c_vals,
+                        paramx_idx_row,
+                        paramx_idx_row,
+                        param_idx_col,
+                        P.shape
+                    )
```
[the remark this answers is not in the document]
SteveDiamond: Yes it should be diagonal in that case, since we don't support matrix quad_form.

phschiele:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=cvxpy_cvxpy&pullRequest=2406) **Quality Gate failed**
> Failed conditions  [3.8% Duplication on New Code](https://sonarcloud.io/component_measures?id=cvxpy_cvxpy&pullRequest=2406&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=cvxpy_cvxpy&pullRequest=2406)

This is only because of some duplication in the tests.

phschiele:
Merging this PR as discussed with @SteveDiamond.

Issue #2408, opened 2024-04-14
`value.A` is deprecated and will be remove in scipy 1.14.0


chofchof:
**Is your feature request related to a problem? Please describe.**
Lines 46 and 48 of `cvxpy/interface/numpy_interface/ndarray_interface.py` contain `value.A`, which is deprecated and will be removed in scipy 1.14.0.
```
        if scipy.sparse.issparse(value):
            result = value.A
        elif isinstance(value, numpy.matrix):
            result = value.A
```

**Describe the solution you'd like**
I suggest to fix them as follows.
```
        if scipy.sparse.issparse(value):
            result = value.toarray()
        elif isinstance(value, numpy.matrix):
            result = numpy.asarray(value)
```

phschiele:
@chofchof I've never found `.A` to be intuitive. Would you be interested in contributing a replacement of all occurrences?

chofchof:
> @chofchof I've never found `.A` to be intuitive. Would you be interested in contributing a replacement of all occurrences?

I could find nine places of `.A` including the two above.

3. Line 259 of `cvxpy/lin_ops/tree_mat.py`:
```
        if isinstance(result, np.matrix):
            result = result.A[0]
```
Since the type of `result` is `np.matrix`, I suggest `result = np.asarray(result)[0]`.

4. Line 137 of `cvxpy/utilities/coeff_extractor.py`:
```
        c = param_coeffs[:-1, :].A
```
Since `param_coeffs` is a sparse matrix, I suggest `c = param_coeffs[:-1, :].toarray()`.

5. Line 350 of `cvxpy/utilities/coeff_extactor.py`:
```
            q = np.vstack([q, constant.A])
```
Since the type of `constant` is `sp.csc_matrix`, I suggest `q = np.vstack([q, constant.toarray()])`.

6. Line 344 of `cvxpy/atoms/affine/binary_operators.py`:
```
                values[i] = values[i].todense().A
```
Since the type of `values[i].todense().A` is `np.ndarray`, I suggest `values[i] = values[i].toarray()`.

7. Line 33 of `cvxpy/atoms/norm_inf.py`:
```
                values = values[0].todense().A.flatten()
```
I suggest `values = values[0].toarray().flatten()`. The reason is the same as the previous one.

8. Line 62 of `cvxpy/atoms/affine/sum.py`:
```
        if intf.is_sparse(values[0]):
            result = np.sum(values[0], axis=self.axis)
[...]

PR #2411, opened 2024-04-16
Fix issue #2408: `value.A` is deprecated and will be removed in scipy 1.14.0

chofchof:
## Description
Fix the `.A` property of sparse matrices which is deprecated and will be removed in SciPy 1.14.0.
Issue link (if applicable): #2408

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -57,7 +57,7 @@ class Sum(AxisAtom, AffAtom):
         """Sums the entries of value.
         """
         if intf.is_sparse(values[0]):
-            result = np.sum(values[0].toarray(), axis=self.axis)
+            result = np.asarray(values[0].sum(axis=self.axis))
             if not self.keepdims and self.axis is not None:
                 result = result.flatten()
         else:
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -57,9 +57,9 @@ class Sum(AxisAtom, AffAtom):
         """Sums the entries of value.
         """
         if intf.is_sparse(values[0]):
-            result = np.sum(values[0], axis=self.axis)
+            result = np.sum(values[0].toarray(), axis=self.axis)
             if not self.keepdims and self.axis is not None:
-                result = result.A.flatten()
+                result = result.flatten()
         else:
             result = np.sum(values[0], axis=self.axis, keepdims=self.keepdims)
         return result
--- a/cvxpy/atoms/prod.py
+++ b/cvxpy/atoms/prod.py
@@ -84,9 +84,9 @@ class Prod(AxisAtom):
         """Takes the product of the entries of value.
         """
         if intf.is_sparse(values[0]):
-            result = np.prod(values[0], axis=self.axis)
+            result = np.prod(values[0].toarray(), axis=self.axis)
             if not self.keepdims and self.axis is not None:
-                result = result.A.flatten()
+                result = result.flatten()
         else:
             result = np.prod(values[0], axis=self.axis, keepdims=self.keepdims)
         return result
--- a/cvxpy/atoms/prod.py
+++ b/cvxpy/atoms/prod.py
@@ -84,9 +84,23 @@ class Prod(AxisAtom):
         """Takes the product of the entries of value.
         """
         if intf.is_sparse(values[0]):
-            result = np.prod(values[0].toarray(), axis=self.axis)
-            if not self.keepdims and self.axis is not None:
-                result = result.flatten()
+            sp_mat = values[0]
+            if self.axis is None:
+                if sp_mat.nnz == sp_mat.shape[0] * sp_mat.shape[1]:
+                    data = sp_mat.data
+                else:
+                    data = np.zeros(1, dtype=sp_mat.dtype)
+                result = np.prod(data)
+            else:
+                assert self.axis in [0, 1]
+                # The following snippet is taken from stackoverflow.
+                # https://stackoverflow.com/questions/44320865/
+                mask = sp_mat.getnnz(axis=self.axis) == sp_mat.shape[self.axis]
+                result = np.zeros(sp_mat.shape[1-self.axis], dtype=sp_mat.dtype)
+                data = sp_mat[:, mask] if self.axis == 0 else sp_mat[mask, :]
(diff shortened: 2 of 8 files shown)


cvxpy/atoms/affine/sum.py
PTNobel, line 1: This one worries me a bit; I'll want to look at it more carefully.

cvxpy/atoms/prod.py
PTNobel, line 1: This one also.

PTNobel (commented):
Overall looks great; two that I'm a little hesitant on and will need to look at when I have a few more minutes tomorrow.

cvxpy/atoms/affine/sum.py
chofchof: If we can assure that `np.sum` works for (scipy) sparse arrays, the following may work. 
```
        if intf.is_sparse(values[0]):
            result = np.sum(values[0], axis=self.axis)
            if not self.keepdims and self.axis is not None:
                result = result.flatten()
```

cvxpy/atoms/prod.py
chofchof: As @phschiele wrote in the issue #2408, `np.prod` does not work for (scipy) sparse arrays.

cvxpy/atoms/affine/sum.py
phschiele: Yes, we should include the snippet from my comment as a test case.

chofchof: @phschiele I agree. I missed that `.flatten()` does not work with `numpy.matrix`.
```
>>> np.matrix([[2, 4]]).flatten()
matrix([[2, 4]])
>>> np.array([[2, 4]]).flatten()
array([2, 4])
```

chofchof: @phschiele SciPy's sparse matrices has the method `.sum()` so that I suggest the following without `np.sum`. Its behaviour is the same as yours.
```
result = np.asarray(values[0].sum(axis=self.axis))
```

cvxpy/atoms/prod.py
chofchof: @PTNobel I agree that `np.prod(values[0].toarray(), axis=self.axis)` is too heavy for sparse matrices. For example, the following code raised `MemoryError` in my notebook.
```
import cvxpy as cp
from scipy.sparse import coo_matrix

A = coo_matrix([100], ([0], [0]), shape=(1000000, 1000000))
cp.prod(A, axis=0).value
```

chofchof: I modified `cvxpy/atoms/prod.py` as follows to work with sparse matrices.
```
    def numeric(self, values):
        """Takes the product of the entries of value.
        """
        if intf.is_sparse(values[0]):
            sp_mat = values[0]
            if self.axis is None:
                if sp_mat.nnz == sp_mat.shape[0] * sp_mat.shape[1]:
                    data = sp_mat.data
                else:
                    data = np.zeros(1, dtype=sp_mat.dtype)
                result = np.prod(data)
            else:
                assert self.axis in [0, 1]
                # The following snippet is taken from stackoverflow.
                # https://stackoverflow.com/questions/44320865/
                mask = sp_mat.getnnz(axis=self.axis) == sp_mat.shape[self.axis]
                result = np.zeros(sp_mat.shape[1-self.axis], dtype=sp_mat.dtype)
                data = sp_mat[:, mask] if self.axis == 0 else sp_mat[mask, :]
                result[mask] = np.prod(data.toarray(), axis=self.axis)
                if self.keepdims:
                    result = np.expand_dims(result, self.axis)
        else:
            result = np.prod(values[0], axis=self.axis, keepdims=self.keepdims)
        return result
```

A sample output is as follows.
```
>>> indices = np.arange(10, dtype=int)
>>> values = np.arange(10, dtype=float)
>>> A = coo_matrix((values, (indices, indices)), shape=(1000000, 1000000))
>>> cp.prod(A).value
0.0
>>> cp.prod(A, axis=0).value
array([0., 0., 0., ..., 0., 0., 0.])
>>> cp.prod(A, axis=1).value
array([0., 0., 0., ..., 0., 0., 0.])
>>> cp.prod(A, axis=0, keepdims=True).value
array([[0., 0., 0., ..., 0., 0., 0.]])
>>> cp.prod(A, axis=1, keepdims=True).value
array([[0.],
       [0.],
       [0.],
       ...,
       [0.],
       [0.],
       [0.]])
```

@PTNobel and @phschiele Could you review this code?

phschiele: @chofchof this looks good. I've created a test case, but I can't push to your fork, even though maintainer edits seem to be enabled for this PR.

Can you add this test case to `test_constant.py` (need to add `import scipy.sparse as sp` at the top.

```py
def test_prod():
    rows = np.concatenate([np.arange(100), np.zeros(100)[1:]])
    cols = np.concatenate([np.zeros(100), np.arange(100)[1:]])
    values = np.ones(199)
    A = sp.coo_matrix((values, (rows, cols)), shape=(100, 100))

    assert np.allclose(cp.prod(A).value, 0.0)
    assert np.allclose(cp.prod(A, axis=0).value, [1] + [0] * 99)
    assert cp.prod(A, axis=0).shape == (100,)
    assert np.allclose(cp.prod(A, axis=1).value, [1] + [0] * 99)
    assert cp.prod(A, axis=1).shape == (100,)
    assert np.allclose(cp.prod(A, axis=0, keepdims=True).value, [[1] + [0] * 99])
    assert cp.prod(A, axis=0, keepdims=True).shape == (1, 100)
    assert np.allclose(cp.prod(A, axis=1, keepdims=True).value, [[1]] + [[0]] * 99)
    assert cp.prod(A, axis=1, keepdims=True).shape == (100, 1)

    B = np.arange(4).reshape(2, 2) + 1
    assert np.allclose(cp.prod(sp.coo_matrix(B)).value, 24)
```

Otherwise, this is good to go from my side.

chofchof: @phschiele It's good to see your review. I'll add your test code to the PR on Sunday after my business trip.

chofchof: @phschiele Updated the PR. Thanks for your test case.

phschiele (approved):
Thanks @chofchof! Merging once CI passes.

PR #2419, opened 2024-04-19
fix: keepdims passed through 'min' correctly

allenlawrence94:
## Description
The `min` atom ignores the `keepdims` flag.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/eliminate_pwl/canonicalizers/min_canon.py
+++ b/cvxpy/reductions/eliminate_pwl/canonicalizers/min_canon.py
@@ -20,8 +20,9 @@ from cvxpy.reductions.eliminate_pwl.canonicalizers.max_canon import max_canon
 
 def min_canon(expr, args):
     axis = expr.axis
+    keepdims = expr.keepdims
     del expr
     assert len(args) == 1
-    tmp = max(-args[0], axis=axis)
+    tmp = max(-args[0], axis=axis, keepdims=keepdims)
     canon, constr = max_canon(tmp, tmp.args)
     return -canon, constr


SteveDiamond:
Good catch!

SteveDiamond:
We should add a test that would fail without this fix.

phschiele:
Thanks @allenlawrence94! Would you mind adding a simple unit test as well?

allenlawrence94:
@phschiele - sure! Are there unit tests for the canonicalizers? Or would we rather just have a test on the `min` atom `keepdims=True`?

SteveDiamond:
> @phschiele - sure! Are there unit tests for the canonicalizers? Or would we rather just have a test on the `min` atom `keepdims=True`?

Do a test on the min atom. i.e. construct a problem then solve it that requires passing through keepdims=True

SteveDiamond:
@allenlawrence94 please update the test_min function in test_atoms to the following code. We are trying to merge all the outstanding bugfix MRs.
```
    def test_min(self) -> None:
        """Test min.
        """
        # One arg, test sign.
        self.assertEqual(cp.min(1).sign, s.NONNEG)
        self.assertEqual(cp.min(-2).sign, s.NONPOS)
        self.assertEqual(cp.min(Variable()).sign, s.UNKNOWN)
        self.assertEqual(cp.min(0).sign, s.ZERO)

        # Test with axis argument.
        self.assertEqual(cp.min(Variable(2), axis=0).shape, tuple())
        self.assertEqual(cp.min(Variable(2), axis=1).shape, (2,))
        self.assertEqual(cp.min(Variable((2, 3)), axis=0).shape, (3,))
        self.assertEqual(cp.min(Variable((2, 3)), axis=1).shape, (2,))

        # Invalid axis.
            cp.min(self.x, axis=4)
        self.assertEqual(str(cm.exception), "Invalid argument for axis.")
        with self.assertRaises(ValueError) as cm:
            cp.min(self.x, self.x)  # a common erroneous use-case
        self.assertEqual(str(cm.exception), cp.min.__EXPR_AXIS_ERROR__)

        # Test canonicalization with keepdims=True
        # https://github.com/cvxpy/cvxpy/pull/2419
        X = cp.Variable((2, 3))
        X_val = np.arange(6).reshape((2, 3))
        c = np.ones((1, 3))
        expr = cp.min(X, axis=0, keepdims=True)
        obj = cp.Maximize(cp.sum(expr + c))
        prob = cp.Problem(obj, [X == X_val])
        prob.solve()
```

SteveDiamond:
I can't push to your fork for some reason, and we like to keep track of contributors so would rather not make a new PR.

SteveDiamond:
Picked up in https://github.com/cvxpy/cvxpy/pull/2431

Issue #2422, opened 2024-04-22
XPRESS release breaks tests


phschiele:
**Describe the bug**
The recent XPRESS [release 9.4.0](https://pypi.org/project/xpress/9.4.0/) breaks our tests:

```
xpress.ModelError: Cannot use problem.addVariable to add existing variables to the problem
```

[CI run](https://github.com/cvxpy/cvxpy/actions/runs/8783711474/job/24100441159)

PR #2423, opened 2024-04-22
Adds warning for constants from nested lists

### LINKED ISSUE #2317: Indexing reverses the orders of nested lists
**Describe the bug**

When I create a constant from a nested array, the shape is not what I expect - If I pass in a array of three sub-arrays, each sub-array having 2 elements, I expect the result to be an array of shape (3,2). Instead, it seems to have shape (2,3).

**To Reproduce**
A [minimal code example](https://stackoverflow.com/help/mcve) that reproduces the bug or unexpected behavior.

Here is an MCE
```
import numpy as np
import cvxpy as cp

foo = [[6292.9718735712195, 12394.971615149743], [188.53574799502698, 1031.7579795558324], [251.0724582117561, 116.460331391552]]

foo_array = np.array(foo)
foo_constant = cp.Constant(foo)
foo_constant2 = cp.Constant(foo_array)

print(foo_constant)
# [[6292.97 188.54 251.07]
#  [12394.97 1031.76 116.46]]
print(foo_constant2)
# [[6292.97 12394.97]
#  [188.54 1031.76]
#  [251.07 116.46]]
```

**Expected behavior**

The `foo_constant` created directly from the nested array should be the same as the numpy array.

**Output**
If applicable, include program output. If reporting a program crash, please include the entire stack trace.

```
[[6292.97 188.54 251.07]
 [12394.97 1031.76 116.46]]
[[6292.97 12394.97]
 [188.54 1031.76]
 [251.07 116.46]]
```

**Version**
 - OS: MacOS 13.3.1 (a) (22E772610a)
 - CVXPY Version: 1.4.1

**Additional context**

phschiele:
## Description
This addresses the discrepancy from the NumPy behaviour mentioned in #2317 by:
- emitting a warning
- adding a test case that makes sure any change in behaviour would be noticed

Issue link (if applicable): closes #2317 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/expressions/constants/constant.py
+++ b/cvxpy/expressions/constants/constant.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import warnings
 from typing import List, Optional, Tuple
 
 import numpy as np
@@ -25,6 +26,9 @@ import cvxpy.settings as s
 import cvxpy.utilities.linalg as eig_util
 from cvxpy.expressions.leaf import Leaf
 from cvxpy.utilities import performance_utils as perf
+
+NESTED_LIST_WARNING =  "Initializing a Constant with a nested list is " \
+    "undefined behavior. Consider using a numpy array instead."
 
 
 class Constant(Leaf):
@@ -45,6 +49,9 @@ class Constant(Leaf):
                 value, convert_scalars=True)
             self._sparse = True
         else:
+            if isinstance(value, list) and any(isinstance(i, list) for i in value):
+                warnings.warn(NESTED_LIST_WARNING)
+
             self._value = intf.DEFAULT_INTF.const_to_matrix(value)
             self._sparse = False
         self._imag: Optional[bool] = None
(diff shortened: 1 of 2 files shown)


cvxpy/expressions/constants/constant.py
```diff
@@ -45,6 +49,9 @@ def __init__(self, value, name: Optional[str] = None) -> None:
                 value, convert_scalars=True)
             self._sparse = True
         else:
+            if isinstance(value, list) and any(isinstance(i, list) for i in value):
+                warnings.warn(NESTED_LIST_WARNING)
```
phschiele, line 53: Looks like we mostly simply use `warnings.warn` (i.e., `UserWarning`) throughout the codebase, but could be changed to `SyntaxWarning`.

SteveDiamond: Either way is fine with me.

SteveDiamond (approved):
Nice!

phschiele:
CI failures for Python 3.11 are because of #2422, merging here.

Issue #2218, opened 2023-08-26
Parameter support for cp.convolve missing


phschiele:
**Describe the bug**
`cp.convolve(a, x)` can give wrong results when `a` is parametrized. Parametrized `x` (and constant `a`) appears to be correct.

**To Reproduce**
```py
n = 3
x = cp.Variable(n)
p = cp.Parameter(n)

problem = cp.Problem(cp.Minimize(cp.sum(cp.convolve(p, x))), [0 <= x, x<= 1])
assert problem.is_dpp()

p.value = np.arange(n)
problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
>>> 0  # correct

p.value = -np.arange(n)
problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
>>> 0  # incorrect

problem = cp.Problem(cp.Minimize(cp.sum(cp.convolve(p, x))), [x == 1])
problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
>>> -9  # correct
```

**Expected behavior**
Solve correctly or have `problem.is_dpp()` return `False`.

**Version**
 - OS: Ubuntu
 - CVXPY Version: master (586ef7c)

**Additional context**
- Similar issue before with `cp.kron`: https://github.com/cvxpy/cvxpy/issues/1656 and links therein
- Python backends raise `AssertionError`, so only C++ is affected

chofchof:
Even for 1-dim case, `cp.convolve(p, x)` does not give a correct answer if the parameter `p` is negative.
```
import cvxpy as cp

x = cp.Variable()
p = cp.Parameter()
problem = cp.Problem(cp.Minimize(cp.convolve(p, x)), [0 <= x, x <= 1])

p.value = -1.0
problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
>>> -1.274313644086471e-09  # 0, not -1
```

I hope to see some warning message in the documentation.

chofchof:
For 2-dim case, cp.convolve(p, x) does not give a correct answer even if the parameter p is nonnegative.
```
import cvxpy as cp

x = cp.Variable(2)
p = cp.Parameter(2)
problem = cp.Problem(cp.Minimize(cp.convolve(p, x)[1]), [0 <= x, x <= 1])

p.value = [0.0, 1.0]
problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
[...]

PR #2427, opened 2024-04-25
Fix convolve bug

SteveDiamond:
## Description
Fixes bug with convolve where the input is a parameter by disabling DPP for convolve. This is the same approach taken with kron.
Issue link (if applicable): #2218

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -1407,7 +1407,8 @@ class SciPyCanonBackend(PythonCanonBackend):
             lhs = lhs.T
 
         rows = lin.shape[0]
-        cols = lin.args[0].shape[0]
+        # lin.args[0] can have shape (), ie, scalar, or (n,), ie, vector
+        cols = lin.args[0].shape[0] if len(lin.args[0].shape) > 0 else 1
         nonzeros = lhs.shape[0]
 
         lhs = lhs.tocoo()
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -1407,7 +1407,6 @@ class SciPyCanonBackend(PythonCanonBackend):
             lhs = lhs.T
 
         rows = lin.shape[0]
-        # lin.args[0] can have shape (), ie, scalar, or (n,), ie, vector
         cols = lin.args[0].shape[0] if len(lin.args[0].shape) > 0 else 1
         nonzeros = lhs.shape[0]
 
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_atoms.py
```diff
@@ -947,6 +947,20 @@ def test_conv(self) -> None:
...
+        p = cp.Parameter()
+        problem = cp.Problem(cp.Minimize(cp.conv(p, x)), [0 <= x, x <= 1])
+
+        p.value = -1.0
+        result = problem.solve(canon_backend=cp.CPP_CANON_BACKEND)
+        self.assertAlmostEqual(result, -1)
+        self.assertAlmostEqual(x.value, 1)
+
+        with pytest.raises(cp.DPPError):
+            problem.solve(canon_backend=cp.CPP_CANON_BACKEND, enforce_dpp=True)
```
phschiele, line 962: @SteveDiamond I've added this test, and it fails. It works if I put it before the first solve. Presumably the `is_atom_convex` should not be cached?

SteveDiamond: I don't fully understand this. ``is_atom_convex`` is cached, but that should be fine?

SteveDiamond: I guess there might be an issue where it's cached in the non-DPP pass but then it needs to return a different value under DPP?

phschiele: Yes, exactly.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -1407,7 +1407,7 @@ def conv(self, lin: LinOp, view: SciPyTensorView) -> SciPyTensorView:
             lhs = lhs.T
 
         rows = lin.shape[0]
-        cols = lin.args[0].shape[0]
+        cols = lin.args[0].shape[0] if len(lin.args[0].shape) > 0 else 1
```
phschiele, line 1410: Didn't handle the scalar case yet, which is explicitly allowed in the docstring.

cvxpy/tests/test_atoms.py
```diff
@@ -947,6 +947,21 @@ def test_conv(self) -> None:
...
+        x = cp.Variable()
+        p = cp.Parameter()
+        problem = cp.Problem(cp.Minimize(cp.conv(p, x)), [0 <= x, x <= 1])
+
+        p.value = -1.0
+        result = problem.solve()
+        self.assertAlmostEqual(result, -1)
+        self.assertAlmostEqual(x.value, 1)
+
+        problem = cp.Problem(cp.Minimize(cp.conv(p, x)), [0 <= x, x <= 1])
```
phschiele, line 961: Recreating the problem now to make the test pass as expected.

phschiele (approved):

PR #2428, opened 2024-04-28
Reduce memory consumption for sum(list)

### LINKED ISSUE #2385: Huge memory consumption in `cp.sum(list)`
**Describe the bug**
Huge memory consumption in `cp.sum(list)`.  I know `cp.sum(cp.hstack(list))` solves the problem, but can `cp.sum(list)` be improved?

**To Reproduce**
``` 
x = [cp.Variable() for i in range(100000)]
expr = cp.sum(x)
```

**Output**
`expr = cp.sum(x)` uses > 30G memory and is super slow.

**Version**
 - OS: Ubuntu 18.04.3 LTS
 - CVXPY Version: 1.1.24

phschiele:
## Description
I memory-profiled the script from #2385
Issue link (if applicable): Closes #2385

```py
x = [cp.Variable() for i in range(20000)]
expr = cp.sum(x)
```

I found the place that results in the quadratic memory consumption, changed in this PR.
The memory consumption is quadratic because we keep track of the (linearly) growing list of `arg_groups` of every subexpression.

It turns out that the only use of `arg_groups` was to make the copy more efficient, passing the groups instead of the (expanded) args. Turns out a neat trick is to simply pass `[self]` as args in the copy function, making it even more efficient than the original speedup.

Before:
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     3  111.594 MiB  111.594 MiB           1   @profile
     4                                         def func():
     5  135.875 MiB   24.281 MiB       20003       x = [cp.Variable() for i in range(20000)]
     6 1694.438 MiB 1558.562 MiB           1       expr = cp.sum(x)
```

After:
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
     3  111.188 MiB  111.188 MiB           1   @profile
     4                                         def func():
     5  135.469 MiB   24.281 MiB       20003       x = [cp.Variable() for i in range(20000)]
     6  136.609 MiB    1.141 MiB           1       expr = cp.sum(x)
```

-> 1558 MiB to 1.1 MiB

@SteveDiamond @rileyjmurray 

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/add_expr.py
+++ b/cvxpy/atoms/affine/add_expr.py
@@ -30,8 +30,6 @@ class AddExpression(AffAtom):
     """
 
     def __init__(self, arg_groups: Iterable[Expression]) -> None:
-        # For efficiency group args as sums.
-        self._arg_groups = arg_groups
         super(AddExpression, self).__init__(*arg_groups)
         self.args = []
         for group in arg_groups:
@@ -46,8 +44,6 @@ class AddExpression(AffAtom):
         """Helper function to extract the arguments from an AddExpression.
         """
         if isinstance(expr, AddExpression):
-            # Remove the _arg_groups attribute to avoid copying it.
-            expr._arg_groups = None
             return expr.args
         else:
             return [expr]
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
Nice job figuring this out!

Issue #2433, opened 2024-05-09
Changing parameter doesn't get propogated to solver for cp.quad_over_lin with Clarabel


AHsu98:
**Describe the bug**
Using Clarabel with a quad_over_lin objective, where the parameter is in the denominator, when I solve the parametrized problem, then change the parameter, the solution doesn't change.
**To Reproduce**
This bug doesn't appear with ECOS, or in other simpler cases of the parameter appearing. Here is code to reproduce the error:
```
x = cp.Variable()
p = cp.Parameter()

loss = cp.quad_over_lin(x,p) + x
prob = cp.Problem(
    cp.Minimize(loss),
)

p.value = 1
prob.solve(solver=cp.CLARABEL)
sol1 = x.value.copy()
p.value = 1000
prob.solve(solver=cp.CLARABEL)
sol2 = x.value.copy()
print(sol1)
print(sol2)
```

This gives the output 
```
-0.4999999999999999
-0.4999999999999999
```

as the solution doesn't change with the two parameters. When I switch the solver to ECOS,
```
x = cp.Variable()
p = cp.Parameter()

loss = cp.quad_over_lin(x,p) + x
prob = cp.Problem(
    cp.Minimize(loss),
)

p.value = 1
prob.solve(solver=cp.ECOS)
sol1 = x.value.copy()
p.value = 1000
prob.solve(solver=cp.ECOS)
sol2 = x.value.copy()
```
the solution does change, and we get 
```
-0.5000006523005858
-500.023639561259
```

Additionally, changing parameter and resolving does work if the parameter appears just in a sum_squares, for example if we set the objective to `cp.sum_squares(x-p)`, re-solving with different p will yield the correct solution.

**Version**
 - OS: macOS 14.4.1 (23E224)
 - CVXPY Version: cvxpy-1.4.3
I'm on clarabel-0.7.1

SteveDiamond:
I confirmed this bug still exists in CVXPY 1.5.

PR #2435, opened 2024-05-09
Fix for issue with quad_over_lin parameters

SteveDiamond:
## Description
When the second argument to quad_over_lin is parameterized, the atom is DPP when canonicalized as an SOCP but not when canonicalized as a quadratic. This PR resolves the bug by making quad_over_lin not DPP in all cases whenever the second argument is parameterized.

Issue link (if applicable): #2433 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/quad_over_lin.py
+++ b/cvxpy/atoms/quad_over_lin.py
@@ -91,7 +91,11 @@ class quad_over_lin(Atom):
     def is_atom_convex(self) -> bool:
         """Is the atom convex?
         """
-        return True
+        # Disable DPP when the second argument is a parameter.
+        if u.scopes.dpp_scope_active():
+            return is_param_free(self.args[1])
+        else:
+            return True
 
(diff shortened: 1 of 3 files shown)


SteveDiamond:
I changed this to the simplest possible approach, which is to disable DPP for all situations when the second argument to quad_over_lin is a parameter.

PTNobel (approved):
LGTM

phschiele:
For DGP, isn't it canonicalized as $\ln \sum \exp(2x_i) - y$?
I seems to have been correct before, so we could have kept the log-log curvature, or added a test case.

SteveDiamond:
Good point, let's do another MR to ad

> For DGP, isn't it canonicalized as ln⁡∑exp⁡(2xi)−y? I seems to have been correct before, so we could have kept the log-log curvature, or added a test case.

Good point, let's add this back.

PR #2438, opened 2024-05-10
[Bug] Fixes QP canonicalization

### LINKED ISSUE #2437: Solving a simple problem is broken on most solvers after 1.4.2
**Describe the bug**
The solution to a simple problem is incorrect in 1.4.3 onwards when solving with PIQP, CLARABEL and OSQP - ECOS still apparently works

**To Reproduce**
```
import cvxpy as cp
import numpy as np

print(f'{cp.__version__}')

np.random.seed(42)
N = 100

ivol = 0.6 * np.ones(N)

alpha = np.random.normal(size=N)
t_cost = np.random.rand(N)
A = np.random.normal(size=(N,N))
sigma = A @ A.T

w = cp.Variable(N,name='w')

risk = 0.02 * (cp.multiply(w, ivol) ** 2).sum()
U = w @ alpha - risk - cp.abs(w) @ t_cost
problem = cp.Problem(cp.Maximize(U), [])

for solver in [cp.ECOS, cp.PIQP, cp.CLARABEL, cp.OSQP]:
    v = problem.solve(verbose=False, solver=solver)
    print(f'{solver}: {v:0.1f}')
```

**Expected behavior**
With 1.4.2 the program outputs
```
1.4.2
ECOS: 1384.0
PIQP: 1384.0
CLARABEL: 1384.0
OSQP: 1384.0
```
But with =>1.4.3
```
1.4.3
ECOS: 1384.0
PIQP: 817.1
CLARABEL: 817.1
OSQP: 817.1
```

**Version**
 - OS: Seen on multiple Linux
 - CVXPY Version: 1.4.3

**Additional context**
There is no problem with
```
risk = 0.02 * cp.quad_form(w,sigma)
```

phschiele:
## Description
See below for bug description in line.

Issue link (if applicable): Closes #2437 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/utilities/coeff_extractor.py
+++ b/cvxpy/utilities/coeff_extractor.py
@@ -168,7 +168,7 @@ class CoeffExtractor:
                     P_tup = TensorRepresentation(
                         c_vals,
                         paramx_idx_row,
-                        paramx_idx_row,
+                        paramx_idx_row.copy(),
                         param_idx_col,
                         P.shape
                     )
@@ -290,7 +290,8 @@ class CoeffExtractor:
         for P in P_list:
             m, n = P.shape
             assert m == n
-    
+            assert P.row is not P.col
+
             # Translate local to global indices within the block diagonal matrix.
             P.row += offset
             P.col += offset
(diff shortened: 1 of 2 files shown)


cvxpy/utilities/coeff_extractor.py
```diff
@@ -290,7 +290,8 @@ def merge_P_list(
         for P in P_list:
             m, n = P.shape
             assert m == n
-    
+            assert P.row is not P.col
+
             # Translate local to global indices within the block diagonal matrix.
             P.row += offset
             P.col += offset
```
phschiele, line 297: Previously, we had `P.row is P.col`, so the offset would be applied twice to each.

Issue #1529, opened 2021-11-05
Allow to send a list of solvers


akshayka:
### Discussed in https://github.com/cvxpy/cvxpy/discussions/1522

This feature has been suggested independently by multiple users.

<div type='discussions-op-text'>

<sup>Originally posted by **erelsgl** November  2, 2021</sup>
In some problems, some solvers may be better than others, but they might also fail with a `SolverError`.  I suggest allow, as the "solver" parameter, to send a list of solvers rather than just one. The function will try them in the order they are given, and return the first solution for which no `SolverError` is raised.</div>

rileyjmurray:
I see one complication with this feature. How would we parse solver options? There aren't consistent names for things like the maximum number of iterations used by a solver or solver termination criteria.

Transurgeon:
completed with PR above :)

PR #2450, opened 2024-05-20
solve problem by multiple solvers

oriyalperin:
## Description
As @erelsgl suggested in discussion #1522,
in some problems, some solvers may be better than others, but they might also fail with a SolverError.
Therefore, we've added a new method that allows sending a list of solvers rather than just one.
This function tries them in the order they are provided, and return the first solution for which no SolverError is raised.
Issue link: #1529

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -416,14 +416,16 @@ class Problem(u.Canonical):
         """
         return self._compilation_time
 
-    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str], *args, **kwargs):
-        """
-        Solve a problem using multiple solvers.
+    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str],
+                                *args, **kwargs):
+        """Solve a problem using multiple solvers.
 
         Arguments
         ---------
         solvers : list of (str, dict) tuples, (str) tuples, or strings.
             The solvers to use. For example, ['SCS', ('CLARABEL',), ('OSQP', {'max_iter':10000})]
+        kwargs : keywords, optional
+            Additional solver specific arguments.
 
         Returns
         -------
@@ -444,13 +446,18 @@ class Problem(u.Canonical):
             try:
                 if isinstance(solver, str):
                     solver_name = solver
-                    solution = self.solve(*args, solver=solver, **kwargs)
-                elif isinstance(solver, tuple) and len(solver) == 2:
-                    solver_name, solver_kwargs = solver
-                    if isinstance(solver_name, str) and isinstance(kwargs, dict):
+                    solution = self.solve(*args, solver=solver_name, **kwargs)
+                elif isinstance(solver, tuple):
+                    if len(solver) == 1:
+                        solver_name, = solver
+                        solution = self.solve(*args, solver=solver_name, **kwargs)
+                    elif len(solver) == 2:
+                        solver_name, solver_kwargs = solver
+                        if not isinstance(solver_name, str) or not isinstance(solver_kwargs, dict):
+                            raise ValueError("Solver tuple input must be (str, dict).")
                         solution = self.solve(*args, solver=solver_name, **solver_kwargs, **kwargs)
                     else:
-                        raise ValueError("Solver tuple input must be (str, dict).")
+                        raise ValueError("Solver tuple input must be (str, dict) or (str)")
                 else:
                     raise ValueError("Solver input must be one of the following:\
                                                 (str, dict) tuple, (str) tuple, or str.")
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -423,7 +423,7 @@ class Problem(u.Canonical):
         Arguments
         ---------
         solvers : list of (str, dict) tuples, (str) tuples, or strings.
-            The solvers to use. For example, ['SCS', ('ECOS'), ('OSQP', {'max_iter':10000})]
+            The solvers to use. For example, ['SCS', ('CLARABEL',), ('OSQP', {'max_iter':10000})]
 
         Returns
         -------
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -451,7 +451,8 @@ class Problem(u.Canonical):
                         else:
                             raise error.ParameterError("Solver tuple input must be (str, dict)")
                     else:
-                        raise error.ParameterError("Solver input must be one of the following: (str, dict) tuple, (str) tuple, or str")
+                        raise error.ParameterError("Solver input must be one of the following:\
+                                                    (str, dict) tuple, (str) tuple, or str")
                     s.LOGGER.info("Solver %s succeeds", solver_name)
                     return solution
                 except error.SolverError as e:
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -417,7 +417,7 @@ class Problem(u.Canonical):
         return self._compilation_time
     
     def _solve_solver_path(self, solve_func, solvers:list[tuple[str, dict] | tuple[str] | str],
-                                *args, **kwargs):
+                                args, kwargs):
         """Solve a problem using multiple solvers.
 
         Arguments
@@ -441,7 +441,7 @@ class Problem(u.Canonical):
         """
             
         if not solvers:
-            raise ValueError("Solvers list must contain at least one solver.")
+            raise ValueError("Solver path must contain at least one solver.")
         for solver in solvers:
             try:
                 if isinstance(solver, str):
@@ -558,7 +558,7 @@ class Problem(u.Canonical):
             if solver is not None:
                 raise ValueError(
                     "Cannot specify both 'solver' and 'solver_path'. Please choose one.")
-            return self._solve_solver_path(solve_func,solver_path,*args,**kwargs)
+            return self._solve_solver_path(solve_func,solver_path, args, kwargs)
         return solve_func(self, *args, **kwargs)
 
     @classmethod
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -416,13 +416,13 @@ class Problem(u.Canonical):
         """
         return self._compilation_time
 
-    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str] = None):
+    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str]):
         """
         Solve a problem using multiple solvers.
 
         Arguments
         ---------
-        solvers : list of (str, dict) tuples, (str) tuples, or strings. optional
+        solvers : list of (str, dict) tuples, (str) tuples, or strings.
             The solvers to use. For example, ['SCS', ('ECOS'), ('OSQP', {'max_iter':10000})]
 
         Returns
@@ -434,31 +434,31 @@ class Problem(u.Canonical):
         ------
         SolverError
             If all solvers fail to find a solution.
-        ParameterError
+        ValueError
             If the input solvers format is incorrect.
         """
             
-        if solvers:
-            for solver in solvers:
-                try:
-                    if isinstance(solver, str):
-                        solver_name = solver
-                        solution = self.solve(solver=solver)
-                    elif isinstance(solver, tuple) and len(solver) == 2:
-                        solver_name, kwargs = solver
-                        if isinstance(solver_name, str) and isinstance(kwargs, dict):
-                            solution = self.solve(solver=solver_name, **kwargs)
-                        else:
-                            raise error.ParameterError("Solver tuple input must be (str, dict)")
+        if not solvers:
+            raise ValueError("Solvers list must contain at least one solver.")
+        for solver in solvers:
+            try:
+                if isinstance(solver, str):
+                    solver_name = solver
+                    solution = self.solve(solver=solver)
+                elif isinstance(solver, tuple) and len(solver) == 2:
+                    solver_name, kwargs = solver
+                    if isinstance(solver_name, str) and isinstance(kwargs, dict):
+                        solution = self.solve(solver=solver_name, **kwargs)
                     else:
-                        raise error.ParameterError("Solver input must be one of the following:\
-                                                    (str, dict) tuple, (str) tuple, or str")
-                    s.LOGGER.info("Solver %s succeeds", solver_name)
-                    return solution
-                except error.SolverError as e:
-                    s.LOGGER.info("Solver %s failed: %s", solver_name, e)
-            raise error.SolverError(f"All solvers failed: {solvers}")
-        return self.solve()
+                        raise ValueError("Solver tuple input must be (str, dict).")
+                else:
+                    raise ValueError("Solver input must be one of the following:\
+                                                (str, dict) tuple, (str) tuple, or str.")
+                s.LOGGER.info("Solver %s succeeds", solver_name)
+                return solution
+            except error.SolverError as e:
+                s.LOGGER.info("Solver %s failed: %s", solver_name, e)
+        raise error.SolverError(f"All solvers failed: {solvers}")
     
     def solve(self, *args, **kwargs):
         """Compiles and solves the problem using the specified method.
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -415,8 +415,8 @@ class Problem(u.Canonical):
                    last time it was compiled.
         """
         return self._compilation_time
-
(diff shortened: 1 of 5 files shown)


cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
         """
         return self._compilation_time
 
+    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str] = None):
```
SteveDiamond, line 419: This is nice functionality! I don't think we need to support an argument of None, since anyone calling this function should be giving some list of solvers.

oriyalperin: Thanks for the review!
I'll remove the support for None.
What about an empty list? Currently, the implementation uses the solve method without arguments in this case.
Do you think I should keep it like that?

SteveDiamond: I would say that an empty list should raise a ValueError.

SteveDiamond: Yes, these are all ValueError. Let me sync a bit with the other devs though.

oriyalperin: Thanks, please let me know if any further changes are needed :)

cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
...
         return self._compilation_time
 
+    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str]):
+        """
+        Solve a problem using multiple solvers.
+
+        Arguments
+        ---------
+        solvers : list of (str, dict) tuples, (str) tuples, or strings.
+            The solvers to use. For example, ['SCS', ('ECOS'), ('OSQP', {'max_iter':10000})]
```
PTNobel, line 426: Heads up, `('ECOS')` is just a string I think you want `('ECOS',)`. Also, plesae use `'CLARABEL'` as an example instead of ECOS

cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
         """
         return self._compilation_time
 
+    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str]):
```
PTNobel, line 419: I feel like there's a lot of `solve` kwargs we should support here, most notably verbose and solver_verbose. Maybe we can add `*args, **kwargs` as arguments here and document it takes all keyword args supported by `.solve` and then call `.solve` as `.solve(solver=solver, *args, **kwargs, **solver_kwargs)`. This will cause errors if someone repeats arguments (which I think is desired).

cvxpy/tests/test_problem.py
```diff
@@ -1416,6 +1416,47 @@ def test_solver_error_raised_on_failure(self) -> None:
...
+        3. Ensures that a ValueError is raised when the inner inputs of the solvers are invalid.
+        
+        """
+
+        A = numpy.random.randn(40, 40)
+        b = cp.matmul(A, numpy.random.randn(40))
+        
+        # valid input, return solution
+        solvers_with_str=[(s.OSQP, {'max_iter':1}), s.CLARABEL]
+        solvers_no_dict=[(s.OSQP, {'max_iter':1}), (s.CLARABEL)]
```
PTNobel, line 1434:
```suggestion
        solvers_no_dict=[(s.OSQP, {'max_iter':1}), (s.CLARABEL,)]
```

PTNobel (changes requested):
Overall I'm positive on this but not a huge fan of the current public API. I left my suggestions. Aditionaly, I don't like `solve_multiple_solvers`, could we call the method `solve_via_chain`, `solve_with_fallback`, `try_solve`, or `try_solve_till_success`? I think Stephen Boyd often uses the term solver chain for this feature; I can ask him next week.

cvxpy/problems/problem.py
oriyalperin: It sounds right to me,
but what if a user wants to use different arguments for each solver, or even the same solver in multiple cases?
How can we determine this?

oriyalperin: Do you mean we should add `*args` and `**kwargs` to the function arguments (and keep the solvers name list)?
It sounds right to me, but what if a user wants to use different arguments for each solver, or even the same solver in multiple cases?
Could you demonstrate how you envision the arguments list, please? It would be greatly appreciated.

PTNobel:
```suggestion
    def solve_multiple_solvers(self, solvers:list[tuple[str, dict] | tuple[str] | str], *args, **kwargs):
```

cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
...
+            If the input solvers format is incorrect.
+        """
+            
+        if not solvers:
+            raise ValueError("Solvers list must contain at least one solver.")
+        for solver in solvers:
+            try:
+                if isinstance(solver, str):
+                    solver_name = solver
+                    solution = self.solve(solver=solver)
```
PTNobel, line 447:
```suggestion
                    solution = self.solve(solver=solver, *args, **kwargs)
```

cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
...
+                elif isinstance(solver, tuple) and len(solver) == 2:
+                    solver_name, kwargs = solver
```
PTNobel, line 449:
```suggestion
                    solver_name, solver_kwargs = solver
```

cvxpy/problems/problem.py
```diff
@@ -416,6 +416,50 @@ def compilation_time(self) -> float | None:
...
+                    if isinstance(solver_name, str) and isinstance(kwargs, dict):
+                        solution = self.solve(solver=solver_name, **kwargs)
```
PTNobel, line 451:
```suggestion
                        solution = self.solve(*args, solver=solver_name, **solver_kwargs, **kwargs)
```

PTNobel (changes requested):

cvxpy/problems/problem.py
PTNobel: Suggested all the changes I think you need!

oriyalperin:
@PTNobel thank you for all your suggestions! 

> Overall I'm positive on this but not a huge fan of the current public API. I left my suggestions. Aditionaly, I don't like `solve_multiple_solvers`, could we call the method `solve_via_chain`, `solve_with_fallback`, `try_solve`, or `try_solve_till_success`? I think Stephen Boyd often uses the term solver chain for this feature; I can ask him next week.

That would be great, thanks!

PTNobel:
Stephen said we should call it a "solver path" in the documentation; he also suggested making it a keyword argument to `.solve` instead of a new public function (maybe `.solve(path=[...])`. I'm wondering if maybe we could just specifialize on if the input is a string or a list to make `.solve(solver=[cp.SCS, cp.CLARABEL])` the API. Thoughts on that @SteveDiamond @phschiele ?

rileyjmurray:
I’m not sure about “path” as a keyword argument. I’d be in favor of
“solver_path” or “solver_priority” for keyword arguments to prob.solve.

PTNobel:
> I’m not sure about “path” as a keyword argument. I’d be in favor of “solver_path” or “solver_priority” for keyword arguments to prob.solve.

Fair! I don't like solver_priority since we fallback to the other solvers instead of giving up, so I think `solver_path` would be a good choice.

erelsgl:
What if both "solver" and "solver_path" are given?

PTNobel:
I think that should raise a runtime error.

rileyjmurray:
That should raise a ValueError

SteveDiamond:
> Stephen said we should call it a "solver path" in the documentation; he also suggested making it a keyword argument to `.solve` instead of a new public function (maybe `.solve(path=[...])`. I'm wondering if maybe we could just specifialize on if the input is a string or a list to make `.solve(solver=[cp.SCS, cp.CLARABEL])` the API. Thoughts on that @SteveDiamond @phschiele ?

How would this work when you have solver specific arguments to the solve method?

phschiele:
> > Stephen said we should call it a "solver path" in the documentation; he also suggested making it a keyword argument to `.solve` instead of a new public function (maybe `.solve(path=[...])`. I'm wondering if maybe we could just specifialize on if the input is a string or a list to make `.solve(solver=[cp.SCS, cp.CLARABEL])` the API. Thoughts on that @SteveDiamond @phschiele ?
> 
> How would this work when you have solver specific arguments to the solve method?

Could be `.solve(solver=[(cp.SCS, scs_args), (cp.CLARABEL, clarabel_args)])` but it might be getting complicated at this point.

oriyalperin:
I've made some changes according to your suggestions. Does this meet your expectations?

PTNobel:
I like this PR with Oriya's changes. Since this is a change to the public API, let's get maintainer consensus on this proposal before writing the new docs and merging.

cvxpy/problems/problem.py
```diff
@@ -415,6 +415,58 @@ def compilation_time(self) -> float | None:
...
+        Raises
+        ------
+        SolverError
+            If all solvers fail to find a solution.
+        ValueError
+            If the input solvers format is incorrect.
+        """
+            
+        if not solvers:
+            raise ValueError("Solvers list must contain at least one solver.")
```
PTNobel, line 444: This should say solver_path

PTNobel, line 444:
```suggestion
            raise ValueError("Solver path must contain at least one solver.")
```

cvxpy/problems/problem.py
```diff
@@ -415,6 +415,58 @@ def compilation_time(self) -> float | None:
                    last time it was compiled.
         """
         return self._compilation_time
+    
+    def _solve_solver_path(self, solve_func, solvers:list[tuple[str, dict] | tuple[str] | str],
+                                *args, **kwargs):
```
PTNobel, line 420:
```suggestion
                                args, kwargs):
```

cvxpy/problems/problem.py
```diff
@@ -500,6 +552,13 @@ def solve(self, *args, **kwargs):
             solve_func = Problem.REGISTERED_SOLVE_METHODS[func_name]
         else:
             solve_func = Problem._solve
+        solver_path = kwargs.pop("solver_path", None)
+        if solver_path is not None:
+            solver = kwargs.get("solver", None)
+            if solver is not None:
+                raise ValueError(
+                    "Cannot specify both 'solver' and 'solver_path'. Please choose one.")
+            return self._solve_solver_path(solve_func,solver_path,*args,**kwargs)
```
PTNobel, line 561:
```suggestion
            return self._solve_solver_path(solve_func,solver_path, args, kwargs)
```

cvxpy/problems/problem.py
```diff
@@ -415,6 +415,58 @@ def compilation_time(self) -> float | None:
...
+                        solver_name, = solver
+                        solution = solve_func(self, *args, solver=solver_name, **kwargs)
+                    elif len(solver) == 2:
+                        solver_name, solver_kwargs = solver
+                        if not isinstance(solver_name, str) or not isinstance(solver_kwargs, dict):
+                            raise ValueError("Solver tuple input must be (str, dict).")
+                        solution = solve_func(
+                            self, *args, solver=solver_name, **solver_kwargs, **kwargs)
+                    else:
+                        raise ValueError("Solver tuple input must be (str, dict) or (str)")
```
PTNobel, line 461:
```suggestion
                        raise ValueError("Solver path entry must be str, tuple[str] or tuple[str, dict[str, Any]]")
```

cvxpy/problems/problem.py
```diff
@@ -415,6 +415,58 @@ def compilation_time(self) -> float | None:
...
+                else:
+                    raise ValueError("Solver input must be one of the following:\
+                                                (str, dict) tuple, (str) tuple, or str.")
```
PTNobel, line 464: Can you define a constant string somewhere and use the same message on line 461 and 463?

cvxpy/tests/test_problem.py
```diff
@@ -1416,6 +1416,52 @@ def test_solver_error_raised_on_failure(self) -> None:
...
+        # invalid input, raise ValueError
+        solvers_invalid_inner_input = [[], [1], [()], [(1)], [(1,{})],[(s.OSQP,[])]]
+
+        for solvers in solvers_invalid_inner_input:
+            with self.assertRaises(ValueError):
+                Problem(cp.Minimize(
+                    cp.sum_squares(cp.matmul(A, cp.Variable(40)) - b))).solve(
+                    solver_path=solvers)
+
+        ## solver_path ans solver arguments
```
PTNobel, line 1459:
```suggestion
        ## Specify both solver_path and solver as arguments
```

PTNobel (changes requested):

SteveDiamond:
This approach looks good to me.

PTNobel (approved):
Thanks Oriya! Everything looks good on my end.

phschiele:
We're still missing the documentation for `solve`. Right now, only the private method is documented.

PTNobel:
This does needs some docs. Certainly it should be added to the .solve docstring and added to this page: https://www.cvxpy.org/tutorial/solvers/index.html Maybe as a new section title "Solver paths" that discusses how this provides a fallback on failure to solve.

oriyalperin:
I hope I've placed the documentation correctly.
If any adjustments are needed, please let me know.
Thank you!

jonathanberthias:
Hi, could you clarify what the `tuple[str]` case is for? Don't the `str` and `tuple[str, dict]` cover all cases, i.e. either no solver specific options, or some options?

oriyalperin:
> Hi, could you clarify what the `tuple[str]` case is for? Don't the `str` and `tuple[str, dict]` cover all cases, i.e. either no solver specific options, or some options?

Hi @jonathanberthias,
Perhaps it should be written as `tuple[str,]` to indicate that the type is a tuple containing only `str`.
The list items can be in three formats:
1. **tuple** of `(str, dict)`, such as `('OSQP', {'max_iter': 10000})`.
2. **tuple** of `(str,)`, such as `('CLARABEL',)`.
3. **string**, such as `'CLARABEL'` or `('CLARABEL')`

jonathanberthias:
Thanks @oriyalperin for answering so quickly! 

What I meant was, when would it be useful to use the 1-str tuple form (option 2. in your list)? As I see things, I either want to pass solver-specific options and would use option 1., or I don't in which case option 3. is perfect. And if I am writing some generic code where I sometimes want solver-specific options and sometimes not, then I can use option 1. with a perhaps empty dictionary. 

But in no case I could think of would it be useful to pass the solver name in a tuple. 🤔

phschiele:
@jonathanberthias it would be useful only if the tuple contains multiple strings, e.g., `("CLARABEL", "ECOS")`.

On that note, though, let's update the documentation of both functions to use `Iterable` instead of tuple/list.

oriyalperin:
> Thanks @oriyalperin for answering so quickly!
> 
> What I meant was, when would it be useful to use the 1-str tuple form (option 2. in your list)? As I see things, I either want to pass solver-specific options and would use option 1., or I don't in which case option 3. is perfect. And if I am writing some generic code where I sometimes want solver-specific options and sometimes not, then I can use option 1. with a perhaps empty dictionary.
> 
> But in no case I could think of would it be useful to pass the solver name in a tuple. 🤔

I agree, it's not necessary at all, it was just to offer an alternative option.
Thanks for pointing that out, I'll remove it.

oriyalperin:
@phschiele,
1. I need some clarification, does that require supporting all iterable types (dict, str...)?
2. Replacing tuple with `Iterable` - the tuple is only used as an inner type. Do you want to replace it in that context?

erelsgl:
> On that note, though, let's update the documentation of both functions to use `Iterable` instead of tuple/list.

@phschiele if we use `iterable`, we also implicitly allow `solvers` to be a string (an iterable of chars), which is of course incorrect. So maybe it is better to allow only `list`?

PTNobel:
> > On that note, though, let's update the documentation of both functions to use `Iterable` instead of tuple/list.
> 
> @phschiele if we use `iterable`, we also implicitly allow `solvers` to be a string (an iterable of chars), which is of course incorrect. So maybe it is better to allow only `list`?

I concur that we should not allow str. I support requiring an explicit list.

phschiele:
I was suggesting 
```
Iterable[tuple[str, dict] | str]
```
which a single string does not satisfy.

Requiring a list is also fine with me.

oriyalperin:
Thank you for your feedback and reviews! I really appreciate it.

PTNobel:
@SteveDiamond did you have outstanding TODOs or can I merge this?

SteveDiamond (approved):
LGTM

PR #2457, opened 2024-05-28
Small optimization for upper_tri_to_full

Transurgeon:
## Description
I was looking into the variable file and found the ``upper_tri_to_full`` function a bit out of place.
It seems like it is only being referenced in one location: ``cvx_attr2constr``, so I decided to move it there.
If the function is further used in the future, I propose that we move it to somewhere in ``cvxpy.utilities``.

Performance difference for (n=100):
```
Original upper_tri_to_full time: 0.052485 seconds
Optimized upper_tri_to_full time: 0.004270 seconds
```
The performance is also more significant when n > 1000.

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel:
I don't follow what is going on in the new implementation, any chance you could add a few more comments?

Transurgeon:
> could you add a test for upper_tri_to_full? I checked and there aren't any tests in the current codebase.

Where should we put this test?

SteveDiamond:
> > could you add a test for upper_tri_to_full? I checked and there aren't any tests in the current codebase.
> 
> Where should we put this test?

I didn't find any great place, you can do it in cvxpy/tests/test_atoms.py

PTNobel:
> > I don't follow what is going on in the new implementation, any chance you could add a few more comments?
> 
> Do you want me to explain what the original function is doing or explain how the optimised version is equivalent to the original one?

Just what the optimized one is doing would be fine with me... Just make it feel a little less like arcane magic would be my goal; I'm also fine with it merging without, I just thikn it would help readability.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -56,6 +56,35 @@
...
+    """
+    entries = n*(n+1)//2
+
+    # Initialize row and col indices from upper triangular matrix
+    rows, cols = np.triu_indices(n)
+
+    # Mask for the symmetric part when i != j
+    mask = rows != cols
+
+    row_idx = np.concatenate([rows * n + cols, cols[mask] * n + rows[mask]])
```
PTNobel, line 81: Why swap rows and cols between the first and second entries here?

PTNobel (approved):
Changes all look good to me! Rereading it, I think the new code makes lots of sense and you don't need to worry too much about any new docs. I did have one small q.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -56,6 +56,35 @@
 ]
 
 
+def upper_tri_to_full(n: int) -> sp.csc_matrix:
+    """
+    Returns a coefficient matrix to create a symmetric matrix.
```
phschiele, line 61: I think also we should improve the docstring.
Maybe something like:
```
Returns coefficient matrix A that, when multiplied with a vector variable and reshaped, creates a symmetric matrix.
That is, 
(A @  v).reshape((n, n)) 
is a symmetric matrix.

```

Transurgeon: done, thanks for the clear and concise explanation.

cvxpy/atoms/affine/upper_tri.py
```diff
@@ -134,19 +130,15 @@ def vec_to_upper_tri(expr, strict: bool = False):
...
-    #       the i-th block of n rows of P gives the entries of the i-th row
-    #       of the upper-triangular matrix associated with expr.
-    # compute expr2 = P @ expr
-    # compute expr3 = reshape(expr2, shape=(n, n)).T
-    #   expr3 is the matrix formed by reading length-n blocks of expr2,
-    #   and letting each block form a row of expr3.
+    """
+    Initialize a coefficient matrix P that creates an upper triangular matrix when 
+    multiplied with a variable vector expr.
+    That is, (P @ expr).reshape((n, n)) is an upper triangular matrix.
```
Transurgeon, line 136: I added a clearer explanation that is similar to what Philipp suggested for the upper_tri_to_full function.

cvxpy/reductions/cvx_attr2constr.py
Transurgeon: So the first entry for both row and col indices are for the upper triangular portion of the coefficient matrix. See the ``vec_to_upper_tri`` function [here](https://github.com/cvxpy/cvxpy/blob/a9447e0f5c6b542a1e2208f91c560d9f61e46b85/cvxpy/atoms/affine/upper_tri.py#L146) for an example. The second entry is for the lower triangular part of the matrix, hence cols are repeating and rows somehow must swap. I am not entirely sure how this results in the correct coefficient matrix, but it follows the original code which had ``row_arr.append(i*n + j)`` and ``row_arr.append(j*n + i)``, respectively.

cvxpy/atoms/affine/upper_tri.py
```diff
@@ -134,19 +130,15 @@ def vec_to_upper_tri(expr, strict: bool = False):
...
-    P_rows = n*row_idx + col_idx
+    P_rows = n * row_idx + col_idx
     P_cols = np.arange(ell)
     P_vals = np.ones(P_cols.size)
-    P = csc_matrix((P_vals, (P_rows, P_cols)), shape=(n**2, ell))
-    expr2 = P @ expr
-    expr3 = reshape(expr2, (n, n)).T
-    return expr3
+    P = csc_matrix((P_vals, (P_rows, P_cols)), shape=(n * n, ell))
+    return reshape(P @ expr, (n, n)).T
```
Transurgeon, line 144: simplified the code here since it is pretty clear with the new comments above.

PTNobel (approved):

cvxpy/tests/test_atoms.py
```diff
@@ -796,6 +796,13 @@ def test_vec_to_upper_tri(self) -> None:
         # works with scalars
         assert np.allclose(cp.vec_to_upper_tri(1, strict=True).value, np.array([[0, 1], [0, 0]]))
 
+    def test_upper_tri_to_full(self) -> None:
+        from cvxpy.reductions.cvx_attr2constr import upper_tri_to_full
```
SteveDiamond, line 800: imports should go at the start of the file.

Transurgeon: sounds good!
I was thinking we could maybe move ``upper_tri_to_full`` to ``cvxpy.atoms.affine.upper_tri`` since it is kind of related.. I feel like it is a bit odd to import a function from a reduction.
What do you think?

SteveDiamond: Agreed, it would be better in cvxpy.atoms.affine.upper_tri

Transurgeon:
This PR looks good on my end and CI are finally all passing :).
Thanks a lot for the reviews guys!

I had a quick question though. It seems like there is another potential speedup (see the stack overflow [answer](https://stackoverflow.com/a/58806626) for reference). 
By doing something like the following:
```
# Initialise matrix X with correct shape
X = np.zeros((n, n))
# Set matrix at upper triangular indices equal to the vector
X[np.triu_indices(n)] = vec
```
However, having tried that, I noticed that cvxpy.Expression values are not always known unlike numpy arrays which doesn't allow this nice simplification. Since I am trying to become more familiar with cvxpy objects, I was just wondering if this was a common occurrence (needing to create a coefficient matrix then performing A @ v) amongst other atoms? Also, do you believe there is any need for adding a fast path, for example if the expression.value is pre-defined by the user?

SteveDiamond (approved):

SteveDiamond:
> This PR looks good on my end and CI are finally all passing :). Thanks a lot for the reviews guys!
> 
> I had a quick question though. It seems like there is another potential speedup (see the stack overflow [answer](https://stackoverflow.com/a/58806626) for reference). By doing something like the following:
> 
> ```
> # Initialise matrix X with correct shape
> X = np.zeros((n, n))
> # Set matrix at upper triangular indices equal to the vector
> X[np.triu_indices(n)] = vec
> ```
> 
> However, having tried that, I noticed that cvxpy.Expression values are not always known unlike numpy arrays which doesn't allow this nice simplification. Since I am trying to become more familiar with cvxpy objects, I was just wondering if this was a common occurrence (needing to create a coefficient matrix then performing A @ v) amongst other atoms? Also, do you believe there is any need for adding a fast path, for example if the expression.value is pre-defined by the user?

expr.value is relevant for numeric evaluation but not for the symbolic problem construction. In fully defined atoms we distinguish between the two but upper_tri_to_full is sort of a half defined atom where both aspects are combined.

phschiele (approved):
Clean implementation! 🎉

PR #2462, opened 2024-05-31
allow M to be a Parameter in the huber function

syanga:
## Description
This change allows the M parameter in the huber function to be a parameter for DPP.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/huber.py
+++ b/cvxpy/atoms/elementwise/huber.py
@@ -20,7 +20,6 @@ import numpy as np
 import scipy.special
 
 from cvxpy.atoms.elementwise.elementwise import Elementwise
-from cvxpy.expressions.constants.parameter import Parameter
 
 # TODO(akshayka): DGP support.
 
--- a/cvxpy/atoms/elementwise/huber.py
+++ b/cvxpy/atoms/elementwise/huber.py
@@ -47,13 +47,13 @@ class huber(Elementwise):
     """
 
     def __init__(self, x, M: int = 1) -> None:
-        self.M = M if isinstance(M, Parameter) else self.cast_to_const(M)
+        self.M = self.cast_to_const(M)
         super(huber, self).__init__(x)
 
     def parameters(self):
         """If M is a Parameter, include it in the list of Parameters"""
         if isinstance(self.M, Parameter):
-            return super().parameters() + [self.M]
+            return super().parameters() + self.M.parameters()
         return super().parameters()
 
     @Elementwise.numpy_numeric
--- a/cvxpy/atoms/elementwise/huber.py
+++ b/cvxpy/atoms/elementwise/huber.py
@@ -22,7 +22,6 @@ import scipy.special
 from cvxpy.atoms.elementwise.elementwise import Elementwise
 from cvxpy.expressions.constants.parameter import Parameter
 
-
 # TODO(akshayka): DGP support.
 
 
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_atoms.py
```diff
@@ -840,6 +840,19 @@ def test_huber(self) -> None:
         self.assertTrue(copy.args[0] is self.y)
         self.assertEqual(copy.get_data()[0].value, atom.get_data()[0].value)
 
+        # Test usage of M as parameter
+        M = Parameter(nonneg=True)
+        expr = cp.huber(1.0, M)
```
phschiele, line 845: Probably we should also add a test case for something like `cp.huber(1.0,  2 * M)` (and decide if we want that to work or not).

cvxpy/atoms/elementwise/huber.py
```diff
@@ -41,68 +42,62 @@ class huber(Elementwise):
...
     x : Expression
         The expression to which the huber function will be applied.
-    M : Constant
+    M : Constant or Parameter
         A scalar constant.
     """
 
     def __init__(self, x, M: int = 1) -> None:
-        self.M = self.cast_to_const(M)
+        self.M = M if isinstance(M, Parameter) else self.cast_to_const(M)
```
SteveDiamond, line 50: This change is not necessary.

cvxpy/atoms/elementwise/huber.py
```diff
@@ -41,68 +42,62 @@ class huber(Elementwise):
...
         super(huber, self).__init__(x)
 
+    def parameters(self):
```
SteveDiamond, line 53: This function is not necessary I believe.

SteveDiamond, line 53: This should be more like ```return super().parameters() + self.M.parameters()```. Your code doesn't handle the case where M is an expression that is parameterized.

cvxpy/tests/test_atoms.py
PTNobel: I think we do want that to work. Do we have any other situation where an affine expression of a parameter wouldn't work?

syanga: I added a test for affine expressions.

cvxpy/atoms/elementwise/huber.py
```diff
@@ -41,68 +42,62 @@ class huber(Elementwise):
...
         A scalar constant.
     """
 
     def __init__(self, x, M: int = 1) -> None:
         self.M = self.cast_to_const(M)
         super(huber, self).__init__(x)
 
+    def parameters(self):
+        """If M is a Parameter, include it in the list of Parameters"""
+        if isinstance(self.M, Parameter):
```
PTNobel, line 55: What if M is a parametrized expression? I don't see why we have an if clause here

PTNobel (changes requested):

PTNobel (approved):

cvxpy/tests/test_atoms.py
```diff
@@ -840,6 +840,27 @@ def test_huber(self) -> None:
...
+        self.assertAlmostEqual(expr.value, 1.0)
+
+        # Test M as affine expression of parameter
+        expr = cp.huber(1.0, 0.25 * M + 0.25)
+        M.value = 1
+        self.assertAlmostEqual(expr.value, 0.75)
+        M.value = 0.5
+        self.assertAlmostEqual(expr.value, 0.609375)
+        M.value = 1
+        self.assertAlmostEqual(expr.value, 0.75)
```
SteveDiamond, line 862: Could we test that huber with parameterized M works properly with DPP? We have had some recent correctness bugs related to DPP so it is important to investigate/cover with tests.

SteveDiamond (approved):
Thanks!!!

SteveDiamond:
@syanga nice job!

PR #2473, opened 2024-06-14
Adding support for n-dimensional sum(axis, keepdims) operation

Transurgeon:
## Description
This PR aims to add support for n-dimensional sum operations with the ability to pass in any axis, including multiple axes as well. In addition, a new function has been added to choose the correct backend depending on the dimensions of the underlying expressions of the problem. 

This PR is also a small step towards enabling full nd support in light of cvxpy 2.0, hence I have commented out the ValueErrors for related instances. 

Future todos to complement this PR:
- parametrized nd-sum
- keep the scipy backend implementation sparse (will need to do some indexing hacks).

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -15,7 +15,7 @@ limitations under the License.
 """
 import builtins
 from functools import wraps
-from typing import List, Optional, Tuple
+from typing import Iterable, List, Optional, Tuple
 
 import numpy as np
 
@@ -40,7 +40,7 @@ class Sum(AxisAtom, AffAtom):
         Whether to drop dimensions after summing.
     """
 
-    def __init__(self, expr, axis: Optional[int] = None, keepdims: bool = False) -> None:
+    def __init__(self, expr, axis: Optional[int | Iterable] = None, keepdims: bool = False) -> None:
         super(Sum, self).__init__(expr, axis=axis, keepdims=keepdims)
 
     def is_atom_log_log_convex(self) -> bool:
@@ -67,7 +67,8 @@ class Sum(AxisAtom, AffAtom):
     def graph_implementation(
         self, arg_objs, shape: Tuple[int, ...], data=None
     ) -> Tuple[lo.LinOp, List[Constraint]]:
-        """Sum the linear expression's entries.
+        """
+        Sum the linear expression's entries.
 
         Parameters
         ----------
@@ -75,8 +76,8 @@ class Sum(AxisAtom, AffAtom):
             LinExpr for each argument.
         shape : tuple
             The shape of the resulting expression.
-        data :
-            Additional data required by the atom.
+        data : [axis, keepdims]
+            The axis and keepdims parameters of the sum expression.
 
         Returns
         -------
@@ -84,6 +85,8 @@ class Sum(AxisAtom, AffAtom):
             (LinOp for objective, list of constraints)
         """
         axis, keepdims = data
+        # Note: added new case for summing with n-dimensional shapes and 
+        # multiple axes. Previous behavior is kept in the else statement.
         if len(arg_objs[0].shape) > 2 or axis not in {None, 0, 1}:
             obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         else:
@@ -107,8 +110,9 @@ class Sum(AxisAtom, AffAtom):
 
 
 @wraps(Sum)
-def sum(expr, axis: Optional[int] = None, keepdims: bool = False):
-    """Wrapper for Sum class.
+def sum(expr, axis: Optional[int | Iterable] = None, keepdims: bool = False):
+    """
+    Wrapper for Sum class.
     """
     if isinstance(expr, list):
         return builtins.sum(expr)
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -86,7 +86,7 @@ class Sum(AxisAtom, AffAtom):
         axis = data[0]
         keepdims = data[1]
         if axis is None:
-            obj = lu.sum_entries(arg_objs[0], shape=shape)
+            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         elif axis == 1:
             if keepdims:
                 const_shape = (arg_objs[0].shape[1], 1)
@@ -101,7 +101,7 @@ class Sum(AxisAtom, AffAtom):
                 const_shape = (arg_objs[0].shape[0],)
             ones = lu.create_const(np.ones(const_shape), const_shape)
             obj = lu.mul_expr(ones, arg_objs[0], shape)
-
+        obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         return (obj, [])
 
 
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -86,7 +86,7 @@ class Sum(AxisAtom, AffAtom):
         axis = data[0]
         keepdims = data[1]
         if axis is None:
-            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
+            obj = lu.sum_entries(arg_objs[0], shape=shape)
         elif axis == 1:
             if keepdims:
                 const_shape = (arg_objs[0].shape[1], 1)
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -22,7 +22,6 @@ import numpy as np
 import cvxpy.interface as intf
 import cvxpy.lin_ops.lin_op as lo
 import cvxpy.lin_ops.lin_utils as lu
-import cvxpy.settings as s
 from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.atoms.axis_atom import AxisAtom
 from cvxpy.constraints.constraint import Constraint
@@ -84,9 +83,10 @@ class Sum(AxisAtom, AffAtom):
         tuple
             (LinOp for objective, list of constraints)
         """
-        if s.DEFAULT_CANON_BACKEND == 'CPP':
-            axis = data[0]
-            keepdims = data[1]
+        axis, keepdims = data
+        if len(arg_objs[0].shape) > 2 or axis not in {None, 0, 1}:
+            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
+        else:
             if axis is None:
                 obj = lu.sum_entries(arg_objs[0], shape=shape)
             elif axis == 1:
@@ -103,9 +103,6 @@ class Sum(AxisAtom, AffAtom):
                     const_shape = (arg_objs[0].shape[0],)
                 ones = lu.create_const(np.ones(const_shape), const_shape)
                 obj = lu.mul_expr(ones, arg_objs[0], shape)
-        else:
-            axis, keepdims = data
-            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         return (obj, [])
 
 
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -15,7 +15,7 @@ limitations under the License.
 """
 import builtins
 from functools import wraps
-from typing import Iterable, List, Optional, Tuple
+from typing import List, Optional, Tuple
 
 import numpy as np
 
@@ -40,7 +40,7 @@ class Sum(AxisAtom, AffAtom):
         Whether to drop dimensions after summing.
     """
 
-    def __init__(self, expr, axis: Optional[int | Iterable] = None, keepdims: bool = False) -> None:
+    def __init__(self, expr, axis: Optional[int] = None, keepdims: bool = False) -> None:
         super(Sum, self).__init__(expr, axis=axis, keepdims=keepdims)
 
     def is_atom_log_log_convex(self) -> bool:
@@ -110,7 +110,7 @@ class Sum(AxisAtom, AffAtom):
 
 
 @wraps(Sum)
-def sum(expr, axis: Optional[int | Iterable] = None, keepdims: bool = False):
+def sum(expr, axis: Optional[int] = None, keepdims: bool = False):
     """
     Wrapper for Sum class.
     """
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -83,24 +83,7 @@ class Sum(AxisAtom, AffAtom):
         tuple
             (LinOp for objective, list of constraints)
         """
-        axis = data[0]
-        keepdims = data[1]
-        if axis is None:
-            obj = lu.sum_entries(arg_objs[0], shape=shape)
-        elif axis == 1:
-            if keepdims:
-                const_shape = (arg_objs[0].shape[1], 1)
-            else:
-                const_shape = (arg_objs[0].shape[1],)
-            ones = lu.create_const(np.ones(const_shape), const_shape)
-            obj = lu.rmul_expr(arg_objs[0], ones, shape)
-        else:  # axis == 0
-            if keepdims:
-                const_shape = (1, arg_objs[0].shape[0])
-            else:
-                const_shape = (arg_objs[0].shape[0],)
-            ones = lu.create_const(np.ones(const_shape), const_shape)
-            obj = lu.mul_expr(ones, arg_objs[0], shape)
+        axis, keepdims = data
         obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         return (obj, [])
 
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -22,6 +22,7 @@ import numpy as np
 import cvxpy.interface as intf
 import cvxpy.lin_ops.lin_op as lo
 import cvxpy.lin_ops.lin_utils as lu
+import cvxpy.settings as s
 from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.atoms.axis_atom import AxisAtom
 from cvxpy.constraints.constraint import Constraint
@@ -83,8 +84,28 @@ class Sum(AxisAtom, AffAtom):
         tuple
             (LinOp for objective, list of constraints)
         """
-        axis, keepdims = data
-        obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
+        if s.DEFAULT_CANON_BACKEND == 'CPP':
+            axis = data[0]
+            keepdims = data[1]
+            if axis is None:
+                obj = lu.sum_entries(arg_objs[0], shape=shape)
+            elif axis == 1:
+                if keepdims:
+                    const_shape = (arg_objs[0].shape[1], 1)
+                else:
+                    const_shape = (arg_objs[0].shape[1],)
+                ones = lu.create_const(np.ones(const_shape), const_shape)
+                obj = lu.rmul_expr(arg_objs[0], ones, shape)
+            else:  # axis == 0
+                if keepdims:
+                    const_shape = (1, arg_objs[0].shape[0])
+                else:
+                    const_shape = (arg_objs[0].shape[0],)
+                ones = lu.create_const(np.ones(const_shape), const_shape)
+                obj = lu.mul_expr(ones, arg_objs[0], shape)
+        else:
+            axis, keepdims = data
+            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
         return (obj, [])
 
 
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -49,10 +49,6 @@ class Atom(Expression):
         self.args = [Atom.cast_to_const(arg) for arg in args]
         self.validate_arguments()
         self._shape = self.shape_from_args()
-        """
-        if len(self._shape) > 2:
-            raise ValueError("Atoms must be at most 2D.")
-        """
 
     def name(self) -> str:
         """Returns the string representation of the function call.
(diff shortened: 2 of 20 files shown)


cvxpy/atoms/affine/sum.py
```diff
@@ -83,25 +84,28 @@ def graph_implementation(
...
-            obj = lu.rmul_expr(arg_objs[0], ones, shape)
-        else:  # axis == 0
-            if keepdims:
-                const_shape = (1, arg_objs[0].shape[0])
-            else:
-                const_shape = (arg_objs[0].shape[0],)
-            ones = lu.create_const(np.ones(const_shape), const_shape)
-            obj = lu.mul_expr(ones, arg_objs[0], shape)
-
+        if s.DEFAULT_CANON_BACKEND == 'CPP':
```
Transurgeon, line 87: The following changes are to deal with the disparity between the CPP and python backends.
However, it doesn't work because we can only access the default backend, not the actual backend that is used for canonicalization. 
I was wondering what's the best way to approach this? @phschiele any thoughts? Should we add another global variable for this?

cvxpy/atoms/axis_atom.py
```diff
@@ -36,33 +36,48 @@ def __init__(self, expr, axis: Optional[int] = None, keepdims: bool = False) ->
...
-        elif not self.keepdims and self.axis is None:
-            shape = []
+        if self.axis is None:
+            return (1,) * len(shape) if self.keepdims else ()
+        elif isinstance(self.axis, int):
+            if self.keepdims:
+                shape[self.axis] = 1
+            else:
+                shape = shape[:self.axis] + shape[self.axis+1:]
         else:
```
Transurgeon, line 50: This is the case when axis is a Tuple (multiple axes) and we need to iterate through all of them setting the values to one (if keepdims is True) or removing those axes (if keepdims is False)

cvxpy/atoms/axis_atom.py
```diff
@@ -36,33 +36,48 @@ def __init__(self, expr, axis: Optional[int] = None, keepdims: bool = False) ->
...
+        Checks that each axis is within the valid range.
+        """
+        if self.axis is not None:
+            axes = [self.axis] if isinstance(self.axis, int) else self.axis
+            dim = self.args[0].ndim
+            for axis in axes:
+                if axis < 0:
+                    axis += dim
+                if axis >= dim or axis < 0:
+                    raise ValueError(f"axis {axis} is out of bounds for array of dimension {dim}")
```
Transurgeon, line 75: I have generalized this function to check for multiple axes.
I also added a case for negative indices by summing dim to it (this is what's happening in Python behind the scenes). I also update the error message to be similar to NumPy's.

cvxpy/expressions/expression.py
```diff
@@ -115,7 +115,8 @@ class Expression(u.Canonical):
     # Handles arithmetic operator overloading with Numpy.
     __array_priority__ = 100
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
```
Transurgeon, line 119: This way of defining an abstract property is deprecated. See [here](https://stackoverflow.com/a/48710068)

cvxpy/tests/test_atoms.py
```diff
@@ -366,14 +366,16 @@ def test_max(self) -> None:
 
         # Test with axis argument.
         self.assertEqual(cp.max(Variable(2), axis=0, keepdims=True).shape, (1,))
-        self.assertEqual(cp.max(Variable(2), axis=1).shape, (2,))
```
Transurgeon, line 369: This test is being replaced below because the behavior will be deprecated in the future. 
See comment [here](https://github.com/cvxpy/cvxpy/issues/1963#issuecomment-1331151141)

cvxpy/tests/test_dqcp.py
```diff
@@ -695,9 +695,10 @@ def test_scalar_sum(self) -> None:
         problem.solve(SOLVER, qcp=True)
         self.assertAlmostEqual(problem.value, 0, places=3)
 
-        problem = cp.Problem(cp.Minimize(cp.cumsum(1/x)))
-        problem.solve(SOLVER, qcp=True)
-        self.assertAlmostEqual(problem.value, 0, places=3)
+        with self.assertRaises(Exception) as cm:
+            problem = cp.Problem(cp.Minimize(cp.cumsum(1/x)))
+            problem.solve(SOLVER, qcp=True)
+        self.assertEqual(str(cm.exception), "axis 0 is out of bounds for array of dimension 0")
```
Transurgeon, line 701: I encountered a weird bug for cumsum because of the ``validate_arguments`` function changes. 
I will have a further look at the cumsum atom in a future PR. It has yet to be updated for n-dimensional support. 

Also, the cumsum atom seems to behave quite differently than other axis atoms in that the axis that we specify in the argument is actually the axis that is being kept rather than removed (or set to 1, if keepdims is True).

cvxpy/tests/test_expressions.py
```diff
@@ -1532,3 +1514,51 @@ def test_wraps(self) -> None:
...
+        prob.solve(canon_backend=cp.SCIPY_CANON_BACKEND)
+        assert np.allclose(expr.value, self.target)
+
+    @pytest.mark.parametrize("axis", [(0),(1),(2),((0,1)),((0,2)),((2,1))])
+    def test_nd_sum(self, axis) -> None:
+        expr = cp.sum(self.x, axis=axis, keepdims=True)
+        y = self.target.sum(axis=axis, keepdims=True)
+        prob = cp.Problem(self.obj, [expr == y])
+        prob.solve(canon_backend=cp.SCIPY_CANON_BACKEND)
+        assert np.allclose(expr.value, y)
```
Transurgeon, line 1564: These are the main tests that show the new features introduced in this PR.
We are able to mimic NumPy in many different ways for nd tensors, especially for sum we are able to specify multiple axes.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -761,7 +761,21 @@ def sum_entries(_lin: LinOp, view: NumPyTensorView) -> NumPyTensorView:
...
+                n = x.shape[-1]
+                p = x.shape[0]
+                if isinstance(axis, tuple):
+                    d = np.prod([shape[i] for i in axis], dtype=int)
+                    axis = tuple([a + 1 for a in axis])
+                else:
+                    d = shape[axis]
+                    axis += 1
+                x = x.reshape((p,)+(shape)+(n,), order='F').sum(axis=axis)
+                return x.reshape((p, n//d, n), order='F')
```
Transurgeon, line 778: This function contains the main changes for cvxpy to support nd sum.
The desired output is explained in greater detail in the test_python_backends file.
The goal is to turn an identity to a reduced matrix representing the summands of the original expression in column major order.
The idea came from @SteveDiamond to do the following:
```py
# Suppose we want to sum Variable(2,2,2)
x = np.eye(8)
out = x.reshape(2,2,2,8).sum(axis=axis).reshape(4,8)
```
Which produces the desired result for general cases. The idea is to keep the columns of the tensor fixed and reshape the remaining dimensions into the shape of the expression, then we simply perform the desired NumPy sum operation and reshape the result back into the desired output shape.
In the code above, p is the number of parameters, d is the output row (basically np.prod the remaining indices after summing) and n is the number of columns (hence ``tensor.shape[-1]``).

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -377,6 +378,43 @@ def construct_solving_chain(problem, candidates,
...
+        The canonicalization backend to use.
+    """
+    max_var_dim = max([p.ndim for p in problem.variables()] or [0])
+    max_param_dim = max([p.ndim for p in problem.parameters()] or [0])
+    if max(max_var_dim, max_param_dim) > 2:
+        if canon_backend is None:
+            warnings.warn(UserWarning(
+                "The problem has an expression with dimension greater than 2. "
+                "Defaulting to the 'SCIPY' backend for canonicalization."))
+            return "SCIPY"
```
SteveDiamond, line 411: use the constants from settings.py for SCIPY and CPP, not raw strings.

Transurgeon: ok will make a change now.

cvxpy/atoms/axis_atom.py
Transurgeon: Code is pretty straightforward to understand. Didn't feel the need to add more changes to the docstrings.

Transurgeon:
This PR looks good on my end. 
Any feedback, questions or suggestions are very welcome. 
Ping for reviews: @PTNobel @SteveDiamond @phschiele (when you are back from Mexico 😅 ).

Also we should have a discussion about our plans for documenting all this in the future.

cvxpy/cvxcore/python/cppbackend.py
```diff
@@ -175,7 +175,7 @@ def make_linC_from_linPy(linPy, linPy_to_linC) -> None:
     linC = cvxcore.LinOp(typ, shape, lin_args_vec)
     linPy_to_linC[linPy] = linC
 
-    if linPy.data is not None:
+    if linPy.data is not None and linPy.type != "sum_entries":
```
SteveDiamond, line 178: Can you add a comment about why this is here?

cvxpy/atoms/affine/sum.py
```diff
@@ -83,25 +83,26 @@ def graph_implementation(
...
-                const_shape = (1, arg_objs[0].shape[0])
-            else:
-                const_shape = (arg_objs[0].shape[0],)
-            ones = lu.create_const(np.ones(const_shape), const_shape)
-            obj = lu.mul_expr(ones, arg_objs[0], shape)
-
+        axis, keepdims = data
+        if len(arg_objs[0].shape) > 2 or axis not in {None, 0, 1}:
+            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
+        else:
```
SteveDiamond, line 89: Could you add a comment about why we have the two cases?

cvxpy/tests/test_dqcp.py
```diff
@@ -695,9 +695,10 @@ def test_scalar_sum(self) -> None:
         problem.solve(SOLVER, qcp=True)
         self.assertAlmostEqual(problem.value, 0, places=3)
 
-        problem = cp.Problem(cp.Minimize(cp.cumsum(1/x)))
-        problem.solve(SOLVER, qcp=True)
-        self.assertAlmostEqual(problem.value, 0, places=3)
+        with self.assertRaises(Exception) as cm:
```
SteveDiamond, line 698: Can you add a comment that we want this to pass and need to develop a fix?

Transurgeon: done.

cvxpy/atoms/affine/sum.py
Transurgeon: done.

cvxpy/cvxcore/python/cppbackend.py
Transurgeon: done. please let me know if the comments are accurate.

SteveDiamond (approved):

SteveDiamond:
This looks good to me.

cvxpy/atoms/affine/sum.py
```diff
@@ -67,47 +67,52 @@ def numeric(self, values):
...
-                const_shape = (arg_objs[0].shape[0],)
-            ones = lu.create_const(np.ones(const_shape), const_shape)
-            obj = lu.mul_expr(ones, arg_objs[0], shape)
-
+        axis, keepdims = data
+        # Note: added new case for summing with n-dimensional shapes and 
+        # multiple axes. Previous behavior is kept in the else statement.
+        if len(arg_objs[0].shape) > 2 or axis not in {None, 0, 1}:
+            obj = lu.sum_entries(arg_objs[0], shape=shape, axis=axis, keepdims=keepdims)
+        else:
```
PTNobel, line 92: Is there a reason to have the else case? Couldn't we just replace all this code with the ndarray code?

It would improve coverage and benchmarking on the new code.

cvxpy/atoms/atom.py
```diff
@@ -49,8 +49,6 @@ def __init__(self, *args) -> None:
         self.args = [Atom.cast_to_const(arg) for arg in args]
         self.validate_arguments()
         self._shape = self.shape_from_args()
-        if len(self._shape) > 2:
-            raise ValueError("Atoms must be at most 2D.")
```
PTNobel, line 53: I am a little hesitant to get rid of the error case in case we release 1.6.0 without finishing ndarrays. Can we leave it in but have it read a global constant for whether to support ndarrays?

cvxpy/atoms/axis_atom.py
PTNobel: Awesome! I assume the negative index behavior is the same as NumPy s?

cvxpy/expressions/constants/constant.py
```diff
@@ -67,14 +66,15 @@ def __init__(self, value, name: Optional[str] = None) -> None:
         super(Constant, self).__init__(intf.shape(self.value))
 
     def name(self) -> str:
-        """The value as a string.
+        """
+        The value of the constant as a string.
```
PTNobel, line 70:
```suggestion
         The value of the constant as a string.
```

cvxpy/expressions/leaf.py
```diff
@@ -56,7 +56,6 @@ class Leaf(expression.Expression):
     shape : Iterable of ints or int
         The leaf dimensions. Either an integer n for a 1D shape, or an
         iterable where the semantics are the same as NumPy ndarray shapes.
-        **Shapes cannot be more than 2D**.
```
PTNobel, line 59: I'm a little skeptical of updating the docs now

cvxpy/expressions/leaf.py
```diff
@@ -97,7 +96,7 @@ class Leaf(expression.Expression):
     __metaclass__ = abc.ABCMeta
 
     def __init__(
-        self, shape: int | Iterable[int, ...], value=None, nonneg: bool = False,
+        self, shape: int | Iterable, value=None, nonneg: bool = False,
```
PTNobel, line 99: Shouldn't it just be int | tuple[int, ...]

cvxpy/expressions/leaf.py
```diff
@@ -106,9 +105,6 @@ def __init__(
     ) -> None:
         if isinstance(shape, numbers.Integral):
             shape = (int(shape),)
-        elif len(shape) > 2:
-            raise ValueError("Expressions of dimension greater than 2 "
-                             "are not supported.")
```
PTNobel, line 111: Ditto above about deleting errors

cvxpy/lin_ops/canon_backend.py
```diff
@@ -226,7 +226,6 @@ def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorVie
         # Leaf nodes
         if lin_op.type == "variable":
             assert isinstance(lin_op.data, int)
-            assert len(lin_op.shape) in {0, 1, 2}
```
PTNobel, line 229: Ditto

PTNobel (commented):
I only got through about the first 3rd tonight. Will keep coming back to it. Would appreciate responses to some of my qs tho

cvxpy/atoms/affine/sum.py
SteveDiamond: The issue is that if we switch over to the ndarray code and remove the else statement, the CPP backend can no longer handle sum_entries. We wanted to keep the CPP backend working for the cases it currently supports. In the long run it would be worthwhile to update the CPP backend so it can handle sum_entries with axis and keepdims arguments, but that was too big a lift for this MR.

cvxpy/atoms/atom.py
SteveDiamond: I told @Transurgeon not to use global constants. But we can discuss more.

cvxpy/expressions/leaf.py
Transurgeon: resolved.

cvxpy/expressions/leaf.py
Transurgeon: yea I agree,  I reverted it for now.

cvxpy/atoms/atom.py
Transurgeon: I added a global variable ``ALLOW_ND_EXPR``. and for all the ValueError assertions, I first check if allow_nd_expr is false. @PTNobel does this look good to you know?

cvxpy/atoms/affine/sum.py
Transurgeon: Yeah I think eventually we would like to simplify this code without the else case.
However, it needs to stay for now to support the CPP backend which doesn't support nd sum, let alone passing in the axis and keepdims params.

PTNobel: I think I'm misunderstanding something then. What is `lu`?  I don't see it as a parameter to this function and I think I'm missing where it fits in. Is this part of the backend API? I had thought we had encapsulated those.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -226,7 +227,8 @@ def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorVie
...
+            if not s.ALLOW_ND_EXPR:
+                 assert len(lin_op.shape) in {0, 1, 2}
```
PTNobel, line 231:
```suggestion
             assert s.ALLOW_ND_ARRAY or len(lin_op.shape) in {0, 1, 2}
```
And maybe drop the if statement? More readable personally.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -758,10 +760,36 @@ def func(x):
...
+                n = x.shape[-1]
+                p = x.shape[0]
+                if isinstance(axis, tuple):
+                    d = np.prod([shape[i] for i in axis], dtype=int)
+                    # adding offset of 1 to every axis because of param axis.
+                    axis = tuple([a + 1 for a in axis])
+                else:
+                    d = shape[axis]
+                    axis += 1
+                x = x.reshape((p,)+(shape)+(n,), order='F').sum(axis=axis)
```
PTNobel, line 791: Is the parentheses around shape intentional? Looks like you're trying to make a tuple but I don't think you are

cvxpy/lin_ops/canon_backend.py
```diff
@@ -1175,13 +1203,35 @@ def parametrized_mul(x):
...
+                if axis is None:
+                    return sp.csr_matrix(x.sum(axis=0))
+                else:
+                    shape = _lin.args[0].shape
+                    n = x.shape[-1]
+                    if isinstance(axis, tuple):
+                        d = np.prod([shape[i] for i in axis], dtype=int)
+                    else:
+                        d = shape[axis]
+                    x = x.toarray().reshape((shape)+(n,), order='F').sum(axis=axis)
```
PTNobel, line 1233: The toarray seems concerningly expensive. Am I missing something?

cvxpy/atoms/affine/sum.py
PTNobel: Never mind. I read down to lin_utils. I get the issue now.

cvxpy/problems/problem.py
```diff
@@ -257,6 +257,16 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
           expr.is_dcp(dpp) for expr in self.constraints + [self.objective])
 
+    @perf.compute_once
+    def _max_ndim(self) -> int:
+        """
+        Returns the maximum number of dimensions of any argument in the problem.
+        """
+        prob_max_ndim = self.objective.expr._max_ndim()
+        for con in self.constraints:
+            prob_max_ndim = max([prob_max_ndim] + [arg._max_ndim() for arg in con.args])
```
PTNobel, line 267: I'm not a huge fan of the list addition. Could we write it as `max(prob_max_ndim, max(arg. ...))`?

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -377,6 +385,42 @@ def construct_solving_chain(problem, candidates,
                           ", ".join([cone.__name__ for cone in cones])))
 
 
+def _get_canon_backend(problem, canon_backend):
+    """
+    This function checks if the problem has expressions of dimension greater
+    than 2, then raises a warning if the default backend is not specified or 
+    raises an error if the backend is specified as 'CPP'.
+
+    TODO: The correct way to do this in the future is to check all nodes of
```
PTNobel, line 394: How is this difference than what you're doing?

PTNobel (changes requested):

cvxpy/reductions/solvers/solving_chain.py
Transurgeon: Oh yea, nice catch. Steven helped me fix this TODO. I'll remove it.

cvxpy/lin_ops/canon_backend.py
Transurgeon: nice catch! removed the parentheses in the newest commit.

cvxpy/problems/problem.py
Transurgeon: yea, I agree it's probably better than storing the values in a list. Changed

cvxpy/lin_ops/canon_backend.py
Transurgeon: I just added a TODO to keep the tensor sparse. The main issue is that there are no >2d sparse representations in SciPy.

cvxpy/atoms/affine/sum.py
```diff
@@ -67,47 +67,52 @@ def numeric(self, values):
     def graph_implementation(
         self, arg_objs, shape: Tuple[int, ...], data=None
```
PTNobel, line 68: Could we pass in an optional argument that says what canonicalization backend is being used and then if we're using SciPy use the new code regardless of axis and keepdim?

PTNobel (approved):
I have one issue left! Otherwise, this looks great and I'm promerging.

cvxpy/atoms/affine/sum.py
Transurgeon: I think that's a great idea, but we would need to pass down the canon backend parameter which I am not sure it is easy to do. I am fine with keeping this as is.

SteveDiamond: Will and I discussed differentiating more based on the backend during the canonicalization but we chose not to go down that route in this PR.

PTNobel: Then I'm good to merge!

PR #2479, opened 2024-06-17
Changing solver in tests from ECOS to CLARABEL

Transurgeon:
## Description
Now that we are transitioning from ECOS to CLARABEL, I think it makes sense to change all our tests to use cp.CLARABEL by default. 

P.S. Some tests are failing (mostly in ``test_DQCP``), I believe this could be due to precision or some other missing parameter? 
I am not too familiar with Clarabel, @PTNobel could you have a look please?
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -1227,13 +1227,13 @@ class TestAtoms(BaseTest):
         # Solve the (simple) two-stage problem by "combining" the two stages
         # (i.e., by solving a single linear program)
         p1 = Problem(Minimize(x+y), [x+y >= 3, y >= 4, x >= 5])
-        p1.solve(solver=cp.ECOS)
+        p1.solve(solver=cp.CLARABEL)
 
         # Solve the two-stage problem via partial_optimize
         p2 = Problem(Minimize(y), [x+y >= 3, y >= 4])
         g = partial_optimize(p2, [y], [x], solver='ECOS')
         p3 = Problem(Minimize(x+g), [x >= 5])
-        p3.solve(solver=cp.ECOS)
+        p3.solve(solver=cp.CLARABEL)
         self.assertAlmostEqual(p1.value, p3.value)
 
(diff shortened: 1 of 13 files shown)


PTNobel:
Happy to, but probably not for a few days.

PTNobel:
The DQCP bug is honestly quite concerning. We have a regression from swapping from ECOS to CLARABEL. I'll try and minimize it and explain what's wrong to send to Paul later tonight.

PTNobel:
Just opened oxfordcontrol/Clarabel.rs#120 with an example of a problem that is broken by the differences. It seems that Clarabel is much worse about detecting primal infeasible problems and that has regressed these methods significantly.

PTNobel:
If we want to, we could keep ECOS for DQCP and then merge this PR for now, but I think we should definitely have a longer term plan around fixing DQCP.

SteveDiamond:
Clarabel has improved quite quickly in the past so I don't think we necessarily need to do anything besides coordinate with Paul about this. It's normal for cvxpy to expose different issues with solvers since there's so much coverage.

Transurgeon:
@PTNobel I propose to update this PR to remove all the failing tests with DQCP.
Then we can open another PR specifically for the remaining tests as well as drop ECOS support, once the clarabel issue is resolved.

PTNobel:
Sounds like a plan. Let's do it @Transurgeon

Transurgeon:
@PTNobel a few tests are still failing because of precision errors (see below).
I tried looking in the clarabel [docs](https://clarabel.org/stable/api_settings/#Clarabel.Settings) but I'm still not sure what solver setting to pass to fix this issue.
image

Transurgeon:
@PTNobel I fixed this PR by removing the precision error tests mentioned above. 
I think the failing benchmark (one of them has a "significant" change) is really weird and definitely unrelated to the changes in this PR. 
Let's move forward with this PR if you are ok with it.

PTNobel (approved):

PR #2484, opened 2024-06-19
Deal with NumPy 2.0 compatibility breaker in gurobi interface

rluce:
## Description
The Gurobi QP interface has an unnecessary `np.array(..., copy=False)` invocation that doesn't play well with NumPy 2.0

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
@@ -181,7 +181,7 @@ class GUROBI(QpSolver):
                 x_grb[idx].start = data['init_value'][idx]
         model.update()
 
-        x = np.array(model.getVars(), copy=False)
+        x = np.array(model.getVars())
 
         if A.shape[0] > 0:
             if hasattr(model, 'addMConstr'):
--- a/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
@@ -181,7 +181,7 @@ class GUROBI(QpSolver):
                 x_grb[idx].start = data['init_value'][idx]
         model.update()
 
-        x = np.array(model.getVars())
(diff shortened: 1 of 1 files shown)


cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -181,7 +181,7 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
                 x_grb[idx].start = data['init_value'][idx]
         model.update()
 
-        x = np.array(model.getVars(), copy=False)
+        x = np.array(model.getVars())
```
phschiele, line 184: The [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword) suggests:
```suggestion
        x = np.asarray(model.getVars())
```

rluce: Yes that's right, but Model.getVars is guaranteed to return a list, so these two are equivalent.  I made the decision based on
```
cvxpy/cvxpy (gurobi_numpy2) % git grep 'np.array' | wc -l
     707
cvxpy/cvxpy (gurobi_numpy2) % git grep 'np.asarray' | wc -l
      10
```
Totally fine to change -- your call!

PTNobel (approved):
@phschiele feel free to merge this is you accept the explanation in the comments.

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
phschiele: @rluce my understanding is that `asarray` only makes a copy when necessary (equivalent to the `copy=False` behavior). To be cautions and assuming the `copy=False` was put here on purpose, I'm changing it before merging.

PTNobel:
Waiting for CI to succeed before merging.

rluce:
Thanks for handling!

PR #2488, opened 2024-06-27
updating shapes.util to accomodate for n-dimensional broadcasting

Transurgeon:
## Description
This PR aims to greatly simplify the current sum_shapes utils by taking advantage of the ``np.broadcast_shapes`` function. 

This also generalizes the mul_shapes function to be able to adhere to n-dimensional broadcasting rules. Previously it required that all outer dimensions (i.e. any dimension other than the last-two) be the same for both operands. 
However, NumPy allows its typical broadcasting rules on the outer dimensions. Ex: (2,2,2) @ (2,2) is allowed because the RHS is broadcast to (1,2,2). 

TODO: add some tests using hypothesis and maybe improve some of the docstrings.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


Transurgeon:
Ok, I noticed a slight issue with the changes in this PR. 
``np.broadcast_shapes`` is a relatively new API function (see image below) and hence fails the builds for Python 3.9 and 3.8. 
image

What should we do about this?

PTNobel:
Could we check the NumPy version and then have a fallback to the old code? Maybe just error if someone is trying to use ndarrays with an old NumPy version?

Alternatively, Python 3.8 is EOL in October, so we could just plan to not support it. This PR won't be backported to 1.4.x or 1.5.x. Python 3.9 is EOL in October '25 (so a while from now :( ), but the oldest supported NumPy is 1.19.3, so it doesn't feel crazy to me to just bump our minimum NumPy version to 1.20. and tell people to stay on 1.5.x when 1.6.x comes out.

cvxpy/tests/test_shape.py
```diff
@@ -14,96 +14,50 @@
...
 from cvxpy.atoms.affine.reshape import reshape
 from cvxpy.expressions.variable import Variable
 from cvxpy.utilities import shape
 
 
-class TestShape(unittest.TestCase):
-    """ Unit tests for the expressions/shape module. """
+class TestShape():
+    
+    @given(s=mutually_broadcastable_shapes(num_shapes=7))
```
SteveDiamond, line 29: what are mutually_broadcastable_shapes? could you add a comment?

Transurgeon: see the hypothesis documentation [here](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes). This test basically generates 7 random shapes that are mutually broadcastable. The test ensures that the ``sum_shapes`` function returns the same thing as NumPy.

Transurgeon: For matrix multiplication, it passes a signature to have two shapes that follow the matrix multiplication rules of numpy (i.e. ``# np.matmul.signature == "(m?,n),(n,p?)->(m?,p?)"``

SteveDiamond (approved):

PR #2500, opened 2024-07-09
generalized indexing for n-dimensional expressions

Transurgeon:
## Description
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -432,14 +432,16 @@ class PythonCanonBackend(CanonBackend):
         """
         Given (A, b) in view, select the rows corresponding to the elements of the expression being
         indexed.
-        """
+        """        
         indices = [np.arange(s.start, s.stop, s.step) for s in lin.data]
-        if len(indices) == 1:
-            rows = indices[0]
-        elif len(indices) == 2:
-            rows = np.add.outer(indices[0], indices[1] * lin.args[0].shape[0]).flatten(order="F")
-        else:
-            raise ValueError
+        rows = indices[0]
+        cum_prod = np.cumprod([lin.args[0].shape])
+        # For each additional dimension, calculate the offset and add it to the rows
+        for i in range(1, len(indices)): 
+            size_product = cum_prod[i - 1]
+            offset = np.add.outer(rows, indices[i] * size_product).flatten(order="F")
+            rows = offset
+
         view.select_rows(rows)
         return view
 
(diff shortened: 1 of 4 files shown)


SteveDiamond (approved):
Nice! This looks good to me. Boolean/array indexing already is supported, so no need to worry about that. Edit: Thought maybe there will be issues with boolean/array indexing into n-dimensional objects. I doubt it, but we should add tests at least.

Transurgeon:
I am not sure why the PR benchmarks are failing, it is a bit weird.
Could be because of the nd-sum sparse to dense conversion?

Transurgeon:
> Nice! This looks good to me. Boolean/array indexing already is supported, so no need to worry about that. Edit: Thought maybe there will be issues with boolean/array indexing into n-dimensional objects. I doubt it, but we should add tests at least.

Just added some tests for it in the newest commit. I'll spend some time exploring the benchmark failures today. It could be related to the updated indexing implementation, but I kind of doubt it.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -431,16 +431,19 @@ def div(self, lin: LinOp, view: TensorView) -> TensorView:
...
         """
         Given (A, b) in view, select the rows corresponding to the elements of the expression being
-        indexed. Supports arbitrary number of dimensions.
-        """    
+        indexed. Supports an arbitrary number of dimensions.
+        """
         indices = [np.arange(s.start, s.stop, s.step) for s in lin.data]
         rows = indices[0]
-        cum_prod = np.cumprod([lin.args[0].shape])
+        if len(indices) == 2:
```
SteveDiamond, line 438: what if len(indices) < 2?

Transurgeon: it's going to set rows = indices[0]. So it's basically the case for ``len(indices)==1`` from before.

Transurgeon: I am doing this because in the case for ``len(indices) > 2`` we need to instantiate rows to ``indices[0]`` and I didn't want to duplicate code.

cvxpy/lin_ops/canon_backend.py
```diff
@@ -434,17 +434,17 @@ def index(lin: LinOp, view: TensorView) -> TensorView:
         indexed. Supports an arbitrary number of dimensions.
         """
         indices = [np.arange(s.start, s.stop, s.step) for s in lin.data]
-        rows = indices[0]
-        if len(indices) == 2:
-            rows = np.add.outer(indices[0], indices[1] * lin.args[0].shape[0]).flatten(order="F")
-        # For each new dimension, calculate the cumulative offset and add it to the rows
-        elif len(indices) > 2:
```
SteveDiamond: add `assert len(indices) > 0`

SteveDiamond (approved):

PR #2501, opened 2024-07-12
sparse implementation for nd sums

Transurgeon:
## Description
This PR addresses a TODO in the nd sum PR. It implements a function ``get_row_sum_idx`` which computes the summands of the sum operation (similar to NumPy but symbolic). The column indices are obtained by flattening a range in fortran order. 

The idea of ``get_sum_row_idx`` is to group the entries that would correspond to a sum with the corresponding axis input. First we compute the out_dims by deleting the dims corresponding to the axis argument. Then we compute the indices corresponding to these out_dims. There will be duplicate values corresponding to the size of axis, those are the values that will be summed together. Hence the goal is to assign the same row values to all the duplicate entries in ``indices[out_dims]``. NumPy provides a function ``ravel_multi_index`` that does exactly that. 

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -1198,8 +1198,7 @@ class SciPyCanonBackend(PythonCanonBackend):
             func = parametrized_mul
         return view.accumulate_over_variables(func, is_param_free_function=is_param_free_lhs)
 
-    @staticmethod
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
Nice job! It is more clear with these comments.

PR #2509, opened 2024-07-17
Avoid integer overflow error for ``nconstr`` in ``canonInterface.py``

Transurgeon:
## Description
This PR fixes a bug that caused some benchmarks to fail. 
See the logs from asv here:
```
 canonInterface.reduce_problem_data_tensor(
File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/cvxcore/python/canonInterface.py", line 130, in reduce_problem_data_tensor
 n_constr = A.shape[0] // (n_cols)
            ~~~~~~~~~~~^^~~~~~~~~~
OverflowError: Python integer 2738598423566 out of bounds for int32
asv: benchmark failed (exit status 1)
 ```
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -127,7 +127,7 @@ def reduce_problem_data_tensor(A, var_length, quad_form: bool = False):
     # add one more column for the offset if not quad_form
     if not quad_form:
         n_cols += 1
-    n_constr = np.divide(A.shape[0], n_cols, dtype=np.int64)
+    n_constr = np.divide(A.shape[0], int(n_cols), dtype=np.int64)
     shape = (n_constr, n_cols)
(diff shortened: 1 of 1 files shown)


Transurgeon:
All benchmarks are passing once again!
image
CI failure is due to the previous commit having failed, which is to be expected.

SteveDiamond:
I feel like this should use np.divmod or something like that. It doesn't seem like quite the right fix.

Transurgeon:
> I feel like this should use np.divmod or something like that. It doesn't seem like quite the right fix.

yea agreed, it's probably better. Just did the fix.

SteveDiamond (approved):

Issue #2519, opened 2024-07-30
Type mismatch in `cvxpy/cvxcore/src/ProblemData.cpp`


bnaras:
**C++ type mismatch in `ProblemData.hpp`**
It is odd that `int` values are being stuffed into a `double` array in the functions [`ProblemData.getI(double *values, int num_values)`](https://github.com/cvxpy/cvxpy/blob/19bd785ae430d5e81aaf1cea3b52812edc21cc56/cvxpy/cvxcore/src/ProblemData.hpp#L90) and [`ProblemData.getJ(double *values, int num_values)`](https://github.com/cvxpy/cvxpy/blob/19bd785ae430d5e81aaf1cea3b52812edc21cc56/cvxpy/cvxcore/src/ProblemData.hpp#L100). 

**Expected behavior**
I'd expect that the first argument should be `int *values` rather than `double* values` since the underlying tensor types are [`int`s](https://github.com/cvxpy/cvxpy/blob/19bd785ae430d5e81aaf1cea3b52812edc21cc56/cvxpy/cvxcore/src/ProblemData.hpp#L28).

Am I missing something?

SteveDiamond:
I agree with you. This looks like a mistake in the code.

bnaras:
I'll make a PR.

PR #2521, opened 2024-07-31
Fix type mismatch in `cvxcore` file `src/ProblemData.hpp`

bnaras:
## Description
Fix type mismatch in `cvxcore` file `ProblemData.hpp` mentioned in issue below.
Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/2519

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/cvxcore/python/cvxcore.i
+++ b/cvxpy/cvxcore/python/cvxcore.i
@@ -42,6 +42,7 @@
 /* Typemap for the getV, getI, getJ, and getConstVec C++ routines in
 	 problemData.hpp */
 %apply (double* ARGOUT_ARRAY1, int DIM1) {(double* values, int num_values)}
+%apply (int* ARGOUT_ARRAY1, int DIM1) {(int* values, int num_values)}
 %include "ProblemData.hpp"
 
 /* Useful wrappers for the LinOp class */
--- a/cvxpy/cvxcore/python/cvxcore.i
+++ b/cvxpy/cvxcore/python/cvxcore.i
@@ -42,7 +42,7 @@
 /* Typemap for the getV, getI, getJ, and getConstVec C++ routines in
 	 problemData.hpp */
 %apply (double* ARGOUT_ARRAY1, int DIM1) {(double* values, int num_values)}
-%apply (int* ARGOUT_ARRAY1, int DIM1) {(int* values, int num_values)}
+%apply (unsigned int* ARGOUT_ARRAY1, int DIM1) {(unsigned int* values, int num_values)}
 %include "ProblemData.hpp"
 
 /* Useful wrappers for the LinOp class */
--- a/cvxpy/cvxcore/python/cvxcore_wrap.cxx
+++ b/cvxpy/cvxcore/python/cvxcore_wrap.cxx
@@ -3168,34 +3168,34 @@ SWIG_Python_NonDynamicSetAttr(PyObject *obj, PyObject *name, PyObject *value) {
 #define SWIGTYPE_p_char swig_types[7]
 #define SWIGTYPE_p_difference_type swig_types[8]
 #define SWIGTYPE_p_double swig_types[9]
-#define SWIGTYPE_p_int swig_types[10]
-#define SWIGTYPE_p_key_type swig_types[11]
-#define SWIGTYPE_p_mapped_type swig_types[12]
-#define SWIGTYPE_p_operatortype swig_types[13]
-#define SWIGTYPE_p_p_PyObject swig_types[14]
-#define SWIGTYPE_p_size_type swig_types[15]
-#define SWIGTYPE_p_std__allocatorT_LinOp_const_p_t swig_types[16]
-#define SWIGTYPE_p_std__allocatorT_LinOp_p_t swig_types[17]
-#define SWIGTYPE_p_std__allocatorT_double_t swig_types[18]
-#define SWIGTYPE_p_std__allocatorT_int_t swig_types[19]
-#define SWIGTYPE_p_std__allocatorT_std__pairT_int_const_int_t_t swig_types[20]
-#define SWIGTYPE_p_std__allocatorT_std__vectorT_double_t_t swig_types[21]
-#define SWIGTYPE_p_std__allocatorT_std__vectorT_int_t_t swig_types[22]
-#define SWIGTYPE_p_std__invalid_argument swig_types[23]
-#define SWIGTYPE_p_std__lessT_int_t swig_types[24]
-#define SWIGTYPE_p_std__mapT_int_Eigen__SparseMatrixT_double_t_std__lessT_int_t_std__allocatorT_std__pairT_int_const_Eigen__SparseMatrixT_double_t_t_t_t swig_types[25]
-#define SWIGTYPE_p_std__mapT_int_int_t swig_types[26]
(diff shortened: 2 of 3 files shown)


bnaras:
@SteveDiamond I made the changes to two files, `cvxcore/src/ProblemData.hpp` and also `cvxcore/python/cvxcore.i` and regenerated `cvxcore_wrap.cxx`. No idea why pr_benchmarks are failing? Is there a time-limit that is being hit? 
Incidentally, the `cvxcore/README.md` suggests adding `#undef_macros = [ "NDEBUG" ]` for the `cvxcore` extension. However, if one does this, even the current master fails tests.

SteveDiamond:
@Transurgeon any idea what's going on with the benchmarks?

rileyjmurray:
The data type probably needs to be int64, not just int. You can access the type as std::int64_t in the cstdint header.

Transurgeon:
> @Transurgeon any idea what's going on with the benchmarks?
> 
> @bnaras I would not put much stock in the cvxcore/REAMDE.md, that must be ancient. Definitely don't compile it in debug mode.

I just tried running the benchmarks on this PR and encounter the following error:
```
[ 52.00%] ··· finance.CVaRBenchmark.time_compile_problem                  failed
[ 52.00%] ···· Traceback (most recent call last):
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 1435, in <module>
                   main()
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 1428, in main
                   commands[mode](args)
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 1167, in main_run
                   result = benchmark.do_run()
                            ^^^^^^^^^^^^^^^^^^
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 573, in do_run
                   return self.run(*self._current_params)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 669, in run
                   samples, number = self.benchmark_timing(timer, min_repeat, max_repeat,
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/site-packages/asv/benchmark.py", line 705, in benchmark_timing
                   timing = timer.timeit(number)
                            ^^^^^^^^^^^^^^^^^^^^
                 File "/opt/anaconda3/envs/cvxpy_bench/lib/python3.11/timeit.py", line 180, in timeit
                   timing = self.inner(it, self.timer)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "<timeit-src>", line 6, in inner
                 File "/Users/willizz/Documents/benchmarks/benchmark/finance.py", line 82, in time_compile_problem
                   self.problem.get_problem_data(solver=cp.SCS)
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/problems/problem.py", line 766, in get_problem_data
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/reductions/chain.py", line 76, in apply
                                  ^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py", line 379, in apply
                   params_to_problem_data = extractor.affine(expr_list)
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/utilities/coeff_extractor.py", line 72, in affine
                   return canonInterface.get_problem_matrix(op_list,
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/cvxcore/python/canonInterface.py", line 297, in get_problem_matrix
                   return build_matrix(id_to_col, param_to_size, param_to_col, var_length, constr_length, linOps)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/cvxpy/cvxcore/python/cppbackend.py", line 102, in build_matrix
                   A = sp.csc_matrix((V, (I, J)), shape=output_shape)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/scipy/sparse/_compressed.py", line 55, in __init__
                   coo = self._coo_container(arg1, shape=shape, dtype=dtype)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/scipy/sparse/_coo.py", line 99, in __init__
                   self._check()
                 File "/Users/willizz/Documents/benchmarks/env/d7b296eeadace6ef7218f773ed071d5d/lib/python3.11/site-packages/scipy/sparse/_coo.py", line 208, in _check
                   raise ValueError(f'negative axis {i} index: {idx.min()}')
               ValueError: negative axis 0 index: -2147474013
               asv: benchmark failed (exit status 1)
```

Transurgeon:
also from the CI it says the following commit is causing issues: https://github.com/cvxpy/cvxpy/commit/f131f5b7.
Which links to this PR. I am not sure how to go about fixing the issue though.
image

bnaras:
This seems to suggest that stuffing the `int`s into `double`s was deliberate: the `int` seems to have overflowed. As @rileyjmurray suggests, perhaps I should try `int64`.

bnaras:
@Transurgeon I was able to reproduce the errors you highlighted above in the screenshot locally. So I made a cast to `unsigned int` and that fixes them. (@rileyjmurray I don't see a way to use `int64_t` because [numpy/swig interface doesn't support it](https://numpy.org/doc/stable/user/basics.types.html)). But now the CI shows some failing actions due to other reasons which to me seems like a false positive.

rileyjmurray:
I think that using a 64 bit integer type is really important here. We’re
guaranteed to run into issues with DPP problems with a 32 bit integer, even
if we’re using 32 bit unsigned.

Transurgeon:
> @Transurgeon I was able to reproduce the errors you highlighted above in the screenshot locally. So I made a cast to `unsigned int` and that fixes them. (@rileyjmurray I don't see a way to use `int64_t` because [numpy/swig interface doesn't support it](https://numpy.org/doc/stable/user/basics.types.html)). But now the CI shows some failing actions due to other reasons which to me seems like a false positive.

The errors seem unrelated to your PR. (probably something to do with a solver update)
image

bnaras:
Couldn't agree more @rileyjmurray . As the pr_benchmark failure showed, someone probably foresaw this and therefore stuck in doubles as a quick fix. Also, the int64 fixes throughout will also have a performance hit since every index array will have need to have `np.int64` if either an index goes beyond 2^32 or a matrix has nnzs beyond index 2^32.  I'm going to close this PR since I don't have the bandwidth for that extensive a fix at the moment---I merely came across this as Anqi and I are updating CVXR. Perhaps a link to the closed PR can then be added in the `cvxpy/cvxcode/README.md` .

rileyjmurray:
A comment for posterity:

The first integer that will round incorrectly when represented as an IEEE
double precision number is 2^53 + 1. So packing the integer index into a
double does get us much farther than int32, and in it does so in a very
platform-agnostic way.

See
https://stackoverflow.com/questions/3793838/which-is-the-first-integer-that-an-ieee-754-float-is-incapable-of-representing-e
for
relevant discussion.

bnaras:
While on the topic, using doubles this way was discussed by R developers early on. See: https://cran.r-project.org/doc/manuals/r-release/R-ints.html#Long-vectors and slide 31 in the linked pdf.

PR #2524, opened 2024-08-02
fix issue in xpress interface for ``prob_.attribute``

Transurgeon:
## Description
Explanation in title. Noticed this in the most recent CI builds.
Related [dcoumentation](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/COLS.html) for the changed attribute.

ADDITIONAL CHANGE: deleting sonarcloud build because it is causing problems. 
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -224,26 +224,3 @@ jobs:
         with:
           name: wheels-base-${{ matrix.os }}-${{ matrix.python-version }}
           path: ./wheelhouse
-
-  sonarcloud:
-    needs: build
-    name: SonarCloud
-    runs-on: ubuntu-latest
-    env:
-      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-    steps:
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):

SteveDiamond:
We could also bump the minimum xpress in pyproject.toml

Transurgeon:
> We could also bump the minimum xpress in pyproject.toml

which version should we put? I am not sure when they introduced this new change (but it must have been relatively recent).

SteveDiamond:
> > We could also bump the minimum xpress in pyproject.toml
> 
> which version should we put? I am not sure when they introduced this new change (but it must have been relatively recent).

idk. let's just merge this fix, too much trouble to figure out.

PTNobel:
Should this be backported?

Transurgeon:
Yea I think so, why not. I will add it to the list.

PR #2525, opened 2024-08-02
adding warning for eventual change of default order

Transurgeon:
## Description
adding warning for not specifying order in reshape, vec and flatten. Also updated the cvxpy internal codebase to ensure that every operation above is specifying the order and wouldn't raise a warning for the user.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -864,10 +864,13 @@ class Expression(u.Canonical):
         from cvxpy import ptp
         return ptp(self, axis, keepdims)
 
-    def reshape(self, shape, order='F'):
+    def reshape(self, shape, order = None):
         """
         Equivalent to `cp.reshape(self, shape, order)`.
         """
+        if order is None:
+            warnings.warn(DEFAULT_ORDER_DEPRECATION_MSG, FutureWarning)
+            order = 'F'
         from cvxpy import reshape
         return reshape(self, shape, order)
 
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -17,7 +17,7 @@ limitations under the License.
 import abc
 import warnings
 from functools import wraps
-from typing import List, Optional, Tuple
+from typing import List, Literal, Tuple
 
 import numpy as np
 
@@ -488,8 +488,9 @@ class Expression(u.Canonical):
         """
         return len(self.shape)
 
-    def flatten(self, order: Optional[str] = None):
-        """Vectorizes the expression.
+    def flatten(self, order: Literal["F", "C", None] = None):
+        """
+        Vectorizes the expression.
 
         order: column-major ('F') or row-major ('C') order.
         """
@@ -865,7 +866,7 @@ class Expression(u.Canonical):
         from cvxpy import ptp
         return ptp(self, axis, keepdims)
 
-    def reshape(self, shape, order: Optional[str] = None):
+    def reshape(self, shape, order: Literal["F", "C", None] = None):
         """
         Equivalent to `cp.reshape(self, shape, order)`.
         """
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -488,7 +488,7 @@ class Expression(u.Canonical):
         """
         return len(self.shape)
 
-    def flatten(self, order = None):
+    def flatten(self, order: str | None = None):
         """Vectorizes the expression.
 
         order: column-major ('F') or row-major ('C') order.
@@ -864,7 +864,7 @@ class Expression(u.Canonical):
         from cvxpy import ptp
         return ptp(self, axis, keepdims)
 
-    def reshape(self, shape, order = None):
+    def reshape(self, shape, order: str | None = None):
         """
         Equivalent to `cp.reshape(self, shape, order)`.
         """
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -85,8 +85,8 @@ DEFAULT_ORDER_DEPRECATION_MSG = (
     """
     You didn't specify the order of the FUNC_NAME expression. The default order
     used in CVXPY is Fortran ('F') order. This default will change to match NumPy's
-    default order ('C') in a future version.
-    To suppress this warning, specify the order explicitly.
+    default order ('C') in a future version of CVXPY.
+    To suppress this warning, please specify the order explicitly.
     """
 )
 
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -81,6 +81,15 @@ You're calling the built-in abs function on a CVXPY expression. This is not
 supported. Consider using the abs function provided by CVXPY.
 """
 
+DEFAULT_ORDER_DEPRECATION_MSG = (
+    """
+    You didn't specify the order of the reshaped expression. The default order
+    used in CVXPY is Fortran ('F') order. This default will change to match NumPy's
+    default order ('C') in a future version.
+    To suppress this warning, specify the order explicitly.
+    """
+)
+
 __BINARY_EXPRESSION_UFUNCS__ = {
         np.add: lambda self, a: self.__radd__(a),
         np.subtract: lambda self, a: self.__rsub__(a),
@@ -442,7 +451,8 @@ class Expression(u.Canonical):
         """
         raise NotImplementedError()
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def shape(self) -> Tuple[int, ...]:
         """tuple : The expression dimensions.
         """
@@ -453,7 +463,8 @@ class Expression(u.Canonical):
         """
         return not self.is_complex()
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def is_imag(self) -> bool:
         """Is the Leaf imaginary?
         """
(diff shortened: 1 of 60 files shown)


cvxpy/tests/test_expression_methods.py
```diff
@@ -237,85 +229,14 @@ def test_reshape_negative_one(self) -> None:
...
-                        "axis 4 is out of bounds for array of dimension 1")
-
-        A = sp.eye(3)
-        self.assertEqual(Constant(A).sum().value, 3)
-
-        A = sp.eye(3)
-        self.assertItemsAlmostEqual(Constant(A).sum(axis=0).value, [1, 1, 1])
-    def test_trace(self) -> None:
-        """Test the trace atom.
-        """
```
Transurgeon, line 309: these tests seem to be duplicated from ``test_expressions``

cvxpy/tests/test_expression_methods.py
```diff
@@ -195,29 +188,28 @@ def test_reshape(self) -> None:
         ])
         b_reshaped = b.reshape((2, 6), order='F')
         X = cp.Variable(b.shape)
-        X_reshaped = X.reshape((2, 6))
+        with pytest.warns(FutureWarning):
+            X_reshaped = X.reshape((2, 6))
```
Transurgeon, line 192: new test to make sure there is a warning when calling reshape without specifying order.

cvxpy/expressions/expression.py
```diff
@@ -477,11 +488,14 @@ def ndim(self) -> int:
         """
         return len(self.shape)
 
-    def flatten(self, order: str = 'F'):
+    def flatten(self, order = None):
```
SteveDiamond, line 491: The type of order is now `order: str | None = None`

cvxpy/expressions/expression.py
```diff
@@ -81,6 +81,15 @@ def cast_op(self, other):
 supported. Consider using the abs function provided by CVXPY.
 """
 
+DEFAULT_ORDER_DEPRECATION_MSG = (
+    """
+    You didn't specify the order of the reshaped expression. The default order
```
SteveDiamond, line 86: Will this error be confusing for vec/flatten, since it says reshape? They are all sort of reshapes so maybe it's fine.

Transurgeon: Yea I agree, we can definitely improve the error message. But I think getting the ``__str__`` of the function name would be a bit too complicated. @PTNobel what do you think?

cvxpy/expressions/expression.py
Transurgeon: yea sounds good.

cvxpy/expressions/expression.py
PTNobel: We can just have a FUNC_NAME and then do a string replace in each of the places we use it with context.

PTNobel (changes requested):
I think we need to change the warning message depending on context. Otherwise, I'm a bit confused about the testing changes but it is otherwise good!

Transurgeon:
> I think we need to change the warning message depending on context. Otherwise, I'm a bit confused about the testing changes but it is otherwise good!

Thanks for the review parth! I just added the changes for FUNC_NAME as you suggested. Please let me know if it looks good now.

As for the test changes, it is a bit unrelated to the PR but I have always wanted to change from unittest to pytest, because the debugging/developer experience is much better. I can always make a separate PR for that if you think it's necessary.

SteveDiamond (approved):
LGTM!

cvxpy/tests/test_expression_methods.py
```diff
@@ -183,41 +176,37 @@ def test_reshape(self) -> None:
...
+            cp.vec(X)
+        with pytest.warns(FutureWarning):
+            cp.reshape(X, (2, 6))
         prob = cp.Problem(cp.Minimize(0), [X_reshaped == b_reshaped])
         prob.solve(solver=cp.SCS)
-        self.assertItemsAlmostEqual(b_reshaped, X_reshaped.value)
-        self.assertItemsAlmostEqual(b, X.value)
-
+        np.allclose(b_reshaped, X_reshaped.value)
+        np.allclose(b, X.value)
```
PTNobel, line 195: This no longer asserts anything. Is this an intended change?

cvxpy/tests/test_expression_methods.py
```diff
@@ -183,41 +176,37 @@ def test_reshape(self) -> None:
         X_reshaped = X.reshape((2, 6), order='C')
         prob = cp.Problem(cp.Minimize(0), [X_reshaped == b_reshaped])
         prob.solve(solver=cp.SCS)
-        self.assertItemsAlmostEqual(b_reshaped, X_reshaped.value)
-        self.assertItemsAlmostEqual(b, X.value)
+        np.allclose(b_reshaped, X_reshaped.value)
+        np.allclose(b, X.value)
```
PTNobel, line 180: DOesn't assert anything?

cvxpy/tests/test_expression_methods.py
```diff
@@ -129,47 +125,44 @@ def test_all_expressions(self) -> None:
...
         a_cp = A_cp.reshape(10, order='C')
-
-        self.assertItemsAlmostEqual(a_np, a_cp.value)
+        np.allclose(a_np, a_cp.value)
 
         x = cp.Variable(10)
         prob = cp.Problem(cp.Minimize(0), [x == a_cp])
         prob.solve(solver=cp.SCS)
-        self.assertItemsAlmostEqual(a_np, x.value)
+        np.allclose(a_np, x.value)
```
PTNobel, line 165: Similar issue?

cvxpy/tests/test_expression_methods.py
```diff
@@ -129,47 +125,44 @@ def test_all_expressions(self) -> None:
...
         A_np = np.reshape(a, (5, 2), order='C')
         A_cp = Constant(a).reshape((5, 2), order='C')
-        self.assertItemsAlmostEqual(A_np, A_cp.value)
+        np.allclose(A_np, A_cp.value)
 
         X = cp.Variable((5, 2))
         prob = cp.Problem(cp.Minimize(0), [X == A_cp])
         prob.solve(solver=cp.SCS)
-        self.assertItemsAlmostEqual(A_np, X.value)
+        np.allclose(A_np, X.value)
```
PTNobel, line 156: Behaviour change that I don't get

cvxpy/tests/test_expression_methods.py
```diff
@@ -129,47 +125,44 @@ def test_all_expressions(self) -> None:
...
-        self.assertEqual(str(cm.exception),
-                         "Invalid reshape dimensions (5, 4).")
+        assert str(cm.value) == "Invalid reshape dimensions (5, 4)."
 
         # Test C-style reshape.
         a = np.arange(10)
         A_np = np.reshape(a, (5, 2), order='C')
         A_cp = Constant(a).reshape((5, 2), order='C')
-        self.assertItemsAlmostEqual(A_np, A_cp.value)
+        np.allclose(A_np, A_cp.value)
```
PTNobel, line 151: Behavior change.

PTNobel (changes requested):
I think a number of errors were made in the test conversion. As a general matter I would prefer PRs at least mention when/why they are changing things beyond their title.

cvxpy/tests/test_expression_methods.py
Transurgeon: no, it's not. I tried to automate things and it caused some mishaps. Sorry about that and nice catch!

Transurgeon:
> I think a number of errors were made in the test conversion. As a general matter I would prefer PRs at least mention when/why they are changing things beyond their title.

Yes, agreed. I will try to separate PRs in the future. You are right that the test changes are not related to the title of the PR.

SteveDiamond:
We should add the order argument to the internal reshapes in cvxpy.

cvxpy/expressions/expression.py
```diff
@@ -850,10 +865,14 @@ def ptp(self, axis=None, *, keepdims=False):
         from cvxpy import ptp
         return ptp(self, axis, keepdims)
 
-    def reshape(self, shape, order='F'):
+    def reshape(self, shape, order: Optional[str] = None):
```
PTNobel, line 868: Can we switch to using the typing.Literal to specify what orders are allowed?

PTNobel:
William can you address Steven's suggestion about updating our usage of reshape?

Also, I have more questions about the test changes. Are we running pytest in the CI? Is the idea to change slowly over time? I'm still quite confused about what's happening there.

cvxpy/expressions/expression.py
Transurgeon: yea will do. Thanks for the suggestion.

Transurgeon:
> William can you address Steven's suggestion about updating our usage of reshape?

Yes. I just made a few commits to address this. I've scanned the whole codebase including tests for instances of reshape, vec and flatten where the order is not specified by default. 

> Also, I have more questions about the test changes. Are we running pytest in the CI? Is the idea to change slowly over time? I'm still quite confused about what's happening there.

I think we are indeed running pytest in the CI, however unittest is also being covered, so no worries on that end. In my opinion, it is just older and less pythonic than pytest (also I might just be used to the later). I decided to revert the changes for now and we could revisit this at a future time, if needed. 

P.S. Sorry about the absurd amount of file changes, hopefully it's not too bad to review.

PTNobel (approved):
Looks great! Do we want to add any docs for this? My gut is no, the warning is clear enough.

Transurgeon:
> Looks great! Do we want to add any docs for this? My gut is no, the warning is clear enough.

No I think it's fine. We should definitely remember to add a bullet point in the changelog though (once we release 1.6).

SteveDiamond (approved):
Very thorough!

PR #2527, opened 2024-08-06
Remove Python 2 compatibility code

jonathanberthias:
Hi there 👋 

I noticed there was still quite a bit of code that uses some constructs from Python 2 in the code base, even though Python 2 is not supported anymore. I used the 2to3 tool (while it still exists) to make the bulk of the changes, and then I adapted a little the result.

The most noticeable change comes from the Abstract Base Classes. The old syntax:
```python
class X:
    __metaclass__ = abc.ABCMeta
```
has no effect in Python 3, so all of these were just plain classes. Fixing that revealed a couple errors with some unimplemented methods, so I added some dummy implementations.

I wanted to add a test that the abstract-ness is working, something like:
```python
for expected_abc in (Canonical, Expression, Atom, Leaf, Reduction, Solver, Constraint):
    assert inspect.isabstract(expected_abc)
```
All of these fail before this PR. Do you have an idea where this test could fit? Looking at the `tests` directory, I couldn't figure out if there were similar tests somewhere or not.

Let me know your thoughts about this!

### DIFF
--- a/cvxpy/atoms/affine/affine_atom.py
+++ b/cvxpy/atoms/affine/affine_atom.py
@@ -27,7 +27,7 @@ from cvxpy.expressions.constants import Constant
 from cvxpy.utilities import performance_utils as perf
 
 
-class AffAtom(Atom, metaclass=abc.ABCMeta):
+class AffAtom(Atom):
     """ Abstract base class for affine atoms. """
     _allow_complex = True
 
--- a/cvxpy/atoms/affine/affine_atom.py
+++ b/cvxpy/atoms/affine/affine_atom.py
@@ -27,9 +27,8 @@ from cvxpy.expressions.constants import Constant
 from cvxpy.utilities import performance_utils as perf
 
 
-class AffAtom(Atom):
+class AffAtom(Atom, metaclass=abc.ABCMeta):
     """ Abstract base class for affine atoms. """
-    __metaclass__ = abc.ABCMeta
     _allow_complex = True
 
     def sign_from_args(self) -> Tuple[bool, bool]:
--- a/cvxpy/atoms/affine/affine_atom.py
+++ b/cvxpy/atoms/affine/affine_atom.py
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-import abc
 from typing import Any, List, Tuple
 
 import scipy.sparse as sp
--- a/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
+++ b/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
@@ -165,7 +165,8 @@ class ConicSolver(Solver):
         # Default is identity.
         return sp.eye(constr.size, format='csc')
 
-    def format_constraints(self, problem, exp_cone_order):
+    @classmethod
+    def format_constraints(cls, problem, exp_cone_order):
         """
         Returns a ParamConeProg whose problem data tensors will yield the
         coefficient "A" and offset "b" for the constraint in the following
@@ -249,7 +250,7 @@ class ConicSolver(Solver):
                     arg_mats.append(space_mat)
                 restruct_mat.append(sp.hstack(arg_mats))
             elif type(constr) == PSD:
-                restruct_mat.append(self.psd_format_mat(constr))
+                restruct_mat.append(cls.psd_format_mat(constr))
             else:
                 raise ValueError("Unsupported constraint type.")
 
--- a/cvxpy/tests/test_cone2cone.py
+++ b/cvxpy/tests/test_cone2cone.py
@@ -49,7 +49,7 @@ class TestDualize(BaseTest):
 
         # Dualize the problem, reconstruct a high-level cvxpy problem for the dual.
         # Solve the problem, invert the dualize reduction.
-        cone_prog = ConicSolver().format_constraints(cone_prog, exp_cone_order=[0, 1, 2])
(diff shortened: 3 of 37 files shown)


cvxpy/atoms/affine/affine_atom.py
```diff
@@ -29,7 +29,6 @@
 
 class AffAtom(Atom):
     """ Abstract base class for affine atoms. """
-    __metaclass__ = abc.ABCMeta
```
SteveDiamond, line 32: AffAtom here should be an abstract base class. There are a bunch of other similar cases below where __metaclass__ ... was deleted but the class was not made abstract.

SteveDiamond:
Maybe the test could go in test_problem? Or you could make a small new file test_base_classes.py

SteveDiamond:
Thanks for catching these issues!

cvxpy/atoms/affine/affine_atom.py
jonathanberthias: `AffAtom` inherits from `Canonical` which is an ABC, so `AffAtom`'s metaclass is also `abc.ABCMeta` which makes it unnecessary to repeat on all child classes. However, adding the metaclass explicitly does it make clearer what the intent is. Would you prefer I add `metaclass=abc.ABCMeta` on all ABCs, i.e. all classes which purposefully don't implement all abstract methods from their parent(s)?

cvxpy/transforms/indicator.py
```diff
@@ -94,6 +94,11 @@ def shape(self) -> Tuple[int, ...]:
         """
         return ()
 
+    def is_dpp(self, context: str = 'dcp') -> bool:
+        """The expression is a disciplined parameterized expression.
+        """
+        return False
```
jonathanberthias, line 100: Not too sure about this one, but since it wasn't implemented I guess nobody needs to use `indicator`s in DPP

cvxpy/utilities/canonical.py
```diff
@@ -161,6 +159,10 @@ def _max_ndim(self) -> int:
         """
         return max([self.ndim] + [arg._max_ndim() for arg in self.args])
 
+    @abc.abstractmethod
+    def __str__(self) -> str:
+        raise NotImplementedError()
```
jonathanberthias, line 164: This was added just so that `Canonical` is also an abstract class. Beside this one, all other methods have implementations, and a class is considered an ABC only if its metaclass is `ABCMeta` __and__ it has at least one `abstractmethod`.

cvxpy/atoms/affine/affine_atom.py
SteveDiamond: Ah ok. No need to repeat for the child classes.

cvxpy/reductions/solvers/conic_solvers/conic_solver.py
```diff
@@ -249,7 +250,7 @@ def format_constraints(self, problem, exp_cone_order):
                     arg_mats.append(space_mat)
                 restruct_mat.append(sp.hstack(arg_mats))
             elif type(constr) == PSD:
-                restruct_mat.append(self.psd_format_mat(constr))
+                restruct_mat.append(cls.psd_format_mat(constr))
```
Transurgeon, line 253: can you explain what this change does and why it's necessary?

cvxpy/tests/test_cone2cone.py
```diff
@@ -49,7 +49,7 @@ def simulate_chain(in_prob):
 
         # Dualize the problem, reconstruct a high-level cvxpy problem for the dual.
         # Solve the problem, invert the dualize reduction.
-        cone_prog = ConicSolver().format_constraints(cone_prog, exp_cone_order=[0, 1, 2])
+        cone_prog = ConicSolver.format_constraints(cone_prog, exp_cone_order=[0, 1, 2])
```
Transurgeon, line 52: nice catch here! But I'm wondering if this will behave the same as the old code? They seem to be different (one looks like a callable while the other a class), but your implementation seems to be correct.

Transurgeon (commented):
Thanks for your contribution @jonathanberthias ! Had a few questions but otherwise looks quite good. This is a great step towards cleaning up the codebase.

cvxpy/reductions/solvers/conic_solvers/conic_solver.py
jonathanberthias: `psd_format_mat` and `format_constraints` are both helper methods which don't depend on a specific solver instance, so I changed `format_constraints` to a class method (`psd_format_mat` is already a static method). It was necessary so that `format_constraints` can be called directly on the class without needing an instance. This is linked to your other question :arrow_down:

cvxpy/tests/test_cone2cone.py
jonathanberthias: The `ConicSolver` class is now truly abstract so we can't instanciate it anymore. Thankfully, the `format_constraints` function does not need a solver instance, so here we are calling it directly on the class.

cvxpy/transforms/indicator.py
SteveDiamond: It should be
```
    def is_dpp(self, context='dcp') -> bool:
        if context.lower() == 'dcp':
            return self.is_dcp(dpp=True)
        elif context.lower() == 'dgp':
            return self.is_dgp(dpp=True)
        else:
            raise ValueError('Unsupported context ', context)
```
but we also need to add a test for indicator.is_dpp()

jonathanberthias: Perhaps this should be done another PR no?

PTNobel (approved):
Looks great! Thank you. Will merge in the morning barring anyone raising more questions.

Issue #566, opened 2018-08-21
allow Variables, Constants, Parameters to have multiple special attributes.


rileyjmurray:
It seems desirable to allow someone to (for example) declare that a square matrix is both nonnegative, and PSD.

Was support for this feature deliberately left out, or was it simply a matter of not having the time to implement it?

SteveDiamond:
The projection becomes much more complicated. It's intentionally left out for that reason. Is there really a good case for it?

rileyjmurray:
You are very right that projection would become much more complicated. However, the way that we use projection right now is only to validate input data (so it's not a user-facing function). To validate input data for a Parameter with multiple attributes, we would just project once for each attribute type (i.e. project onto R^(n * n)_+, the PSD cone, cone of diagonal matrices, etc...) and accept the input iff each of these projection operations had sufficiently small residual.

Making the necessary changes may be more trouble than it's worth, but here are some benefits I see (provided that a user wants multiple attributes in their mathematical model):

1. Improved code readability for complicated optimization models.
2. Improved compile-times.
   - cvxpy can probably bypass much of the overhead in parsing a constraint if that constraint is known to be one of a handful of attributes available to a Variable.
   - If nothing else, it would ensure that constraints are vectorized. (I.e. X == X.T instead of a bunch of X[i,j] == X[j,i]).
3. Smaller and more tractable optimization problems passed to the solver.
    - For example, if you want to declare a symmetric matrix with integer entries, then the only way to do that is ``X = cvxpy.Variable((n,n), integer=True)`` and ``X == X.T``. 
    - This syntax creates almost twice as many integer variables as are necessary for the problem, and adds n^2 equality constraints that are otherwise unnecessary.

r-zip:
I have a use case for this: a generalized geometric programming problem where the objective and constraints involve affine combinations of diagonals of PSD matrices with nonnegative variables.

Because I can't specify a PSD constant, I'm forced to precompute these diagonals so that the DGP rules are satisfied. I know a priori that the diagonals will be nonnegative, but I have no way to prove that to CVXPY without doing it myself. The code to do this is much more verbose and error-prone than it would be otherwise.

E.g., have to write (simplifying my actual problem and ignoring efficiency for the sake of clarity)
```python
import cvxpy as cp
import numpy as np

x = cp.Variable(S.shape[0], pos=True)
[...]

PR #2528, opened 2024-08-07
Allow expressions with multiple attributes

Transurgeon:
## Description
This PR attempts to tackle a long-standing TODO for CVXPY, which is enabling users to specify more than one special attribute for variables and parameters. 
Note: I don't think the PR is ready by any means yet, but I would like some feedback and potentially start a discussion on what tests can be added (which combinations of attributes, what sorts of feasibility problems, etc.).

I have imported the ``cvxattr2constr`` reduction to ensure that the attributes are in fact able to create all the expected new constraints. It seems like many things are working by default (for example, bounds and integer/boolean and also PSD with nonneg). 
Issue link (if applicable): #566 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -142,15 +142,6 @@ class Leaf(expression.Expression):
         else:
             self.integer_idx = []
 
-        # Only one attribute be True (except can be boolean and integer).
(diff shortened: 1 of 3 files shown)


SteveDiamond:
We decided to pause work on multiple attributes until ND-array support is finished.

Issue #2038, opened 2023-02-13
Unclear how to use sparsity pattern on variables


1ucasvb:
In the [Advanced Features](https://www.cvxpy.org/tutorial/advanced/index.html#attributes) listing of attributes for Leaf objects, the following is given:
* "sparsity (list of tuplewith) – Fixed sparsity pattern for the variable." (Note the "tuplewith" typo as well)

I assume this is a list of tuples [(i1,j1), (i2,j2), ...] specifying which values are non-zero. However, there's no further elaboration on how this actually works in practice, and it does not seem to really have any effect if used.

SteveDiamond:
sparsity is not yet implemented. It works through CVXPYgen https://github.com/cvxgrp/cvxpygen but not in the main cvxpy.

Transurgeon:
this issue can be closed. There is a new sparsity attribute in cvxpy 1.6 and some documentation on how to use it.

Issue #1159, opened 2020-10-28
Sparse parameters break matrix stuffing


akshayka:
**Describe the bug**
When a parameter is a sparse matrix, the application of the ASA tensor fails.

**To Reproduce**
```python3
import cvxpy as cp                                                              
import numpy as np                                                              
import scipy.sparse                                                             
                                                                                
m, n = 10, 10                                                                   
A = cp.Parameter((m, n))                                                        
b = np.random.randn(m)                                                          
x = cp.Variable(m)                                                              
                                                                                
A.value = scipy.sparse.csc_matrix((m, n))                                       
A.value[0, 0] = 1.                                                              
problem = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b)), [])                
problem.solve()  
```

**Output**
```
  File "test.py", line 13, in <module>
    problem.solve()
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/problems/problem.py", line 396, in solve
    return solve_func(self, *args, **kwargs)
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/problems/problem.py", line 750, in _solve
    data, solving_chain, inverse_data = self.get_problem_data(
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/problems/problem.py", line 525, in get_problem_data
    data, inverse_data = solving_chain.apply(self)
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/reductions/chain.py", line 71, in apply
    problem, inv = r.apply(problem)
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/reductions/solvers/qp_solvers/qp_solver.py", line 74, in apply
    P, q, d, AF, bg = problem.apply_parameters()
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/reductions/qp2quad_form/qp_matrix_stuffing.py", line 173, in apply_parameters
    param_vec = canonInterface.get_parameter_vector(
  File "/home/akshay/src/cvxgrp/cvxpy/cvxpy/cvxcore/python/canonInterface.py", line 60, in get_parameter_vector
    param_vec[col:col + size] = value
```

**Version**
 - CVXPY Version: 1.1.7

**Additional context**
[...]

PR #2532, opened 2024-08-12
Sparsity attribute

Transurgeon:
## Description
This PR attempts to introduce a new attribute for cvxpy expressions: sparsity.
The input must be in a form similar to [np.indices](https://numpy.org/doc/stable/reference/generated/numpy.indices.html) but instead of numpy arrays we only support a list of tuples. 
Another option is to pass a tuple of ndarrays analogous to ``np.where(condition)`` or ``np.nonzero()``.
I added a few input validations in a function called ``_validate_sparsity``.
Issue link (if applicable):
#1159 , #2038

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/tests/test_attributes.py
```diff
@@ -0,0 +1,49 @@
...
+        X = cp.Variable((3, 3), sparsity=[(0, 1), (0, 2)])
+        prob = cp.Problem(cp.Minimize(cp.sum(X)), [X >= -1, X <= 1])
+        prob.solve()
+        z = np.zeros((3, 3))
+        z[X.sparse_idx] = -1
+        assert np.allclose(X.value, z)
+    
+    def test_sparsity_invalid_input(self):
+        with pytest.raises(ValueError, match="Sparsity should have 2 dimensions."):
+            cp.Variable((3, 3), sparsity=[(0, 1), (0, 1), (0, 1)])
```
SteveDiamond, line 19: I thought we were taking the input as ndarrays?

Transurgeon: yea sorry about that. Let me try to fix it now.

Transurgeon: actually I just tried with list of ndarrays as input and it is also supported. I've added a new test for that in the newest commit.

cvxpy/expressions/leaf.py
```diff
@@ -127,37 +127,61 @@ def __init__(
...
+        
+        if isinstance(indices, list):
+            indices = np.array(indices)
+
+        if indices.shape[0] != len(self._shape):
+            raise ValueError(f"Sparsity should have {len(self._shape)} dimensions.")
+        
+        # Check if all indices are within bounds
+        for dim in range(len(self._shape)):
+            if np.any(indices[:, dim] < 0) or np.any(indices[:, dim] >= self._shape[dim]):
```
PTNobel, line 180: I think it would be better to do `np.array(self._shape)` and then just eliminate the loop

cvxpy/expressions/leaf.py
```diff
@@ -127,37 +127,61 @@ def __init__(
...
+                raise ValueError(
+                    f"Sparsity is out of bounds for expression with shape {self._shape}.")
+        self.sparse_idx = tuple(indices)
```
PTNobel, line 183: Why not keep it an array?

cvxpy/expressions/leaf.py
```diff
@@ -428,10 +456,14 @@ def save_value(self, val) -> None:
 
     @property
     def value(self):
-        """NumPy.ndarray or None: The numeric value of the parameter.
-        """
+        """np.ndarray or None: The numeric value of the expression."""
```
PTNobel, line 459: Can we move the return type to a type annotation?

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -37,7 +37,8 @@
     'diag',
     'PSD',
     'NSD',
-    'bounds'
+    'bounds',
+    'sparsity'
```
PTNobel, line 41: I think we should have a trailing comma but maybe that's just my preference

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -151,7 +148,17 @@ def apply(self, problem):
                     id2new_var[var.id] = upper_tri
                     fill_coeff = Constant(upper_tri_to_full(n))
                     full_mat = fill_coeff @ upper_tri
-                    obj = reshape(full_mat, (n, n), order='F')
+                    obj = reshape(full_mat, (n, n))
```
PTNobel, line 151: Why no explicit order?

PTNobel (changes requested):
I don't look at the test changes, but I'm overall pretty positive on this!

cvxpy/expressions/leaf.py
Transurgeon: done.

cvxpy/reductions/cvx_attr2constr.py
Transurgeon: done.

cvxpy/reductions/cvx_attr2constr.py
Transurgeon: done.

cvxpy/expressions/leaf.py
Transurgeon: because it allows us to directly index the expression with the sparsity like this: ``expr[sparsity]``.
also the documentation for [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html#numpy.nonzero) returns a tuple of np.ndarrays.

cvxpy/expressions/leaf.py
Transurgeon: yea great catch.. I was using chatgpt for some initial prototyping and missed this, thanks!

cvxpy/expressions/leaf.py
```diff
@@ -178,7 +161,7 @@ def __init__(
         self.args = []
         self.bounds = bounds
 
-    def _validate_indices(self, indices: List[tuple | np.ndarray]) -> None:
+    def _validate_indices(self, indices: List[tuple] | tuple[np.ndarray]) -> None:
```
PTNobel, line 164:
```suggestion
    def _validate_indices(self, indices: list[tuple[int]] | tuple[np.ndarray]) -> None:
```

PTNobel (approved):
Looks good to me I think

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -57,14 +58,14 @@
 ]
 
 
-def convex_attributes(variables):
+def convex_attributes(variables) -> List[str]:
```
PTNobel, line 61:
```suggestion
def convex_attributes(variables) -> list[str]:
```

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -57,14 +58,14 @@
...
     """Returns a list of the (constraint-generating) convex attributes present
        among the variables.
     """
     return attributes_present(variables, CONVEX_ATTRIBUTES)
 
 
-def attributes_present(variables, attr_map):
+def attributes_present(variables, attr_map) -> List[str]:
```
PTNobel, line 68:
```suggestion
def attributes_present(variables, attr_map) -> list[str]:
```

PTNobel:
Ah sorry missed that the reduction isn't implemented yet

Transurgeon:
> Ah sorry missed that the reduction isn't implemented yet

I did implement it [here](https://github.com/cvxpy/cvxpy/pull/2532/files#diff-473769730e5348777c84850d453e9aa50cda2d3cf48cdb2562d541c5ff019069R152-R161)
but I still need to add more tests.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -81,13 +82,14 @@ def recover_value_for_variable(variable, lowered_value, project: bool = True):
         idxs = np.triu_indices(n)
         value[idxs] = lowered_value.flatten(order='F')
         return value + value.T - np.diag(value.diagonal())
+    #TODO add case for sparsity
```
Transurgeon, line 85: Ok I think the PR is ready for a more thorough review.
@SteveDiamond does this need to be implemented for this PR? I am not entirely sure what ``recover_value_for_variable`` does here. 
I am planning to add a follow-up PR to this one, for sparse parameters/coo_tensors, maybe I can implement this in the future?

SteveDiamond: I'm not sure. Maybe the project case handles it? The purpose is to set the ._value field of the original sparse variable object.

Transurgeon: yea I think the project case did handle it. I did reimplement the same thing just in case.
Also added a TODO for returning a sparse coo_tensor instead of a dense array.

SteveDiamond (approved):
lgtm

PR #2535, opened 2024-08-17
update nd-sum to be general for sparse variables

Transurgeon:
## Description
nd sums were not working in conjunction with sparse variables and it was because I had previously assumed that the tensor was always an identity matrix. @PTNobel gave the suggestion to form a sparse coefficient matrix to multiply the tensor with, thanks!
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond (approved):

SteveDiamond:
This makes sense to me.

PR #2541, opened 2024-08-19
Avoid unnecessarily creating numpy array of Var objects

simonbowly:
The array is unused in the code path for any currently supported version of Gurobi (>= v9.5) so can be removed to avoid unnecessary overhead. Additionally, in Python 3.12, creating this array triggers a [memory leak due to a bug in gurobipy](https://support.gurobi.com/hc/en-us/articles/27635574446353-Memory-leak-in-gurobipy-with-Python-3-12), so it would be good to remove it from cvxpy to reduce user impact.

Figured I'd start with a minimal change, but I'm happy to do a larger rewrite to remove the old workarounds in `gurobi_qpif.py`.

## Description
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.
- [x] Run the unittests and check that they’re passing (`pytest cvxpy/tests/test_qp_solvers.py` passes with gurobipy v11.0.3)

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
@@ -179,70 +179,21 @@ class GUROBI(QpSolver):
             # Set the start value of Gurobi vars to user provided values.
             for idx in range(len(x_grb)):
                 x_grb[idx].start = data['init_value'][idx]
-        model.update()
-
-        x = np.asarray(model.getVars())
 
         if A.shape[0] > 0:
-            if hasattr(model, 'addMConstr'):
-                # We can pass all of A @ x == b at once, use stable API
-                # introduced with Gurobi v9.5
-                model.addMConstr(A, None, grb.GRB.EQUAL, b)
-            elif hasattr(model, 'addMConstrs'):
-                # We can pass all of A @ x == b at once, use (now) deprecated
-                # API introduced with Gurobi v9.0
-                model.addMConstrs(A, None, grb.GRB.EQUAL, b)
-            else:
-                # Add equality constraints: iterate over the rows of A
-                # adding each row into the model
-                for i in range(A.shape[0]):
-                    start = A.indptr[i]
-                    end = A.indptr[i+1]
-                    variables = x[A.indices[start:end]]
-                    coeff = A.data[start:end]
-                    expr = grb.LinExpr(coeff, variables)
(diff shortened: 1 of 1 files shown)


Transurgeon (commented):
LGTM. also gonna add a backport tag to this PR.

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
```diff
@@ -181,7 +181,12 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
                 x_grb[idx].start = data['init_value'][idx]
         model.update()
 
-        x = np.asarray(model.getVars())
+        if hasattr(model, 'addMConstr') and hasattr(model, 'setMObjective'):
```
PTNobel, line 184: Could we just check the version number?

PTNobel (commented):
If it's not needed for anything we support, any reason to not just delete the code?

cvxpy/reductions/solvers/qp_solvers/gurobi_qpif.py
simonbowly: Sure, but I would go one way or the other for all the following branches.

simonbowly:
> If it's not needed for anything we support, any reason to not just delete the code?

I should clarify: by "supported" I mean versions that Gurobi still officially supports. As of now that's gurobi 9.5 and later. I'm not sure what cvxpy still intends to support in terms of Gurobi versions.

PTNobel:
> I'm not sure what cvxpy still intends to support in terms of Gurobi versions.

We only support what Gurobi supports is our policy as of today.

Transurgeon:
@simonbowly To move forward with this PR, I propose that we delete the old code since cvxpy only supports the same versions as Gurobi. What do you think?

simonbowly:
> To move forward with this PR, I propose that we delete the old code since cvxpy only supports the same versions as Gurobi. What do you think?

Works for me! This is now done with deafe23. The tests pass against gurobipy 9.5.2, 10.0.3, and 11.0.3.

Note that I also removed the update() calls. They're no longer needed in between adding variables and constraints in recent versions of Gurobi.

Transurgeon (approved):
LGTM.

PTNobel (approved):

PTNobel:
Thank you! I set the CI to run, should be good to merge once it passes!

PR #2542, opened 2024-08-21
`np.reshape` needs order as keyword argument

jonathanberthias:
## Description
In the latest release of NumPy (v2.1.0), the `order` argument to `np.reshape` became keyword-only, so the `cp.reshape` atom is failing.

The `newshape` argument is also deprecated, `shape` is preferred to align with the Array API.

### Note

I think the `order` change was unintended and might be reverted in a patch release (reported in numpy/numpy#27256). However, it is still worth fixing this in cvxpy to guard against future changes.

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/reshape.py
+++ b/cvxpy/atoms/affine/reshape.py
@@ -102,7 +102,7 @@ class reshape(AffAtom):
     def numeric(self, values):
         """Reshape the value.
         """
-        return np.reshape(values[0], self.shape, self.order)
+        return np.reshape(values[0], shape=self.shape, order=self.order)
 
     def validate_arguments(self) -> None:
         """Checks that the new shape has the same number of entries as the old.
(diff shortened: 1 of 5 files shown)


SteveDiamond (approved):

SteveDiamond:
The tests are failing. Maybe the change to reshape isn't backwards compatible?

Transurgeon:
> The tests are failing. Maybe the change to reshape isn't backwards compatible?

I just checked and the tests are failing because our builds are using an older version of NumPy (I think all of them are < 2.0). I am not sure if that counts as a breaking change.

@jonathanberthias could you add conditions on the new code you added to only be applied if the NumPy version is greater or equal than 2.1? Otherwise, I think it would be fine for us to wait until NumPy fixes the issue as well.

jonathanberthias:
Indeed the ˋshapeˋ keyword is also new in 2.1.0. It should be enough to just pass the shape as a positional argument to get it to work across versions.

jonathanberthias:
There is a pending fix to allow `order` to be passed by position in v2.1.1: https://github.com/numpy/numpy/pull/27262, I'll let you decide if it's still worth merging this PR.

SteveDiamond:
It is still worth merging IMO because it removes the newshape keyword arguments.

PR #2543, opened 2024-08-22
adding support for nd sum in parametrized expressions

Transurgeon:
## Description
This PR adapts the SciPy backend sum function to support parametrized sums. 
I introduced a new helper function ``get_sum_coeff_matrix`` since I noticed that it was being reused for both cases (non-dpp and dpp). 
This PR keeps the old functionality the same and adds a general case for when ``axis != None``.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -1574,7 +1574,7 @@ class TestND_Expressions():
     @pytest.mark.parametrize("axis", [(0),(1),(2),((0,1)),((0,2)),((2,1))])
     def test_nd_parametrized_sum(self, axis) -> None:
         param = cp.Parameter((2,2,2))
-        param.value = np.ones((2,2,2))
+        param.value = np.arange(8).reshape(2,2,2)
         expr = cp.multiply(self.x, param).sum(axis=axis)
         target = self.target.sum(axis=axis)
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_expressions.py
```diff
@@ -1571,6 +1571,16 @@ def test_nd_sum(self, axis) -> None:
         prob.solve(canon_backend=cp.SCIPY_CANON_BACKEND)
         assert np.allclose(expr.value, y)
 
+    @pytest.mark.parametrize("axis", [(0),(1),(2),((0,1)),((0,2)),((2,1))])
+    def test_nd_parametrized_sum(self, axis) -> None:
+        param = cp.Parameter((2,2,2))
+        param.value = np.ones((2,2,2))
```
SteveDiamond, line 1577: can you do it with the value np.reshape(np.arange(8), (2,2,2))? using all ones might mask bugs.

Transurgeon: ok let me change that.

Transurgeon: should be resolved once all the CI passes.

SteveDiamond (approved):
Nice job!

PR #2546, opened 2024-08-26
Improving indice processing for boolean and integer attributes

Transurgeon:
## Description
This PR aims to improve the processing of boolean and integer indices by avoiding an expensive list of tuple unpacking.
Results below:
```py
import time

import numpy as np


def new(shape):
    idx = np.arange(np.prod(shape))
    multi_idx = np.unravel_index(idx, shape, order='F')
    return np.ravel_multi_index(multi_idx, shape, order='F')

def prev(shape):
    idx = np.ndindex(shape)
    multi_idx = list(zip(*idx))
    return np.ravel_multi_index(multi_idx, shape, order='F')

# compare prev with new for time
shape = (10000, 1000)
start = time.time()
prev(shape)
print('prev:', time.time() - start)
start = time.time()
new(shape)
print('new:', time.time() - start)
```
Output:
```
prev: 3.726440906524658
new: 0.06124114990234375
```
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/matrix_stuffing.py
+++ b/cvxpy/reductions/matrix_stuffing.py
@@ -77,28 +77,22 @@ def extract_upper_bounds(variables: list, var_size: int) -> Optional[np.ndarray]
 
 
 def extract_mip_idx(variables) -> Tuple[List[int], List[int]]:
-    """Coalesces bool, int indices for variables.
-
-       The indexing scheme assumes that the variables will be coalesced into
-       a single one-dimensional variable, with each variable being reshaped
-       in Fortran order.
     """
-    def ravel_multi_index(multi_index, x, vert_offset):
-        """Ravel a multi-index and add a vertical offset to it.
-        """
-        ravel_idx = np.ravel_multi_index(multi_index, max(x.shape, (1,)), order='F')
-        return [(vert_offset + idx,) for idx in ravel_idx]
-    boolean_idx = []
-    integer_idx = []
-    vert_offset = 0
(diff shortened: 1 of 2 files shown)


cvxpy/reductions/matrix_stuffing.py
```diff
@@ -77,28 +77,22 @@ def extract_upper_bounds(variables: list, var_size: int) -> Optional[np.ndarray]
...
+    in Fortran order.
+    """
+    boolean_idx, integer_idx, offset = [], [], 0
     for x in variables:
+        ravel_shape = max(x.shape, (1,))
         if x.boolean_idx:
-            multi_index = list(zip(*x.boolean_idx))
-            boolean_idx += ravel_multi_index(multi_index, x, vert_offset)
+            ravel_idx = np.ravel_multi_index(x.boolean_idx, ravel_shape, order='F')
+            boolean_idx += [(idx + offset,) for idx in ravel_idx]
```
Transurgeon, line 91: @SteveDiamond I am thinking if it's possible to not have to call ravel_multi_index here and directly store the indices in vectorized format? If so, we can remove the call to ``np.unravel_index`` in ``leaf.py`` and then just apply the offset to each idx.

Transurgeon: nevermind, I just tried and some tests are failing. Let's keep it as is for now.

PTNobel (approved):

SteveDiamond (approved):

PR #2552, opened 2024-09-02
New CVaR atom

davidppineiro:
## Description
This PR creates the `cvar` atom (conditonal value at risk), which is implemented as a wrapper around the existing `dotsort` atom.

## Type of change
- [ ] Bug fix

- [X] Write unittests.

### DIFF
--- a/cvxpy/atoms/cvar.py
+++ b/cvxpy/atoms/cvar.py
@@ -16,7 +16,6 @@ limitations under the License.
 
 import numpy as np
 
-import cvxpy as cp
 from cvxpy.atoms.dotsort import dotsort
 
 
@@ -37,21 +36,18 @@ def cvar(x, beta):
 
     Parameters
     ----------
-    x : cvxpy.Expression
+    x : Expression or numeric constant.
         The vector of samples.
     beta : float
         The probability level, must be in the range :math:`[0, 1)`.
 
     Returns
     -------
-    cvxpy.Expression
+    Expression
         The CVaR of :math:`x` at probability level :math:`\beta`.
     """
     if not 0 <= beta < 1:
         raise ValueError(f"The probability level beta must be in the range [0, 1), got {beta}")
-
-    if not isinstance(x, cp.Expression):
-        raise TypeError("Input must be a CVXPY expression")
 
     if len(x.shape) > 1 and x.shape[1] != 1:
         raise ValueError(f"Input must be a vector, got shape {x.shape}")
--- a/cvxpy/atoms/cvar.py
+++ b/cvxpy/atoms/cvar.py
@@ -14,13 +14,41 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from cvxpy.atoms.dotsort import dotsort
 import numpy as np
+
+from cvxpy.atoms.dotsort import dotsort
 
 
 def cvar(x, beta):
-    """
-    The conditional value at risk (CVaR) of x at quantile level (1 - beta).
+    r"""The Conditional Value at Risk (CVaR) of a discrete random variable.
+
+    CVaR at level beta is a risk measure that captures the expected value over 
+    the worst (1-beta) fraction of outcomes of a real-valued random variable. 
+    For a random variable X representing losses, CVaR at level beta is defined as:
+
+    .. math::
+        \phi_\beta(X) = \mathbb{E}[X | X \geq \psi_\beta(X)]
+
+    where psi_beta(X) is the Value at Risk (VaR) at level beta.
+
+    For a discrete distribution represented by samples z_1, ..., z_m, 
+    CVaR can be computed as:
+
+    .. math::
+        \phi_\beta(z) = \inf_{\alpha \in \mathbb{R}} \left\{ \alpha + 
+        \frac{1}{(1-\beta)m}\sum_{i=1}^m(z_i-\alpha)_+ \right\}
+
+    where (z-alpha)_+ = max(z-alpha, 0) is the positive part of z-alpha.
+
+    This function is a wrapper around the dotsort atom, providing an efficient 
+    computation of CVaR for discrete distributions.
+
+    Parameters
+    ----------
+    x : cvxpy.Expression
+        The vector of sample losses.
+    beta : float
+        The probability level, between 0 and 1.
     """
     k = (1 - beta) * x.shape[0]
     w = np.append(np.ones(int(k)), k - int(k))
--- a/cvxpy/atoms/cvar.py
+++ b/cvxpy/atoms/cvar.py
@@ -1,0 +1,28 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


PTNobel:
Hey David! I'd encourage that you install and setup the pre-commit hook to make sure your files are properly formatted.

Overall seems like a fine change, but we'll need tests and docs to go with it too.

Tests can be whatever we have for dotsort, just a couple quick tests, and docs can just be an additional row in the atoms table.

Transurgeon:
Hello @davidppineiro thank you for your contribution. 
Just reiterating on Parth's suggestion and giving you more pointers.
- code style instructions can be found [here](https://github.com/cvxpy/cvxpy/blob/master/CONTRIBUTING.md#code-style).
- tests should be added in ``test_atoms.py`` below the ``TestDotSort`` class. You could try solving the wikipedia definition of cvar and do an assertion that the cvar atom gives a similar answer (up to some very small precision). 
- docs will mostly involve a better description in the docstrings of the function. Put some math, and explain the type of the inputs (x must be an vector expression, alpha must be a float between 0 and 1, representing a percentage), etc. Then to add it to the website, you would need to add one line below ``cp.cummax`` [here](https://github.com/cvxpy/cvxpy/blob/master/doc/source/api_reference/cvxpy.atoms.other_atoms.rst#cummax) so that sphinx can generate the docs for the new cvar atom.

PTNobel:
> docs will mostly involve a better description in the docstrings of the function. Put some math, and explain the type of the inputs (x must be an vector expression, alpha must be a float between 0 and 1, representing a percentage), etc. Then to add it to the website, you would need to add one line below cp.cummax [here](https://github.com/cvxpy/cvxpy/blob/master/doc/source/api_reference/cvxpy.atoms.other_atoms.rst#cummax) so that sphinx can generate the docs for the new cvar atom.

It also needs to be added to the Atoms table! David, you can find me on Tuesday and I can show you where that goes.

SteveDiamond (approved):
Nice!

doc/source/tutorial/functions/index.rst
```diff
@@ -81,6 +81,15 @@ and returns a scalar.
      - Curvature |_|
      - Monotonicity
 
+  * - :ref:`cvar(x, beta) <cvar>`
+      - average of the :math:`(1-\beta)` fraction of worst outcomes in :math:`x`
```
PTNobel, line 85: @Transurgeon Is this too long for the new format?

Transurgeon: yea most certainly will, good catch.
I think we might need to separate it into two lines.
Let met try to generate the docs and see how it looks.

Transurgeon: ok, there was actually a formatting error and the table wasn't showing at all.. good that we checked before merging.
@PTNobel or @davidppineiro , can you change the atoms table change from line 84-92 to the following (make sure the star and dashes are aligned with everything else)?:
```py
   * - :ref:`cvar(x, beta) <cvar>`
     - average of the :math:`(1-\beta)`
      
       fraction of worst
       
       outcomes in :math:`x`
     - :math:`x \in \mathbf{R}^m`
      
       :math:`\beta \in (0,1)`
     - depends on :math:`x`
     - |convex| convex
     - |incr| incr.
```

davidppineiro: Done! Does it show up now?

doc/source/tutorial/functions/index.rst
```diff
@@ -81,6 +81,19 @@ and returns a scalar.
      - Curvature |_|
      - Monotonicity
 
+   * - :ref:`cvar(x, beta) <cvar>`
+     - average of the :math:`(1-\beta)`
+      
+       fraction of worst
```
PTNobel, line 87: Last change, what does worst mean? Can we either say smallest or largest?

PTNobel (changes requested):

Transurgeon (approved):
LGTM on condition of addressing Parth's suggestion.
Thanks for doing this @davidppineiro !

doc/source/tutorial/functions/index.rst
davidppineiro: If we go with largest/smallest, we have to specify if the vector x represents losses or profits. 
Which version do you prefer?
1. average of the :math:`(1-\beta)` fraction of worst values in :math:`x`
2. average of the :math:`(1-\beta)` fraction of largest losses in :math:`x`
Let me know if you have a better alternative.

SteveDiamond: +1 for option 3

doc/source/tutorial/functions/index.rst
```diff
@@ -83,17 +83,19 @@ and returns a scalar.
...
      - average of the :math:`(1-\beta)`
-      
-       fraction of worst
        
-       outcomes in :math:`x`
+       fraction of largest values
+       
+       in :math:`x`, where :math:`x`
+       
+       is a vector of losses
```
SteveDiamond, line 91: do we need to say what `x` is, given we are saying cvar is averaging the largest values?

SteveDiamond: in other words I don't think we need the part about `where x is a vector of losses`

davidppineiro: Yeah, I think you are right. It's redundant. Will take it out.

doc/source/tutorial/functions/index.rst
PTNobel: Definitely 3

davidppineiro:
@PTNobel @SteveDiamond I have made some changes to the docstring in the CVaR atom and the table, and included some checks. Let me know if you have more suggestions.

cvxpy/atoms/cvar.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Returns
+    -------
+    Expression
+        The CVaR of :math:`x` at probability level :math:`\beta`.
+    """
+    if not 0 <= beta < 1:
+        raise ValueError(f"The probability level beta must be in the range [0, 1), got {beta}")
+
+    if len(x.shape) > 1 and x.shape[1] != 1:
```
SteveDiamond, line 52: I don't think this is the condition we want. It should be `len(x.shape) == 1`.

davidppineiro: But that will fail for x = cp.Variable((n,1)). Is this what we want?

SteveDiamond: Yes, that should fail. It's not consistent with numpy style to allow `x = cp.Variable((n,1))` but not allow `x = cp.Variable((1,n))`. If we allow 2D inputs we need to start thinking about having an axis argument, which would be much more complicated (though maybe doable?)

davidppineiro: Makes sense. Just changed the assert to raise an error if it's not a 1D array.

davidppineiro:
@SteveDiamond @PTNobel Also added another test for the CVaR atom: solving a small LP with a CVaR constraint.

SteveDiamond (approved):
Nice job!

PTNobel (approved):

PR #2556, opened 2024-09-07
improve sum_largest to use partitionsort instead of standard sort

Transurgeon:
## Description
This PR takes advantage of numpy's [argpartition](https://numpy.org/doc/stable/reference/generated/numpy.argpartition.html) function to partially sort the expression X. 
Since we only care about the $k$ largest (resp. smallest) values, we just need to make sure that the k-th element is sorted and all the values to its right (resp. left) are larger (resp. smaller). 

Algorithmic improvement (worst-case): ``O(nlogn)`` to ``O(nlogk)``.
Some small benchmarks:
 ```py
import time
import numpy as np
def new(x, k):
    indices = np.argpartition(-x, k)[:k]
    return x[indices].sum()
def prev(x, k):
    indices = np.argsort(-x)[:k]
    return x[indices].sum()

# compare prev with new for time
x = np.random.rand(10_000_000)
for k in [100, 1000, 10_000]:
    print('k:', k)
    start = time.time()
    prev(x, k)
    print('prev:', time.time() - start)
    start = time.time()
    new(x, k)
    print('new:', time.time() - start)
```
Output:
```
k: 100
prev: 1.1751160621643066
new: 0.10212492942810059
k: 1000
prev: 1.080510139465332
new: 0.10003519058227539
k: 10000
prev: 1.0831398963928223
new: 0.10057806968688965
```
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/sum_largest.py
+++ b/cvxpy/atoms/sum_largest.py
@@ -42,8 +42,6 @@ class sum_largest(Atom):
     def numeric(self, values):
         """
         Returns the sum of the k largest entries of the matrix.
-
-        New in 1.6.0: sum_largest uses np.argpartition instead of np.argsort
         """
         value = values[0].flatten()
         k = int(self.k)
--- a/cvxpy/atoms/sum_largest.py
+++ b/cvxpy/atoms/sum_largest.py
@@ -24,10 +24,11 @@ from cvxpy.atoms.atom import Atom
 
 
 class sum_largest(Atom):
-    """Sum of the largest k values in the matrix X.
+    """
+    Sum of the largest k values in the expression X
     """
 
-    def __init__(self, x, k) -> None:
+    def __init__(self, x, k: int) -> None:
         self.k = k
         super(sum_largest, self).__init__(x)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cvxpy/atoms/sum_largest.py [hunk 1]
```diff
@@ -39,10 +40,14 @@ def validate_arguments(self) -> None:
         super(sum_largest, self).validate_arguments()
 
     def numeric(self, values):
-        """Returns the sum of the k largest entries of the matrix.
+        """
+        Returns the sum of the k largest entries of the matrix.
+
+        New in 1.6.0: sum_largest uses np.argpartition instead of np.argsort
```
PTNobel, line 46: I don't see what this line is needed. We generally don't document internal implmenetation details.

PTNobel (approved):
One nit on docs, then it looks good to merge to me.

cvxpy/atoms/sum_largest.py
Transurgeon: done

PTNobel (approved):

PR #2562, opened 2024-09-14
improving performance of diff_matrix for cumsum

Transurgeon:
## Description
This PR introduces some slight speedups for the formation of the diff_matrix used in the cumsum atom. 
Also added tests and improved something for _grad of the atom as well. 
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/cumsum.py
+++ b/cvxpy/atoms/affine/cumsum.py
@@ -43,25 +43,10 @@ def get_diff_mat(dim: int, axis: int) -> sp.csc_matrix:
     SciPy CSC matrix
         A square matrix representing first order difference.
     """
-    # Construct a sparse matrix representation.
-    val_arr = []
-    row_arr = []
-    col_arr = []
-    for i in range(dim):
-        val_arr.append(1.)
-        row_arr.append(i)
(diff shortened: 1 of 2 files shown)


Transurgeon:
Ok, this PR seems ready to me. I found a weird thing where the axis input to cumsum defaults to zero, see below: https://github.com/cvxpy/cvxpy/blob/d6350ddd17ab436d128e3fc77f095fffaa777b3b/cvxpy/atoms/affine/cumsum.py#L63
This is going to behave unexpectedly if a user passes a matrix and ``axis=None``. Also this deviates from NumPy's behavior (see image below).
image
@PTNobel , @SteveDiamond , Albeit outside the scope of this PR, would fixing this default axis argument for cumsum constitute a breaking API change?

SteveDiamond:
I guess this would be a breaking API change. Let's keep a list of these inconsistencies with numpy.

PTNobel:
> I guess this would be a breaking API change. Let's keep a list of these inconsistencies with numpy.

I know I started a doc somewhere... Let me look for it.

SteveDiamond (approved):

Issue #2559, opened 2024-09-09
Supported solvers missing from "install" part of web docs


rileyjmurray:
Web docs for CVXPY 1.4 (and most lower versions) list the solvers we support and how to get them:

https://www.cvxpy.org/version/1.4/install/index.html

The web docs for CVXPY 1.5 do not:

https://www.cvxpy.org/version/1.5/install/index.html

It would be great if we could fix this ASAP.

rileyjmurray:
I love the new theme!

How about we merge the sign and monotonicity columns under a new "DCP Properties" column?

We could also make the "meaning" column more narrow for a few atoms. Here are screenshots with how I'd format things in a LaTeX (MathJax) align environment. 

<img width="470" alt="image" src="https://github.com/user-attachments/assets/d427aab6-3592-4516-93a0-1695243ee796">

could become
```
& \text{dot product of} \\
& \operatorname{sort}\operatorname{vec}(X) \text{ and } \\
& \operatorname{sort}\operatorname{vec}(X)  \\
```

<img width="452" alt="image" src="https://github.com/user-attachments/assets/648ffb5b-549a-4f77-9806-52f7c479ede6">

could be like the following (although it seems that this is already intended and just implemented incorrectly):
```
&\text{sum of } k \text{ largest} \\
&\text{eigenvalues of } X
```

<img width="446" alt="image" src="https://github.com/user-attachments/assets/7b9cc107-409d-4ccc-9be3-6fcd9e88ce4d">

could be the following (where numpy.std is a clickable link [to this page](https://numpy.org/doc/stable/reference/generated/numpy.std.html#numpy-std))
```
cvxpy analog
to numpy.std
```

<img width="419" alt="image" src="https://github.com/user-attachments/assets/428fd20d-5384-4c1f-add4-eb02890a0a9b">

could be analogous to with the std atom:
```
cvxpy analog
to numpy.var
```

rileyjmurray:
@Transurgeon this table also should have the columns narrowed. Right now it looks like the whole table fits but that's because the MIP column is getting hidden:

<img width="802" alt="image" src="https://github.com/user-attachments/assets/5e99f798-0f3c-4b4a-9c75-8300ca112d21">

SteveDiamond:
Probably we should roll forward with the new theme and make the tables work as best we can. With @rileyjmurray's improvements hopefully the tables will fit.

Transurgeon:
this issue is resolved from the two PRs above.

PR #2563, opened 2024-09-14
shortening meaning for scalar atoms to shrink table width

Transurgeon:
## Description
Related to discussion in #2559 .
[Link](https://cvxpy--60.org.readthedocs.build/en/60/tutorial/functions/index.html#scalar-functions) for preview of the new table.
I made another change on top of @rileyjmurray's suggestions (thanks for the feedback!).
For the total variation atom, it also took quite a bit of space so I separated the meaning into two parts (see image below). Please let me know what you think.
image
The next steps to fix the atoms table will be to combine monotonicity and sign under a new column called DCP properties (or maybe just properties). And also we need to show the MIP column for the solver capability table.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond (approved):
Thanks!

PTNobel:
We should remove the readthe docs files before merging right?

Transurgeon:
> We should remove the readthe docs files before merging right?

done.

PR #2585, opened 2024-10-08
Re-enable sdist publishing + clean up build workflows

### LINKED ISSUE #2584: Source Distribution (sdist) publishing broken
**Describe the bug**
As of the 1.5.3 release, source distributions are no longer being published to pypi.

It appears this was broken in 69e3e6c95b3908a6d3e897d7ba665379c20191af, which removed the `single_action_config` matrix entries that were being used to build/publish sdists.

Sdist builds are keyed on the `SINGLE_ACTION_CONFIG` env var
https://github.com/cvxpy/cvxpy/blob/fc92b6937e0fb95ce96f10cd56f4c67cc5045e18/.github/workflows/build.yml#L114-L118

Which is set from the similarly-named matrix variable:
https://github.com/cvxpy/cvxpy/blob/fc92b6937e0fb95ce96f10cd56f4c67cc5045e18/.github/workflows/build.yml#L82

**To Reproduce**
Visit [pypi](https://pypi.org/project/cvxpy/#files) and observe no files available under the "Source Distributions header.

**Expected behavior**
Source distribution expected to be available, e.g. [this example from the 1.5.2 release](https://pypi.org/project/cvxpy-base/1.5.2/#files)

**Version**
 - OS: Any
 - CVXPY Version: 1.5.3

**Additional context**

bkurtz:
## Description
Reduce duplication in github actions workflow for builds, and re-add "single action" configurations that produce coverage uploads and Source Distribution (sdist) files.

Fixes #2584.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,8 @@ jobs:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
         include:
+          # These are intended to just add their extra parameter to existing matrix combinations;
+          # see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
           - os: ubuntu-20.04
             python-version: 3.12
             openmp: "True"
@@ -84,6 +86,8 @@ jobs:
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
         build-cvxpy-base: [ true, false ]  # whether to build cvxpy-base (true) or regular cvxpy (false)
         include:
+          # This is intended to just add the single_action_config parameter to one existing combination;
+          # see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
           - os: ubuntu-20.04
             python-version: 3.12
             single_action_config: "True"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,7 +150,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.19.1
 
       - name: Build source
-        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
+        if: ${{env.SINGLE_ACTION_CONFIG == 'True'}}
         run: |
           pip install build
           python -m build --sdist -o wheelhouse
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,7 +150,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.19.1
 
       - name: Build source
-        if: ${{env.SINGLE_ACTION_CONFIG == 'True'}}
+        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
         run: |
           pip install build
           python -m build --sdist -o wheelhouse
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,6 +101,7 @@ jobs:
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
       PYPI_PASSWORD: ${{ matrix.build-cvxpy-base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
+      CVXPY_BASE_SUFFIX: ${{ matrix.build-cvxpy-base && "base-" || "" }}
 
     steps:
       - uses: actions/checkout@v4
@@ -126,7 +127,6 @@ jobs:
           sed -i.bak '/scs >= /d' pyproject.toml
           sed -i.bak -e 's/name="cvxpy",/name="cvxpy-base",/g' setup.py
           rm -rf pyproject.toml.bak setup.py.bak
-          echo "CVXPY_BASE_SUFFIX=base-" >> "$GITHUB_ENV"
 
       - name: Verify that we can build by pip-install on each OS.
         if: ${{matrix.build-cvxpy-base && env.PIP_INSTALL == 'True'}}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,7 +97,7 @@ jobs:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
       SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config == 'True' }}"
-      PIP_INSTALL: "${{ matrix.python-version == '3.12' }}"
+      PIP_INSTALL: "${{ matrix.python-version == '3.11' }}"
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
       PYPI_PASSWORD: ${{ matrix.build-cvxpy-base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,13 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
+        exclude:
+          - os: ubuntu-20.04
+            python-version: 3.12
+        include:
+          - os: ubuntu-20.04
+            python-version: 3.12
+            single_action_config: "True"
 
     env:
       RUNNER_OS: ${{ matrix.os }}
@@ -76,12 +83,21 @@ jobs:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
         base: [ true, false ]
+        exclude:
+          - os: ubuntu-20.04
+            python-version: 3.12
+            base: true
+        include:
+          - os: ubuntu-20.04
+            python-version: 3.12
+            base: true
+            single_action_config: "True"
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
       SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config == 'True' }}"
-      PIP_INSTALL: "${{ matrix.pip_install == 'True' }}"
+      PIP_INSTALL: "${{ matrix.python-version == '3.12' }}"
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
       PYPI_PASSWORD: ${{ matrix.base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,14 +148,14 @@ jobs:
           python -m build --sdist -o wheelhouse
 
       - name: Check wheels
-        if: ${{github.event_name == 'push' &&  env.USE_OPENMP != 'True'}}
+        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
         shell: bash
         run: |
           python -m pip install --upgrade twine
           twine check wheelhouse/*
 
-      - name:  Release to pypi
-        if: ${{env.DEPLOY == 'True' &&  env.USE_OPENMP != 'True'}}
+      - name: Release to pypi
+        if: ${{env.DEPLOY == 'True' && env.USE_OPENMP != 'True'}}
         shell: bash
         run: |
           twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_PASSWORD
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,7 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
-        base: [ true, false ]
+        build-cvxpy-base: [ true, false ]  # whether to build cvxpy-base (true) or regular cvxpy (false)
         include:
           - os: ubuntu-20.04
             python-version: 3.12
@@ -95,7 +95,7 @@ jobs:
       PIP_INSTALL: "${{ matrix.python-version == '3.12' }}"
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
-      PYPI_PASSWORD: ${{ matrix.base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
+      PYPI_PASSWORD: ${{ matrix.build-cvxpy-base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
 
     steps:
       - uses: actions/checkout@v4
@@ -110,7 +110,7 @@ jobs:
           echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
 
       - name: Adapt pyproject.toml
-        if: ${{matrix.base}}
+        if: ${{matrix.build-cvxpy-base}}
         shell: bash
         run: |
           # Mac has a different syntax for sed -i, this works across oses
@@ -124,7 +124,7 @@ jobs:
           echo "CVXPY_BASE_SUFFIX=base-" >> "$GITHUB_ENV"
 
       - name: Verify that we can build by pip-install on each OS.
-        if: ${{matrix.base && env.PIP_INSTALL == 'True'}}
+        if: ${{matrix.build-cvxpy-base && env.PIP_INSTALL == 'True'}}
         run: |
           pip install . pytest cplex
           pytest cvxpy/tests/test_conic_solvers.py -k 'TestCPLEX'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,9 +30,6 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
-        exclude:
-          - os: ubuntu-20.04
-            python-version: 3.12
         include:
           - os: ubuntu-20.04
             python-version: 3.12
@@ -83,14 +80,9 @@ jobs:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
         base: [ true, false ]
-        exclude:
-          - os: ubuntu-20.04
-            python-version: 3.12
-            base: true
         include:
           - os: ubuntu-20.04
             python-version: 3.12
-            base: true
             single_action_config: "True"
 
     env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,9 @@ jobs:
         include:
           - os: ubuntu-20.04
             python-version: 3.12
+            openmp: "True"
+          - os: macos-12
+            python-version: 3.12
             single_action_config: "True"
 
     env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,14 +36,15 @@ jobs:
           - os: ubuntu-20.04
             python-version: 3.12
             openmp: "True"
+            single_action_config: false
           - os: macos-12
             python-version: 3.12
-            single_action_config: "True"
+            single_action_config: true
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -75,14 +82,25 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
+        base: [ true, false ]
```
bkurtz, line 85: Combined the below `build_cvxpy-base` job into this one using an extra matrix parameter to keep things DRY.  That mostly meant using the `build_wheels` job definition, with the addition of a couple steps, but git seems to detect the change as mostly re-using the `build_cvxpy-base` job definition with a bunch of minor changes.

(if it makes life easier when reviewing, that refactor is mostly contained in the first commit of the PR).

.github/workflows/build.yml
```diff
@@ -75,14 +82,25 @@ jobs:
...
+        exclude:
```
bkurtz, line 86: Since, (AFAICT) the "single action" configurations _also_ build all the normal steps, I went ahead and set this up to _delete_ one normal matrix item and add back a new one with "single action config" so we don't end up with duplication.

.github/workflows/build.yml
```diff
@@ -75,14 +82,25 @@ jobs:
...
+          - os: ubuntu-20.04
+            python-version: 3.12
+            base: true
+        include:
+          - os: ubuntu-20.04
+            python-version: 3.12
+            base: true
```
bkurtz, line 93: As I think through this with the fresh light of a new day, I think we might actually want to do this for both base=true and base=false, so that we end up with sdists for both packages?

.github/workflows/build.yml
```diff
@@ -206,21 +156,21 @@ jobs:
...
-          twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_BASE_PASSWORD
+          twine upload --skip-existing --repository-url $PYPI_SERVER wheelhouse/* -u $PYPI_USER -p $PYPI_PASSWORD
 
       - name: Upload artifacts to github
-        if: ${{github.event_name == 'push'}}
+        if: ${{github.event_name == 'push' && env.USE_OPENMP != 'True'}}
         uses: actions/upload-artifact@v4
         with:
-          name: wheels-base-${{ matrix.os }}-${{ matrix.python-version }}
+          name: wheels-${CVXPY_BASE_SUFFIX}${{ matrix.os }}-${{ matrix.python-version }}
```
Transurgeon, line 175: does this suffix exist?

.github/workflows/build.yml
```diff
@@ -75,14 +82,25 @@ jobs:
...
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
       SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config == 'True' }}"
+      PIP_INSTALL: "${{ matrix.python-version == '3.12' }}"
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
-      PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      PYPI_PASSWORD: ${{ matrix.base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
```
Transurgeon, line 103: if my understanding is correct, we want to check if matrix.base is True and if so, we set PYPI_PASSWORD to be ``secrets.PYPI_BASE_PASSWORD``, is this code implementing this functionality?

.github/workflows/build.yml
```diff
@@ -177,9 +126,10 @@ jobs:
           sed -i.bak '/scs >= /d' pyproject.toml
           sed -i.bak -e 's/name="cvxpy",/name="cvxpy-base",/g' setup.py
           rm -rf pyproject.toml.bak setup.py.bak
+          echo "CVXPY_BASE_SUFFIX=base-" >> "$GITHUB_ENV"
 
       - name: Verify that we can build by pip-install on each OS.
-        if: ${{env.PIP_INSTALL == 'True'}}
+        if: ${{matrix.base && env.PIP_INSTALL == 'True'}}
```
Transurgeon, line 132: why are we checking if matrix.base is True here?

.github/workflows/build.yml
```diff
@@ -75,14 +82,25 @@ jobs:
...
+          - os: ubuntu-20.04
+            python-version: 3.12
+            base: true
+            single_action_config: "True"
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
       SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config == 'True' }}"
+      PIP_INSTALL: "${{ matrix.python-version == '3.12' }}"
```
Transurgeon, line 100: what does this do? Is it only being triggered when using python3.12?

.github/workflows/build.yml
Transurgeon: yea that part makes sense. But I don't particularly like the delete then adding it back with single action config. Maybe you can revert to using the previous setup [here](https://github.com/cvxpy/cvxpy/commit/69e3e6c95b3908a6d3e897d7ba665379c20191af#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L33-L43) but change everything to python3.12 instead of 3.8 (and also remove 3.12 from the matrix.python-version). What do you think?

.github/workflows/build.yml
Transurgeon: yeah that's probably a good idea, maybe if you just remove the ``base: true`` line it would work for both cases?

.github/workflows/build.yml
bkurtz: I guess I'd feel like that mostly depends on whether you want _other_ one-off configs added (e.g. previously there was also an "openmp" version of some sort.  If it's just for one config, it sortof seems silly to have to add _four_ manual entries rather than two to me, but... it's not my repo.

bkurtz: Actually, I think on a more careful read of the [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude) we should be able to just add the "include" clause and it will add keys to one of the existing matrix entries, and use similar shortcuts below.

.github/workflows/build.yml
bkurtz: Read through the docs in more detail, and I think you're right.

.github/workflows/build.yml
bkurtz: Previously this was only done in the `cvxpy-base` job of the workflow.  I don't have a strong opinion about when it should be done, but was trying to preserve previous behavior.

.github/workflows/build.yml
bkurtz: I don't actually know what y'all have set up with the secrets, but _I assume_ based on the way the job was already coded that there are separate passwords for the `cvxpy` and `cvxpy-base` packages being submitted to pypi.  This is supposed to condition which one of those ends up in the `PYPI_PASSWORD` environment variable.

Several other places in this file that check for true/false values seem to assume that the matrix outputs are strings.  I was hoping maybe they can be just directly used as booleans, but I'd want to test that!

.github/workflows/build.yml
bkurtz: On the "cvxpy-base" configuration, this was being used to test that packages successfully `pip install` after building.  This sortof seems like overkill to me, but it was supported before, so thought I'd add it back.  The previous setup only did this for one version of python (I assume on the grounds that if it works on one, it works on all?), so I thought I'd do the same here, and it seemed easier to just key it directly on a value than to add matrix entries.  Though now I've learned about the better syntax for adding matrix keys, that presumably wouldn't be too bad either.

.github/workflows/build.yml
bkurtz: The environment variable would (AFAICT) be undefined in that case, resulting in the shell expanding it as an empty string.

.github/workflows/build.yml
```diff
@@ -30,6 +30,13 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
+        exclude:
+          - os: ubuntu-20.04
+            python-version: 3.12
+        include:
```
bkurtz, line 36: this matrix previously included some extra configs that checked openmp compatibility.  I haven't added them back, but maybe we should?

.github/workflows/build.yml
Transurgeon: ok I think you are right, I am fine with keeping it as is.

.github/workflows/build.yml
Transurgeon: ok this seems fine to me.

.github/workflows/build.yml
Transurgeon: ok, thanks for the explanation. I think we should definitely keep the old behavior the same, so let's keep it this way.

.github/workflows/build.yml
Transurgeon: I understand what you mean here, but the current code has a condition on ``matrix.base && secrets.PYPI_BASE`` which doesn't seem to set the ``PYPI_PASSWORD`` to the correct value (doesn't it return a boolean instead now?)

.github/workflows/build.yml
Transurgeon: ok seems fine to me, as long as it doesn't create an error.

.github/workflows/build.yml
Transurgeon: yea I think that's a good idea.

Transurgeon:
@bkurtz the CI is failing due to an error with pytest when ``single_action_config == True``. It seems to be something related to the ``cov`` flag (code is [here](https://github.com/cvxpy/cvxpy/blob/master/continuous_integration/test_script.sh#L22-L26). I don't think this test was failing before.. and it seems quite weird that is is failing now, any idea why this is happening?

Transurgeon:
> @bkurtz the CI is failing due to an error with pytest when `single_action_config == True`. It seems to be something related to the `cov` flag (code is [here](https://github.com/cvxpy/cvxpy/blob/master/continuous_integration/test_script.sh#L22-L26) ). I don't think this test was failing before.. and it seems quite weird that it is failing now, any idea why this is happening?

NVM, I just made a PR with only the single action config changes in #2587 and it seems to be failing with the same error. I'll look into it.

bkurtz:
> @bkurtz the CI is failing due to an error with pytest when `single_action_config == True`. It seems to be something related to the `cov` flag (code is [here](https://github.com/cvxpy/cvxpy/blob/master/continuous_integration/test_script.sh#L22-L26) ). I don't think this test was failing before.. and it seems quite weird that it is failing now, any idea why this is happening?

That'll be the different dependencies that we used to install for 3.8 here:
https://github.com/cvxpy/cvxpy/blob/3ccafc2d9ae15a0967011a019986aa03de7c3384/continuous_integration/install_dependencies.sh#L14

(in particular, `pytest-cov` is relevant)

.github/workflows/build.yml
bkurtz: I think (as is true with short-circuit logical operators in python) the JS evaluation of these template fields doesn't coerce to boolean values.

See, for example, the existing code here:
https://github.com/cvxpy/cvxpy/blob/273e0cfa65a90440b0e5cf05164a76a501edf6c0/.github/workflows/build.yml#L41

Or compare to python:
```python
>>> False and "hi" or "bye"
'bye'
>>> True and "hi" or "bye"
'hi'
```

Transurgeon: ohhh I see, it's some javascript trick.. looks good then

Transurgeon:
> > @bkurtz the CI is failing due to an error with pytest when `single_action_config == True`. It seems to be something related to the `cov` flag (code is [here](https://github.com/cvxpy/cvxpy/blob/master/continuous_integration/test_script.sh#L22-L26) ). I don't think this test was failing before.. and it seems quite weird that it is failing now, any idea why this is happening?
> 
> That'll be the different dependencies that we used to install for 3.8 here:
> 
> https://github.com/cvxpy/cvxpy/blob/3ccafc2d9ae15a0967011a019986aa03de7c3384/continuous_integration/install_dependencies.sh#L14
> 
> (in particular, `pytest-cov` is relevant)

Yeah really nice catch! I was not careful when removing support for python 3.8.
can you add pytest-cov to the install dependencies on this [line](https://github.com/cvxpy/cvxpy/blob/297278e2a88db3c0084750052a16e60672074da3/continuous_integration/install_dependencies.sh#L12)?
If the CI passes, I think the PR makes sense and looks good to me :)

.github/workflows/build.yml
Transurgeon: could you add a comment next to base stating that it's for building cvxpy and cvxpy_base simultaneously? Would be helpful for future reference.

.github/workflows/build.yml
bkurtz: One thing I realized from this conversation that I do still want to verify is whether matrix values can actually be native booleans, or if I might need this to be
```suggestion
      PYPI_PASSWORD: ${{ matrix.base == 'true' && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
```
I'm going to push a commit that tests this on the regular "build" section of the workflow so that it's something we can test before merging the PR.
(Actually, we should talk more about what we _can_ do to test this more thoroughly before merging?)

.github/workflows/pr_benchmark_comment.yml
```diff
@@ -8,4 +8,4 @@ on:
 
 jobs:
   comment_pr:
-    uses: cvxpy/actions/.github/workflows/pr_benchmark_comment.yml@main
+    uses: cvxpy/actions/.github/workflows/benchmark_comment.yml@main
```
bkurtz, line 11: This is totally unrelated, but I've been getting warnings about this after most pushes on this PR, and it seemed like an easy thing to fix while I was here, assuming there aren't other issues in that pipeline that need fixing (if so, I'll probably just revert this change and leave it for another day).

.github/workflows/pr_benchmark_comment.yml
```diff
@@ -8,4 +8,4 @@ on:
 
 jobs:
   comment_pr:
-    uses: cvxpy/actions/.github/workflows/pr_benchmark_comment.yml@main
+    uses: cvxpy/actions/.github/workflows/bechmark_comment.yml@main
```
bkurtz, line 11: Aand... there's a typo in the filename in the other repo.  Should obviously be spelled be**n**chmark with an N.

bkurtz: Looks like these trigger from the main branch anyway, so getting the spelling right in this PR only helps so much.  Given that, and given that it only use AFAICT is in this repo, I'd be inclined to just rename the other end, rather than trying to fix here.  I've reverted these changes.

bkurtz:
Okay, I think from my point of view this is ready for final comments.  The "pr_benchmarks" job has a spurious failure (it says performance is worse, but it's been passing normally and I obviously didn't change any related code in this PR), so I assume it will pass on a re-run.

I tested that it looks to me like all the env vars are being set (in particular, that removing the restriction that sdists are only built on "push" got one to build), so I think everything is tested and working.

Do you all use "squash" merges or regular merges?  The commit history here is pretty gross and if it's not all going to squash down into one commit, I'd prefer to clean it up a bit before merging, but otherwise from my point of view this is ready to merge.

Transurgeon:
> Okay, I think from my point of view this is ready for final comments. The "pr_benchmarks" job has a spurious failure (it says performance is worse, but it's been passing normally and I obviously didn't change any related code in this PR), so I assume it will pass on a re-run.

Yeah this PR benchmark failure has happened a lot before, and it's clear that it has nothing to do with your changes.

> Do you all use "squash" merges or regular merges? The commit history here is pretty gross and if it's not all going to squash down into one commit, I'd prefer to clean it up a bit before merging, but otherwise from my point of view this is ready to merge.

yes we do squash our commits. Thanks again for your work!

Transurgeon (approved):
This PR looks good to me. @SteveDiamond @PTNobel could you do a quick sanity check and merge please?

Transurgeon:
> This PR looks good to me. @SteveDiamond @PTNobel could you do a quick sanity check and merge please?

Actually I had one last concern, while taking a second look at this. It seems like a few of the steps are being skipped due to the OPENMP environment variable?
image

.github/workflows/build.yml
```diff
@@ -30,11 +30,21 @@ jobs:
...
+          - os: macos-12
+            python-version: 3.12
+            single_action_config: true
 
     env:
       RUNNER_OS: ${{ matrix.os }}
       PYTHON_VERSION: ${{ matrix.python-version }}
-      SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config == 'True' && 'True' || 'False' }}"
+      SINGLE_ACTION_CONFIG: "${{ matrix.single_action_config && 'True' || 'False' }}"
       USE_OPENMP: "${{ matrix.openmp == 'True' && 'True' || 'False' }}"
```
Transurgeon, line 48: maybe we would need to remove the ``== 'True'`` similar to the change you did on line 47?

Transurgeon (changes requested):
Ideally the actions steps shouldn't be skipped (Build wheels/source, check wheels, etc.), although now that I check our previous CI behavior, it seems like it was already like this. Could you please address my final comments and I think we would be good to merge after that?

.github/workflows/build.yml
```diff
@@ -75,14 +85,23 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
+        build-cvxpy-base: [ true, false ]  # whether to build cvxpy-base (true) or regular cvxpy (false)
+        include:
+          # This is intended to just add the single_action_config parameter to one existing combination;
+          # see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
+          - os: ubuntu-20.04
+            python-version: 3.12
+            single_action_config: "True"
```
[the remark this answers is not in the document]
bkurtz: Can be done either way.  They were originally both strings and I changed the other one to test the way I was doing boolean values in the matrix, but didn't bother to change this one.  Let me make them consistent.

.github/workflows/build.yml
```diff
@@ -30,11 +30,21 @@ jobs:
       matrix:
         os: [ ubuntu-20.04, macos-12, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12" ]
+        include:
+          # These are intended to just add their extra parameter to existing matrix combinations;
+          # see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
+          - os: ubuntu-20.04
+            python-version: 3.12
+            openmp: "True"
```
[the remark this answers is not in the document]
bkurtz: No, this is working as expected.  This is a little confusing because the only uses of the `OPENMP` env var in _this_ file are in the second half (the "build wheels" job) which doesn't ever set the `OPENMP` environment variable.  The first half sets the environment variable (you can check it's value in the github actions output; I'll post a screenshot) and then it's used in the `install_dependencies.sh` and `test_script.sh` files I think.

It _probably_ makes sense to clean up the references to OPENMP in the second half, but they aren't doing any harm right now and I wanted to leave them in case I'm missing somehow that they're being set (and because I felt like before deleting them I should do the leg work to understand when they used to be used and I didn't feel like taking the time this week).

.github/workflows/build.yml
bkurtz: No, this is working correctly, I'll post a screenshot in the other thread illustrating this.
(On line 47, the value for `single_action_config` is a bool and so can be used in the logical statement directly; here, it's a string, so you have to compare to a string first to convert to a boolean.)

.github/workflows/build.yml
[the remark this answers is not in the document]
bkurtz: Here's an example showing ubuntu on python 3.12 with the env variable ending up set to `"True"` and on 3.11 showing it set to `"False"`
image
image

bkurtz (commented):
> Ideally the actions steps shouldn't be skipped (Build wheels/source, check wheels, etc.), although now that I check our previous CI behavior, it seems like it was already like this. Could you please address my final comments and I think we would be good to merge after that?

These are being skipped because of the `github.event_name == 'push'` condition on those steps.  I did a test (see state prior to bde70a4d0a3e3f5479752f202130c7a674d2586a) with one of those clauses removed to validate that things were working as expected for the other conditions here.

Essentially, this workflow is triggered two ways:
1. on PR updates (what we're doing here)
2. when there's a push event to `master`, a release branch, or any tag

I guess the rationale is that you only want to bother with wheel builds when updating master or a release branch, and TBH we could probably add a condition that skips the "build_wheels" jobs _entirely_ rather than only some steps, but that seemed like a bigger change than I set out to take on here!  It looks like we go even further and only actually push to pypi for "push" events to a tag (which makes sense).

bkurtz:
> These are being skipped because of the `github.event_name == 'push'` condition on those steps. I did a test (see state prior to [bde70a4](https://github.com/cvxpy/cvxpy/commit/bde70a4d0a3e3f5479752f202130c7a674d2586a)) with one of those clauses removed to validate that things were working as expected for the other conditions here.

Specifically, commit 5ccdc70291f9b26fa55094f6b4960fc285804ae1 is the one with valid tests for this, e.g. see https://github.com/cvxpy/cvxpy/actions/runs/11313988982/job/31463507745 which completed the "build source" step successfully.

Transurgeon (approved):
ok I think this LGTM now, thanks for the detailed PR @bkurtz , I have learned a lot about our CI system through this. @SteveDiamond could you have a quick look then merge please?

SteveDiamond (approved):
Thanks so much @bkurtz ! We are a very small and busy team, so this kind of help goes a long way.

Issue #2567, opened 2024-09-18
Adding support for more array manipulation atoms


Transurgeon:
As we are entering the age of ND arrays for cvxpy, it would be nice to have extended support for analog array manipulation atoms than those of NumPy's. A few that come to mind are the following:
- make vstack work with 3d arrays, raise an error for anything greater
- [numpy.stack](https://numpy.org/doc/stable/reference/generated/numpy.stack.html)
- [numpy.concatenate](https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html)

yemine0x01:
I am not sure if this is remain of interest. But I want to give `cp.stack` a try.

Transurgeon:
> I am not sure if this is remain of interest. But I want to give `cp.stack` a try.

it definitely still is!

For ``stack`` I would probably refer to ``vstack`` and other indice manipulation atoms (maybe ``broadcast`` can help?). The main difficulty comes from dealing with multiple array inputs and validating them. 

Let's chat on the discord if you have any questions.

yemine0x01:
Of course let's continue on discord. Can you ping me on it?

ClayCampaigne:
`cp.stack` is implemented in https://github.com/cvxpy/cvxpy/pull/2956. 
I may move on to `block` at some point soon.

Transurgeon:
closing this issue since it has been mostly taken care of.

PR #2597, opened 2024-10-23
Support concatenate on atoms

andresliszt:
## Description
Thank you for this amazing project!, I've been using it for a while and I decided to contribute. This PR aims to add `cp.concatenate` support as mentioned in here #2567. Note that so far I added it to the atoms and to the numpy/scipy backends. I haven't added it yet to the cpp/rust backends. I'd love to add it there, but before going too far, i'd like to know if this is needed/wanted and if my PR is in a good spot. I tried to follow numpy concatenate convention. I added several unit tests and also docstring documentation you can check to follow my ideas. 

**I pushed my changes after runnning ruff format command**, so as you can see there are some format changes that i'd like to avoid --- Do you have a command to do formatting that will sync with master branch?

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/concatenate.py
+++ b/cvxpy/atoms/affine/concatenate.py
@@ -25,7 +25,7 @@ from cvxpy.constraints.constraint import Constraint
 
 
 def concatenate(arg_list, axis: Optional[int] = 0):
-    assert axis is None or (isinstance(axis,int) and axis >=0)
+    assert axis is None or (isinstance(axis, int) and axis >= 0)
     return Concatenate(*(arg_list + [axis]))
 
 
@@ -56,36 +56,12 @@ class Concatenate(AffAtom):
         return [self.axis]
 
     def validate_arguments(self) -> None:
-        # If axis is None, arrays are flattened, so no validation is necessary.
-        if self.axis is None:
-            return
-        axis = self.axis
-        ref_shape = self.args[0].shape
-        # Zero-dimensional arrays cannot be concatenated along a specified axis.
-        if ref_shape == ():
-            raise ValueError("Zero-dimensional arrays cannot be concatenated along an axis")
-        ndim = len(ref_shape)
-        if axis >= ndim:
-            raise ValueError(f"Axis {axis} is out of bounds for array of dimension {ndim}")
-
-        for idx, arg in enumerate(self.args[1:], start=1):
-            arg_shape = arg.shape
-            if arg_shape == ():
-                raise ValueError("Zero-dimensional arrays cannot be concatenated along an axis")
-            if len(arg_shape) != ndim:
-                raise ValueError(
-                    f"all the input arrays must have same number of dimensions, but the array "
-                    f"at index 0 has {ndim} dimension(s) and the array at index {idx} has "
-                    f"{len(arg_shape)} dimension(s)"
-                )
-            for i in range(ndim):
-                if i != axis and ref_shape[i] != arg_shape[i]:
-                    raise ValueError(
-                        "All the input array dimensions except for the concatenation axis "
-                        f"must match exactly, but along dimension {i}, the array at index 0 "
-                        f"has size {ref_shape[i]} and the array at index {idx} has "
-                        f"size {arg_shape[i]}"
-                    )
+        # Validates that the input shapes in `self.args` are suitable for
+        # concatenation along a specified axis using numpy API with empty arrays
+        np.concatenate(
+            [np.empty(arg.shape, dtype=np.dtype([])) for arg in self.args],
+            axis=self.axis,
+        )
 
     def shape_from_args(self) -> Tuple[int, ...]:
         if self.axis is None:
--- a/cvxpy/atoms/affine/concatenate.py
+++ b/cvxpy/atoms/affine/concatenate.py
@@ -1,5 +1,5 @@
 """
-Copyright 2013 Steven Diamond
+Copyright, the CVXPY authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -24,7 +24,7 @@ from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.constraints.constraint import Constraint
 
 
-def concatenate(arg_list, axis: Optional[int] = None):
+def concatenate(arg_list, axis: Optional[int] = 0):
     assert axis is None or (isinstance(axis,int) and axis >=0)
     return Concatenate(*(arg_list + [axis]))
 
--- a/cvxpy/atoms/affine/concatenate.py
+++ b/cvxpy/atoms/affine/concatenate.py
@@ -64,17 +64,10 @@ class Concatenate(AffAtom):
         )
 
     def shape_from_args(self) -> Tuple[int, ...]:
-        if self.axis is None:
-            # Flatten all arrays and sum their sizes
-            total_size = sum(arg.size for arg in self.args)
-            return (total_size,)
-        axis = self.axis
-        ref_shape = self.args[0].shape
-        # Initialize the output shape with the reference shape
-        output_shape = list(ref_shape)
-        # Sum sizes along the specified axis
-        output_shape[axis] = sum(arg.shape[axis] for arg in self.args)
-        return tuple(output_shape)
+        return np.concatenate(
+            [np.empty(arg.shape, dtype=np.dtype([])) for arg in self.args],
+            axis=self.axis,
+        ).shape
 
     def graph_implementation(
         self,
--- a/cvxpy/atoms/affine/concatenate.py
+++ b/cvxpy/atoms/affine/concatenate.py
@@ -1,0 +1,125 @@
+"""
+Copyright 2013 Steven Diamond
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+from typing import List, Optional, Tuple
+
+import numpy as np
+
+import cvxpy.lin_ops.lin_op as lo
+import cvxpy.lin_ops.lin_utils as lu
+from cvxpy.atoms.affine.affine_atom import AffAtom
+from cvxpy.constraints.constraint import Constraint
+
+
+def concatenate(arg_list, axis: Optional[int] = None):
+    assert axis is None or (isinstance(axis,int) and axis >=0)
+    return Concatenate(*(arg_list + [axis]))
+
+
+class Concatenate(AffAtom):
+    """Concatenate along an existing axis"""
+
+    def __init__(self, *args) -> None:
+        if isinstance(args[-1], int) or args[-1] is None:
+            # Assume the last positional argument is axis
+            axis = args[-1]
+            args = args[:-1]
+            self.axis = axis
+        else:
+            self.axis = None
+        super().__init__(*args)
+
+    def is_atom_log_log_convex(self) -> bool:
+        return True
+
+    def is_atom_log_log_concave(self) -> bool:
+        return True
+
+    # Returns the concatenation of the values along the specified axis.
+    def numeric(self, values):
+        return np.concatenate(values, axis=self.axis)
+
+    def get_data(self) -> List[Optional[int]]:
+        return [self.axis]
+
+    def validate_arguments(self) -> None:
+        # If axis is None, arrays are flattened, so no validation is necessary.
+        if self.axis is None:
+            return
+        axis = self.axis
+        ref_shape = self.args[0].shape
+        # Zero-dimensional arrays cannot be concatenated along a specified axis.
+        if ref_shape == ():
+            raise ValueError("Zero-dimensional arrays cannot be concatenated along an axis")
+        ndim = len(ref_shape)
+        if axis >= ndim:
+            raise ValueError(f"Axis {axis} is out of bounds for array of dimension {ndim}")
+
+        for idx, arg in enumerate(self.args[1:], start=1):
+            arg_shape = arg.shape
+            if arg_shape == ():
+                raise ValueError("Zero-dimensional arrays cannot be concatenated along an axis")
+            if len(arg_shape) != ndim:
+                raise ValueError(
+                    f"all the input arrays must have same number of dimensions, but the array "
+                    f"at index 0 has {ndim} dimension(s) and the array at index {idx} has "
+                    f"{len(arg_shape)} dimension(s)"
+                )
+            for i in range(ndim):
+                if i != axis and ref_shape[i] != arg_shape[i]:
+                    raise ValueError(
+                        "All the input array dimensions except for the concatenation axis "
+                        f"must match exactly, but along dimension {i}, the array at index 0 "
+                        f"has size {ref_shape[i]} and the array at index {idx} has "
+                        f"size {arg_shape[i]}"
+                    )
+
+    def shape_from_args(self) -> Tuple[int, ...]:
+        if self.axis is None:
+            # Flatten all arrays and sum their sizes
+            total_size = sum(arg.size for arg in self.args)
+            return (total_size,)
+        axis = self.axis
+        ref_shape = self.args[0].shape
+        # Initialize the output shape with the reference shape
+        output_shape = list(ref_shape)
+        # Sum sizes along the specified axis
+        output_shape[axis] = sum(arg.shape[axis] for arg in self.args)
+        return tuple(output_shape)
+
+    def graph_implementation(
+        self,
+        arg_objs,
+        shape: Tuple[int, ...],
+        data=None,
+    ) -> Tuple[lo.LinOp, List[Constraint]]:
+        """Concatenate the expressions along an existing axis.
+
+        Parameters
+        ----------
+        arg_objs : list
+            LinOp for each argument.
+        shape : tuple
+            The shape of the resulting expression.
+        data :
+            Additional data required by the atom. In this case data wraps axis
+
+        Returns
+        -------
+        tuple
+            (LinOp for the objective, list of constraints)
+        """
+        return (lu.concatenate(arg_objs, shape, data[0]), [])
--- a/cvxpy/atoms/__init__.py
+++ b/cvxpy/atoms/__init__.py
@@ -31,7 +31,7 @@ from cvxpy.atoms.affine.partial_transpose import partial_transpose
 from cvxpy.atoms.affine.promote import promote
 from cvxpy.atoms.affine.real import real
 from cvxpy.atoms.affine.reshape import deep_flatten, reshape
-# TODO: Remove comment once concatenated is enabledf
+# TODO: Remove comment once concatenated is enabled
 # from cvxpy.atoms.affine.concatenate import concatenate
 from cvxpy.atoms.affine.sum import sum
 from cvxpy.atoms.affine.trace import trace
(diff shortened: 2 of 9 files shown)


SteveDiamond:
This is a great feature! We have essentially stopped adding atoms to the CPP backend due to the difficulty of updating it, though it's certainly better to have concatenate in the CPP backend than not have it. Don't worry about the rust backend. @Transurgeon can you take a look at this PR when you have time?

Transurgeon:
> **I pushed my changes after runnning ruff format command**, so as you can see there are some format changes that i'd like to avoid --- Do you have a command to do formatting that will sync with master branch?

yes, we usually use ``pre-commit``. You can find some instructions [here](https://www.cvxpy.org/contributing/index.html#code-style) on how to get setup. 

P.S. Thanks a lot for doing this, I'll give a proper review once the linting errors are fixed :).

andresliszt:
> > **I pushed my changes after runnning ruff format command**, so as you can see there are some format changes that i'd like to avoid --- Do you have a command to do formatting that will sync with master branch?
> 
> yes, we usually use `pre-commit`. You can find some instructions [here](https://www.cvxpy.org/contributing/index.html#code-style) on how to get setup.
> 
> P.S. Thanks a lot for doing this, I'll give a proper review once the linting errors are fixed :).

Thanks, I ran the commands and also removed some redundancy. I think it's ready to be reviewed

cvxpy/atoms/affine/concatenate.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self, *args) -> None:
+        if isinstance(args[-1], int) or args[-1] is None:
+            # Assume the last positional argument is axis
+            axis = args[-1]
+            args = args[:-1]
+            self.axis = axis
+        else:
+            self.axis = None
+        super().__init__(*args)
```
andresliszt, line 43: I'm not sure if this is the best way to define `__init__` , but in many places the instances are created using `type(self)(*(args + data)) `, so it doesn't let me to define `axis` as positional. An alternative could be define something like `def __init__(self, expr_l, expr_r, axis)` and we iteratively handle `arg_list` in the main concatenate method

SteveDiamond: Yes, this is not correct. There should be an explicit axis atom. This must be communicated to the internals via the get_data function, which you already have.

andresliszt: there is an axis atom class `AxisAtom(Atom)`, but it was thought (afik) for single expressions, for example, tensor sum on a specific axis

SteveDiamond: sorry, typo. I meant axis argument.

SteveDiamond: I do sort of see though what you mean. Let us get back to you on the right approach.

Transurgeon (changes requested):
Thanks for doing @andresliszt , great work overall!
Apart from the review comments, I had two things to mention:
1. I have some ideas for more tests we could do (potentially generating some using hypothesis).
2. I'll need to spend more time looking at what you wrote for the docstrings of the ``python_backend`` test.
I'll probably do another round of review to address these two issues.

andresliszt:
> Thanks for doing @andresliszt , great work overall!
> 
> Apart from the review comments, I had two things to mention:
> 
> 1. I have some ideas for more tests we could do (potentially generating some using hypothesis).
> 
> 2. I'll need to spend more time looking at what you wrote for the docstrings of the ``python_backend`` test.
> 
> I'll probably do another round of review to address these two issues.

Thanks for the review, yes will be addresing your comments soon!

cvxpy/atoms/affine/concatenate.py
andresliszt: I think that probably axis as **keyword only** might not work. The problem with this is that we define an arbitrary number of expression at init `def __init__(self, *args, axis=0)`  is that axis becomes a keyword only argument and this doesn't work with this cvxpy [code](https://github.com/andresliszt/cvxpy/blob/297278e2a88db3c0084750052a16e60672074da3/cvxpy/reductions/eval_params.py#L70):

```python
def apply(self, problem):
      """Replace parameters with constant values."""

         # code omitted for simplicity ...

        data = c.get_data()
        if data is not None:
            constraints.append(type(c)(*(args + data))) # This has problems with arbitrary args mixed with keywords
        else:
            constraints.append(type(c)(*args))
      return problems.problem.Problem(objective, constraints), []
```
The evaluation of `type(c)(*(args + data))` will never pass the keyword only values to the constructor, so axis will be passed as positional.

Possible solutions:

1.  Define a `get_data_kw` which allows to define extra data as keyword only as well. This is more complicated because there are several references to `get_data` in the source code.
2. Define `Concatenate.__init__` to work with 2 expressions  `Concatenate.__init__(expr_l, expr_r, axis = 0)`. In this case `axis` works as positional or keyword argument and `type(c)(*(args + data))` will correct setup the values for the expressions and axis. In this case the main function `concatenate`, which I understand is the only function that the user uses, will still having an `arg_list` as argument but we have to `iteratively`  concatenate pairs of expressions. For example in the case of arg_list = `[expr1, expr2, expr3]` we should return `Concatenate(expr3, Concatenate(expr1, expr2, axis = axis), axis = axis)
 `

SteveDiamond: I think the way you did it originally is fine for now. We can revisit the topic in a separate PR.

andresliszt:
Iteresting, some tests are failing due to `XPRESS`, i did not notice it locally because the test is skipped, now I could reproduce it installing `XPRESS`, I will be poking arround!

SteveDiamond:
Don't worry about Xpress. That was fixed recently in a different MR. We discussed internally and there is interest in including this MR in the upcoming 1.6 release. Do you need help to implement the C++ version of concatenate? It only needs to work for up to 2D inputs. Whenever you have a value with >2 dimensions then cvxpy will ensure the SciPy backend is used.

andresliszt:
> Don't worry about Xpress. That was fixed recently in a different MR. We discussed internally and there is interest in including this MR in the upcoming 1.6 release. Do you need help to implement the C++ version of concatenate? It only needs to work for up to 2D inputs. Whenever you have a value with >2 dimensions then cvxpy will ensure the SciPy backend is used.

Cool! Yes i can do it, i was waiting for your confirmation, ty

Transurgeon:
> Cool! Yes i can do it, i was waiting for your confirmation, ty

Thanks a lot @andresliszt , just wanted to say everything else you did looks good as well. We might make another PR for more tests at a future point. If you want to make the CI pass, you can update this branch (my personal preference is to use ``git merge master`` from the branch after syncing your fork with cvxpy master). But it is not necessary. 

P.S. can you check if vstack and hstack work with 3d arrays also? I remember last time I was testing, one of them was failing.

cvxpy/atoms/affine/concatenate.py
```diff
@@ -0,0 +1,125 @@
...
+                    f"{len(arg_shape)} dimension(s)"
+                )
+            for i in range(ndim):
+                if i != axis and ref_shape[i] != arg_shape[i]:
+                    raise ValueError(
+                        "All the input array dimensions except for the concatenation axis "
+                        f"must match exactly, but along dimension {i}, the array at index 0 "
+                        f"has size {ref_shape[i]} and the array at index {idx} has "
+                        f"size {arg_shape[i]}"
+                    )
```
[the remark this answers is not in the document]
Transurgeon: I just asked someone from the NumPy community (through the scientific python discord) about this and he recommended us to use [np.lib.side_tricks.as_strided](https://numpy.org/devdocs/reference/generated/numpy.lib.stride_tricks.as_strided.html) to mimic arrays of arbitrary dimensions. Let me get back to you on the best way we should go about this.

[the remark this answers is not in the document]
Transurgeon: ok so actually we can use ``np.empty(shape, dtype=np.dtype([]))`` to create truly empty arrays.
Then we can call np.concatenate on them to validate the inputs :). 
But if this changes a lot of the tests, we can always do this in a separate PR.

andresliszt:
I might be wrong but, is there a way to get the `data` from `linOp`, in this case, `data = [axis]` in the C++ side? . I was looking at the code, and this [line](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/cvxcore/python/cppbackend.py#L179) took my attention. The `sum_entries` defines a data in the same way i'm defining data for my `concatenate` operator. You can see it in [here](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/lin_ops/lin_utils.py#L386). I was looking the C++ side of that operator and I don't see references to `axis` or `keepdims`.

cvxpy/atoms/affine/concatenate.py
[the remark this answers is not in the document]
andresliszt: cool!

SteveDiamond:
> I might be wrong but, is there a way to get the `data` from `linOp`, in this case, `data = [axis]` in the C++ side? . I was looking at the code, and this [line](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/cvxcore/python/cppbackend.py#L179) took my attention. The `sum_entries` defines a data in the same way i'm defining data for my `concatenate` operator, and according to that line _I think is being ignored_. You can see it in [here](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/lin_ops/lin_utils.py#L386). I was looking the C++ side of that operator and I don't see references to `axis` or `keepdims`.

There is not a clean way to get the data right now. You can see for sum_entries that in certain case the ``axis`` argument produces different graph_implementations 
https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/atoms/affine/sum.py#L104
This is because the C++ interface doesn't handle ``axis`` arguments, though the data is passed down so it is possible to extract.

@Transurgeon I think there is a strong case for not adding ``concatenate`` to the C++ backend. After we release 1.6 and people start using the N-dimensional functionality, there will be many requests/MRs for new linear operators. It's not really feasible to keep adding more functionality to the C++ interface.

@andresliszt what do you think?

andresliszt:
> > I might be wrong but, is there a way to get the `data` from `linOp`, in this case, `data = [axis]` in the C++ side? . I was looking at the code, and this [line](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/cvxcore/python/cppbackend.py#L179) took my attention. The `sum_entries` defines a data in the same way i'm defining data for my `concatenate` operator, and according to that line _I think is being ignored_. You can see it in [here](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/lin_ops/lin_utils.py#L386). I was looking the C++ side of that operator and I don't see references to `axis` or `keepdims`.
> 
> There is not a clean way to get the data right now. You can see for sum_entries that in certain case the `axis` argument produces different graph_implementations
> 
> https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/atoms/affine/sum.py#L104
> 
> 
> This is because the C++ interface doesn't handle `axis` arguments, though the data is passed down so it is possible to extract.
> @Transurgeon I think there is a strong case for not adding `concatenate` to the C++ backend. After we release 1.6 and people start using the N-dimensional functionality, there will be many requests/MRs for new linear operators. It's not really feasible to keep adding more functionality to the C++ interface.
> 
> @andresliszt what do you think?

So if this is not added to C++, the `get_backend` function in photo should change?  For example, search in all problem's atoms, and if some atom is in a _non cpp supported list_, so far `non_cpp_supported = [concatenate]`, we use scipy or are you thinking in get rid build matrix C++ completely?
image

Transurgeon:
> > I might be wrong but, is there a way to get the `data` from `linOp`, in this case, `data = [axis]` in the C++ side? . I was looking at the code, and this [line](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/cvxcore/python/cppbackend.py#L179) took my attention. The `sum_entries` defines a data in the same way i'm defining data for my `concatenate` operator, and according to that line _I think is being ignored_. You can see it in [here](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/lin_ops/lin_utils.py#L386). I was looking the C++ side of that operator and I don't see references to `axis` or `keepdims`.
> 
> There is not a clean way to get the data right now. You can see for sum_entries that in certain case the `axis` argument produces different graph_implementations
> 
> https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/atoms/affine/sum.py#L104
> 
> This is because the C++ interface doesn't handle `axis` arguments, though the data is passed down so it is possible to extract.
> 
> @Transurgeon I think there is a strong case for not adding `concatenate` to the C++ backend. After we release 1.6 and people start using the N-dimensional functionality, there will be many requests/MRs for new linear operators. It's not really feasible to keep adding more functionality to the C++ interface.

Yea I totally agree. I forgot about this major difference between backends. Honestly, I think this PR can be merged for now, but maybe we just don't advertise it in the 1.6 release yet?

@andresliszt as you can see the CPP backend has been a bit of a historical pain to maintain. We are actively trying to get rid of it by improving the performance on the SciPy backend. Please let us know if you would like to help! We can coordinate on [discord](https://discord.gg/4urRQeGBCr) if needed :).
Thanks again for this great PR!

cvxpy/atoms/affine/concatenate.py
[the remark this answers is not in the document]
andresliszt: Hey @Transurgeon  i was thinking in this approach, `np.empty` uses memory, so if the shape is big, it could cause some memory issues. I tested it locally and for big arrays numpy raises a `MemoryAllocationError`. What do you think?

> Unlike other array creation functions (e.g. [zeros](https://numpy.org/devdocs/reference/generated/numpy.zeros.html#numpy.zeros), [ones](https://numpy.org/devdocs/reference/generated/numpy.ones.html#numpy.ones), [full](https://numpy.org/devdocs/reference/generated/numpy.full.html#numpy.full)), [empty](https://numpy.org/devdocs/reference/generated/numpy.empty.html#numpy.empty) does not initialize the values of the array, and may therefore be marginally faster. However, the values stored in the newly allocated array are arbitrary. For reproducible behavior, be sure to set each element of the array before reading.

Transurgeon:
> So if this is not added to C++, the `_get_cannon_backend` function in photo should change? For example, search in all problem's atoms, and if some atom is in a _non cpp supported list_, so far `non_cpp_supported = [concatenate]`, we use scipy or are you thinking in get rid build matrix C++ completely? 

Your suggestion is great but might be a bit too complicated. I think anyways, users won't be using cp.concatenate with vectors and matrices (you can already do all of that with vstack/hstack I think), so we hope that users will only use concatenate on >3d expressions which already defaults to the scipy backend. So there's probably no need to make any further changes.

cvxpy/atoms/affine/concatenate.py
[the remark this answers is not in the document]
andresliszt: EDIT: Adding dtype as you suggested works fine, no issues. I will add this change to my PR

andresliszt:
> > > I might be wrong but, is there a way to get the `data` from `linOp`, in this case, `data = [axis]` in the C++ side? . I was looking at the code, and this [line](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/cvxcore/python/cppbackend.py#L179) took my attention. The `sum_entries` defines a data in the same way i'm defining data for my `concatenate` operator, and according to that line _I think is being ignored_. You can see it in [here](https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/lin_ops/lin_utils.py#L386). I was looking the C++ side of that operator and I don't see references to `axis` or `keepdims`.
> 
> > 
> 
> > There is not a clean way to get the data right now. You can see for sum_entries that in certain case the `axis` argument produces different graph_implementations
> 
> > 
> 
> > https://github.com/cvxpy/cvxpy/blob/fc66f1449835580cc6f0d47dea2051b6428520f0/cvxpy/atoms/affine/sum.py#L104
> 
> > 
> 
> > This is because the C++ interface doesn't handle `axis` arguments, though the data is passed down so it is possible to extract.
> 
> > 
> 
> > @Transurgeon I think there is a strong case for not adding `concatenate` to the C++ backend. After we release 1.6 and people start using the N-dimensional functionality, there will be many requests/MRs for new linear operators. It's not really feasible to keep adding more functionality to the C++ interface.
> 
> 
> 
> Yea I totally agree. I forgot about this major difference between backends. Honestly, I think this PR can be merged for now, but maybe we just don't advertise it in the 1.6 release yet?
> 
> 
> 
> @andresliszt as you can see the CPP backend has been a bit of a historical pain to maintain. We are actively trying to get rid of it by improving the performance on the SciPy backend. Please let us know if you would like to help! We can coordinate on [discord](https://discord.gg/4urRQeGBCr) if needed :).
> 
> Thanks again for this great PR!

Would love to help! I will need to bring back my math notebooks haha

Transurgeon:
> > @andresliszt as you can see the CPP backend has been a bit of a historical pain to maintain. We are actively trying to get rid of it by improving the performance on the SciPy backend. Please let us know if you would like to help! We can coordinate on [discord](https://discord.gg/4urRQeGBCr) if needed :).
> > Thanks again for this great PR!
> 
> Would love to help! I will need to bring back my math notebooks haha

It would actually be more software involved, but definitely some math to understand the backend (although you seem to have grasped many things already on your own :) ). Could you send a message on the cvxpy discord general channel? I will add you from there.

rileyjmurray:
A formal way to not “advertise” is to not have the function appear in the
top level namespace.

andresliszt:
Alright, as @rileyjmurray suggested I took off `concatenate` from the main api for now, i.e `cp.concatenate` doesn't exist.  I think this is reasonable and must be enabled once this is fully supported. Also @Transurgeon I used numpy API for validation and shape set up.  Note same validation could be implemented in many cvxpy methods such as hstack or vstack.

Also would be nice to have custom errors to something like this `raise CustomCvxPyError from numpy_exc
` for having more info where the error is happening

cvxpy/atoms/affine/concatenate.py
```diff
@@ -0,0 +1,94 @@
...
+    def get_data(self) -> List[Optional[int]]:
+        return [self.axis]
+
+    def validate_arguments(self) -> None:
+        # Validates that the input shapes in `self.args` are suitable for
+        # concatenation along a specified axis using numpy API with empty arrays
+        np.concatenate(
+            [np.empty(arg.shape, dtype=np.dtype([])) for arg in self.args],
+            axis=self.axis,
+        )
```
andresliszt, line 64: this could be saved to a private attribute to be used below in `shape_from_args` but i don't think is worth it @Transurgeon

Transurgeon (approved):
I think this PR is ready for now. We will definitely come back to this at some future point to polish things and deal with the CPP backend. Thanks again @andresliszt

SteveDiamond (approved):

cvxpy/atoms/affine/concatenate.py
andresliszt: @SteveDiamond @Transurgeon  a con of this approach is that numpy error msgs have changed across different verisons.  I set up the tests with the error msgs i got locally with my specific numpy version. What do you suggest to handle this issue?

Transurgeon: Is it possible to just match the last part of the error message?
i.e. ``but along dimension 0, the array at index 0...``

SteveDiamond:
Looks like there are some test failures. This approach is fine for now, but we should do a follow up MR to ensure concatenate is not used with the CPP backend. It shouldn't be too difficult. Then we can add it back into the main namespace.

cvxpy/atoms/affine/concatenate.py
SteveDiamond: Yeah don't sweat it too much, whatever works and is at least testing there is an exception of some kind is fine.

andresliszt: ready to go !

SteveDiamond:
@andresliszt thanks for doing this!

Issue #2583, opened 2024-10-07
inverse product of one variable


aszekMosek:
``inv_prod`` does something incorrect when the argument has length 1 so one has to use ``inv_pos`` for this special case. Compare these two models of ``1/x[0] + 1/(x[0]*x[1])``:

```
import cvxpy as cp

x = cp.Variable(2)
prob = cp.Problem(cp.Minimize(cp.inv_prod(x[:1])+cp.inv_prod(x[:2])), [cp.sum(x)==2])
prob.solve(verbose=True)
print(x.value)
print(prob.value) # Wrong

x = cp.Variable(2)
prob = cp.Problem(cp.Minimize(cp.inv_pos(x[0])+cp.inv_prod(x[:2])), [cp.sum(x)==2])
prob.solve(verbose=True)
print(x.value)
print(prob.value) # Correct
```

Transurgeon:
I digged a bit into this and it seems like both ``geo_mean`` and ``inv_pos`` are not able to deal with scalar inputs. 
A fix in the source code of ``atoms/inv_prod.py`` could be something like the following:
Before:
```py
    return power(inv_pos(geo_mean(value)), int(sum(value.shape)))
```
After:
```py
    p = int(sum(value.shape))
    if p in [0, 1]:
        return inv_pos(value)
    return power(inv_pos(geo_mean(value)), p)
```
I think this could fix the issue with a minimal amount of changes, but it might also be worth trying to generalize ``geo_mean``. Any thoughts?

SteveDiamond:
> I digged a bit into this and it seems like both `geo_mean` and `inv_pos` are not able to deal with scalar inputs. A fix in the source code of `atoms/inv_prod.py` could be something like the following: Before:
> 
> ```python
>     return power(inv_pos(geo_mean(value)), int(sum(value.shape)))
> ```
> 
> After:
> 
> ```python
>     p = int(sum(value.shape))
>     if p in [0, 1]:
>         return inv_pos(value)
>     return power(inv_pos(geo_mean(value)), p)
> ```
> 
> I think this could fix the issue with a minimal amount of changes, but it might also be worth trying to generalize `geo_mean`. Any thoughts?

It would be better to fix geo_mean so it works with scalar inputs.

UweGensheimer:
It seems that when we pass an argument of shape (1, ) to `cp.inv_prod` then `gm_constrs` returns an empty constraint.
Would it make sense to add an if block in `gm_constrs` like below?

[...]

PR #2609, opened 2024-10-31
Bug/geo mean

UweGensheimer:
## Description
Issue link (if applicable): [Issue #2583](https://github.com/cvxpy/cvxpy/issues/2583)
As discussed in the linked issue, when a variable x of shape (1,) is passed to inv_prod, it is not properly modeled it as 1/x.
This issue seems to be caused by the geo_mean (in `gm_constrs` function) which for x variable with the shape of (1,) it does not include a constraint to bind the epigraph variable t with x.

@Transurgeon has nicely provided some tests which I added to `test_dgp.py ` to make sure we properly support scalar inputs.
Currently, test_geo_mean_scalar2 is not passing.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_dgp.py
+++ b/cvxpy/tests/test_dgp.py
@@ -54,7 +54,7 @@ class TestDgp(BaseTest):
         div = x / y
         self.assertTrue(div.is_log_log_affine())
 
-        div = posynomial / (3.0 * x * y**(-0.1))
+        div = posynomial / (3.0 * x * y ** (-0.1))
         self.assertTrue(div.is_log_log_convex())
         self.assertFalse(div.is_log_log_concave())
         self.assertTrue(div.is_dgp())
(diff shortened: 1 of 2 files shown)


cvxpy/atoms/geo_mean.py
```diff
@@ -213,7 +214,10 @@ def __init__(self, x, p: Optional[List[int]] = None, max_denom: int = 1024) -> N
         elif p is not None and hasattr(p, '__getitem__'):
             p = np.array(p)
             idxs = p > 0
-            x = Expression.cast_to_const(x)[idxs]
```
UweGensheimer, line 216: Special thanks to @Transurgeon for suggesting this fix here

cvxpy/atoms/geo_mean.py
```diff
@@ -213,7 +214,10 @@ def __init__(self, x, p: Optional[List[int]] = None, max_denom: int = 1024) -> N
...
+            if isinstance(x, list):
+                x = np.array(x)
+            shape = (1,) if x.ndim == 0 else x.shape
+            x = Expression.cast_to_const(promote(x, shape=shape))[idxs]
```
Transurgeon, line 220: so I just noticed that promote raises an error on non-scalar expressions.
maybe you can change this to be something like this:
```py
if x.ndim == 0:
      x = Expression.cast_to_const(promote(x, shape=shape))[idxs]
else:
     x = Expression.cast_to_const(x)[idxs]
```

cvxpy/utilities/power_tools.py
```diff
@@ -86,6 +86,12 @@ def gm_constrs(t, x_list, p):
         if 1 not in elem:
             constraints += [gm(d[elem], d[children[0]], d[children[1]])]
 
+    # Handle single-variable case
+    if len(x_list) == 1:
+        # Assuming p[0] = 1 for geometric mean
+        x = x_list[0]
+        constraints += [t == x]
+
```
Transurgeon, line 94: nice!

Transurgeon (changes requested):

cvxpy/tests/test_dgp.py
```diff
@@ -160,6 +160,41 @@ def test_geo_mean(self) -> None:
         self.assertTrue(geo_mean.is_log_log_convex())
         self.assertTrue(geo_mean.is_log_log_concave())
 
+    def test_geo_mean_scalar1(self) -> None:
+        x = cvxpy.Variable(1, pos=True)
+        p = np.array([2])
+        geo_mean = cvxpy.geo_mean(x, p)
+        self.assertTrue(geo_mean.is_dgp())
```
SteveDiamond, line 167: Can you test that if you maximize cvxpy.geo_mean(x, p) here with the constraint x == 2 then you get an optimal value of 2?

UweGensheimer: Done!

cvxpy/atoms/geo_mean.py
```diff
@@ -213,7 +214,12 @@ def __init__(self, x, p: Optional[List[int]] = None, max_denom: int = 1024) -> N
         elif p is not None and hasattr(p, '__getitem__'):
             p = np.array(p)
             idxs = p > 0
-            x = Expression.cast_to_const(x)[idxs]
+            if isinstance(x, list):
+                x = np.array(x)
```
SteveDiamond, line 218: What case is being covered here?

SteveDiamond (approved):
LGTM, had one question about list input.

cvxpy/atoms/geo_mean.py
UweGensheimer: In some cases, x is passed as a list. For example, in `test_problem.py::TestProblem::test_geo_mean` when we try to solve
```python
        x = Variable(5)
        prob = Problem(cp.Maximize(cp.geo_mean(x, p)), [cp.sum(x) <= 1])
        prob.solve(solver=cp.SCS, eps=1e-5)
        x = np.array(x.value).flatten()
        x_true = p/sum(p)
    
        self.assertTrue(np.allclose(prob.value, cp.geo_mean(list(x), p).value))
```

Transurgeon (approved):
LGTM thanks!

SteveDiamond:
Thanks for fixing this @UweGensheimer !

Issue #2598, opened 2024-10-24
Include ruff format in `pre-commit`


andresliszt:
I noticed that ruff formatting is not being done on commits. The `pre-commit.yaml` file is

```yaml
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    # Ruff version.
    rev: 'v0.1.11'
    hooks:
      - id: ruff
        args: [--fix, --exit-non-zero-on-fix]
```

According to [ruff official docs](https://docs.astral.sh/ruff/formatter/#sorting-imports) they do not support (yet: _A unified command for both linting and formatting is planned_) formatting and linting in a single command. Running ruff format in the source code shows several files that must be formatted.

It makes it a bit difficult to create a PR if the code formatting is out of sync with the master

Transurgeon:
I am not too familiar with Ruff format, does it make changes on all the files or only those that have changes in the pre-commit? Also, feel free to open a PR for this, we would probably be happy to add this feature.

andresliszt:
> I am not too familiar with Ruff format, does it make changes on all the files or only those that have changes in the pre-commit? Also, feel free to open a PR for this, we would probably be happy to add this feature.

Happy to help with. The steps are:

1. Add the format rule in the `pre-commit`
2. The very first commit with this rule will format all the files that ruff finds that needs formatting according to the rules defined in the `pyproject.toml`. I ran it locally and are several python files.
3. New commits will make formatting mandatory, done automatically by pre-commit

Transurgeon:
i'll close this since it caused way too many changes and also we are not sure what rules we want for ruff just yet.

PR #2611, opened 2024-11-01
Support/ruff formatter

andresliszt:
## Description
This PR adds support for ruff formatting. Also upgrades `pre-commit` ruff version to 0.7.2.  In this PR `ruff format` command was executed in the source code that's why there are several file changes. This is a one time thing, with this the code is formatted, and we must enforce actions [here](https://github.com/cvxpy/actions/blob/main/.github/workflows/linting.yml) to check the format style too. 

**There are not code changes other than running ruff format**

Issue link : #2598 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


cvxpy/atoms/axis_atom.py
```diff
@@ -43,7 +43,7 @@ def shape_from_args(self) -> Tuple[int, ...]:
             if self.keepdims:
                 shape[self.axis] = 1
             else:
-                shape = shape[:self.axis] + shape[self.axis+1:]
+                shape = shape[: self.axis] + shape[self.axis + 1 :]
```
Transurgeon, line 46: I am not sure I agree with this change being better. But that's just my personal preference.

cvxpy/atoms/elementwise/kl_div.py
```diff
@@ -80,17 +75,15 @@ def _grad(self, values) -> List[Optional[csc_matrix]]:
             # Non-differentiable.
             return [None, None]
         else:
-            div = values[0]/values[1]
+            div = values[0] / values[1]
```
Transurgeon, line 78: ditto. I don't think this makes any difference in terms of readability.

cvxpy/constraints/constraint.py
```diff
@@ -234,8 +229,7 @@ def __bool__(self):
 
     @property
     def dual_value(self):
-        """NumPy.ndarray : The value of the dual variable.
-        """
+        """NumPy.ndarray : The value of the dual variable."""
```
Transurgeon, line 232: we should probably change this to a type hint.

cvxpy/constraints/nonpos.py
```diff
@@ -90,8 +90,9 @@ def residual(self):
     def violation(self):
         res = self.residual
         if res is None:
-            raise ValueError("Cannot compute the violation of an constraint "
-                             "whose expression is None-valued.")
+            raise ValueError(
+                'Cannot compute the violation of an constraint ' 'whose expression is None-valued.'
+            )
```
Transurgeon, line 95: this looks kind of weird. Why is there a ``' '`` after the word constraint?

cvxpy/constraints/zero.py
```diff
@@ -83,8 +81,7 @@ def residual(self):
     # The value of the dual variable.
     @property
     def dual_value(self):
-        """NumPy.ndarray : The value of the dual variable.
-        """
+        """NumPy.ndarray : The value of the dual variable."""
```
Transurgeon, line 84: ditto, should move np.ndarray to a type hint.

cvxpy/expressions/constants/constant.py
```diff
@@ -67,35 +68,34 @@ def __init__(self, value, name: Optional[str] = None) -> None:
...
         return [self]
 
     def is_constant(self) -> bool:
         return True
 
     @property
     def value(self):
-        """NumPy.ndarray or None: The numeric value of the constant.
-        """
+        """NumPy.ndarray or None: The numeric value of the constant."""
```
Transurgeon, line 94: ditto: this should be optional[np.ndarray]

pyproject.toml
```diff
@@ -17,6 +17,11 @@ exclude = [
 # The minimum Python version that should be supported
 target-version = "py39"
 
+ignore = ["E721"]
+
+[tool.ruff.format]
+quote-style = "single"
```
Transurgeon, line 23: is there a reason we are using single-quotes? it doesn't seem like a standard across other repos. I'd like others opinion on this, @PTNobel @SteveDiamond .

Transurgeon (changes requested):
There are some really clean improvements to the readability of the code with these changes! But some changes also don't feel so necessary, so I think we should get other maintainers feedback first.

andresliszt:
> There are some really clean improvements to the readability of the code with these changes! But some changes also don't feel so necessary, so I think we should get other maintainers feedback first.

Yes, some changes seem to make little difference in terms of readability, but the idea is to have code formatted by the formatter; that way, we have a standard that everyone is required to follow. For example, the difference between a/b and a / b is minimal; there’s only one space in between. This is where the formatter comes in handy, as I don’t have to worry about these details and let the tool handle it for me.

pyproject.toml
andresliszt: Yes this was my preference, can be double without problem

cvxpy/constraints/nonpos.py
andresliszt: I think it put both strings in the same line. Also it suggests that is no need to use 2 strings (msg is short), but the formartter never changes your code so that must be manual

pyproject.toml
PTNobel: I greatly prefer double as I'm used to double being for strings and single for chars. I'll let BDFL Steven make the call tho

SteveDiamond: Agreed should be double quotes

andresliszt:
Yes for sure! In the future happy to help with this

jonathanberthias:
> We discussed more internally and are not ready yet to switch to autoformatting. We will look into it more on our side.

If this is due to worries regarding the formatting of math expressions, there currently is an initiative to put together some guidelines: scientific-python/specs#326. It's from some Scipy maintainers, and it seems they caught the attention of the Ruff developers, so you might be able to expect _something_ better suited to your needs in the future 🤞

Transurgeon:
closing this since it became out-of-date.. maybe to be revisited later on though.

PR #2631, opened 2024-11-14
Enable derivative of dual variables w.r.t. problem parameters

ericlux:
## Description
Enable cp.Problem.derivative() to write deltas to the dual variables.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -1423,12 +1423,12 @@ class Problem(u.Canonical):
         dvars = param_prog.split_solution(
             dx, [v.id for v in self.variables()])
 
-        dcons = {}
-        constraint_offsets = inverse_data[0].constraint_offsets
-        for c in self.constraints:
-            col = constraint_offsets[c.id]
-            value = dy[col:c.size+col]
-            dcons[c.id] = value
+        # dcons = {}
+        # constraint_offsets = inverse_data[0].constraint_offsets
+        # for c in self.constraints:
+        #     col = constraint_offsets[c.id]
+        #     value = dy[col:c.size+col]
+        #     dcons[c.id] = value
 
         for variable in self.variables():
             # variable.delta = solution.primal_vars[variable.id]
@@ -1440,8 +1440,8 @@ class Problem(u.Canonical):
                 variable.delta = variable.delta * variable.value
 
         for cons in self.constraints:
-            # val = solution.dual_vars[cons.id]
-            val = dcons[cons.id]
+            val = solution.dual_vars[cons.id]
+            # val = dcons[cons.id]
             dual_var = cons.dual_variables[0]
             dual_var.delta = val
 
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -42,9 +42,8 @@ from cvxpy.reductions.dqcp2dcp import dqcp2dcp
 from cvxpy.reductions.eval_params import EvalParams
 from cvxpy.reductions.flip_objective import FlipObjective
 from cvxpy.reductions.solution import INF_OR_UNB_MESSAGE
-from cvxpy.reductions.solvers import bisection
+from cvxpy.reductions.solvers import bisection, conic_solvers
 from cvxpy.reductions.solvers import defines as slv_def
-from cvxpy.reductions.solvers.conic_solvers import DIFFCP
 from cvxpy.reductions.solvers.conic_solvers.conic_solver import ConicSolver
 from cvxpy.reductions.solvers.defines import SOLVER_MAP_CONIC, SOLVER_MAP_QP
 from cvxpy.reductions.solvers.qp_solvers.qp_solver import QpSolver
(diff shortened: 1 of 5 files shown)


cvxpy/problems/problem.py
```diff
@@ -44,6 +44,7 @@
 from cvxpy.reductions.solution import INF_OR_UNB_MESSAGE
 from cvxpy.reductions.solvers import bisection
 from cvxpy.reductions.solvers import defines as slv_def
+from cvxpy.reductions.solvers.conic_solvers import DIFFCP
```
SteveDiamond, line 47: We can't import DIFFCP here because it might not be installed.

ericlux: Moved into the derivative method, which can't be called without diffcp I believe

cvxpy/problems/problem.py
```diff
@@ -1394,17 +1393,56 @@ def derivative(self) -> None:
...
+                "iter" : self.solver_stats.num_iters,
+                "status_val" : 1, # see SCS.STATUS_MAP
+                "pobj" : self.value,
+            },
+            "x" : dx,
+            "y" : dy,
+            "solve_method": s.SCS,
+        }
+
+        diffcp = conic_solvers.DIFFCP()
```
PTNobel, line 1417: Shouldn't there already be an instance of this class in the reduction chain? As the last reduction?

cvxpy/problems/problem.py
```diff
@@ -1394,17 +1393,56 @@ def derivative(self) -> None:
...
         start = time.time()
-        dx, _, _ = D(-dA, db, dc)
+        dx, dy, _ = D(-dA, db, dc)
         end = time.time()
         backward_cache['D_TIME'] = end - start
+
+        chain = self._cache.solving_chain
+        inverse_data = self._cache.inverse_data
+
+        solution = {
```
PTNobel, line 1403: What's happening with this dictionary? Is the use of SCS here necessary? I'm worried if we're breaking derivatives with Clarabel/ECOS support.

cvxpy/problems/problem.py
PTNobel: Could we have cached the solution instead?

cvxpy/problems/problem.py
```diff
@@ -1394,17 +1393,56 @@ def derivative(self) -> None:
...
+            },
+            "x" : dx,
+            "y" : dy,
+            "solve_method": s.SCS,
+        }
+
+        diffcp = conic_solvers.DIFFCP()
+        _, solver_inverse_data = self._cache.solving_chain.solver.apply(self._cache.param_prog)
+        solution = diffcp.invert(solution, solver_inverse_data)
+        solution = chain.invert(solution, inverse_data)
```
PTNobel, line 1420: Huh, I'm surprised the chain doesn't have diffcp in it already; doesn't `requires_grad` require `solver=cp.DIFFCP`?

cvxpy/problems/problem.py
```diff
@@ -1394,17 +1393,56 @@ def derivative(self) -> None:
...
+
         dvars = param_prog.split_solution(
             dx, [v.id for v in self.variables()])
+
+        # dcons = {}
```
PTNobel, line 1425: Style nit: no commented out code without like a good explanation for it

cvxpy/problems/problem.py
```diff
@@ -1394,17 +1393,56 @@ def derivative(self) -> None:
...
+        # constraint_offsets = inverse_data[0].constraint_offsets
+        # for c in self.constraints:
+        #     col = constraint_offsets[c.id]
+        #     value = dy[col:c.size+col]
+        #     dcons[c.id] = value
+
         for variable in self.variables():
+            # variable.delta = solution.primal_vars[variable.id]
```
PTNobel, line 1433: Same nit as above.

cvxpy/reductions/inverse_data.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, problem) -> None:
         varis = problem.variables()
         self.id_map, self.var_offsets, self.x_length, self.var_shapes = \
             InverseData.get_var_offsets(varis)
-
+
```
PTNobel, line 27: Should not be adding white space at the end of a line

cvxpy/reductions/inverse_data.py
```diff
@@ -53,4 +53,4 @@ def get_var_offsets(variables):
             var_offsets[x.id] = vert_offset
             id_map[x.id] = (vert_offset, x.size)
             vert_offset += x.size
-        return (id_map, var_offsets, vert_offset, var_shapes)
+        return (id_map, var_offsets, vert_offset, var_shapes)
```
PTNobel, line 56: Please end your text files with a `\n`

PTNobel (changes requested):
Micellaneous unsubstantial concerns + some confusion about what's happening in the code. Some of this might be me not remembering what the derivative codepath in CVXPY is like.

cvxpy/reductions/inverse_data.py
PTNobel:
```suggestion

```

Transurgeon:
I'll close this PR since its been a while. Let's revisit this if its still of interest at some point.

PR #2635, opened 2024-11-17
Fixes assigning to `.value` for sparse leafs and adds `.value_sparse`

maxschaller:
## Description
This PR fixes a bug when assigning the value of a `Leaf` with the `sparsity` attribute. Previously, the `_validate_value` function was demanding that the shape of the assigned value is the same as the shape of the `Leaf`. At the same time, the `project` function is expecting the value to be a flat array of only the nonzero entries. This PR eliminates this contradiction and changes `_validate_value` such that the value can be passed as a flat array. This is relevant especially for `Parameter` with `sparsity` (as used by `cvxpygen`, for example). *It adds a warning if you are instantiating an unnecessarily dense array. --- Parth*

*In addition, it adds an attribute `value_sparse` to leaves, and only leaves, to read the sparse value of a matrix from. Additionally, we add a compat layer to support patch around old SciPy lacking support for ND coo arrays. --- Parth*

No benchmarks were run because it seems they disappeared from `cvxpy/tests`. The [contributing guide](https://github.com/cvxpy/cvxpy/blob/master/CONTRIBUTING.md#benchmarks) still refers to `cvxpy/tests/test_benchmarks.py`.

## Type of change
- [x] Bug fix

- [x] Write unittests.


Transurgeon (approved):
LGTM thanks for addressing this issue with the new sparsity attribute!
Do you think it would be worth allowing users to specify a sparse value with a sparse array at some point? (The only issue would be when users deal with 3d variables or greater, we might need to create a custom datatype to support this).

Transurgeon:
also about the benchmarks, the contributing guide in the repo is outdated..
we have a benchmarks [repo](https://github.com/cvxpy/benchmarks) that gets run on every PR.
I think the contributing guide on the [docs](https://www.cvxpy.org/contributing/index.html#benchmarks) reflects this.

maxschaller:
> LGTM thanks for addressing this issue with the new sparsity attribute! Do you think it would be worth allowing users to specify a sparse value with a sparse array at some point? (The only issue would be when users deal with 3d variables or greater, we might need to create a custom datatype to support this).

I think so, yes. Currently, if a user wants to assign a `scipy.sparse.csc_matrix` S to a sparse `Parameter` P, they can simply say `P.value = S.data`. If their sparse array is row-major or nd, then things are less simple. I guess this would go into a separate PR.

maxschaller:
The new design expects a numpy nd array. This generalizes to any number of dimensions, and is closer to the previous API.

Transurgeon:
> The new design expects a numpy nd array. This generalizes to any number of dimensions, and is closer to the previous API.

hmm I think this is great with minimal changes. But then we are not able to do what you mentioned above with setting ``P.value = S.data``. Let's get back to this in another PR though. 
Like you said, this fixes a current bug with setting parameter values of sparse leaves.
@SteveDiamond , what do you think about this?

PTNobel:
I'll write a design doc tomorrow

SteveDiamond (approved):

PTNobel:
Design doc is here: https://github.com/PTNobel/PTNobel/blob/main/cvxpy_design_docs/sparsity-api.md

Decision was to go with Option 3, but raise a warning if we're accessing a sparse variable through its dense representation.

PTNobel:
@maxschaller @SteveDiamond @Transurgeon if y'all want to look at this PR, and confirm you're okay with this approach, I'll write tests tomorrow or the weekend.

cvxpy/expressions/leaf.py
```diff
@@ -437,18 +441,41 @@ def save_value(self, val) -> None:
...
         """The numeric value of the expression."""
-        return self._value
+        if self.sparse_idx is None:
+            return self._value
+        else:
+            warnings.warn('Reading from a sparse CVXPY expression via `.value` is discouraged.'
+                          ' Use `.value_sparse` instead', RuntimeWarning, 1)
+            val = np.zeros(self.shape)
+            val[self.sparse_idx] = self._value[self.sparse_idx]
+            return self._value
```
Transurgeon, line 451: shouldn't this return val?

cvxpy/expressions/leaf.py
```diff
@@ -423,9 +425,11 @@ def project(self, val):
                     return val
                 w[bad] = 0
             return (V * w).dot(V.T)
-        elif self.attributes['sparsity']:
+        elif self.attributes['sparsity'] and not sparse_path:
+            warnings.warn('Assigning to a sparse CVXPY expression via `.value` is discouraged.'
+                          ' Use `.value_sparse` instead', RuntimeWarning, 3)
```
Transurgeon, line 430: nice, was just curious what the parameter 3 does? Is it for the stack trace? Why is it different than the warning below?

cvxpy/expressions/leaf.py
```diff
@@ -437,18 +441,41 @@ def save_value(self, val) -> None:
...
+    @property
+    def value_sparse(self) -> Optional[...]:
+        """The numeric value of the expression if it is a sparse variable."""
+        return self._value
+
+    @value_sparse.setter
+    def value_sparse(self, val) -> None:
+        if not isinstance(val.coords, (scipy_coo_array, sp.coo_array)):
+            raise ValueError('Invalid type for assigning value_sparse.'
+                             f'type should be {scipy_coo_array_name}')
```
Transurgeon, line 466: we should add some documentation so the users knows how to define a ``cvxpy.compat.scipy_coo_array``.
Also im concerned that we are gonna get errors for ``invalid attribute coords``. does a numpy array have a coords attribute?

cvxpy/expressions/leaf.py
```diff
@@ -439,18 +441,41 @@ def save_value(self, val) -> None:
...
     def value(self) -> Optional[np.ndarray]:
         """The numeric value of the expression."""
-        return self._value
+        if self.sparse_idx is None:
+            return self._value
+        else:
+            warnings.warn('Reading from a sparse CVXPY expression via `.value` is discouraged.'
+                          ' Use `.value_sparse` instead', RuntimeWarning, 1)
+            val = np.zeros(self.shape)
+            val[self.sparse_idx] = self._value[self.sparse_idx]
```
maxschaller, line 450: Given that `self._value` holds only the nonzeros, shouldn't we say

`val[self.sparse_idx] = self._value`

cvxpy/expressions/leaf.py
```diff
@@ -423,7 +425,7 @@ def project(self, val):
                     return val
                 w[bad] = 0
             return (V * w).dot(V.T)
-        elif self.attributes['sparsity']:
+        elif self.attributes['sparsity'] and not sparse_path:
             warnings.warn('Assigning to a sparse CVXPY expression via `.value` is discouraged.'
                           ' Use `.value_sparse` instead', RuntimeWarning, 3)
             new_val = np.zeros(self.shape)
```
maxschaller, line 431: This seems to assign a dense array to `self._value`. I think if we want to avoid storing a dense array, we should warn here if there are nonzeros outside of the pattern, return just `val[self.sparse_idx]`, and bypass the `delta` check in `_validate_value`.

cvxpy/expressions/leaf.py
```diff
@@ -439,18 +441,41 @@ def save_value(self, val) -> None:
...
+        return self._value
+
+    @value_sparse.setter
+    def value_sparse(self, val) -> None:
+        if not isinstance(val.coords, (scipy_coo_array, sp.coo_array)):
+            raise ValueError('Invalid type for assigning value_sparse.'
+                             f'type should be {scipy_coo_array_name}')
+        if val.coords != self.sparse_idx:
+            raise ValueError(f'Indexes differ between {val} and variable.')
+        self.save_value(self._validate_value(val.data, True))
```
maxschaller, line 469: How about we say

`self.save_value(self._validate_value(val, True).data)` 

This way, the shape check inside `_validate_value` still works. Inside `project`, we could add another case under `self.attributes['sparsity']` that just passes `val` through if `sparse_path=True`.

cvxpy/expressions/leaf.py
PTNobel: Yes. Yes it should.

cvxpy/expressions/leaf.py
PTNobel: It is for the stack trace, the 3 means that it goes up above this function into self._validate_value and then up from there to .value

cvxpy/expressions/leaf.py
PTNobel: Definitely need to write some docs. We should check for a NumPy array first.

cvxpy/expressions/leaf.py
PTNobel: Yes. Yes we should.

cvxpy/expressions/leaf.py
PTNobel: Hmmm so you think we should move the warning up too? Can you comment where you want it?

cvxpy/expressions/leaf.py
PTNobel: I think that we will break some of the other code and we want to allow mixing attributes other than PSD/NSD and sparse I think

cvxpy/expressions/leaf.py
maxschaller: Agreed, let's move it to the `value` setter.

cvxpy/expressions/leaf.py
maxschaller: Hmm I see. Maybe we'll need to skip or check the shape differently.

cvxpy/expressions/leaf.py
PTNobel: Wait we have a type check first? If it isn't a known good type, then we error about the types.

PTNobel:
@maxschaller were you going to do the docs?

PTNobel:
Update: I need to figure out what's happening with `cvxpy.interface` to figure out the most maintainable approach here.

Transurgeon:
> Update: I need to figure out what's happening with `cvxpy.interface` to figure out the most maintainable approach here.

are we planning to support both ``sp.*_matrix`` and ``sp.*_array``? My feeling is yes.
But then we need to think about compatibility between the two. I feel like we can replicate the code for numpy arrays and matrices in the interface. (also we can probably remove any instance of ``np.matrix`` as it is officially deprecated).

PTNobel:
> are we planning to support both sp.*_matrix and sp.*_array? My feeling is yes.

My preference is *no*, we should only support `sp.coo_array`.

1. `coo_array` has nearly the same internal data structure as our API: I think this is really valuable for helping us be flexible on changes to our implementation as we still finish it. 
2. `coo_array` is the only nD sparse format in scipy so far.

Transurgeon:
> My preference is _no_, we should only support `sp.coo_array`.
> 
> 1. `coo_array` has nearly the same internal data structure as our API: I think this is really valuable for helping us be flexible on changes to our implementation as we still finish it.
> 2. `coo_array` is the only nD sparse format in scipy so far and I think we should be nd first.

Just to clarify, I meant from the users perspective, since we currently support ``*_matrix`` as sparse constants, we would need to atleast raise a warning before deprecating it right? Otherwise, I agree with you that internally we should replace all instances of ``*_matrix`` to ``*_array`` in the backend.

Transurgeon:
I think the next steps for this feature are the following:
1. add docs (I guess we can add a small section next to the sparsity attribute [docs](https://www.cvxpy.org/tutorial/constraints/index.html#sparsity-attribute))
2. figure out what's going on in cvxpy.interface, some WiP in #2663 , maybe we should merge that PR first
3. add more tests to this PR then merge

What are your thoughts?

maxschaller:
My suggestion for a small section next to the sparsity attribute [docs](https://www.cvxpy.org/tutorial/constraints/index.html#sparsity-attribute):

**Reading and writing the value of a variable or parameter with the sparsity attribute**

To avoid storing entries that are known to be zero, we provide the `.value_sparse` field, which stores only the nonzero entries as a `scipy.sparse.coo_array`. For details on this data structure, please see the [scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_array.html). When passing a value, it is expected to be a `scipy.sparse.coo_array`.

```
sparsity = ([0, 1, 2, 2], [0, 2, 1, 2])
P = cp.Parameter((3, 3), sparsity=sparsity)
P.value_sparse = coo_array(([1.3, 2.1, 0.7, 3.2], sparsity))
```

Similarly, the value of a sparse variable or parameter is read via `.value_sparse`.

```
X = cp.Variable((3, 3), sparsity=[(0, 1), (0, 2)])
prob = cp.Problem(cp.Minimize(cp.sum(X)), [...])
prob.solve()
print(X.value_sparse)
```

doc/source/tutorial/constraints/index.rst
```diff
@@ -145,11 +145,29 @@ as defined in `np.indices <https://numpy.org/doc/stable/reference/generated/nump
 
     prob = cp.Minimize(cp.norm(X) + cp.sum(X))
 
-.. warning::
+Reading and writing the value of a variable or parameter with the sparsity attribute:
```
Transurgeon, line 148:
```suggestion
Reading and writing the value of a sparse expression
----------------------------------------------------
```

doc/source/tutorial/constraints/index.rst
```diff
@@ -145,11 +145,29 @@ as defined in `np.indices <https://numpy.org/doc/stable/reference/generated/nump
...
 
-    The sparsity attribute is not yet supported for setting the value of a variable or parameter.
-    In a future release, we plan to have a projection method for sparsity attributes using a custom datatype
-    based on PyTorch's `sparse_coo format <https://pytorch.org/docs/stable/generated/torch.sparse_coo_tensor.html>`_.
+To avoid storing entries that are known to be zero, we provide the ``.value_sparse`` field,
+which stores only the nonzero entries as a ``scipy.sparse.coo_array``.
+For details on this data structure, please see the `scipy documentation <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_array.html>`_.
+When passing a value, it is expected to be a ``scipy.sparse.coo_array``.
```
Transurgeon, line 153:
```suggestion
For details on this data structure, please see the `SciPy documentation <https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_array.html>`_.
When passing a sparse value, it is expected to be a ``scipy.sparse.coo_array``.
```

doc/source/tutorial/constraints/index.rst
```diff
@@ -145,11 +145,29 @@ as defined in `np.indices <https://numpy.org/doc/stable/reference/generated/nump
...
+
+.. code:: python
+
+    # Use the sparsity pattern for both the parameter and its assigned value
+    sparsity = ([0, 1, 2, 2], [0, 2, 1, 2])
+    P = cp.Parameter((3, 3), sparsity=sparsity)
+    P.value_sparse = coo_array(([1.3, 2.1, 0.7, 3.2], sparsity))
```
Transurgeon, line 160:
```suggestion
    data = [1.3, 2.1, 0.7, 3.2]
    P = cp.Parameter((3, 3), sparsity=sparsity)
    P.value_sparse = coo_array((data, sparsity))
```

Transurgeon (changes requested):
Just had a few suggestions for doc improvements @maxschaller , otherwise this PR looks ready to merge to me. I think the stuff about the scipy.sparse_array interface can maybe be done in another PR. 

Just to make sure, does the changes in this PR fix the existing issue with the sparsity attribute not being compatible with cvxpygen? 

Pinging @PTNobel and @SteveDiamond for your approval as well.

PTNobel:
I'm going to add sparse matrix support and then do a full review and hopefully merge.

maxschaller:
I wrote a couple more tests and fixed some bugs. Since the `value` getter and setter of the `Paramteter` class would override those of the `Leaf` class, they are commented out until @PTNobel will have a closer look.

cvxpy/expressions/constants/parameter.py
```diff
@@ -76,15 +76,15 @@ def is_constant(self) -> bool:
...
-        self._value = self._validate_value(val)
+    #@property
+    #def value(self):
+    #    """NumPy.ndarray or None: The numeric value of the parameter.
+    #    """
+    #    return self._value
+
+    #@value.setter
+    #def value(self, val):
+    #    self._value = self._validate_value(val)
```
PTNobel, line 87: Blocked on deciding whether to just delete this setter and getter

PTNobel (changes requested):
Just one thing that I consider blocking. I'm investigating what caused this issue.

PTNobel (approved):

SteveDiamond (approved):
lgtm

Transurgeon (approved):
LGTM, we can merge after addressing Steven's comment above. 

Quick question though, don't we need to add changes to the matrix [interface](https://github.com/cvxpy/cvxpy/tree/master/cvxpy/interface) too?

PTNobel:
Quick question though, don't we need to add changes to the matrix interface too?

I believe that was only necessary if we wanted to get rid of the `sparse_path` parameter. But I'm convinced it's worth keeping.

PR #2637, opened 2024-11-17
adding variable bounds to MOSEK

Transurgeon:
## Description
This PR is work in progress. I have outlined a bit of what I envision needs to be added for this feature, but haven't gotten to testing if it works or not just yet (efficiency is also another question, but shouldn't be too bad since we are using masks).
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -257,7 +257,11 @@ def apply(self, problem):
             else:
                 data['A_bar_data'] = []
                 data['c_bar_data'] = []
-
+        if problem.x._bounds:
+            lower_bounds = problem.x._bounds[0]
+            data['lb'] = lower_bounds
+            upper_bounds = problem.x._bounds[1]
+            data['ub'] = upper_bounds
```
rileyjmurray, line 264: @Transurgeon this won't work in its current form. The issue arises if we take the ``else`` branch of the second if-else statement, in which case the entries in ``data`` should be in terms of the dual problem rather than the primal problem. 

There are a number of potential paths forward to get things to work. One option is to only use this logic for mixed-integer problems. Another option is to force the ``Slacks`` reduction in the presence of bound constraints. A third option (the most comprehensive) would be to change the ``Dualize.apply`` reduction so it has special logic for bound constraints.

If you take the second option then some changes will be needed in the``MOSEK(ConicSolver).invert`` method. The third option has two aspects to it: correctly dualizing primal variable bounds (necessary for correctness and very doable) and figuring out which constraints on dual variables can be mapped to simple bound constraints (harder, optional, useful).

Transurgeon: ohhh I see, thank you for your detailed comment. Steven had mentioned that it was complicated to add variable bounds and I wasn't sure why, but now it makes sense. Also, I would like this to work for all problems, so let's skip option one.
I think option 2 could be the easiest, but I have a feeling (and also think you would agree) that its not ideal to change existing behavior to introduce these variable bounds to Mosek.
So I would like to tackle option 3. Is there any pointers you could give in terms of correctly dualizing primal variable bounds? I am not familiar with the whole dualization logic. Also I am not sure what you mean by figuring out the dual variable constraints which are mapped to simple bound constraints (what constitutes a "simple" bound)? An example on a simple problem would be great!

rileyjmurray: It's great to see your enthusiasm!

This is fundamentally a matter of conic duality. The best reference for cone programming is the MOSEK Modeling Cookbook.

Transurgeon:
some updates on this. We decided it would be better to have the feature only work for MIPs for now.
This could avoid having to dualize the bounds. Also, while trying to pass the bounds information, I noticed that the functions ``extract_lower/upper_bounds`` located in ``matrix_stuffing.py`` would sometimes erroneously think that the variables in question had no bounds. I will investigate further why this is happening.

Good news though, the logic for creating mosek boundkeys seems to work.

SteveDiamond:
@Transurgeon I did some work on this and got all the tests passing, as well as a new test that uses variable bounds to communicate constraints.

Transurgeon (commented):
I think this looks good, could you approve yourself since I made the PR originally?

cvxpy/tests/test_conic_solvers.py
```diff
@@ -433,40 +433,6 @@ def setUp(self) -> None:
         self.B = cp.Variable((2, 2), name='B')
         self.C = cp.Variable((3, 2), name='C')
 
-    def test_clarabel_parameter_update(self) -> None:
```
SteveDiamond, line 436: Not sure why this disappeared

SteveDiamond (approved):

PR #2654, opened 2024-12-13
removing some test deprecation warnings

Transurgeon:
## Description
This PR shortens the output of running ``pytest`` and is a small QoL improvement for developers.
Example of output caused by the tests in ``test_KKT.py``.
```bash
cvxpy/tests/test_KKT.py::TestKKT_QPs::test_qp_0
  /opt/anaconda3/envs/cvxpy_env/lib/python3.12/unittest/case.py:690: DeprecationWarning: It is deprecated to return a value that is not None from a test case (<bound method TestKKT_QPs.test_qp_0 of <cvxpy.tests.test_KKT.TestKKT_QPs testMethod=test_qp_0>>)
    return self.run(*args, **kwds)

cvxpy/tests/test_KKT.py::TestKKT_SOCPs::test_socp_0
  /opt/anaconda3/envs/cvxpy_env/lib/python3.12/unittest/case.py:690: DeprecationWarning: It is deprecated to return a value that is not None from a test case (<bound method TestKKT_SOCPs.test_socp_0 of <cvxpy.tests.test_KKT.TestKKT_SOCPs testMethod=test_socp_0>>)
    return self.run(*args, **kwds)
```
Also removed tests for the ``np.matrix`` interface since there were also deprecation warnings there from NumPy.
Finally, there is one last fix on a missed internal usage of reshape with explicit order.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/dgp2dcp/canonicalizers/mulexpression_canon.py
+++ b/cvxpy/reductions/dgp2dcp/canonicalizers/mulexpression_canon.py
@@ -37,5 +37,5 @@ def mulexpression_canon(expr, args):
         rows.append(row)
     mat = bmat(rows)
     if mat.shape != expr.shape:
-        mat = reshape(mat, expr.shape)
+        mat = reshape(mat, expr.shape, order='F')
     return mat, []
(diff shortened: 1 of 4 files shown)


Transurgeon:
@aryamanjeendgar could you confirm that removing ``return sth`` from the tests doesn't affect the result? My understanding was that the methods like ``check_...`` are already testing what is required.

SteveDiamond (approved):

PR #2658, opened 2024-12-22
Updating scipy and numpy versions for python < 3.12

Transurgeon:
## Description
CI is failing on master for python 3.10 and 3.9 on MacOS. The error is odd because the tests crashes without any error messages on python's side. I did a bit of digging and found that the most likely solver that is causing this is [highspy](https://pypi.org/project/highspy/) which new version 1.9.0 was released two days ago (December 20th, 2024).

Update: instead of pinning highspy to 1.8, updating the scipy and numpy versions somehow fixes the issue.
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/continuous_integration/install_dependencies.sh
+++ b/continuous_integration/install_dependencies.sh
@@ -13,7 +13,7 @@ conda install mkl pip pytest pytest-cov hypothesis openblas "setuptools>65.5.1"
 
 if [[ "$PYTHON_VERSION" != "3.13" ]]; then
   conda install ecos scs cvxopt proxsuite daqp
-  python -m pip install coptpy==7.1.7 gurobipy piqp clarabel osqp highspy==1.8.1
+  python -m pip install coptpy==7.1.7 gurobipy piqp clarabel osqp highspy
 else
   # only install the essential solvers for Python 3.13.
   conda install scs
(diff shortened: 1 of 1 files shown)


Transurgeon:
also, we can probably remove Python 3.9 from the CI. We will most likely make a new minor release after it becomes end of life (october 2025).

SteveDiamond (approved):

PTNobel:
> also, we can probably remove Python 3.9 from the CI. We will most likely make a new minor release after it becomes end of life (october 2025).

I don't follow this logic. I'm sure we'll make a release in the next 10 months...

Transurgeon:
> > also, we can probably remove Python 3.9 from the CI. We will most likely make a new minor release after it becomes end of life (october 2025).
> 
> I don't follow this logic. I'm sure we'll make a release in the next 10 months...

ok no worries, I think we should hold on for this PR then, I'll open an issue on highspy for them to have a look.

Transurgeon:
@SteveDiamond @PTNobel I was playing around with the sparse_arrays interface in #2663 and magically the CI was passing by upgrading the numpy/scipy versions for older versions of python (basically 3.9, 3.10 and 3.11).
Looking at scipy's [supported versions](https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy) (see image below) it seems like a fine change to me. What do you think?

image

PR #2667, opened 2025-01-09
Allow multiple attributes

Transurgeon:
## Description
This PR aims to remove the error being raised when declaring more than one attribute for cvxpy expressions.
The tradeoff for being able to declare such expressions is that we disable projections onto the sets defined by the multiple attributes.
Issue link (if applicable):
#566 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel:
What's the plan for projecting onto multiple attributes?

SteveDiamond:
> What's the plan for projecting onto multiple attributes?

To start, projection will be disabled when there are multiple attributes. In the long run there may be a few combinations of attributes that are still worth projecting onto.

cvxpy/expressions/leaf.py
```diff
@@ -372,8 +371,9 @@ def project(self, val, sparse_path=False):
         numeric type
             The value rounded to the attribute type.
         """
-        # Only one attribute can be active at once (besides real,
-        # nonpos/nonneg, and bool/int).
+        # Skip the projection operation for more than one attribute
+        if self.num_attributes > 1:
+            return val
```
PTNobel, line 376: This should be below the complex check

PTNobel (commented):
Looks really good

cvxpy/expressions/leaf.py
Transurgeon: yea agreed. sorry for the late reply.

Transurgeon:
ok I had another look at this and was a bit more thoughtful with tests, cases like sparse integer, sparse bounded, and integer bounded variables all seem to work.

One thing that concerns me is people might want to pass sparse values as bounds (or only the nonzeros), so we will need to add a bit of code to cover that. But maybe we can do that in another PR if this seems fine for now?

SteveDiamond:
> ok I had another look at this and was a bit more thoughtful with tests, cases like sparse integer, sparse bounded, and integer bounded variables all seem to work.
> 
> One thing that concerns me is people might want to pass sparse values as bounds (or only the nonzeros), so we will need to add a bit of code to cover that. But maybe we can do that in another PR if this seems fine for now?

Yeah, we can worry about sparse bounds later.

SteveDiamond (commented):
We also need tests for Parameters with multiple attributes.

SteveDiamond:
I added tests for parameters with multiple attributes. These tests exposed an issue where parameter input validation doesn't work for multiple attributes. The issue is too complex to fix in this MR.

Transurgeon:
> I added tests for parameters with multiple attributes. These tests exposed an issue where parameter input validation doesn't work for multiple attributes. The issue is too complex to fix in this MR.

ok sounds good, should we add some expected to fail decorators to those tests? And add TODOs for later?

Transurgeon (commented):
LGTM, should we approve and merge this? We can iterate on this for the other things we had in mind. 
@PTNobel it would be nice to have your review on this as well, if you have time.

cvxpy/expressions/leaf.py
```diff
@@ -147,17 +147,16 @@ def __init__(
...
-            true_attr -= 1
-        if true_attr > 1:
-            raise ValueError("Cannot set more than one special attribute in %s."
-                             % self.__class__.__name__)
+        # count number of attributes
+        self.num_attributes = sum(1 for k, v in self.attributes.items() if v)
+        dim_reducing_attr = ['diag', 'symmetric', 'PSD', 'NSD', 'hermitian', 'sparsity']
+        if sum(1 for k in dim_reducing_attr if self.attributes[k]) > 1:
+            raise ValueError(
+                "A CVXPY Variable cannot have more than one of the following attributes be true: "
```
PTNobel, line 155:
```suggestion
                "A CVXPY Variable cannot have more than one of the following attributes: "
```

cvxpy/tests/test_attributes.py
```diff
@@ -205,6 +210,167 @@ def test_variable_bounds(self):
...
+        target = np.zeros((2,2))
+        prob = cp.Problem(cp.Minimize(cp.sum(x)), [])
+        prob.solve()
+        assert np.allclose(prob.value, 0)
+        assert np.allclose(x.value, target)
+    
+    def test_sparse_symmetric_variable(self) -> None:
+        with pytest.raises(
+            ValueError, 
+            match="A CVXPY Variable cannot have more than one of the following attributes be true"
```
PTNobel, line 270:
```suggestion
            match="A CVXPY Variable cannot have more than one of the following attributes"
```

cvxpy/tests/test_attributes.py
```diff
@@ -205,6 +210,167 @@ def test_variable_bounds(self):
...
+        p.value = -np.ones((2, 2))
+        x = cp.Variable(shape=(2, 2))
+        prob = cp.Problem(cp.Minimize(cp.sum(x)), [x >= p])
+        prob.solve()
+        assert np.allclose(x.value, -np.ones((2, 2)))
+
+        # TODO make parameter validation work for multiple attributes.
+        # # Invalid assignment should raise ValueError
+        # with pytest.raises(ValueError, match="Parameter value must be nonpositive."):
+        #     p.value = np.ones((2, 2))
```
PTNobel, line 302: Is there a plan to uncomment these tests?

PTNobel (approved):
Overall looks good. I'm curious for the story around the commented out tests.

cvxpy/tests/test_attributes.py
Transurgeon: the issue is that parameter validation is not working for multiple attributes. We will come back to this in another PR.

Transurgeon:
made a PR to fix the failing (and unrelated) tests #2784 . 
also based on a new comment in discussion #2664, I think we should also test if sparse complex variables work. However, I can't really think of a good test. copilot suggested something but I am not sure its making any sense.

SteveDiamond:
> made a PR to fix the failing (and unrelated) test #2784 . also based on a new comment in discussion #2664, I think we should also test if sparse complex variables work. However, I can't really think of a good test. copilot suggested something but I am not sure its making any sense.

I added complex + sparse tests. They actually caught a bug.

cvxpy/expressions/leaf.py
```diff
@@ -467,7 +467,7 @@ def value(self) -> Optional[np.ndarray]:
                           ' Use `.value_sparse` instead', RuntimeWarning, 1)
             if self._value is None:
                 return None
-            val = np.zeros(self.shape)
+            val = np.zeros(self.shape, dtype=self._value.dtype)
```
Transurgeon, line 470: oh really nice catch!

PR #2671, opened 2025-01-13
Added workflow for installing and testing optional conic solvers

govindchari:
## Description
Added CI workflow for installing and testing optional conic solvers

## Type of change
- [x] Other (CI)
 
- [ ] Write unittests.


Transurgeon (commented):
@govindchari thanks for doing this, I really like the idea and its making me think how we should proceed. This change could allow us to completely separate the optional solvers from the default ones in the CI (which is sometimes problematic because it could fail before the build wheels step). Could you have a look at the current failures? For the MOSEK missing license, maybe you can add the check [here](https://github.com/cvxpy/cvxpy/blob/6bc1a1ef838e3bf2a8ea2142fa04457551a8c89b/continuous_integration/install_dependencies.sh#L60) before installing it?

PR #2684, opened 2025-01-22
Introduce “make pre-commit"

tschm:
## Description
Here we simplify the application of pre-commit hooks via ‘make pre-commit’. This will install the hooks and apply them to all files. If a user run this on their side they can be sure the pre-commit workflow is not complaining about formatting, etc.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


Transurgeon:
I am closing this since we are not sure if we want to move forward with it or not.

PR #2689, opened 2025-01-28
Support Scipy sparse arrays (sparray) while still supporting spmatrix

dschult:
## Description
Prepares for SciPy changes to sparse.

This PR implements a set of 'pass 1' changes as described in the [migration guide from SciPy spmatrix to sparray](https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html). The spmatrix interface is not deprecated, but will eventually be. These changes allow the code to work for both spmatrix and sparray apis.

Basically, that means changing:

`* -> @` only for matmul involving scipy sparse. Scalar multiplication of sparse uses *.
`spdiags -> diags`. This change still creates sparse matrix, but using the more modern array-api syntax for spmatrix. 
`A.H -> A.T.conjugate()`
I also added `np.array(...)` inside `cvxpy.Constant` in 3 places in `test_dgp2dcp.py` where warnings are currently generated indicating to do so.

Sorry for the length of the diff here. Each change is hopefully straightforward. If this should be split into parts, let me know.

**Versions**: sparrays were first introduced in Scipy 1.8 and were effectively full featured in 1.14. So anyone using them will be using those Scipy versions or later (current release is 1.15). I believe anyone using spmatrix should be fine even with older scipy, but I have not checked how far back. It might make sense to bump the required version for Scipy. Given that `numpy >= 2.0.0` is required  in `pyproject.toml`, I would recommend `scipy >= 1.14`.  But you will know better than me on this question.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


PTNobel:
> Given that numpy >= 2.0.0 is required in pyproject.toml, I would recommend scipy >= 1.14

Can you explain a bit more about why you suggest this? Does NumPy 2.0 require a recent SciPy? Do old scipy not support NumPy 2.0?

dschult:
Thanks for catching that `_getnnz` doesn't work for Scipy 1.9.  
Unfortunately the switch from `_getnnz` to `getnnz` won't work for sparrays.
So I will find a workaround for that. We've had other packages rely on `getnnz(axis=...)` too.

I've pushed that. Hopefully it is OK...

Transurgeon:
> I've pushed that. Hopefully it is OK...

thanks a lot for this PR @dschult ! We've decided it might be better to bump the SciPy version in our builds and CI as you suggested. I made a PR for it.

dschult:
I've reverted the extra code to replace using `_getnnz` based on comments here and the change to supported version of scipy. I added a comment that we can replace using a private method `_getnnz` with the public method `count_nonzero` for scipy v1.15+ once that is the min supported version. `count_nonzero` exists before 1.15 but supports the `axis` keyword arg starting in 1.15.

Transurgeon (approved):
We've merged the PR updating our scipy dependency.
LGTM and thanks for doing this @dschult !

Transurgeon:
@PTNobel is this good to merge?

PTNobel:
I'll do my review tonight. Thanks Will.

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
```diff
@@ -285,7 +285,8 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
                 nlinrow = dims[s.EQ_DIM] + dims[s.LEQ_DIM]
                 nlincol = A.shape[1]
 
-                diag = sp.spdiags(np.ones(nconedim), -nlinrow, A.shape[0], nconedim)
+                diag = sp.diags(np.ones(nconedim), offsets=-nlinrow,
+                                shape=(A.shape[0], nconedim))
                 A = sp.csc_matrix(sp.hstack([A, diag]))
```
PTNobel, line 290: Should this be a `.tocsc` instead of an explicit matrix?

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
```diff
@@ -313,7 +314,8 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
                     nlinrow += sum(dims[s.SOC_DIM])
                 nlincol = A.shape[1]
 
-                diag = sp.spdiags(np.ones(nexpconedim), -nlinrow, A.shape[0], nexpconedim)
+                diag = sp.diags(np.ones(nexpconedim), offsets=-nlinrow,
+                                shape=(A.shape[0], nexpconedim))
                 A = sp.csc_matrix(sp.hstack([A, diag]))
```
PTNobel, line 319: Ditto

PTNobel (approved):
Overall looks good. We're still using csc_matrix explicitly a lot, so maybe I'm missing some of the point of this PR. Are we trying to get away from matrices now, or is this just laying the groundwork for the future?

dschult:
Thanks for the question -- sorry I wasn't clear earlier. This PR represents a "pass 1" of a 2-pass process of migration to sparray (as described in the [migration to sparray guide](https://docs.scipy.org/doc/scipy/reference/sparse.migration_to_sparray.html)). 

Pass 1 updates existing code to work with the latest spmatrix interface, which is also the sparray interface. So this enables the code to work with sparrays wherever it currently uses spmatrix. I haven't worked through how much cvxpy relies on sparse user input as opposed to creating sparse within the package. But this PR should allow all sparse inputs to be sparray and it should all work (the outputs would still be spmatrix in most cases).  

Pass 2 updates all internal use/creation to sparray.  That's where all the `csr_matrix` -> `csr_array` changes are made. And pass 2 has a deeper issue to be solved: should sparse return values be spmatrix or sparray? For most functions, there is a sparse input as well as the sparse output. So we can return sparray/spmatrix dependent on what was input to the function.  But some packages have chosen to just always return sparray with the user responsible for converting with e.g. `csr_matrix` if the user wants a spmatrix.  

For functions that don't have a sparse input, there are multiple approaches depending on how much users manipulate the sparse result of functions on their own, and how quickly the package wants the migration to be. Some packages are just shifting directly to returning sparray and ask users to wrap it in `csr_matrix` if desired. Others create two separate functions for spmatrix and sparray (e.g. `scipy.sparse.eye` and `scipy.sparse.eye_array`). And others provide a keyword argument to indicate which return type (e.g. `scipy.io.mmread(filename, spmatrix=False)`).

Looking at sparse return values will be part of the pass-2 PR. The timing of pass 2 could be: "soon" (I can put one together if that helps), "delayed", or "not-planned". The point is that after pass 1 your code can work with both. So that's important. The pass 2 shift to using sparray internally is about moving toward the eventual deprecation of spmatrix. That doesn't need to be right now, but SciPy has resources allocated to helping packages do that as a way to help speed up the shift.

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
dschult: Yes -- though even better than `tocsc()` would be to change the line starting `A = ` to use the `format="csc"` keyword arg to `sp.hstack`. I'll make that change.

dschult:
I made the suggested change and I rebased on master after #2690 
So maybe the CI will be less red. I think this is ready to merge unless something else comes up on CI.

PR #2694, opened 2025-02-02
Assigning to and reading from sparse leaves

maxschaller:
## Description
This PR fixes two bugs. First, when assigning a sparse value, any order of nonzero indices is accepted, as long as the set of nonzero indices is correct. Previously, the sparsity pattern was rejected as invalid when the order of nonzero indices was not identical to that used when the `Leaf` was constructed. Second, when reading a sparse value and it is `None`, `None` will be returned. Previously, there was an error when attempting to read `_value.data`.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -161,13 +161,17 @@ class Leaf(expression.Expression):
         self.args = []
         self.bounds = self._ensure_valid_bounds(bounds)
 
-    def _validate_indices(self, indices: list[tuple[int]] | tuple[np.ndarray]) -> None:
+    def _validate_indices(self, indices: list[tuple[int]] | tuple[np.ndarray]) -> tuple[np.ndarray]:
         """
         Validate the sparsity pattern for a leaf node.
 
         Parameters:
         indices: List or tuple of indices indicating the positions of non-zero elements.
         """
+        if self._shape == ():
+            if indices != []:
+                raise ValueError("Indices should have 0 dimensions.")
+            return []
         validator = sp.coo_array((np.empty(len(indices[0])), indices), shape=self._shape)
         validator.sum_duplicates()
         return validator.coords
@@ -427,8 +431,10 @@ class Leaf(expression.Expression):
 
     # Getter and setter for parameter value.
     def save_value(self, val, sparse_path=False) -> None:
-        if self.sparse_idx and not sparse_path:
+        if self.sparse_idx is not None and not sparse_path:
             self._value = sp.coo_array((val[self.sparse_idx], self.sparse_idx), shape=self.shape)
+        elif self.sparse_idx is not None and sparse_path:
+            self._value = val.data
         else:
             self._value = val
 
@@ -456,7 +462,12 @@ class Leaf(expression.Expression):
     @property
     def value_sparse(self) -> Optional[...]:
         """The numeric value of the expression if it is a sparse variable."""
-        return self._value
+        if self._value is None:
+            return None
+        if isinstance(self._value, np.ndarray):
+            return sp.coo_array((self._value, self.sparse_idx), shape=self.shape)
+        else:
+            return self._value
 
     @value_sparse.setter
     def value_sparse(self, val) -> None:
@@ -468,8 +479,7 @@ class Leaf(expression.Expression):
                 raise ValueError(
                     'Invalid type for assigning value_sparse.'
                     'Try using `'
-                    'expr.value_sparse = scipy.sparse.coo_array((values, expr.sparse_idx))'
-                    '` instead.')
+                    'expr.value_sparse = scipy.sparse.coo_array((values, expr instead.')
             raise ValueError(
                 'Invalid type for assigning value_sparse.'
                 f'Recieved: {type(val)} Expected scipy.sparse.coo_array.'
@@ -508,7 +518,8 @@ class Leaf(expression.Expression):
             if sparse_path:
                 val.sum_duplicates()
                 coords_val = get_coords(val)
-                if coords_val != self.sparse_idx:
+                if len(coords_val) != len(self.sparse_idx) or \
+                     any((a != b).any() for a, b in zip(coords_val, self.sparse_idx)):
                     raise ValueError(
                         'Invalid sparsity pattern %s for %s value.' %
                         (get_coords(val), self.__class__.__name__)
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -168,7 +168,7 @@ class Leaf(expression.Expression):
         Parameters:
         indices: List or tuple of indices indicating the positions of non-zero elements.
         """
-        validator = sp.coo_array((np.empty(len(indices[0]), dtype=np.void), indices))
+        validator = sp.coo_array((np.empty(len(indices[0])), indices))
         validator.sum_duplicates()
         return validator.coords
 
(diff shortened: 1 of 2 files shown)


cvxpy/expressions/leaf.py
```diff
@@ -513,11 +515,14 @@ def _validate_value(self, val, sparse_path=False):
...
                     (intf.shape(val), self.__class__.__name__)
                 )
-            if sparse_path and (np.array(get_coords(val)) != np.array(self.sparse_idx)).any():
-                raise ValueError(
-                    'Invalid sparsity pattern %s for %s value.' %
-                    (get_coords(val), self.__class__.__name__)
-                )
+            if sparse_path:
+                coords_val = set(coord for coord in zip(*get_coords(val)))
+                coords_sparse_idx = set(coord for coord in zip(*self.sparse_idx))
```
PTNobel, line 520: Don't we need to reorder the entries of values if the orders differ?

cvxpy/tests/test_attributes.py
```diff
@@ -81,6 +81,11 @@ def test_sparsity_assign_value(self):
         prob.solve()
         assert np.allclose(X.value, z)
         
+        sparsity_rotated = [(1, 2, 2, 0), (2, 2, 1, 0)]
+        A.value_sparse = sp.coo_array((-np.ones(4), sparsity_rotated))
```
PTNobel, line 85: Please add a test with nonuniform value

PTNobel (commented):
First fix looks good, I'm less sure about the second

cvxpy/tests/test_attributes.py
maxschaller: Done

cvxpy/tests/test_attributes.py
```diff
@@ -81,6 +81,12 @@ def test_sparsity_assign_value(self):
         prob.solve()
         assert np.allclose(X.value, z)
         
+        z = np.zeros((3, 3))
+        z[A.sparse_idx] = [-1, -2, -3, -4]
+        A.value_sparse = sp.coo_array(([-1, -3, -2, -4], [(0, 1, 2, 2), (0, 2, 1, 2)]))
```
PTNobel, line 86: I think this should raise an error because the coordinates are in a different order and I don't want to commit to sorting them.

maxschaller: The data and coordinates are in a consistent order. Isn't the `coo_array` equivalent to

`sp.coo_array(([-1, -2, -3, -4], [(0, 2, 1, 2), (0, 1, 2, 2)]))`?

If assigning two equivalent sparse arrays (identical when converted to dense) to parameter `A` gives different solutions `X`, then there is a bug that we need to fix (and this test revealed the bug).

Maybe you had something else in mind - which behavior do you suggest? After a parameter is specified with sparsity `pattern`, any value assignment should follow the same order of coordinates?

PTNobel: > consistent order

I'm not certain what this means

> any value assignment should pattern without any re-ordering of coordinates

Correct. if we remove your change in this PR to the error, this line errors saying that the coordinates are invalid. We could give a better error message that makes it clear that the order is wrong, but I think we should raise an error.

Otherwise, we are committing to sorting the coordinates, and I don't think we should do that.

PTNobel:
> Previously, the sparsity pattern was rejected as invalid when the order of nonzero indices was not identical to that used when the Leaf was constructed.

I think this is correct behavior, and I  think that's part of our confusion.

maxschaller:
I see your point that you prefer not to commit to sorting coordinates. However, I think this is not very user-centric. It basically breaks an existing `cvxpygen` test. There the sparsity pattern of an MPC dynamics matrix is set via a list comprehension (traversing the main diagonal and some off-diagonal). Then, the sparse value is assigned by writing down the matrix as an `ndarray`, performing some dense calculations, and ultimately casting it as a `coo_array`. The order of nonzeros is different between the `Parameter` and its assgined `value_sparse`, and things break.

maxschaller:
Just thinking out loud: How about using `project` to sort the coordinates?

PTNobel:
Proposal, let's just require coordinates all be in the "Canonical form" for coo arrays. SciPy then does the sorting for us. I pushed a commit to do this change.

cvxpy/expressions/leaf.py
```diff
@@ -161,26 +161,20 @@ def __init__(
...
-        if np.any(indices < 0) or np.any(indices >= np.array(self._shape).reshape(-1, 1)):
-            raise ValueError(
-                f"Indices are out of bounds for expression with shape {self._shape}.")
-        return tuple(indices)
+        if self._shape == ():
+            if indices != []:
+                raise ValueError("Indices should have 0 dimensions.")
+            return []
+        validator = sp.coo_array((np.empty(len(indices[0])), indices), shape=self._shape)
+        validator.sum_duplicates()
```
Transurgeon, line 176: I am not sure what this is doing and why it's necessary

cvxpy/expressions/leaf.py
```diff
@@ -476,8 +479,7 @@ def value_sparse(self, val) -> None:
                 raise ValueError(
                     'Invalid type for assigning value_sparse.'
                     'Try using `'
-                    'expr.value_sparse = scipy.sparse.coo_array((values, expr.sparse_idx))'
-                    '` instead.')
+                    'expr.value_sparse = scipy.sparse.coo_array((values, expr instead.')
```
Transurgeon, line 482:
```suggestion
                    'expr.value_sparse = scipy.sparse.coo_array((values, expr)) instead.')
```

cvxpy/expressions/leaf.py
```diff
@@ -464,7 +462,12 @@ def value(self, val) -> None:
     @property
     def value_sparse(self) -> Optional[...]:
         """The numeric value of the expression if it is a sparse variable."""
-        return self._value
+        if self._value is None:
+            return None
+        if isinstance(self._value, np.ndarray):
+            return sp.coo_array((self._value, self.sparse_idx), shape=self.shape)
```
Transurgeon, line 468: so if my understanding is correct we now allow ``value_sparse`` to be a dense array of data. Then we would reconstruct the sparse.coo_array here.

cvxpy/expressions/leaf.py
```diff
@@ -513,11 +515,15 @@ def _validate_value(self, val, sparse_path=False):
...
-            if sparse_path and (np.array(get_coords(val)) != np.array(self.sparse_idx)).any():
-                raise ValueError(
-                    'Invalid sparsity pattern %s for %s value.' %
-                    (get_coords(val), self.__class__.__name__)
-                )
+            if sparse_path:
+                val.sum_duplicates()
+                coords_val = get_coords(val)
+                if len(coords_val) != len(self.sparse_idx) or \
+                     any((a != b).any() for a, b in zip(coords_val, self.sparse_idx)):
```
Transurgeon, line 522: is this essentially ensuring that ``coords_val`` is equal to ``self.sparse_idx``?

Transurgeon (approved):
Changes all look good! I don't fully understand what you guys mean by the "consistent order" of the indices and why does sorting get involved? Some clarifications could be helpful. 
Otherwise had some questions for my personal understanding, but nothing to say on the changes you have made.

cvxpy/expressions/leaf.py
PTNobel: I'll add some comments

cvxpy/expressions/leaf.py
PTNobel: `_value` can be just an array with the non zero entries.

cvxpy/expressions/leaf.py
PTNobel: Exactly! Makes sure they are defining the same coordinates

rileyjmurray (approved):
No red flags from what I can tell. Plenty of tests. I support merging as-is.

PR #2695, opened 2025-02-02
Improve type hinting for expressions

smachen42:
## Description
Add type hints to several `Expression` methods.

These type hints should reduce the number of instances where the mypy static type checker incorrectly raises type warnings, for example in this snippet
```
import cvxpy as cp

x = cp.Variable()
prob = cp.Problem(objective=cp.Minimize(x), constraints=[x == 1])
```
one would get the warning `Expected type 'list[Constraint] | None', got 'list[bool]' instead`.

In this case, we can fix the static analysis by adding `Equality` as the return type hint of the `Expression.__eq__` method. This PR implements this simple change and several other similar changes in the `Expression` class.

[Link to short discord discussion.](https://discordapp.com/channels/845323972962549790/1335428795192184832/1335428795192184832)

## Type of change
- [ ] Bug fix

- [ ] Write unittests. **Figured it's probably not worth it for this change, but I'm happy to look into it if that's preferred. I'd probably have to add the [*pytest-mypy*](https://pypi.org/project/pytest-mypy/) plugin as a new test dependency.**
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression. **I noticed the `test_benchmarks.py` was removed.**

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -808,63 +808,63 @@ class Expression(u.Canonical):
     def __abs__(self):
         raise TypeError(__ABS_ERROR__)
 
-    def conj(self):
+    def conj(self) -> "Expression":
         """
         Equivalent to `cp.conj(self)`.
         """
         from cvxpy import conj
         return conj(self)
 
-    def conjugate(self):
+    def conjugate(self) -> "Expression":
         """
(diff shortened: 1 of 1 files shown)


PTNobel (approved):
LGTM.

Will merge once CI passes.

PR #2696, opened 2025-02-02
experiment with sp.csc_array for the SciPy canon backend

PTNobel:
## Description

This should probably be thrown out eventually, but I wanted to just poke around with this new array interface in scipy.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/canon_backend.py
+++ b/cvxpy/lin_ops/canon_backend.py
@@ -1142,7 +1142,7 @@ class SciPyCanonBackend(PythonCanonBackend):
         if is_param_free_lhs:
             reps = view.rows // lhs.shape[-1]
             if reps > 1:
-                stacked_lhs = (sp.kron(sp.eye(reps, format="csr"), lhs))
+                stacked_lhs = (sp.kron(sp.eye_array(reps, format="csr"), lhs))
             else:
                 stacked_lhs = lhs
 
@@ -1150,7 +1150,7 @@ class SciPyCanonBackend(PythonCanonBackend):
                 if p == 1:
                     return (stacked_lhs @ x).tocsr()
                 else:
-                    return ((sp.kron(sp.eye(p, format="csc"), stacked_lhs)) @ x).tocsc()
+                    return ((sp.kron(sp.eye_array(p, format="csc"), stacked_lhs)) @ x).tocsc()
         else:
             reps = view.rows // next(iter(lhs.values())).shape[-1]
             if reps > 1:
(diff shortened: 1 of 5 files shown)


Transurgeon:
I think we can keep these changes! It seems like we just need to update the tests to reflect these changes now.
I can try to fix some of them.
Although this 1d-slice one might be weird to fix.
```
NotImplementedError: We have not yet implemented 1D sparse slices; please index using explicit indices, e.g. `x[:, [0]]`
```

Transurgeon:
ok I changed a few instances of ``eye`` to ``eye_array`` which were causing issues.
However, we are still encountering errors because we are sometimes expecting the matrix to be 2d in the backend.
Two instances I noticed:
1. we index when selecting rows of a "tensor" like so: ``x[rows, :]``, which assumes that ``x`` is 2d.
2. we sometimes pass vectors to kron as (n,1) matrices. We are now getting errors because kron is expecting 2d matrices as input. 
@dschult I was wondering if you had any suggestions on how we should proceed? I have a feeling like maybe cvxpy did depend on the old behavior of sparse matrices (especially in regards to the dimension always being 2).

dschult:
Slicing of 1d objects is implemented in 1.15. There are workarounds (usually avoiding 1d) for previous versions. 

> 1) we index when selecting rows of a "tensor" like so: x[rows, :], which assumes that x is 2d.

You can index a 2D object and keep the 2D nature of the result using [advanced indexing](https://numpy.org/doc/2.2/user/basics.indexing.html#advanced-indexing) like the following (The `None`/`np.newaxis` index also works like in Numpy: `x[None, row, :]`.)
```python
x[[row], :]
x[:, [col]]
```

If you already have a 1D vector `v` that you want to make 2D, reshape it or use 1D indexing starting in v1.15:
`v.reshape((v.shape[0], 1))` or in v1.15: `v[None, :]` or `v[:, None]`.

> 2) we sometimes pass vectors to kron as (n,1) matrices. We are now getting errors because kron is expecting 2d matrices as input.

This is related to the previous item. If you have a vector `lhs` that you want to provide to `kron` you need to make it 2D:
```python
# numpy would be:
lhs = np.array([1,2,3])
result = np.kron(np.eye(3), lhs[:, None])

# so scipy v1.15 we can use:
lhs = csr_array([1,2,3])
result = np.kron(np.eye(3), lhs[:, None])
# before 1.15
result = np.kron(sp.eye_array(3), lhs.reshape((lhs.shape[0], 1))
```
Your sparse 1D vectors are probably coming from indexing. If so, the first bullet can avoid creating 1D only convert to 2D.

Transurgeon:
we can probably close this PR once #2710 gets merged

PR #2697, opened 2025-02-02
Updates for clarabel v0.10.0. 

goulart-paul:
Updates for clarabel v0.10.0. Adds warm start functionality to wrapper.

## Description
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
@@ -14,7 +14,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
-This interface borrows heavily from the one in scs_conif.py
 """
 import numpy as np
 import scipy.sparse as sp
@@ -87,6 +86,7 @@ def triu_to_full(upper_tri, n):
     full[np.tril_indices(n, k=-1)] /= np.sqrt(2)
     full[np.triu_indices(n, k=1)] /= np.sqrt(2)
     return np.reshape(full, n*n, order="F")
+
 
 def clarabel_psdvec_to_psdmat(vec: Expression, indices: np.ndarray) -> Expression:
     """
@@ -161,7 +161,7 @@ class CLARABEL(ConicSolver):
         import clarabel  # noqa F401
 
     def supports_quad_obj(self) -> bool:
-        """Clarabel supports quadratic objective with any combination 
+        """Clarabel supports quadratic objective with any combination
         of conic constraints.
         """
         return True
@@ -256,10 +256,12 @@ class CLARABEL(ConicSolver):
             return failure_solution(status, attr)
 
     @staticmethod
-    def parse_solver_opts(verbose, opts):
+    def parse_solver_opts(verbose, opts, settings=None):
         import clarabel
 
-        settings = clarabel.DefaultSettings()
+        if settings is None:
+            settings = clarabel.DefaultSettings()
+
         settings.verbose = verbose
 
         # use_quad_obj is only for canonicalization.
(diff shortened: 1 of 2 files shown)


PTNobel:
Should this be backported to 1.6.1 or just for CVXPY 1.7?

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
```diff
@@ -299,27 +300,55 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
-        def solve(_solver_opts):
+        def new_solver():
+
+            _settings = CLARABEL.parse_solver_opts(verbose, solver_opts)
+            _solver = clarabel.DefaultSolver(P, q, A, b, cones, _settings)
+            return _solver
+
+        def updated_solver():
+
+            if (not warm_start) or (solver_cache is None) or (self.name() not in solver_cache):
```
PTNobel, line 321: Where is solver_cache defined?

PTNobel (changes requested):
Can you add a test of this functionality? Just that we aren't erroring or anything rather than needin to test that we're avoiding the memory allocation.

goulart-paul:
> Should this be backported to 1.6.1 or just for CVXPY 1.7?

I believe it will work for both.   The only change in functionality is that the solver's python wrapper now exposes internal data updating functions in v0.10.0.    If those don't exist (tested via `hasattr(_solver, "update")`), then it should behave as before.

It would maybe be clearer to test directly against the version number, but I think that would require taking on `packaging` as a dependency.

goulart-paul:
> Should this be backported to 1.6.1 or just for CVXPY 1.7?

Should this be backported to 1.6.1 or just for CVXPY 1.7?

I believe it will work for both. The only change in functionality is that the solver's python wrapper now exposes internal data updating functions in v0.10.0. If those don't exist (tested via hasattr(_solver, "update")), then it should behave as before.

It would maybe be clearer to test directly against the version number, but I think that would require taking on packaging as a dependency.

goulart-paul:
> Can you add a test of this functionality? Just that we aren't erroring or anything rather than needin to test that we're avoiding the memory allocation.

It was unclear to me whether this was being tested already somehow.   Warmstarting is also supported by scs and osqp, but I couldn't find specific tests targeting warm start capabilities for those specifically, or for solvers generally.

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
goulart-paul: It is an argument to `solve_via_data`, which `updated_solver` is nested within.   The solver_cache is populated with a previously constructed solver just before `solve_via_data` exits.

It's my understanding that the `solver_cache` has no particular requirements about its content.   I think both SCS and OSQP save prior results for warm starting, and OSQP also holds the data used to construct its problem instance so it can compare the sparsity pattern of any update.    I didn't do either of those here because we don't use the prior solution as a warm start point, and sparsity matching is verified internally (erroring if there is a mismatch).    I don't think cvxpy can give problem data updates with sparsity that deviates from what was passed at problem construction though.

SteveDiamond:
> > Can you add a test of this functionality? Just that we aren't erroring or anything rather than needin to test that we're avoiding the memory allocation.
> 
> It was unclear to me whether this was being tested already somehow. Warmstarting is also supported by scs and osqp, but I couldn't find specific tests targeting warm start capabilities for those specifically, or for solvers generally.

Each solver that supports warm start has a test for it in cvxpy/tests/test_conic_solvers.py

Transurgeon:
@PTNobel is this good to merge? The changes seem fine with the new warmstart tests that Paul added

cvxpy/reductions/solvers/conic_solvers/clarabel_conif.py
PTNobel: That all sounds right to me

PTNobel (approved):

PR #1053, opened 2020-06-10
Adding reciprocal product as a new convex atom

angeris:
Adding in the (somewhat surprising!) new atom `inv_prod` equal to

```
(Π x)⁻¹
```

for some vector `x`, as per @akshayka 's recommendation.

PR #2702, opened 2025-02-03
Fixes `cp.inv_prod(cp.Variable()).curvature`

PTNobel:
## Description
This fixes the following broken snippet:
```python
x = cp.Variable()
t = cp.inv_prod(x)
assert t.curvature == 'CONSTANT'
```

I'm open to other ways to fix this and overall I'm not sure the implementation is correct at all:

1. Why are we summing the shape and not producting it together?

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/inv_prod.py
+++ b/cvxpy/atoms/inv_prod.py
@@ -35,4 +35,4 @@ def inv_prod(value):
 
         where :math:`n` is the length of :math:`x`.
     """
-    return power(inv_pos(geo_mean(value)), int(sum(value.shape)))
+    return power(inv_pos(geo_mean(value)), int(sum(value.shape) if len(value.shape) > 0 else 1))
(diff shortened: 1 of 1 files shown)


Transurgeon:
I found the original PR here: #1053 . 
My feeling is that the input is supposed to be a vector so summing the shape just takes the length of the vector?
But yea it is definitely weird.. since even something like shape=(n,1) seems to be wrong?

SteveDiamond:
I see. The product of the shape seems more correct.

PTNobel:
I think it should just `np.prod(value.shape)` and that will always be correct since `np.prod(()) == 1.0`

Transurgeon:
> I think it should just `np.prod(value.shape)` and that will always be correct since `np.prod(()) == 1.0`

yea makes sense.
Is inv_prod well-defined for matrices? Maybe we can add some assertions to ensure that the expression is a vector?

PTNobel:
> Is inv_prod well-defined for matrices? Maybe we can add some assertions to ensure that the expression is a vector?

Sure, it should do the exact same thing like sum would

Transurgeon (approved):
LGTM, nice fix to cover the scalar case!

PR #2703, opened 2025-02-05
BUGFIX: Adds new broadcast_to atom for nd broadcasting

Transurgeon:
## Description
Added some tests, but they aren't passing yet. Will tinker more and see what is the general behavior.
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -17,7 +17,7 @@ limitations under the License.
 import abc
 import warnings
 from functools import wraps
-from typing import List, Literal, Tuple, Union
+from typing import List, Literal, Optional, Tuple, Union
 
 import numpy as np
 
@@ -124,12 +124,12 @@ class Expression(u.Canonical):
 
     @property
     @abc.abstractmethod
-    def value(self):
-        """NumPy.ndarray or None : The numeric value of the expression.
+    def value(self) -> Optional[np.ndarray]:
+        """Returns: The numeric value of the expression.
         """
         raise NotImplementedError()
 
-    def _value_impl(self):
+    def _value_impl(self) -> Optional[np.ndarray]:
         """Implementation of .value.
         """
         return self.value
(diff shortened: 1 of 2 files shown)


cvxpy/expressions/expression.py
```diff
@@ -591,19 +591,13 @@ def broadcast(lh_expr: "Expression", rh_expr: "Expression"):
...
-            # Broadcast along dim 1.
-            if lh_expr.shape[1] == 1 and lh_expr.shape[1] < dims[1]:
-                lh_expr = lh_expr @ np.ones((1, dims[1]))
-            if rh_expr.shape[1] == 1 and rh_expr.shape[1] < dims[1]:
-                rh_expr = rh_expr @ np.ones((1, dims[1]))
+            output_shape = np.broadcast_shapes(lh_expr.shape, rh_expr.shape)
+            if lh_expr.shape != output_shape:
+                lh_expr = np.ones(np.prod(output_shape)/np.prod(lh_expr.shape)) @ lh_expr
+            if rh_expr.shape != output_shape:
+                rh_expr = np.ones(np.prod(output_shape)/np.prod(rh_expr.shape)) @ rh_expr
```
PTNobel, line 600: Does this correctly handle cases where we have multiple dimensions added in broadcasting?

e.g. if we're broadcasting (3, 3, 3) with (3,), is this correct?

Transurgeon: no I don't think so. I am still trying to fully figure out the behavior of the previous code. (which did mul or rmul depending on the axis being broadcast)

SteveDiamond: We may want to make a broadcast atom and linop rather than relying on multiplication to solve everything.

Transurgeon: oh yea that would definitely be the best solution.

Transurgeon:
closing as I had some issues with branches from the fork.

PR #2711, opened 2025-02-12
docs: add links to interactive marimo examples

akshayka:
## Description

This change adds links to interactive marimo examples for:

1. least squares
2. linear programs
3. quadratic programs
4. generic convex optimization (which points the reader to the DCP docs)
5. SDPs
6. minimum fuel optimal control
7. markowitz portfolio optimization

Examples 2, 3, 6, and 7 have interesting interactive components and visualizations that hopefully help the reader better understand the topic at hand.

Additionally, this change splits up "basic examples" into "basic examples" and "basic applications", to make the simpler application notebooks more discoverable. The SOCP example was moved into Advanced; it manually uses an `SOC` constraint, which feels very not basic to me.

## Type of change
- [ ] Bug fix

This is purely a docs change, so these are vacuously true.

- [x] Write unittests.

### DIFF
--- a/doc/source/examples/index.rst
+++ b/doc/source/examples/index.rst
@@ -13,31 +13,50 @@ These examples show many different ways to use CVXPY.
 
 There are also application-specific sections.
 
+* The :ref:`basic-applications` section shows real-world applications of
+  common optimization problems.
 * The :ref:`machine-learning` section is a tutorial on convex optimization in
   machine learning.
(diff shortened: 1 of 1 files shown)


Transurgeon (approved):
Looks good to me, thanks for doing this!

also I agree with moving the SOC example to advanced

SteveDiamond (approved):
Looks good!

PR #2714, opened 2025-02-14
Swap order of `passOptions` and `passModel` in HiGHS interface

tasseff:
## Description
Currently, when using the CVXPY interface to `highspy`, the message `Running HiGHS 1.9.0 (git hash: fa40bdf): Copyright (c) 2024 HiGHS under MIT licence terms` is logged to standard output even if `output_flag` is set to `False`. Here's a minimum working example:
```python
import cvxpy as cp

x = cp.Variable(nonneg=True)
problem = cp.Problem(cp.Minimize(x))
problem.solve(solver="HIGHS", output_flag=False)
```
This pull request swaps the order of `passOptions` and `passModel` to ensure that options are set prior to model instantiation, when the logging seems to be triggered.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
@@ -211,8 +211,8 @@ class HIGHS(ConicSolver):
             setattr(options, key, value)
 
         solver = hp.Highs()
-        solver.passModel(model)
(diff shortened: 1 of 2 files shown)


SteveDiamond (approved):
Good catch!

PR #2720, opened 2025-02-18
Fix failing optional solvers

govindchari:
## Description
After the sparse_array PR #2710 merged, qoco and ecos/ecos_bb unit tests started failing.

QOCO unit tests failed, since the python wrapper expects the row index and column pointer arrays to be int32, but with sparse_arrays they are int64. This PR uses the `safely_cast_index_arrays` function to cast these indices. Unfortunately, this function first appears in scipy v1.15.0, but cvxpy requires only scipy  >=1.11.0.

The ecos failures are due to the expectation that matrices passed to the ecos python wrapper are of type csc_matrix rather than csc_array. There are two potential solutions: open a PR in [ecos-python](https://github.com/embotech/ecos-python) to expect csc_array, but this might be an invasive change and break existing projects csc_matrix is passed into ecos-python. The second potential solution is to convert the csc_array into a csc_matrix before passing problem data to `ecos_conif.py`. I would love to get some feedback on which solution is preferred.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/ecos_bb_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/ecos_bb_conif.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from scipy import sparse
 
 import cvxpy.interface as intf
 import cvxpy.settings as s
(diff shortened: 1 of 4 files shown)


Transurgeon (approved):
changes look good, thanks for this fast fix @govindchari !
merging once CI is passing.

PR #2724, opened 2025-02-21
Adds new broadcast_to atom for nd broadcasting

Transurgeon:
## Description
This PR introduces the broadcast_to atom in hopes of fixing issues with broadcasting in higher dimensions. 
The broadcast_to atom is used internally inside of the expression class in its ``broadcast`` method.
This method is called in many places, including binary ops, basic ops and others.

The implementation of the atom follows other similar shape manipulation atoms such as vstack and concatenate. The idea is to form a set of indices which are reshaped to the expression's shape, then using numpy's API we are able to mimic the behavior of ``np.broadcast_to`` on cvxpy's tensor views in the backend. 

### CPP backend compatibility
In the #2728 discussion, I brought up some issues with the difference in behavior for the backends. Those issues were resolved as follows: 
- only call ``cp.broadcast_to`` when one of the operands has dim > 2, which ensures that the SCIPY backend is used.
- full broadcasting is supported in those cases, even the edge cases mentioned in the discussion above.
- in hopes of keeping the CPP backend behavior the same, both the SCIPY and CPP backends don't do full broadcasting when both operands are <= 2. However, this only omits some edge cases, as most practical scenarios are covered by existing code. 

### Improved error messages
Recent changes to the codebase (notably adding the cp.concatenate atom) showed that we are able to instantiate empty numpy arrays (using ``dtype=np.empty([])``) and then perform analogous operations on those arrays to get the error handling from numpy for free. This technique has been reused here to improve on the existing multiply atom error handling. 

Issue link (if applicable): #2739 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -232,21 +232,18 @@ class MulExpression(BinaryOperator):
 
 
 class multiply(MulExpression):
-    """ Multiplies two expressions elementwise.
-    """
+    """Multiplies two expressions elementwise."""
 
     def __init__(self, lh_expr, rh_expr) -> None:
         lh_expr, rh_expr = self.broadcast(lh_expr, rh_expr)
         super(multiply, self).__init__(lh_expr, rh_expr)
 
     def is_atom_log_log_convex(self) -> bool:
-        """Is the atom log-log convex?
-        """
+        """Is the atom log-log convex?"""
         return True
 
     def is_atom_log_log_concave(self) -> bool:
-        """Is the atom log-log concave?
-        """
+        """Is the atom log-log concave?"""
         return True
 
     def is_atom_quasiconvex(self) -> bool:
@@ -262,8 +259,7 @@ class multiply(MulExpression):
             arg.is_nonpos() for arg in self.args)
 
     def numeric(self, values):
-        """Multiplies the values elementwise.
-        """
+        """Multiplies the values elementwise."""
         if sp.issparse(values[0]):
             return values[0].multiply(values[1])
         elif sp.issparse(values[1]):
@@ -271,10 +267,13 @@ class multiply(MulExpression):
         else:
             return np.multiply(values[0], values[1])
 
+    def validate_arguments(self):
+        """Validate that the arguments are broadcastable."""
+        np.broadcast_shapes(self.args[0].shape, self.args[1].shape)
+
     def shape_from_args(self) -> Tuple[int, ...]:
-        """The sum of the argument dimensions - 1.
-        """
-        return u.shape.sum_shapes([arg.shape for arg in self.args])
+        """Call np.broadcast on multiply arguments."""
+        return np.broadcast_shapes(self.args[0].shape, self.args[1].shape)
 
     def is_psd(self) -> bool:
         """Is the expression a positive semidefinite matrix?
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -269,11 +269,10 @@ class multiply(MulExpression):
 
     def validate_arguments(self):
         """Validate that the arguments are broadcastable."""
-        if not self.args[0].is_scalar() and not self.args[1].is_scalar():
-            output_shape = np.broadcast_shapes(self.args[0].shape, self.args[1].shape)
-            if self.args[0].shape != output_shape and self.args[1].shape != output_shape:
-                raise ValueError("Cannot multiply expressions with dimensions %s and %s" %
-                                (self.args[0].shape, self.args[1].shape))
+        np.broadcast(
+            np.empty(self.args[0].shape, dtype=np.dtype([])),
+            np.empty(self.args[1].shape, dtype=np.dtype([]))
+        )
 
     def shape_from_args(self) -> Tuple[int, ...]:
         """Call np.broadcast on multiply arguments."""
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -269,10 +269,11 @@ class multiply(MulExpression):
 
     def validate_arguments(self):
         """Validate that the arguments are broadcastable."""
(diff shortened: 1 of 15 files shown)


cvxpy/tests/test_expressions.py
```diff
@@ -1713,3 +1711,61 @@ def test_nd_transpose(self) -> None:
...
+                                        ((15, 1, 5), (15, 3, 5)),
+                                        ((3, 5), (15, 3, 5)),
+                                        ((3, 1), (15, 3, 5))])
+    def test_nd_multiply_broadcast(self, shapes) -> None:
+        x = cp.Variable(shapes[0])
+        y = np.arange(np.prod(shapes[1])).reshape(shapes[1])
+        expr = cp.multiply(x, y)
+        prob = cp.Problem(cp.Minimize(cp.sum(expr)), [x == 1])
+        prob.solve(canon_backend=cp.SCIPY_CANON_BACKEND)
+        assert np.allclose(expr.value, y)
```
Transurgeon, line 1762: this test shows that nd broadcasting with multiply doesn't raise an unwanted error. 
But I am still not sure if the solution is correct. Any idea for better tests? and how to validate that its working as expected (maybe compare solution to numpy multiply?)

cvxpy/expressions/expression.py
```diff
@@ -604,6 +601,13 @@ def broadcast(lh_expr: "Expression", rh_expr: "Expression"):
                 lh_expr = lh_expr @ np.ones((1, dims[1]))
             if rh_expr.shape[1] == 1 and rh_expr.shape[1] < dims[1]:
                 rh_expr = rh_expr @ np.ones((1, dims[1]))
+        # Broadcasting.
+        if lh_expr.ndim >= 3 or rh_expr.ndim >= 3:
+            output_shape = np.broadcast_shapes(lh_expr.shape, rh_expr.shape)
+            if lh_expr.shape != output_shape:
+                lh_expr = cp.broadcast_to(lh_expr, output_shape)
+            if rh_expr.shape != output_shape:
+                rh_expr = cp.broadcast_to(rh_expr, output_shape)
```
Transurgeon, line 610: added this as special case to ensure that the SCIPY backend is used to call cp.broadcast_to, otherwise it will raise an error that the CPP backend didn't implement this LinOp.

cvxpy/atoms/affine/binary_operators.py
```diff
@@ -262,19 +259,25 @@ def is_atom_quasiconcave(self) -> bool:
...
             return np.multiply(values[0], values[1])
 
+    def validate_arguments(self):
+        """Validate that the arguments are broadcastable."""
+        if not self.args[0].is_scalar() and not self.args[1].is_scalar():
+            output_shape = np.broadcast_shapes(self.args[0].shape, self.args[1].shape)
+            if self.args[0].shape != output_shape and self.args[1].shape != output_shape:
+                raise ValueError("Cannot multiply expressions with dimensions %s and %s" %
+                                (self.args[0].shape, self.args[1].shape))
+
```
Transurgeon, line 277: this function adds validation for better error messages when calling multiply.

Transurgeon (commented):
I think this is ready for a first round of reviews. I have still yet to add backend specific tests which could help understand the operator (although it is mostly about duplicating existing tensors along the rows). 
Things that are open for discussion, do we want to support broadcasting for both operands? (i.e. cp.multiply((3,1), (5)) is allowed and will give an (3,5) expression)
this means we will need to broadcast both the constant data and the variable/expr and differs from the current behavior (although matches the NumPy API for [broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html#broadcastable-arrays)

cvxpy/tests/test_expressions.py
```diff
@@ -1713,3 +1705,85 @@ def test_nd_transpose(self) -> None:
...
+        error_str = "operands could not be broadcast together"
+        with pytest.raises(Exception, match=error_str):
+            x = cp.Variable(shapes[0])
+            y = cp.broadcast_to(x, shape=shapes[1])
+            assert y.shape is not shapes[1]
+
+    @pytest.mark.parametrize("shapes", [((5),(3, 1)),
+                                        ((15, 1),(8)),
+                                        ((3), (2, 1))])
+    def test_segfault_multiply(self, shapes) -> None:
```
SteveDiamond, line 1737: Can you say more about this test? Would these shapes be compatible with numpy broadcasting rules?

SteveDiamond (approved):

SteveDiamond:
This is a really elegant solution for the backend! To be clear, in this MR we are not handling cases such as (5) + (5,1) -> (5,5)? We should still handle these cases but it can be in a separate MR.

SteveDiamond:
There is a bug in ``_aggregate_metrics``. It's not recursive. It should be
```
    @pu.compute_once
    def _aggregate_metrics(self) -> dict:
        """
        Aggregates and caches metrics for expression trees in self.args. So far
        metrics include the maximum dimensionality ('max_ndim') and whether
        all sub-expressions support C++ ('all_support_cpp').

        """
        max_ndim = self.ndim
        cpp_support = self._supports_cpp()

        for arg in self.args:
            max_ndim = max(max_ndim, arg._max_ndim())
            cpp_support = cpp_support and arg._all_support_cpp()

        metrics = {
            "max_ndim": max_ndim,
            "all_support_cpp": cpp_support
        }
        return metrics
```

cvxpy/expressions/expression.py
```diff
@@ -605,7 +605,7 @@ def broadcast(lh_expr: "Expression", rh_expr: "Expression"):
             if rh_expr.shape[1] == 1 and rh_expr.shape[1] < dims[1]:
                 rh_expr = rh_expr @ np.ones((1, dims[1]))
         # Broadcasting.
-        elif lh_expr.ndim >= 3 or rh_expr.ndim >= 3 or lh_expr.ndim != rh_expr.ndim:
+        else:
```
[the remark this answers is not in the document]
Transurgeon: yeah agreed, will do.

PR #2747, opened 2025-03-08
Draft: fix #2503 , turn `cvxpy` into meta-package that depends on `cvxpy-base`

### LINKED ISSUE #2503: Can you make `cvxpy` a meta-package that depends on `cvxpy-base` and the solvers?
It is not possible with the current `pyproject.toml` syntax to list alternative dependencies, something like `cvxpy|cvxpy-base` is not supported. In fact, it's currently possible to install both `cvxpy` and `cvxpy-base` in the same environment, which causes the `cvxpy` folder in `site-packages` to be written twice. If a package has a dependency on `cvxpy-base` and is installed in an environment which has another dependency on `cvxpy` this causes issues.

I think it would make more sense to turn `cvxpy` into a meta-package that depends on `cvxpy-base`, and your chosen selection of solvers. This can be done by fixing your `pyproject.toml` file without other changes in your codebase.

**Describe alternatives you've considered**
I researched long and found no way to list alternative dependencies, see above.

**Additional context**
This would make developing libraries that use `cvxpy` or `cvxpy-base` safer.

enzbus:
## Description
See discussion in #2503, improve compatibility between `cvxpy` and `cvxpy-base` dependents.

## Type of change
- [ X] New feature (backwards compatible)
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,7 +85,6 @@ jobs:
       matrix:
         os: [ ubuntu-22.04, macos-13, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12", "3.13" ]
-        build-cvxpy-base: [ true, false ]  # whether to build cvxpy-base (true) or regular cvxpy (false)
         include:
           # This is intended to just add the single_action_config parameter to one existing combination;
           # see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
@@ -100,8 +99,8 @@ jobs:
       PIP_INSTALL: "${{ matrix.python-version == '3.11' }}"
       PYPI_SERVER: ${{ secrets.PYPI_SERVER }}
       PYPI_USER: ${{ secrets.PYPI_USER }}
-      PYPI_PASSWORD: ${{ matrix.build-cvxpy-base && secrets.PYPI_BASE_PASSWORD || secrets.PYPI_PASSWORD }}
-      CVXPY_BASE_SUFFIX: ${{ matrix.build-cvxpy-base && 'base-' || '' }}
+      PYPI_BASE_PASSWORD: ${{ secrets.PYPI_BASE_PASSWORD }}
+      PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
 
     steps:
       - uses: actions/checkout@v4
@@ -114,27 +113,6 @@ jobs:
         run: |
           echo "PYTHON_SUBVERSION=$(echo $PYTHON_VERSION | cut -c 3-)" >> $GITHUB_ENV
           echo "DEPLOY=$( [[ $GITHUB_EVENT_NAME == 'push' && $GITHUB_REF == 'refs/tags'* ]] && echo 'True' || echo 'False' )" >> $GITHUB_ENV
-
-      - name: Adapt pyproject.toml
-        if: ${{matrix.build-cvxpy-base}}
-        shell: bash
-        run: |
-          # Mac has a different syntax for sed -i, this works across oses
-          sed -i.bak -e 's/name = "cvxpy"/name = "cvxpy-base"/g' pyproject.toml
-          sed -i.bak '/clarabel >= /d' pyproject.toml
-          sed -i.bak '/osqp >= /d' pyproject.toml
-          sed -i.bak '/ecos >= /d' pyproject.toml
-          sed -i.bak '/scs >= /d' pyproject.toml
-          sed -i.bak -e 's/name="cvxpy",/name="cvxpy-base",/g' setup.py
-          rm -rf pyproject.toml.bak setup.py.bak
-
-      - name: Verify that we can build by pip-install on each OS.
-        if: ${{matrix.build-cvxpy-base && env.PIP_INSTALL == 'True'}}
-        run: |
-          pip install . pytest cplex
-          pytest cvxpy/tests/test_conic_solvers.py -k 'TestCPLEX'
-          # ensure a fresh state for cibuildwheel
-          rm -r build
 
       - name: Set up QEMU  # For aarch64, see https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation
         if: runner.os == 'Linux'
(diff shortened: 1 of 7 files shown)


.github/workflows/test_optional_solvers.yml
```diff
@@ -37,6 +37,7 @@ jobs:
         run: |
           source continuous_integration/install_optional_solvers.sh
           pip install . pytest hypothesis
+          pip install "./meta-package"
```
PTNobel: I don't think this is necessary, we only need the optional solvers for this test.

meta-package/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+OSQP = []
+PDLP = ["ortools>=9.7,<9.12"]
+PIQP = ["piqp"]
+PROXQP = ["proxsuite"]
+QOCO = ["qoco"]
+SCIP = ["PySCIPOpt"]
+SCIPY = ["scipy"]
+SCS = []
+XPRESS = ["xpress"]
+DAQP = ["daqp"]
```
PTNobel, line 49: Need to rebase to add MPAX

PTNobel (changes requested):
The release logic needs to be rewritten in `build.yml`

Transurgeon:
@PTNobel all the CI seems to pass except for the benchmarks (I presume its because no solvers is being installed). 

Also, could you explain the changes you made to the PR? I don't quite follow the CI meta-package logic.

.github/workflows/build.yml
```diff
@@ -157,22 +135,41 @@ jobs:
           pip install build
           python -m build --sdist -o wheelhouse
 
+      - name: Build metapackage source
+        if: ${{github.event_name == 'push' && env.SINGLE_ACTION_CONFIG == 'True'}}
+        run: |
+          cd ./meta-package
+          pip install build
```
PTNobel: This isn't needed; build should be isntalled by the last step

PTNobel:
> Also, could you explain the changes you made to the PR? I don't quite follow the CI meta-package logic.

Basically, for each step of the build process, we also do it in the meta-package directory. However, we only make a sdist for the meta package, since there's no code to be compiled

Transurgeon (approved):
LGTM

jonathanberthias:
I think this will change what gets installed when installing from git, it will only install cvxpy-base now. Not sure if that's an issue though.

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-22.04, macos-13, windows-2022 ]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set Additional Envs
+        run: |
+          echo "PYTHON_SUBVERSION=$(echo $PYTHON_VERSION | cut -c 3-)" >> $GITHUB_ENV
+          echo $MOSEK_CI_BASE64 | base64 -d > mosek.lic
+          echo "MOSEKLM_LICENSE_FILE=$( [[ $RUNNER_OS == 'macOS' ]] && echo $(pwd)/mosek.lic || echo $(realpath mosek.lic) )" >> $GITHUB_ENV
```
jonathanberthias, line 26: Is this needed?

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          python-version: 3.12
+          channels: conda-forge,anaconda
```
jonathanberthias, line 34: Same comment, everything is installed from pip below

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/checkout@v4
```
jonathanberthias, line 35: There was already a checkout a couple steps above

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Install cvxpy dependencies and all optional solvers
+        run: |
+          cd meta-package
+          pip install .
```
jonathanberthias, line 39: This should be enough
```suggestion
          pip install "./meta-package"
```

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Print installed solvers
+        run : |
+          cd meta-package
```
jonathanberthias, line 42: I don't think you need to do this, you are testing the installed package so it doesn't matter where you run it from

.github/workflows/test_metapackage.yml
```diff
@@ -0,0 +1,50 @@
...
+          python -c "import cvxpy; print(cvxpy.installed_solvers())" 
+          python -c "import clarabel" 
+          python -c "import scs" 
+          python -c "import osqp" 
+
+    env:
+      RUNNER_OS: ${{ matrix.os }}
+      PYTHON_VERSION: 3.12
```
jonathanberthias, line 50: Is this needed?

meta-package/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+    "cvxpy-base" # TODO: use dynamic dependencies
+]
+requires-python = ">=3.9"
+urls = {Homepage = "https://github.com/cvxpy/cvxpy"}
+license = {text = "Apache License, Version 2.0"}
+authors = [{name = "Steven Diamond", email = "stevend2@stanford.edu"},
+           {name = "", email = "akshayka@cs.stanford.edu"},
+           {name = "Eric Chu", email = "echu508@stanford.edu"},
+           {name = "Stephen Boyd", email = "boyd@stanford.edu"}]
+dynamic = ["version"]
```
jonathanberthias, line 17: The version is never set, it gives 0.0.0 when installing

pyproject.toml
```diff
@@ -66,29 +63,6 @@ authors = [{name = "Steven Diamond", email = "stevend2@stanford.edu"},
 dynamic = ["version"]
 
 [project.optional-dependencies]
-# Solver names as in cvxpy.settings = pip-installable distribution providing it
-CBC = ["cylp>=0.91.5"]
```
jonathanberthias, line 70: Shouldn't these stay? So that we can `pip install cvxpy-base[CBC]` to only install `cylp`, as it is today

.github/workflows/test_backends.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - name: Install cvxpy dependencies
         run: |
           pip install .
```
jonathanberthias, line 21: It is already installed through `meta-package`
```suggestion
```

jonathanberthias (commented):
The test workflow seems a little bloated. Also, it seems there is something missing for the version to be set correctly.

PTNobel:
Thanks for the code review @jonathanberthias ! I'm starting to worry that this PR needs more testing on master before it gets released.

jonathanberthias:
> Thanks for the code review @jonathanberthias ! I'm starting to worry that this PR needs more testing on master before it gets released.

Glad to help! Let me know if/how I can contribute to this feature, I think it's a great improvement for downstream libraries.

PTNobel:
> Glad to help! Let me know if/how I can contribute to this feature, I think it's a great improvement for downstream libraries.

If you want to take it over, I know nothing about Python packaging and seem to be lost enough

But otherwise, any doc pointers on how to do dynamic versioning would be appreciated

pyproject.toml
enzbus: @jonathanberthias I would vote against that, downstream library that requires cvxpy-base and solvers x and y would just have to specify `["cvxpy-base", "solver_x", "solver_y"]` in their pyproject, not a huge extra cost. Advantage is that all solver-specific pyproject syntax is not duplicated on the cvxpy side, it lives in the cvxpy (meta-package) pyproject.

enzbus:
@PTNobel @jonathanberthias as per discussion in the issue #2503 one thing that would help a lot here is to switch CVXPY versioning from custom code to standard [`setuptools_scm`,](https://setuptools-scm.readthedocs.io/en/latest/config/) which I have strong suspicion does pretty much the same thing. There is also need to write a custom setuptools hook to write the dependencies in the meta-package, because we want a dependency `cvxpy-base==X.Y.Z` where X.Y.Z are resolved dynamically. Just pushed a commit for just the meta-package pyproject.

meta-package/pyproject.toml
enzbus: yes see discussion in #2503 cvxpy currently uses custom versioning code that is hard to adapt to what is needed here

Transurgeon:
I think this PR can be closed.. we haven't thought about this in a long time and it didn't seem to be blocking at the time.
Feel free to reopen this if this is still something we could revisit @enzbus !

PR #2756, opened 2025-03-17
Improve string representation of bounded variables

Transurgeon:
## Description
Variable with bounds currently print the entire arrays which is not ideal.
This PR aims to improve the string representation of this case by using the ``np.array2string`` method to shorten the bounded variable repr.
Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -185,8 +185,21 @@ class Leaf(expression.Expression):
         """Get a string representing the attributes."""
         attr_str = ""
         for attr, val in self.attributes.items():
-            if attr != 'real' and val:
-                attr_str += ", %s=%s" % (attr, val)
+            if val is not False and val is not None:
+                if isinstance(val, bool):
+                    attr_str += ", %s=%s" % (attr, val)
+                elif attr == 'bounds' and val is not None:
+                    lower = np.array2string(val[0],
+                                    edgeitems=s.PRINT_EDGEITEMS,
+                                    threshold=s.PRINT_THRESHOLD,
+                                    formatter={'float': lambda x: f'{x:.2f}'})
+                    upper = np.array2string(val[1],
+                                    edgeitems=s.PRINT_EDGEITEMS,
+                                    threshold=s.PRINT_THRESHOLD,
+                                    formatter={'float': lambda x: f'{x:.2f}'})
+                    attr_str += ", %s=(%s, %s)" % (attr, lower, upper)
+                elif attr in ('sparsity', 'boolean', 'integer') and isinstance(val, Iterable):
+                    attr_str += ", %s=%s" % (attr, val)
         return attr_str
 
     def copy(self, args=None, id_objects=None):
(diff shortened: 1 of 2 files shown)


cvxpy/expressions/leaf.py
```diff
@@ -185,8 +185,21 @@ def _get_attr_str(self) -> str:
         """Get a string representing the attributes."""
         attr_str = ""
         for attr, val in self.attributes.items():
-            if attr != 'real' and val:
-                attr_str += ", %s=%s" % (attr, val)
```
Transurgeon, line 189: I don't think the `real` attributes exists anymore? So decided to remove this check.

cvxpy/expressions/leaf.py
```diff
@@ -185,8 +185,21 @@ def _get_attr_str(self) -> str:
...
+                    attr_str += ", %s=%s" % (attr, val)
+                elif attr == 'bounds' and val is not None:
+                    lower = np.array2string(val[0],
+                                    edgeitems=s.PRINT_EDGEITEMS,
+                                    threshold=s.PRINT_THRESHOLD,
+                                    formatter={'float': lambda x: f'{x:.2f}'})
+                    upper = np.array2string(val[1],
+                                    edgeitems=s.PRINT_EDGEITEMS,
+                                    threshold=s.PRINT_THRESHOLD,
+                                    formatter={'float': lambda x: f'{x:.2f}'})
```
Transurgeon, line 199: this could be problematic if the passed bounds are sparse matrices.

cvxpy/expressions/leaf.py
```diff
@@ -185,8 +185,21 @@ def _get_attr_str(self) -> str:
...
+                    attr_str += ", %s=(%s, %s)" % (attr, lower, upper)
+                elif attr in ('sparsity', 'boolean', 'integer') and isinstance(val, Iterable):
+                    attr_str += ", %s=%s" % (attr, val)
```
Transurgeon, line 202: we should also clean up the string representation of these three cases with index array (coords). Not sure what is the best way to show this though.. will think about it.

cvxpy/tests/test_attributes.py
```diff
@@ -204,3 +204,28 @@ def test_variable_bounds(self):
...
+        assert y.__repr__() == (
+            "Variable((10, 10), y, bounds=([[0 0 ... 0 0]\n"
+            " [0 0 ... 0 0]\n"
+            " ...\n"
+            " [0 0 ... 0 0]\n"
+            " [0 0 ... 0 0]], [[10 10 ... 10 10]\n"
+            " [10 10 ... 10 10]\n"
+            " ...\n"
+            " [10 10 ... 10 10]\n"
+            " [10 10 ... 10 10]]))"
```
Transurgeon, line 224: this makes the output much shorter. Previously it would output the full 10 by 10 matrices for both upper and lower bound.

cvxpy/expressions/leaf.py
SteveDiamond: That's ok. We can assume bounds are dense for now.

SteveDiamond (approved):

Issue #2674, opened 2025-01-17
More information when solving with ``verbose=True``


Transurgeon:
Recently some activity from the community raised two interesting stats that could be added to the verbose log.

1. print the number of nodes in the expression tree (can help to identify large DPP problems). see discussion [here](https://github.com/cvxpy/cvxpy/discussions/1912#discussioncomment-11707137). 
2. (from discord), show the time it took to verify DCP, it can be useful when verifying convexity of "moderately large" quad_forms. I think we can add this log output after the grammar compliance log output [here](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/problems/problem.py#L1089-L1091).

If you have any other ideas please comment below.

rileyjmurray:
I’ve long thought it would be cool to print BibTeX info for the relevant
technologies that are used in a given solve. This includes the grammar
(DCP, DGP, DQCP) and the solver. Probably unwise to print out by default,
but we could have a message print saying “solve with bibtex=True to have
bibtex displayed for the academic papers that made this possible.”

Issue #1341, opened 2021-04-18
Citation generator


rileyjmurray:
It would be cool if there could be a keyword argument like ``prob.solve(citations=True)``, which printed BibTeX for relevant academic papers that went into making a certain reduction chain possible. This would print relevant CVXPY papers (e.g. the original 1.0 release, and the DGP release) and solver-related papers. 

The choice of printing solver papers can be determined by problem structure. If you use ECOS at all then you get BibTeX for the original paper. If you use ECOS to solve an exponential cone program then you also get BibTeX for Santiago Serrano's PhD thesis. If you used MOSEK for power cones or exponential cones then we'd give you BibTeX for [this paper](https://link.springer.com/article/10.1007/s10107-021-01631-4). The [OSQP website](https://osqp.org/docs/citing/index.html) lists several papers that could be cited depending on features used.

phschiele:
GitHub also provides the possibility to upload a citation file (see [docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) and [blog post](https://github.blog/2021-08-19-enhanced-support-citations-github/)) for convenience. 

Your suggestion seems better suited for CVXPY, as there are potentially multiple papers relevant to a given problem.
Perhaps that feature is still interesting, either for CVXPY or other packages in the ecosystem.

Transurgeon:
closing this, thanks @govindchari again for doing this!

PR #2761, opened 2025-03-19
Add Bibtex Citation Info 

govindchari:
## Description
As per the comment in this issue ticket: https://github.com/cvxpy/cvxpy/issues/2674, this PR adds bibtex citation information for cvxpy, the grammars used, and the solver.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -1162,6 +1162,10 @@ class Problem(u.Canonical):
                     s.LOGGER.info(
                             'Reducing DQCP problem to a one-parameter '
                             'family of DCP problems, for bisection.')
+                    if bibtex:
+                        print(_CITATION_STR)
+                        print(CITATION_DICT["CVXPY"])
+                        print(CITATION_DICT["DQCP"])                             
                 reductions = [dqcp2dcp.Dqcp2Dcp()]
                 start = time.time()
                 if type(self.objective) == Maximize:
@@ -1197,13 +1201,11 @@ class Problem(u.Canonical):
             # Cite CVXPY papers.
             print(CITATION_DICT["CVXPY"])
 
-            # Cite problem grammars.
+            # Cite problem grammar.
             if self.is_dcp():
                 print(CITATION_DICT["DCP"]) 
             if gp:
                 print(CITATION_DICT["DGP"]) 
-            if qcp:
-                print(CITATION_DICT["DQCP"]) 
 
             # Cite solver.
             print(solving_chain.reductions[-1].cite(data))
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -55,6 +55,7 @@ from cvxpy.reductions.solvers.solving_chain import (
 from cvxpy.settings import SOLVERS
 from cvxpy.utilities import debug_tools
 from cvxpy.utilities.deterministic import unique_list
+from cvxpy.utilities.citations import cite
 
 SolveResult = namedtuple(
     'SolveResult',
@@ -85,6 +86,14 @@ _NUM_SOLVER_STR = (
     '\n' +
     '-'*_COL_WIDTH
 )
+_CITATION_STR = (
+    '-'*_COL_WIDTH +
+    '\n' +
+    ('Citations').center(_COL_WIDTH) +
+    '\n' +
+    '-'*_COL_WIDTH
+)
+
 _FOOTER = (
     '-'*_COL_WIDTH +
     '\n' +
@@ -1030,6 +1039,7 @@ class Problem(u.Canonical):
                solver: str = None,
                warm_start: bool = True,
                verbose: bool = False,
+               bibtex: bool = False,
                gp: bool = False,
                qcp: bool = False,
                requires_grad: bool = False,
@@ -1180,6 +1190,10 @@ class Problem(u.Canonical):
         solver_verbose = kwargs.pop('solver_verbose', verbose)
         if solver_verbose and (not verbose):
             print(_NUM_SOLVER_STR)
+        if verbose and bibtex:
+            print(_CITATION_STR)
+            breakpoint()
+            cite(self.is_dcp(), gp, qcp, solving_chain.reductions[-1].name(), data['dims'].exp > 0)
         solution = solving_chain.solve_via_data(
             self, data, warm_start, solver_verbose, kwargs)
         end = time.time()
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -1097,7 +1097,6 @@ class Problem(u.Canonical):
 
         if verbose:
             print(_HEADER)
-
         for parameter in self.parameters():
             if parameter.value is None:
                 raise error.ParameterError(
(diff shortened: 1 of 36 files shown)


SteveDiamond:
This is quite interesting! I'll share with the other maintainers.

Transurgeon (commented):
looks really good, thanks @govindchari . Could you fix the linting issues? Maybe we should add some documentation somewhere that this is available?

rileyjmurray (commented):
Thanks so much for stepping up to work on this, @govindchari! 

I'd like to suggest a slightly different approach. Rather than ``cite`` being a free function that's called within ``solve``, I'd suggest that ``cite`` be an instance method method of Solver objects. This method can accept the data generated by ``Solver.apply`` and return the citation data in whatever format you'd like. (See "small clarification" below.) The solver-specific citation info would then be combined with citation info stemming from the grammar ruleset like being DCP/DQCP.

The benefit of this approach is that it becomes easier for solvers to return papers most appropriate to the problem that was actually solved. For instance, there are a few papers that describe MOSEK's numerical algorithm. One of them was published back in 2020 and is essential for MOSEK's efficient handling of power cone or exponential cone problems. In this PR you wouldn't have to get all the references perfect. Getting things started with baseline citations and the infrastructure would be a huge help. This is the kind of feature that would improve over time with more fine-grained references.

*Small clarification*: ``Solver.apply`` isn't actually defined. But it is defined in ``ConicSolver`` and ``QpSolver``. So we could make it another abstract base class method like ``solve_via_data``:
https://github.com/cvxpy/cvxpy/blob/19f8eeeee9077bbef24a68d0ec1baa1db19b33ab/cvxpy/reductions/solvers/solver.py#L69-L73

rileyjmurray:
For more context, my interest in this feature is much older than issue #2674. See https://github.com/cvxpy/cvxpy/issues/1341, from way back in 2021!

cvxpy/problems/problem.py
```diff
@@ -1180,6 +1192,22 @@ def _solve(self,
...
+            # Cite problem grammars.
+            if self.is_dcp():
+                print(CITATION_DICT["DCP"]) 
+            if gp:
+                print(CITATION_DICT["GP"]) 
+            if qcp:
+                print(CITATION_DICT["QCP"]) 
+
+            # Cite solver.
+            print(solving_chain.reductions[-1].cite(data))
```
rileyjmurray, line 1210: This is glorious. Thank you!

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -777,3 +778,15 @@ def tolerance_params() -> tuple[str]:
...
+
+        Parameters
+        ----------
+        data : dict
+            Data generated via an apply call.
+        """
+        if data['K_dir']['de'] > 0:
+            return CITATION_DICT["MOSEK"] + CITATION_DICT["MOSEK_EXP"]
+        else:
+            return CITATION_DICT["MOSEK"]
```
rileyjmurray, line 792: Change this function to something like
```python
from cvxpy.reductions.cone2cone.affine2direct import EXP, DUAL_EXP, POW3D, DUAL_POW3D
citation = CITATION_DICT['MOSEK']
# We need another citation if nonsymmetric cones are present.
# The implementation below looks kind of gross.
# It may be better to create a helper function in affine2direct.py
# which performs this check on an input dict. If such
# a helper existed then we'd just call it from here.
nonsym_cones = False
K_dir = data.get('K_dir', dict())
if K_dir.get(DUAL_EXP, 0):
    nonsym_cones = True
if K_dir.get(DUAL_POW3D, 0):
    nonsym_cones = True
K_aff = data.get('K_aff', dict())
if K_aff.get(EXP, 0):
    nonsym_cones = True
if K_aff.get(POW3D, 0):
    nonsym_cones = True
if nonsym_cones:
    citation += CITATION_DICT['MOSEK_EXP']
return citation
```

cvxpy/reductions/solvers/conic_solvers/sdpa_conif.py
```diff
@@ -187,3 +188,13 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
+    
+    def cite(self, data):
+        """Returns the result of the call to the solver.
+
+        Parameters
+        ----------
+        data : dict
+            Data generated via an apply call.
+        """
+        return CITATION_DICT["SDPA"]
```
rileyjmurray, line 200: Ping @usamamuneeb. Once this PR is merged it would be great if you could add logic to check if we should cite SDPA-GMP.

cvxpy/reductions/solvers/constant_solver.py
```diff
@@ -35,3 +35,6 @@ def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
             return Solution(s.OPTIMAL, problem.objective.value, {}, {}, {})
         else:
             return Solution(s.INFEASIBLE, None, {}, {}, {})
+
+    def cite(self, data):
+        return None
```
rileyjmurray, line 40: I suggest this return an empty string instead of None.

rileyjmurray (changes requested):
This looks great! I only have some minor comments. Once they're addressed, I think we can merge :)

cvxpy/problems/problem.py
rileyjmurray: Small request: can you change the keyword argument from ``citation`` to ``bibtex``?

SteveDiamond:
Unit tests cannot create files. If you look in test_problem, there's an example of a test that reads the output.

govindchari:
I added the unit tests and made the suggested changes.

If you take a look at `problem.py` I have the citation functions appear in two places due to how the dqcp case is handled in its own branch. I also do not cite the solver in the dqcp case. 

If anyone has ideas on how to handle this more elegantly and how to extract the solver used in the dqcp branch please let me know.

cvxpy/tests/test_problem.py
```diff
@@ -383,6 +384,52 @@ def test(self, a, b: int = 2):
...
+        backup = sys.stdout
+
+        # Solve Disciplined Convex Program
+        sys.stdout = StringIO()  # capture output
+        p = Problem(cp.Minimize(self.a + self.x[0]),
+                    [self.a >= 2, self.x >= 2])
+        p.solve(verbose=True, bibtex=True)
+        out = sys.stdout.getvalue()  # release output
+        sys.stdout.close()  # close the stream
+        sys.stdout = backup  # restore original stdout
```
jonathanberthias, line 401: You might want to use [`contextlib.redirect_stdout`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout) to handle this more easily

Transurgeon (approved):
I think the tests you've added seem fine, and the rest of the PR looks great. Let's still wait for Riley to review and merge this :).

Transurgeon:
@rileyjmurray could you have a final look at this please?

rileyjmurray (approved):
Looks good to me! Thanks a million, @govindchari!

Issue #2764, opened 2025-03-20
BROADCAST_TO is not supported


langestefan:
**Describe the bug**
PR #2724 adds the `broadcast_to` atom. 

This is causing `vstack` (and perhaps other atomic functions too) to fail with the error:
`NotImplementedError: Type BROADCAST_TO is not supported.`

**To Reproduce**

```
import cvxpy as cp
import numpy as np

# state-space matrices
n = 3  # number of states
m = 2  # number of inputs
N = 10  # number of time steps

A = 0.99 * cp.Constant(np.eye(n))
B = cp.Constant(np.eye(n, m))

# state variable
T = cp.Variable((n, N))

# input variable
u0 = cp.Variable((1, N - 1))
u1 = cp.Parameter((m - 1, N - 1), value=np.ones((m - 1, N - 1)))
u = cp.vstack([u0, u1])

# no issue
# T_0 = 18.0 * np.ones(n)

# NotImplementedError: Type BROADCAST_TO is not supported
T_0 = cp.vstack([cp.Constant(18.0), cp.Constant(18.0), cp.Constant(18.0)])

constraints = []
for t in range(N - 1):
    constraints.append(T[:, t + 1] == A @ T[:, t] + B @ u[:, t])
constraints.append(T[:, 0] == T_0)

# state constraints
constraints.append(T[0, :] <= 22.0)
constraints.append(T[0, :] >= 18.0)

# input constraints
constraints.append(u >= 0)

# objective function
objective = cp.Minimize(cp.sum(u0))
problem = cp.Problem(objective, constraints)
problem.solve()
# print the results
print("Status:", problem.status)
print("Objective value:", problem.value)

```

**Expected behavior**
I expect `T_0 = cp.vstack([cp.Constant(18.0), cp.Constant(18.0), cp.Constant(18.0)])` to just give me a vector `[18.0, 18.0, 18.0]` like it did in previous versions of cvxpy. 

**Output**
```
Traceback (most recent call last):
  File "C:\dev\warmteplanner\model\warmteplanner\broadcast_to.py", line 41, in <module>
    problem.solve()
  File "C:\dev\warmteplanner\model\warmteplanner\modelpy\.venv\Lib\site-packages\cvxpy\problems\problem.py", line 600, in solve
    return solve_func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2765, opened 2025-03-20
Detect broadcasting in constraints

SteveDiamond:
## Description
This MR moves the selection of the canonicalization backend to the point in the canonicalization right before matrix stuffing (i.e. right before the backend is used). The backend is also selected by analyzing precisely the problem that is being passed to the backend. Previously the backend was selected by analyzing the initial problem before any canonicalization. If the canonicalization process introduced linear operators like `broadcast_to` that are not supported by the CPP backend, the backend selection could be incorrect and lead to an error.

This MR has a regression in the matrix stuffing benchmarks (https://github.com/cvxpy/benchmarks/blob/main/benchmark/matrix_stuffing.py) because the backend selection was moved from the start of the canonicalization process to the start of matrix stuffing. The overall compile time should be similar though, so the regression is acceptable IMO.

Issue link (if applicable): #2764 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/constraints/nonpos.py
+++ b/cvxpy/constraints/nonpos.py
@@ -20,7 +20,6 @@ import numpy as np
 
 # Only need Variable from expressions, but that would create a circular import.
 from cvxpy.constraints.constraint import Constraint
-from cvxpy.expressions import cvxtypes
 from cvxpy.utilities import scopes
 
 
@@ -180,7 +179,6 @@ class Inequality(Constraint):
         A unique id for the constraint.
     """
     def __init__(self, lhs, rhs, constr_id=None) -> None:
-        lhs, rhs = cvxtypes.expression().broadcast(lhs, rhs)
         self._expr = lhs - rhs
         if self._expr.is_complex():
             raise ValueError("Inequality constraints cannot be complex.")
--- a/cvxpy/constraints/nonpos.py
+++ b/cvxpy/constraints/nonpos.py
@@ -20,6 +20,7 @@ import numpy as np
 
 # Only need Variable from expressions, but that would create a circular import.
 from cvxpy.constraints.constraint import Constraint
+from cvxpy.expressions import cvxtypes
 from cvxpy.utilities import scopes
 
 
@@ -179,6 +180,7 @@ class Inequality(Constraint):
         A unique id for the constraint.
     """
     def __init__(self, lhs, rhs, constr_id=None) -> None:
+        lhs, rhs = cvxtypes.expression().broadcast(lhs, rhs)
         self._expr = lhs - rhs
         if self._expr.is_complex():
             raise ValueError("Inequality constraints cannot be complex.")
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -579,10 +579,7 @@ class Expression(u.Canonical):
         return expr if isinstance(expr, Expression) else cvxtypes.constant()(expr)
 
     @staticmethod
-    def broadcast(
-            lh_expr: "Expression", 
-            rh_expr: "Expression"
-        ) -> tuple["Expression", "Expression"]:
+    def broadcast(lh_expr: "Expression", rh_expr: "Expression"):
         """Broadcast the binary operator."""
         lh_expr = Expression.cast_to_const(lh_expr)
         rh_expr = Expression.cast_to_const(rh_expr)
@@ -615,19 +612,6 @@ class Expression(u.Canonical):
             if rh_expr.shape != output_shape:
                 rh_expr = cp.broadcast_to(rh_expr, output_shape)
         return lh_expr, rh_expr
-
-    @staticmethod
-    def broadcast_list(expr_list: list["Expression"]) -> list["Expression"]:
-        """Broadcast an arbitrary number of expressions to the same dimensions."""
-        if len(expr_list) <= 1:
-            return expr_list
-
-        reduction_expr = expr_list[0] 
-        output_list = []
-        for expr in expr_list[1:]:
-            reduction_expr, output_expr = Expression.broadcast(reduction_expr, expr)
-            output_list.append(output_expr)
-        return [reduction_expr] + output_list
 
     @_cast_other
     def __add__(self, other: ExpressionLike) -> "Expression":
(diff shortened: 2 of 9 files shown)


cvxpy/expressions/expression.py
```diff
@@ -613,6 +616,19 @@ def broadcast(lh_expr: "Expression", rh_expr: "Expression"):
...
 
+    @staticmethod
+    def broadcast_list(expr_list: list["Expression"]) -> list["Expression"]:
+        """Broadcast an arbitrary number of expressions to the same dimensions."""
+        if len(expr_list) <= 1:
+            return expr_list
+
+        reduction_expr = expr_list[0] 
+        output_list = []
+        for expr in expr_list[1:]:
```
PTNobel, line 627: This method is incorrect. Give it inputs `(1,), (2,), (2,1)` and it will return shapes `(2, 1), (2,), (2, 1)`which is wrong I think. Not sure how to do this without being quadratic

PTNobel (changes requested):
At least one method isn't as documented, but maybe it doesn't need to be fully correct?

It at least needs doc changes

cvxpy/expressions/expression.py
Transurgeon: maybe we should use something like np.broadcast_shapes(*shapes), and then broadcast everything to that target shape?

cvxpy/constraints/nonpos.py
```diff
@@ -179,6 +180,7 @@ class Inequality(Constraint):
         A unique id for the constraint.
     """
     def __init__(self, lhs, rhs, constr_id=None) -> None:
+        lhs, rhs = cvxtypes.expression().broadcast(lhs, rhs)
         self._expr = lhs - rhs
```
Transurgeon, line 184: I feel like this substraction will already call the broadcast function on lhs and rhs, but maybe I could be wrong on that?

SteveDiamond:
Thanks for the feedback! On further reflection, the approach here is not right. The fundamental issue is that linear operators not supported by the CPP backend might be used during canonicalization. The most sustainable way to handle this issue IMO is to check whether the CPP backend is supported right before matrix stuffing. Currently we check whether the CPP backend is supported at the very start of canonicalization.

cvxpy/expressions/expression.py
PTNobel: That makes sense to me

Transurgeon (approved):
nice work @SteveDiamond ! I now understand a bit better what you meant by the constraints happen after the initial canonicalization. I think the changes are fine for now. But we should definitely think about some other tests to add potentially.

cvxpy/tests/test_constraints.py
```diff
@@ -405,6 +405,26 @@ def test_bound_properties(self) -> None:
...
+    def test_broadcasting(self) -> None:
+        """Test interaction of constraints and broadcasting."""
+        x = cp.Variable((3, 1))
+        # c = cp.Constant(np.arange(3))
+        c = cp.Constant(np.ones(3))
+        # Equality constraint.
+        con = (x == c)
+        prob = cp.Problem(cp.Minimize(0), [con])
+        prob.solve(solver=cp.CLARABEL) 
+        assert np.allclose(x.value, c.value)
```
Transurgeon, line 417: so if my understanding is correct this constraint will be of shape (3,3) right?
maybe we can just assert that's the shape of the constraint and add a comment saying that broadcasting happened which switches the backend from CPP to SCIPY.

SteveDiamond:
I ran the benchmarks multiple times and the same problem failed. This may be a real performance regression, which is understandable given the implementation right now.

Transurgeon:
> I ran the benchmarks multiple times and the same problem failed. This may be a real performance regression, which is understandable given the implementation right now.

ah I see, makes sense. It is probably because some benchmarks didn't follow these weird broadcasting rules we introduced. 
I am wondering what the behavior was before? I.e. having (3,)==(3,1)? Did it just create 3 equality constraint instead of 3x3?

SteveDiamond:
> > I ran the benchmarks multiple times and the same problem failed. This may be a real performance regression, which is understandable given the implementation right now.
> 
> ah I see, makes sense. It is probably because some benchmarks didn't follow these weird broadcasting rules we introduced. I am wondering what the behavior was before? I.e. having (3,)==(3,1)? Did it just create 3 equality constraint instead of 3x3?

Unfortunately it's because the new code makes the canonicalization slower. It's possible the old approach of broadcasting in the constraint definitions was better for speed. Let's discuss more over VC.

SteveDiamond:
This MR is ready to merge. I addressed the compilation time regression by removing an unnecessary call to `get_canon_backend`.

cvxpy/tests/test_constraints.py
```diff
@@ -405,6 +405,45 @@ def test_bound_properties(self) -> None:
...
+
+        with pytest.raises(ValueError, match="The CPP backend cannot be used"):
+            prob = cp.Problem(cp.Minimize(0), [con])
+            prob.solve(solver=cp.CLARABEL, canon_backend=cp.CPP_CANON_BACKEND) 
+
+        # Test QP codepath.
+        prob = cp.Problem(cp.Minimize(cp.sum_squares(x)), [con])
+        prob.solve(solver=cp.CLARABEL)
```
PTNobel, line 426: Shouldn't this be OSQP to test the QP code path?

cvxpy/tests/test_constraints.py
```diff
@@ -405,6 +405,45 @@ def test_bound_properties(self) -> None:
...
+            prob.solve(solver=cp.CLARABEL, canon_backend=cp.CPP_CANON_BACKEND) 
+
+        # Test QP codepath.
+        prob = cp.Problem(cp.Minimize(cp.sum_squares(x)), [con])
+        prob.solve(solver=cp.CLARABEL) 
+        assert np.allclose(x.value, c.value)
+
+        with pytest.raises(ValueError, match="The CPP backend cannot be used"):
+            prob = cp.Problem(cp.Minimize(cp.sum_squares(x)), [con])
+            prob.solve(solver=cp.CLARABEL, canon_backend=cp.CPP_CANON_BACKEND)
```
PTNobel, line 431: Ditto

PTNobel (approved):

cvxpy/tests/test_constraints.py
SteveDiamond: Good catch, fixed.

Transurgeon:
sounds good, will merge once CI passes

PR #2769, opened 2025-03-26
Fix OSQP polish parameter name for v1.0+

bstellato:
## Description
    • OSQP v1.0+ renamed the 'polish' parameter to 'polishing'
    • This PR detects the OSQP version and uses the appropriate parameter name
    • Ensures compatibility with both older and newer versions of OSQP

Issue link (if applicable):
https://github.com/osqp/osqp-python/issues/170

@vineetbansal @ianmcinerney

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ cvxpy/version.py
 
 # Hypothesis testing
 .hypothesis/
+.aider*
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -12,7 +12,8 @@ class OSQP(QpSolver):
     """QP interface for the OSQP solver"""
 
     # Map of OSQP status to CVXPY status.
-    STATUS_MAP = {1: s.OPTIMAL,
+    # Note: Status map has changed in versions >= 1
+    STATUS_MAP_PRE_V1 = {1: s.OPTIMAL,
                   2: s.OPTIMAL_INACCURATE,
                   -2: s.SOLVER_ERROR,           # Maxiter reached
                   -3: s.INFEASIBLE,
@@ -22,6 +23,16 @@ class OSQP(QpSolver):
                   -6: s.USER_LIMIT,
                   -5: s.SOLVER_ERROR,           # Interrupted by user
                   -10: s.SOLVER_ERROR}          # Unsolved
+    STATUS_MAP = {1: s.OPTIMAL,
+                  2: s.OPTIMAL_INACCURATE,
+                  3: s.INFEASIBLE,
+                  4: s.INFEASIBLE_INACCURATE,
+                  5: s.UNBOUNDED,
+                  6: s.UNBOUNDED_INACCURATE,
+                  7: s.SOLVER_ERROR,           # Maxiter reached
+                  8: s.USER_LIMIT,
+                  10: s.SOLVER_ERROR,          # Interrupted by user
+                  11: s.SOLVER_ERROR}          # Unsolved
 
     def name(self):
         return s.OSQP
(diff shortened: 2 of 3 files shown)


bstellato:
I believe there is something wrong in `cvxpy/tests/test_attributes.py::TestAttributes::test_infeasible_sparse`

The solver terminates with the correct status but the returned number is not correct. Could it be a discrepancy between  [the solver settings](https://osqp.org/docs/interfaces/status_values.html) and the [status map](https://github.com/cvxpy/cvxpy/blob/95a43f736cace3287e382cf368519152edcf6ae3/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L15) ?

Transurgeon:
> I believe there is something wrong in `cvxpy/tests/test_attributes.py::TestAttributes::test_infeasible_sparse`
> 
> The solver terminates with the correct status but the returned number is not correct. Could it be a discrepancy between [the solver settings](https://osqp.org/docs/interfaces/status_values.html) and the [status map](https://github.com/cvxpy/cvxpy/blob/95a43f736cace3287e382cf368519152edcf6ae3/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L15) ?

yeah seems like that is where the error is coming from. Did you guys change the status values for OSQP recently?

Transurgeon:
we just merged #2768 to fix the issue you mentioned. but I think we should still update the status map.

bstellato:
Yes, the status codes are different for 1.0 OSQP solves that problem accurately. I wonder how many other problems were actually solved correctly. We need to update the status map.

.gitignore
```diff
@@ -61,4 +61,5 @@ tags
 cvxpy/version.py
 
 # Hypothesis testing
-.hypothesis/
\ No newline at end of file
+.hypothesis/
+.aider*
```
Transurgeon, line 65: what is this for?

Transurgeon (approved):
I think this looks good to merge. we can update the status map in another PR

vineetbansal:
@Transurgeon - I've just opened a PR towards this branch `b/osqp1.0` that fixes the status map issue, and also reverts the change that was made earlier to cvxpy on account of the (incorrect) status map being assumed. Let me know if there are any tweaks to make there.

Transurgeon:
thanks @vineetbansal I just merged. We should maybe also update the pyproject.toml to enforce wheels to be built with the latest osqp.

vineetbansal:
Thanks @Transurgeon - regarding your comment about tests failing on `osqp` - if I'm reading the code right, it reduces to the code in `master` if `float(osqp.__version__.split('.')[0]) < 1`. I'm not familiar with the development history of `cvxpy` and cannot say which versions were working w.r.t `osqp`, but we've made no changes to `osqp==0.6.x`, and `osqp>=1.x` would not have worked here before anyway, so I'm not sure..

Transurgeon:
> Thanks @Transurgeon - regarding your comment about your local tests failing on `osqp` - if I'm reading this PR right, it reduces to the code in `master` if `float(osqp.__version__.split('.')[0]) < 1`. I'm not familiar with the development history of `cvxpy` and cannot say which versions were working w.r.t `osqp`, but we've made no changes to `osqp==0.6.x`, and `osqp>=1.x` would not have worked here before anyway, so I'm not sure..

ok sounds good, thanks for your reply. Maybe it had something to do with conflicting versions on my local.. I can't even reproduce these errors now so no worries. 

 I think this is ready to merge in order to fix #2770 .

bstellato:
Fantastic! Thanks @vineetbansal and @Transurgeon.

.gitignore
bstellato: That's [aider.chat](https://aider.chat/) sometimes helping :)

PR #2771, opened 2025-03-26
fixes for PR 2769

vineetbansal:
@bstellato - here are some tweaks to maintain osqp correctness going forward. Not ideal (we should probably be pulling these from `osqp` instead of hard-coding), but that's another discussion, and hard-coding values for now is better than having incorrect ones.

Also reverted the change `cvxpy` made in its tests.


Transurgeon:
@vineetbansal thanks for this! the changes look good. However, as I was testing on my local it seems like some older version of OSQP are failing. 
I was wondering which is the latest version of OSQP that works with the cvxpy interface?

PR #2784, opened 2025-04-08
change OSQP error type when passing non-psd A matrix

Transurgeon:
## Description
Not sure if this is the correct fix.
I think there were some changes to the OSQP interface which causes one of the test to fail.
The test in question uses the wrapper ``assume_PSD=True`` on a clearly non-PSD matrix, which causes the OSQP solver to fail. However, it seems like the error being raised is different with the new OSQP. 
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py
@@ -125,7 +125,7 @@ class OSQP(QpSolver):
             solver = osqp.OSQP()
             try:
                 solver.setup(P, q, A, lA, uA, verbose=verbose, **solver_opts)
-            except ValueError as e:
+            except Exception as e:
                 raise SolverError(e)
 
         results = solver.solve()
--- a/cvxpy/tests/test_quad_form.py
+++ b/cvxpy/tests/test_quad_form.py
@@ -20,6 +20,7 @@ import numpy as np
 import pytest
 import scipy.sparse as sp
 from numpy.testing import assert_allclose, assert_equal
+from osqp.interface import OSQPException
 
 import cvxpy as cp
 from cvxpy.settings import EIGVAL_TOL
@@ -214,6 +215,5 @@ class TestNonOptimal(BaseTest):
 
         prob = cp.Problem(cp.Minimize(expr))
         # Transform to a SolverError.
-        with pytest.raises(cp.SolverError,
-                           match=r"(Workspace allocation error!)|(Setup Error \(Error Code 4\))"):
+        with pytest.raises(OSQPException):
             prob.solve(solver=cp.OSQP)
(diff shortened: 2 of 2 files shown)


SteveDiamond (approved):

SteveDiamond:
lgtm

PTNobel:
Should the OSQP interface catch the exception and change it to a SolverError?

SteveDiamond:
> Should the OSQP interface catch the exception and change it to a SolverError?

That does make sense.

Transurgeon:
yea agreed also. Even better would be an error that the matrix is not PSD but that might be too hard.

Transurgeon:
@vineetbansal could you have a look into this please?
I am not sure what is the best way to proceed.. is it even possible to raise a cp.SolverError in this case? It seems different than the status map solver errors.

vineetbansal:
@Transurgeon - yes, we did update the raised Exceptions from plain `ValueError`s to `OSQPException`s, which have more information as to what went wrong. Here's the relevant part of the [changelog](https://github.com/osqp/osqp-python/releases/tag/v1.0.3):

> The .setup() (and .solve(raise_error=True)) invocations on <solver> would initially have raised ValueErrors. These now raise osqp.OSQPExceptions which are more introspectable.

In the test case that you modified in this PR, it could be used as:
```
with pytest.raises(OSQPException) as e:
    assert e == e==osqp.SolverError.OSQP_NONCVX_ERROR
```

However, `osqp.OSQPException` (and `osqp.SolverError`) only exist in `osqp>=1.0.3`. Since `cvxpy` claims to work with all `osqp` versions, it seems to me like a better fix might be in [osqp_qpif.py](https://github.com/cvxpy/cvxpy/blob/38d16a2a9687e55cfdf512e63f04f25932ff9884/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L126):

Change:
```
            try:
                solver.setup(P, q, A, lA, uA, verbose=verbose, **solver_opts)
            except ValueError as e:
                raise SolverError(e)
```
to:
```
            try:
                solver.setup(P, q, A, lA, uA, verbose=verbose, **solver_opts)
            except ValueError as e:
                raise SolverError(e)
            except Exception as e:
                try:
                    from osqp import OSQPException
                except ImportError:
                    raise
                else:
                    if isinstance(e, OSQPException):
                        # do something useful with osqp.SolverError value?
                        raise SolverError(e)
```

A simpler fix is to simply not insist on a `ValueError` (which was a somewhat arbitrary decision on the part of `osqp` anyway), and simply catch `Exception`:
```
            try:
                solver.setup(P, q, A, lA, uA, verbose=verbose, **solver_opts)
            except Exception as e:
                raise SolverError(e)
```

SteveDiamond:
My preference is for
```
            try:
                solver.setup(P, q, A, lA, uA, verbose=verbose, **solver_opts)
            except Exception as e:
                raise SolverError(e)
```
out of those two options.

Issue #2662, opened 2024-12-31
support (more?) HiGHS options


nlaws-camus:
**Is your feature request related to a problem? Please describe.**
I tried passing in `presolve` like:
```python
problem.solve(verbose=True, solver=cp.HIGHS, presolve=True)
```
but it results in:
```sh
  File "/common_deps_cvxpy/site-packages/cvxpy/problems/problem.py", line 577, in solve
    return solve_func(self, *args, **kwargs)
  File "/common_deps_cvxpy/site-packages/cvxpy/problems/problem.py", line 1160, in _solve
    solution = solving_chain.solve_via_data(
  File "/common_deps_cvxpy/site-packages/cvxpy/reductions/solvers/solving_chain.py", line 524, in solve_via_data
    return self.solver.solve_via_data(data, warm_start, verbose,
  File "/common_deps_cvxpy/site-packages/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py", line 184, in solve_via_data
    setattr(options, key, value)
TypeError: (): incompatible function arguments. The following argument types are supported:
    1. (self: highspy._core.HighsOptions, arg0: str) -> None

Invoked with: <highspy._core.HighsOptions object at 0x7a56e07ef170>, True
```

**Describe the solution you'd like**
A clear and concise description of what you want to happen, and how your solution would solve the problem.
Support [HiGHS options](https://ergo-code.github.io/HiGHS/dev/options/definitions/). Documentation would be nice too.

Transurgeon:
@UweGensheimer could you look into this please?
@nlaws-camus what do you mean by documentation? We have solver options documentation [here](https://www.cvxpy.org/tutorial/solvers/index.html#setting-solver-options), but HiGHS hasn't been added yet.

senhalil:
@nlaws-camus this looks like a type error because HiGHS expects a string ("on"/"off") for `presolve`, at least this is what I gathered from the error and the documentation.

tasseff:
The HiGHS `solver` option (documented [here](https://ergo-code.github.io/HiGHS/dev/solvers/#solvers)) also cannot be used due to overlap with the `solver` argument required by CVXPY's `solve` method. It would be useful if there were some way to specify which HiGHS algorithm is used.

SteveDiamond:
[...]

PR #2785, opened 2025-04-08
Implement highs_options to prevent option/argument name clashes 

senhalil:
## Description

This PR adds a `highs_options` argument to solve (à-la `mosek_params`, `scipy_options`, etc.) to prevent the argument/option name clash issue observed in https://github.com/cvxpy/cvxpy/issues/2662 -- i.e., HiGHS option `solver="simplex"` clashing with solve argument `solver=cp.HIGHS`.

The main logic change/addition is `unpack_highs_options_inplace` function. 

---

(Out-of-scope) This function can be called at a higher level, generalizing this `solvername_options`/`solvername_params` (or more aptly `solver_opts`) unpacking logic and solving this issue for all solvers at the same time. A backwards compatible way to implement this logic would have been a significantly larger PR so I went with the suggestion here https://github.com/cvxpy/cvxpy/issues/2662#issuecomment-2660430851.

---

For some reason `ruff format` show some modifications in the files that I touched, I extracted these changes into a single commit (634740adbb12acc7ea05fd81ab3bb85ed3c739ee). (Out-of-scope) Calling `ruff format cvxpy` modifies a significant portion of the repo -- this might be due to mismatch of ruff versions (`ruff` version is not pinned in the project) or maybe these files were never formatted.

I also refactored TestHIGHS class in 61ae00208533cdb837be70b16857574fa69f6c4e to be based on `pyyest` (instead of `unittest.TestCase`) so that I can use `pytest.parameterize`.

I suspect reviewing the PR commit-by-commit might be easier due to these two commits but if needed, I can split these into another PR, rendering the actual logic change PR more atomic.

---

Issue link (if applicable): Closes https://github.com/cvxpy/cvxpy/issues/2662

## Type of change
- [ ] Bug fix

- [ ] (N/A) Add our license to new files.
- [x] Write unittests.
- [ ] (CI?) Run the benchmarks to make sure your change doesn’t introduce a regression.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
@@ -106,7 +106,9 @@ class HIGHS(QpSolver):
             sol = failure_solution(status, attr)
         return sol
 
-    def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
+    def solve_via_data(
+        self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None
+    ):
         """Returns the result of the call to the solver.
 
         minimize      1/2 x' P x + q' x
--- a/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
@@ -98,9 +98,7 @@ class HIGHS(QpSolver):
             sol = failure_solution(status, attr)
         return sol
 
-    def solve_via_data(
-        self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None
-    ):
+    def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, solver_cache=None):
         """Returns the result of the call to the solver.
 
         minimize      1/2 x' P x + q' x
--- a/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
@@ -24,6 +24,14 @@ from cvxpy.reductions.solution import Solution, failure_solution
 from cvxpy.reductions.solvers.qp_solvers.qp_solver import QpSolver
 
 
+def unpack_highs_options_inplace(solver_opts) -> None:
+    # Users can pass options inside a nested dict -- e.g. to circumvent a name clash
+    highs_options = solver_opts.pop("highs_options", dict())
+
+    # merge via update(dict(...)) is needed to avoid silently over-writing options
+    solver_opts.update(dict(**solver_opts, **highs_options))
+
+
 class HIGHS(QpSolver):
     """QP interface for the HiGHS solver"""
 
@@ -176,6 +184,7 @@ class HIGHS(QpSolver):
             hessian.value_ = P.data
 
         # setup options
+        unpack_highs_options_inplace(solver_opts)
         options = hp.HighsOptions()
         options.log_to_console = verbose
         for key, value in solver_opts.items():
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2172,8 +2172,12 @@ class TestHIGHS:
         highs_options = {
             "solver": "simplex",  # string option -- also `"solver"` shouldn't name-clash
             "output_flag": True,  # bool option
-            "time_limit": 0.001,  # double option
+            "time_limit": 0.123,  # double option
             "random_seed": 1234,  # int option
+            # no need to optimize for real
+            "mip_rel_gap": 1.0,  
+            "primal_feasibility_tolerance": 1e-3,
+            "dual_feasibility_tolerance": 1e-3,
         }
         problem(solver=cp.HIGHS, highs_options=highs_options)
 
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2155,6 +2155,31 @@ class TestHIGHS:
     def test_highs_solving(self, problem) -> None:
         problem(solver=cp.HIGHS)
 
+    @pytest.mark.parametrize(
+        "problem",
+        # it is enough to validate the options with one problem from each type
+        [StandardTestLPs.test_lp_0, StandardTestLPs.test_mi_lp_0],
+    )
+    def test_highs_options(self, problem) -> None:
+        with pytest.raises(AttributeError):
+            problem(solver=cp.HIGHS, highs_options={"invalid_highs_option": None})
+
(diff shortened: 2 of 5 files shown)


senhalil:
I see that `ruff format` fails because `'--preview'` argument is needed to enable `'quote-style = preserve'` (I had to add this `preserve` option to prevent `ruff format` introducing countably-many single/double quote modification but CI marking this "stable" option as "preview" is a clue that I might be write about the `ruff` version mismatch.

> this might be due to mismatch of ruff versions (ruff version is not pinned in the project) 

Please let me know how you want me to fix this or feel free to jump in (as that might be a lot faster due to timezone differences).

SteveDiamond:
We can take a look. This issue has not come up before and we are stretched pretty thin so it might take some time. If the fix is pinning a ruff version, that's fine. The level of formatting changes in the PR is definitely not expected.

jonathanberthias:
AFAIK there is no automatic formatter in this repo and never has been, Ruff is only used for linting. That should explain the numerous changes when you run a formatter @senhalil

senhalil:
Oh, that definitely explains it!  I'll revert the formatting related changes (which will fix the linter) and follow the existing format manually. 

I presume, my PR cannot be the reason for the failed build and backend checks/actions as they fail in `cvxpy/tests/test_quad_form.py::TestNonOptimal::test_assume_psd - osqp.interface.OSQPException: 4`

Transurgeon:
> I presume, my PR cannot be the reason for the failed build and backend checks/actions as they fail in `cvxpy/tests/test_quad_form.py::TestNonOptimal::test_assume_psd - osqp.interface.OSQPException: 4`

yes you are right, we are looking into fixing this in another PR..

cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
```diff
@@ -24,6 +24,14 @@
...
 
 
+def unpack_highs_options_inplace(solver_opts) -> None:
+    # Users can pass options inside a nested dict -- e.g. to circumvent a name clash
+    highs_options = solver_opts.pop("highs_options", dict())
+
+    # merge via update(dict(...)) is needed to avoid silently over-writing options
+    solver_opts.update(dict(**solver_opts, **highs_options))
+
+
```
senhalil, line 34: This violates DRY principle but it looks like this is how its done in other solvers so I kept the logic similar.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2128,42 +2128,56 @@ def test_scip_time_limit_reached(self) -> None:
...
+            StandardTestLPs.test_lp_1,
+            StandardTestLPs.test_lp_2,
+            StandardTestLPs.test_lp_3,
+            StandardTestLPs.test_lp_4,
+            StandardTestLPs.test_lp_5,
+            StandardTestLPs.test_mi_lp_0,
+            StandardTestLPs.test_mi_lp_1,
+            StandardTestLPs.test_mi_lp_2,
+            StandardTestLPs.test_mi_lp_3,
+            StandardTestLPs.test_mi_lp_4,
```
senhalil, line 2148: `StandardTestLPs.test_mi_lp_4` was not present in the old version but the test passes and does not increase the test duration in a perceptible way so I figured, there is no harm including it.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2128,42 +2128,56 @@ def test_scip_time_limit_reached(self) -> None:
...
+        # Duplicate options
+        with pytest.raises(TypeError):
+            problem(solver=cp.HIGHS, presolve="off", highs_options=dict(presolve="off"))
+
+        # Happy path
+        highs_options = {
+            "solver": "simplex",  # string option -- also `"solver"` shouldn't name-clash
+            "output_flag": True,  # bool option
+            "time_limit": 0.001,  # double option
+            "random_seed": 1234,  # int option
```
senhalil, line 2176: I wonder if this very restricted duration would lead to HiGHS stopping too early (i.e., w/o any solution) and leading to a flaky CI. I'll increase it just in case but also increase the optimality tolerances:

```suggestion
            "time_limit": 0.123,  # double option
            "random_seed": 1234,  # int option
            # no need to optimize for real
            "mip_rel_gap": 1.0,  
            "primal_feasibility_tolerance": 1e-3,
            "dual_feasibility_tolerance": 1e-3,
```

doc/source/tutorial/solvers/index.rst
```diff
@@ -780,6 +780,20 @@ Here is the complete list of solver options.
...
+
+    All options of HiGHS solver can be specified within the ``solve`` command directly as keyword arguments. For example `solve(solver=cp.HIGHS, presolve="off")` would switch of the HiGHS presolver.
+
+    **Please note**: If an option name clashes with a `solve` method argument (e.g., `solver` option of HiGHS), it can be modified via the `highs_options` argument below.
+
+    ``highs_options``
+        Solver options can also be passed as a dictionary where the key-value pairs are composed of HiGHS option names and associated values. For example, `solve(solver=cp.HIGHS, highs_options=dict(solver="simplex"))` would set the LP solver of HiGHS to simplex.
+
+    For other options see `HiGHS documentation <https://ergo-code.github.io/HiGHS/dev/options/definitions/>`_.
+
```
senhalil, line 795: Do I need to do anything special to re-generate the docs (or will the CI take care of it)?

senhalil (commented):
Just made the necessary (formatting) updates and this PR is officially open for business.

doc/source/tutorial/solvers/index.rst
Transurgeon: we actually manually run the CI for the website 😅 but don't worry, we will update the website in the newest patch release.

cvxpy/tests/test_conic_solvers.py
Transurgeon: yea that's definitely fine, thanks!

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2128,42 +2128,60 @@ def test_scip_time_limit_reached(self) -> None:
...
-
-    def test_highs_mi_lp_1(self) -> None:
-        StandardTestLPs.test_mi_lp_1(solver='HIGHS')
+# We can't inherit from unittest.TestCase since we access some advanced pytest features.
+# As a result, we use the pytest skipif decorator instead of unittest.skipUnless.
+@pytest.mark.skipif('HIGHS' not in INSTALLED_SOLVERS, reason="HiGHS is not installed.")
+class TestHIGHS:
+    @pytest.mark.parametrize(
+        "problem",
+        [
```
Transurgeon, line 2137: nice! I really like this way of parametrising the tests so that you can easily test the different highs_options.

Transurgeon (approved):
this is looking good so far, thanks @senhalil !

SteveDiamond (approved):

PR #2798, opened 2025-04-30
settings.py: send logging messages to stderr

samuela:
## Description
As is the UNIX convention. See eg https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout. Sending messages to stdout breaks scripts that wish to use cvxpy but also have requirements on their own stdout specification. For example, if `myscript` is required to output valid JSON then one cannot use cvxpy in `myscript`.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/settings.py
+++ b/cvxpy/settings.py
@@ -19,7 +19,7 @@ import sys
 LOGGER = logging.getLogger("__cvxpy__")
 LOGGER.propagate = False
 LOGGER.setLevel(logging.INFO)
-_stream_handler = logging.StreamHandler(sys.stdout)
+_stream_handler = logging.StreamHandler(sys.stderr)
 _stream_handler.setLevel(logging.INFO)
 _formatter = logging.Formatter(
     fmt="(CVXPY) %(asctime)s: %(message)s", datefmt="%b %d %I:%M:%S %p"


samuela:
To be more precise, we're getting
```
(CVXPY) Apr 30 10:06:58 PM: Encountered unexpected exception importing solver GLOP:
RuntimeError('Unrecognized new version of ortools (9.12.4544). Expected < 9.12.0. Please open a feature request on cvxpy to enable support for this version.')
(CVXPY) Apr 30 10:06:58 PM: Encountered unexpected exception importing solver PDLP:
RuntimeError('Unrecognized new version of ortools (9.12.4544). Expected < 9.12.0. Please open a feature request on cvxpy to enable support for this version.')
```
messages on stdout when simply importing cvxpy with `import cvxpy`. Technically speaking, a library shouldn't even be configuring its own logging `Formatter` or `Handler` but that's for a different day (see https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

Transurgeon:
> To be more precise, we're getting
> 
> ```
> (CVXPY) Apr 30 10:06:58 PM: Encountered unexpected exception importing solver GLOP:
> RuntimeError('Unrecognized new version of ortools (9.12.4544). Expected < 9.12.0. Please open a feature request on cvxpy to enable support for this version.')
> (CVXPY) Apr 30 10:06:58 PM: Encountered unexpected exception importing solver PDLP:
> RuntimeError('Unrecognized new version of ortools (9.12.4544). Expected < 9.12.0. Please open a feature request on cvxpy to enable support for this version.')
> ```
> 
could you post a snippet of how to reproduce this error?

samuela:
> could you post a snippet of how to reproduce this error? It seems to be unrelated to the stdout issue you mentioned.

The issue is that this message is being printed to stdout. I'm not concerned with the content of the message in this case.

SteveDiamond:
The snippet you showed clearly should be going to stderr but in general the logging is used for the verbose output, which should go to stdout. So I think this fix is too simple.

samuela:
> The snippet you showed clearly should be going to stderr but in general the logging is used for the verbose output, which should go to stdout. So I think this fix is too simple.

Why isn't all logging sent to stderr?

SteveDiamond:
> > The snippet you showed clearly should be going to stderr but in general the logging is used for the verbose output, which should go to stdout. So I think this fix is too simple.
> 
> Why isn't all logging sent to stderr?

We discussed more internally and did some more research. All logging can go to stderr. We were not familiar with the conventions and were more focused on what was an "error".

SteveDiamond (approved):

PR #2810, opened 2025-05-14
Set variable value after bound sanitization

senhalil:
## Description
If the `value` of a variable is set during object creation but if the `bounds` set to be `[None,None]` then the value setter raises an `AttributeError: 'Variable' object has no attribute 'bounds'` because the bounds are not "sanitized" yet. 

This PR swaps the order of value assignment and the bound cleanup operation in `Leaf.__init__` and adds a tiny non-regression test.

(While at it I also fixed some of the obvious white-space issues.)

## Type of change
- [X] Bug fix

- [X] Write unittests.
- [N/A] Run the benchmarks to make sure your change doesn’t introduce a regression.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -155,10 +155,10 @@ class Leaf(expression.Expression):
                 "A CVXPY Variable cannot have more than one of the following attributes: "
                 f"{dim_reducing_attr}"
             )
-        if value is not None:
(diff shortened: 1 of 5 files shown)


Transurgeon (approved):
thanks! This seems to be a very edge case, but I think it's fine to initialise the value after the bounds.

SteveDiamond (approved):

PR #2811, opened 2025-05-14
Let there be names in LP files

senhalil:
## Description
This is an unsolicited proposal. I am curious what you think and whether there is a cleaner and more automatic way to do this.

Long story short, this sets the variable/column names so that the exported LP file contains the original names set by the user.

This PR is actually on top of https://github.com/cvxpy/cvxpy/pull/2810 but I can't point at that branch as the source of that PR is not in cvxpy/cvxpy repo. Basically the first two commits are from https://github.com/cvxpy/cvxpy/pull/2810 and only the last commit is for this PR.

TODO:
- [ ] There are two questions in NOTES
- [ ] A small test to verify that the variable names do appear in the exported LP files
- [ ] The lines that export the LPs need to be removed

## Type of change
- [ ] Bug fix

- [TODO] Write unittests.
- [TODO] Run the unittests and check that they’re passing.
- [N/A] Run the benchmarks to make sure your change doesn’t introduce a regression.


Transurgeon:
> ## Description
> This is an unsolicited proposal. I am curious what you think and whether there is a cleaner and more automatic way to do this.
> 
I think it's a great idea, but I am not too familiar with LP files so cannot comment on how useful this feature is. Also I am not sure users would want the variable names as columns every time (so maybe it should be a flag?). 
> This PR is actually on top of #2810 but I can't point at that branch as the source of that PR is not in cvxpy/cvxpy repo. Basically the first two commits are from #2810 and only the last commit is for this PR.
> 
that's ok, you can maybe merge the master branch onto this one once #2810 gets merged.

SteveDiamond:
This generally looks fine. There is a lot of demand for better naming in LP files so thanks for working on this. Starting by adding the functionality for one solver interface as you've done is a good approach.

PTNobel:
I went ahead and rebased the PR to clean up the diff. Can you explain the workflow for .lp files? I'm trying to better design the APIs.

senhalil:
> Can you explain the workflow for .lp files?

afaik, cvxpy does not have a way to export .lp/.mps files natively and it uses the underlying solver for exporting the model. 

So the workflow depends on the underlying solver (gurobi, cplex, highs, etc.) but with highs, setting [write_model_file](https://ergo-code.github.io/HiGHS/dev/options/definitions/#write_model_file) option does the trick and the solver creates a model file with the variable names it received.

If the following snippet sparked your comment, these lines were just a temporary hack, and I'll remove it now.

```
        # REMOVE
        solver.writeModel("highs_model.lp")
        solver.presolve()
        solver.writePresolvedModel("highs_presolved_model.lp")
```

senhalil:
> Starting by adding the functionality for one solver interface as you've done is a good approach.

👍  Would `apply` be a better place for this logic? It looks like `PARAM_PROB` related data shenanigans are done in there.


@SteveDiamond any idea about this question:
```
            # NOTE: variable.variable_of_provenance() is a bit of a hack so that variables
            # created by nonneg=True etc. are named correctly.
            # Is this alright?
```
I suspect it should be alright (at least for `nonneg=True` style cases) but I don't know if `variable_of_provenance` is set in another case where this assumption (of using the original variable name) doesn't hold.

senhalil:
> Also I am not sure users would want the variable names as columns every time (so maybe it should be a flag?).

The solvers that support variable names (at least Cplex, Gurobi, Highs) create them internally even if the user doesn't set them as far as I know, like so:

before

By passing the variable names from cvxpy space to the solver space (`lp.col_names_ = var_names`) we respect user's choice but if a user wants to name the variables but doesn't want to pass them to the solver then a flag would indeed be a way to go. 

Variable names (for that matter, constraint names) are not only useful for model exports but they help with warnings originating from the solver when they are set correctly:
```
Warning:  No solution found from 1 MIP starts.
Retaining values of one MIP start for possible repair.
Warning:  Non-integral bounds for integer variables rounded.
Row 'c29993' implies binary variable 'x2283' is greater than 1.
```

senhalil:
> I'm trying to better design the APIs.

@PTNobel was your commen about designing a general interface for model/solutions exports/writers?

SteveDiamond:
> > Starting by adding the functionality for one solver interface as you've done is a good approach.
> 
> 👍 Would `apply` be a better place for this logic? It looks like `PARAM_PROB` related data shenanigans are done in there.
> 
> @SteveDiamond any idea about this question:
> 
> ```
>             # NOTE: variable.variable_of_provenance() is a bit of a hack so that variables
>             # created by nonneg=True etc. are named correctly.
>             # Is this alright?
> ```
> 
> I suspect it should be alright (at least for `nonneg=True` style cases) but I don't know if `variable_of_provenance` is set in another case where this assumption (of using the original variable name) doesn't hold.

The `variable_of_provenance()` approach is fine.

ClayCampaigne:
I've long wanted this feature. Is there anything I can do to help get this PR over the finish line?

PTNobel:
You can help get the CI passing!

SteveDiamond:
@claude map out a path to getting this merged. what changes are needed? review the code fully

SteveDiamond:
Continued in #2944

Transurgeon:
closing this as it is continued in #2944. Thanks for starting it @senhalil !

Issue #2736, opened 2025-02-26
Sparse contradict converts to dense (`quad_form`, `sparse_cholesky`)


deasmhumhna:
**Describe the bug**
CVXPY always converts a large, sparse PSD matrix (by construction) to a dense array. I identified the problem with sparsity checking code. It checks if a wrapped Expression 

**To Reproduce**
```python
import scipy as sp
from cvxpy.atoms.affine.wraps import psd_wrap
from cvxpy.utilities.linalg import sparse_cholesky, spar
from cvxpy.expressions.expression import Expression

M = sp.sparse.diags(np.ones(10))
wrapped_M = Expression.cast_to_const(M) # as in first line of quad_form
type(wrapped_M) # <class 'cvxpy.expressions.constants.constant.Constant'>, not spmatrix
sign, L, p = sparse_cholesky(wrapped_M) # fails
sign, L, p = sparse_cholesky(psd_wrap(wrapped_M)) # assume_PSD = True, fails
```

**Expected behavior**
Identifies sparse matrices, Performs sparse cholesky

**Output**
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-58-4708dd35074b>](https://localhost:8080/#) in <cell line: 0>()
      7 wrapped_M = Expression.cast_to_const(M) # as in first line of quad_form
      8 type(wrapped_M) # <class 'cvxpy.expressions.constants.constant.Constant'>, not spmatrix
----> 9 sign, L, p = sparse_cholesky(wrapped_M) # fails
     10 sign, L, p = sparse_cholesky(psd_wrap(wrapped_M)) # assume_PSD = True, fails

[/usr/local/lib/python3.11/dist-packages/cvxpy/utilities/linalg.py](https://localhost:8080/#) in sparse_cholesky(A, sym_tol, assume_posdef)
    207 
    208     if not isinstance(A, spar.spmatrix):
--> 209         raise ValueError(SparseCholeskyMessages.NOT_SPARSE)
    210     if np.iscomplexobj(A):
    211         raise ValueError(SparseCholeskyMessages.NOT_REAL)

ValueError: Input must be a SciPy sparse matrix.
```

**Version**
 - OS: Linux (Google Colab)
 - CVXPY Version: 1.6.0

**Additional context**

Transurgeon:
[...]

PR #2834, opened 2025-06-16
Update sparse Cholesky to handle Expression inputs

rileyjmurray:
Issue #2736 shows that our sparse_cholesky utility function often gets called with CVXPY Constant objects. This patch resolves that by adding a suitable input handling.

### DIFF
--- a/cvxpy/utilities/linalg.py
+++ b/cvxpy/utilities/linalg.py
@@ -185,7 +185,8 @@ class SparseCholeskyMessages:
     ASYMMETRIC = 'Input matrix is not symmetric to within provided tolerance.'
     INDEFINITE = 'Input matrix is neither positive nor negative definite.'
     EIGEN_FAIL = 'Cholesky decomposition failed.'
-    NOT_SPARSE = 'Input must be a SciPy sparse matrix.'
+    NOT_CONST = 'The only allowed Expression inputs are Constant objects.'
+    NOT_SPARSE = 'Non-Expression inputs must be SciPy sparse matrices.'
     NOT_REAL = 'Input matrix must be real.'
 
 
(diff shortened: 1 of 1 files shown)


Transurgeon (approved):
LGTM!

Transurgeon:
it might be good to add a test also. To ensure that the code snippet from the original issue doesn't fail.

rileyjmurray:
@Transurgeon it looks like tests for sparse cholesky are being skipped. For example, see https://github.com/cvxpy/cvxpy/actions/runs/15742618682/job/44371536262?pr=2834#step:6:126.

This would happen if our C++ sparse cholesky extension wasn't being built. I can understand us choosing to not build that extension in some cases (say, Windows builds), but I'm surprised that this seems to be universal. Can you comment on if our tests avoid building the sparse cholesky extension in some way? Ping @SteveDiamond, in case you have thoughts.

SteveDiamond:
> @Transurgeon it looks like tests for sparse cholesky are being skipped. For example, see https://github.com/cvxpy/cvxpy/actions/runs/15742618682/job/44371536262?pr=2834#step:6:126.
> 
> This would happen if our C++ sparse cholesky extension wasn't being built. I can understand us choosing to not build that extension in some cases (say, Windows builds), but I'm surprised that this seems to be universal. Can you comment on if our tests avoid building the sparse cholesky extension in some way? Ping @SteveDiamond, in case you have thoughts.

It's definitely supposed to be building the C++ extension.

rileyjmurray:
Apparently I was using stupid syntax in checking if the extension was built. Will fix.

rileyjmurray:
Should be fixed now. We can wait till tests pass though.

rileyjmurray:
The pre-commit checks passed and the meaningful tests passed for my last commit, so I'll merge now.

Issue #2836, opened 2025-06-18
Issue with HiGHS storing dual variables with the CVaR atom


Transurgeon:
### Discussed in https://github.com/cvxpy/cvxpy/discussions/2830

<div type='discussions-op-text'>

<sup>Originally posted by **kmcentush-eq** June 12, 2025</sup>
Per the title, I'm trying to maximize profit of a portfolio subject to not exceeding a specified CVaR. I'm trying the code below but getting an error when the solution is being inverted. I imagine this is a result of what I'm trying to do, so any advice is appreciated!

<details><summary>Code that Errors</summary>

```
import numpy as np
import cvxpy as cp

# Generate data
num_stocks = 5
num_samples = 25
pnl_samples = np.random.uniform(low=0.0, high=1.0, size=(num_samples, num_stocks))
pnl_expected = pnl_samples.mean(axis=0)

# Prepare to solve
quantile = 0.05
w = cp.Variable(num_stocks, nonneg=True)
cvar = cp.cvar(pnl_samples @ w, 1 - quantile)
pnl = w @ pnl_expected

# Solve
objective = cp.Maximize(pnl)
constraints = [cvar <= 0.5]
problem = cp.Problem(objective, constraints)
problem.solve(
    solver=cp.HIGHS,
    verbose=True,
)
```
</details> 

<details><summary>Error Traceback</summary>

```
File <command-6934689331428126>, line 19
     17 constraints = [cvar <= 0.5]
     18 problem = cp.Problem(objective, constraints)
---> 19 problem.solve(
     20     solver=cp.HIGHS,
     21     verbose=True,
     22     log_to_console=True,
     23     output_flag=True,
     24 )
File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.11/site-packages/cvxpy/problems/problem.py:600, in Problem.solve(self, *args, **kwargs)
    597         raise ValueError(
    598             "Cannot specify both 'solver' and 'solver_path'. Please choose one.")
    599     return self._solve_solver_path(solve_func,solver_path, args, kwargs)
--> 600 return solve_func(self, *args, **kwargs)
File /local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.11/site-packages/cvxpy/problems/problem.py:1193, in Problem._solve(self, solver, warm_start, verbose, gp, qcp, requires_grad, enforce_dpp, ignore_dpp, canon_backend, **kwargs)
   1191 end = time.time()
[...]

PR #2838, opened 2025-06-18
[HIGHS conif] Maps dual solution back to dual variables

senhalil:
## Description
Fixes https://github.com/cvxpy/cvxpy/issues/2836

## Type of change
- [X] Bug fix

- [X] Write unittests. (Making sure that existing conif tests are used.)

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2154,7 +2154,12 @@ class TestHIGHS:
         ],
     )
     def test_highs_solving(self, problem) -> None:
-        problem(solver=cp.HIGHS)
+        # HACK needed to use the HiGHS conic interface rather than 
+        # the QP interface for LPs.
+        from cvxpy.reductions.solvers.conic_solvers.highs_conif import HIGHS
+        solver = HIGHS()
+        solver.name = lambda: "HIGHS conic"
+        problem(solver=solver)
 
     @pytest.mark.parametrize(
         ["problem", "confirmation_string"],
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2158,7 +2158,7 @@ class TestHIGHS:
         # the QP interface for LPs.
         from cvxpy.reductions.solvers.conic_solvers.highs_conif import HIGHS
         solver = HIGHS()
-        solver.name = lambda: "HIGHS conic"
+        solver.name = lambda: "HIGHS CONIC"
         problem(solver=solver)
 
     @pytest.mark.parametrize(
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2216,6 +2216,33 @@ class TestHIGHS:
         }
         problem(solver=cp.HIGHS, highs_options=highs_options)
 
+    def test_highs_cvar(self) -> None:
+        """Test problem with CVaR constraint from 
+        https://github.com/cvxpy/cvxpy/issues/2836
+        """
+        # Generate data
+        num_stocks = 5
+        num_samples = 25
+        np.random.seed(1)
+        pnl_samples = np.random.uniform(low=0.0, high=1.0, size=(num_samples, num_stocks))
+        pnl_expected = pnl_samples.mean(axis=0)
+
+        # Prepare to solve
+        quantile = 0.05
+        w = cp.Variable(num_stocks, nonneg=True)
+        cvar = cp.cvar(pnl_samples @ w, 1 - quantile)
+        pnl = w @ pnl_expected
+
+        # Solve
+        objective = cp.Maximize(pnl)
+        constraints = [cvar <= 0.5]
+        problem = cp.Problem(objective, constraints)
+        problem.solve(
+            solver=cp.HIGHS,
+            verbose=True,
+        )
+        assert problem.status == cp.OPTIMAL
+
 
 class TestAllSolvers(BaseTest):
 
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2216,33 +2216,6 @@ class TestHIGHS:
         }
         problem(solver=cp.HIGHS, highs_options=highs_options)
 
-    def test_highs_cvar(self) -> None:
-        """Test problem with CVaR constraint from 
-        https://github.com/cvxpy/cvxpy/issues/2836
-        """
-        # Generate data
-        num_stocks = 5
-        num_samples = 25
-        np.random.seed(1)
-        pnl_samples = np.random.uniform(low=0.0, high=1.0, size=(num_samples, num_stocks))
-        pnl_expected = pnl_samples.mean(axis=0)
-
-        # Prepare to solve
-        quantile = 0.05
-        w = cp.Variable(num_stocks, nonneg=True)
-        cvar = cp.cvar(pnl_samples @ w, 1 - quantile)
-        pnl = w @ pnl_expected
-
-        # Solve
-        objective = cp.Maximize(pnl)
-        constraints = [cvar <= 0.5]
-        problem = cp.Problem(objective, constraints)
-        problem.solve(
-            solver=cp.HIGHS,
-            verbose=True,
-        )
-        assert problem.status == cp.OPTIMAL
-
 
 class TestAllSolvers(BaseTest):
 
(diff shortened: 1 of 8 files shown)


senhalil:
- I didn't have a chance to add a test
- I have a question about the primal solution mapping. It looks like current version should not work at all but it evidently works so I am confused about this lack of `inverse_data` mapping in the primal solution dict

Transurgeon:
thanks @senhalil , could you add a test for this? I think you can just copy the failing snippet from the issue.

SteveDiamond:
I'm confused on why the existing HIGHS tests in cvxpy/tests/test_conic_solvers.py didn't catch the lack of dual variables.

rileyjmurray:
@SteveDiamond, [the HiGHS tests](https://github.com/senhalil/cvxpy/blob/d7f0f52944c828b8ce1564906faa45e7f9497541/cvxpy/tests/test_conic_solvers.py#L2138-L2155) use LP0 through LP5. The specifications of those tests can be found [here](https://github.com/senhalil/cvxpy/blob/d7f0f52944c828b8ce1564906faa45e7f9497541/cvxpy/tests/solver_test_helpers.py#L257-L340) (although the _function definitions_ for the tests are later in solver_test_helpers.py).
Here's a summary of those test cases:
* LP0 includes equality constraints (only) and _does not_ specify the optimal dual variables.
* LP1 and LP2 have inequality constraints (only) and _do_ specify the optimal dual variables.
* LP5 has inequality and equality constraints and _does not_ specify the optimal dual variables.
 
I'd expect that LP5 would have triggered the error reported in https://github.com/cvxpy/cvxpy/issues/2836 even though it didn't check for specific values of the dual variables. Logs from a recent run of unit tests on master ([see here](https://github.com/cvxpy/cvxpy/actions/runs/15751624329/job/44397776638#step:9:83)) indicate that HIGHs tests were run (insofar as they were not listed as having been skipped). So I'm unsure of why the we didn't see this earlier.

SteveDiamond:
> @SteveDiamond, [the HiGHS tests](https://github.com/senhalil/cvxpy/blob/d7f0f52944c828b8ce1564906faa45e7f9497541/cvxpy/tests/test_conic_solvers.py#L2138-L2155) use LP0 through LP5. The specifications of those tests can be found [here](https://github.com/senhalil/cvxpy/blob/d7f0f52944c828b8ce1564906faa45e7f9497541/cvxpy/tests/solver_test_helpers.py#L257-L340) (although the _function definitions_ for the tests are later in solver_test_helpers.py). Here's a summary of those test cases:
> 
> * LP0 includes equality constraints (only) and _does not_ specify the optimal dual variables.
> * LP1 and LP2 have inequality constraints (only) and _do_ specify the optimal dual variables.
> * LP5 has inequality and equality constraints and _does not_ specify the optimal dual variables.
> 
> I'd expect that LP5 would have triggered the error reported in #2836 even though it didn't check for specific values of the dual variables. Logs from a recent run of unit tests on master ([see here](https://github.com/cvxpy/cvxpy/actions/runs/15751624329/job/44397776638#step:9:83)) indicate that HIGHs tests were run (insofar as they were not listed as having been skipped). So I'm unsure of why the we didn't see this earlier.

I've been digging into the LP tests, and have exposed several issues:
1. The HiGHS LP tests in test_conic_solvers are using the HiGHS QP interface. So the conic interface dual variables are not being tested. It should be possible to modify the tests to use the conic interface.
2. Regarding https://github.com/cvxpy/cvxpy/issues/2836, cvar (which wraps dotsort) has a bug such that it is not being flagged as piecewise linear. I will push a fix to this MR.

SteveDiamond:
I pushed a fix to the conic solver tests so that the HiGHS conic interface is being tested for LPs. I also pushed from fixes for dotsort. However, #2836 now returns infeasible when I run it, so there may be another bug somewhere. If we decide to split out the fixes to dotsort into another MR that's fine.

SteveDiamond:
Problem #2836 is feasible with other solvers, so looks like a bug in highs_conif.py

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2154,7 +2154,12 @@ class TestHIGHS:
         ],
     )
     def test_highs_solving(self, problem) -> None:
-        problem(solver=cp.HIGHS)
+        # HACK needed to use the HiGHS conic interface rather than 
+        # the QP interface for LPs.
+        from cvxpy.reductions.solvers.conic_solvers.highs_conif import HIGHS
+        solver = HIGHS()
+        solver.name = lambda: "HIGHS conic"
```
senhalil: @SteveDiamond is this name modification needed? It breaks the CI [here](https://github.com/cvxpy/cvxpy/actions/runs/15832068239/job/44626930377?pr=2838#step:9:55) 

I'd expect  `from cvxpy.reductions.solvers.conic_solvers.highs_conif import HIGHS` to be enough to ensure that HiGHS conic interface is used. Is this not the case?

In any case, good find! I was always confused by the fact that some of the HiGHS conif tests calling the QP solver but I assumed that it was the intended behavior and I wrote tests with and without binary variables to make sure that both highs_conif and highs_qpif interfaces were called 😬 I should have asked!

Let me know if you need more from my side but as far as I'm concerned this is ready to be reviewed and can be merged once you and the CI are happy.

SteveDiamond: > @SteveDiamond is this name modification needed? It breaks the CI [here](https://github.com/cvxpy/cvxpy/actions/runs/15832068239/job/44626930377?pr=2838#step:9:55)
> 
> I'd expect `from cvxpy.reductions.solvers.conic_solvers.highs_conif import HIGHS` to be enough to ensure that HiGHS conic interface is used. Is this not the case?

There is some internal accounting that errors if you pass in a solver with the same name as an existing solver. I can use a slightly different hack so that the CI still passes.

> 
> In any case, good find! I was always confused by the fact that some of the HiGHS conif tests calling the QP solver but I assumed that it was the intended behavior and I wrote tests with and without binary variables to make sure that both highs_conif and highs_qpif interfaces were called 😬 I should have asked!
> 
> Let me know if you need more from my side but as far as I'm concerned this is ready to be reviewed and can be merged once you and the CI are happy.

Ok I am concerned though about the correctness of the highs_conif interface. If you base on this branch, you will see that the problem in https://github.com/cvxpy/cvxpy/issues/2836 returns unbounded. For other solvers the problem gives a consistent optimal value, which suggests that the issue is specifically with highs_conif. Do you have bandwidth to investigate?

SteveDiamond: I will add a failing test so this issue is more clear.

Transurgeon:
@SteveDiamond to not clutter this PR, I think we should set the new failing test as skip for now and tackle it in another PR. 

I think the changes to the HiGHS interface fix the original issue about the dual values not being propagated. Thanks again @senhalil !

PTNobel:
>  I think we should set the new failing test as skip for now and tackle it in another PR.

I disagree. I think we should at least diagnose the issue and make sure it isn't an interface bug before merging more changes to the interface.

SteveDiamond:
I dug more into this and uncovered some more gems. The good news is there is no issue with the HiGHS interfaces. We can merge this MR. The bad news is that the QP pathway might be buggy. The dotsort graph implementation was using a variable with a nonneg attribute. This attribute was being ignored because the QP pathway processes attributes before the main  canonicalization/reduction step, whereas the conic pathway processes attributes after canonicalization. I changed the dotsort graph implementation to use an explicit constraint.

@senhalil if you're interested in passing variable bounds directly to HiGHS check out the changes to highs_qpif.py in https://github.com/cvxpy/cvxpy/pull/2838/commits/3f14812baeb47490aa3f03d96afe6663a46103d9 I reverted the change in this MR because it is out of scope.

SteveDiamond:
I looked over all the graph implementations for piecewise linear atoms and there are no other variables with attributes, so the QP reduction pathway is fine for now. We may want to reorder the QP solving chain from
```
        reductions += [
            CvxAttr2Constr(reduce_bounds=not solver_instance.BOUNDED_VARIABLES),
            qp2symbolic_qp.Qp2SymbolicQp(),
            QpMatrixStuffing(canon_backend=canon_backend),
            solver_instance,
        ]
 ```
to 
```
        reductions += [
            qp2symbolic_qp.Qp2SymbolicQp(),
            CvxAttr2Constr(reduce_bounds=not solver_instance.BOUNDED_VARIABLES),
            QpMatrixStuffing(canon_backend=canon_backend),
            solver_instance,
        ]
 ```

PTNobel (approved):

PR #2839, opened 2025-06-20
Only warns about `.value` for sparse leafs when it causes significant fillin

PTNobel:
## Description

This makes us only warn about `.value` for sparse leafs when it causes significant fill-in. I'm not sure about my exact metric for fill-in, right now I'm doing 3 x as many 0s as my cut off. I'm open to making it an absolute number of zeros being added or some mix of the two. (more than 100 0s added and 2x as much perhaps?)

@rileyjmurray this feels like something you'd care about.

Issue link (if applicable): https://github.com/cvxpy/cvxpy/discussions/2782

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -145,7 +145,7 @@ class Leaf(expression.Expression):
             self.integer_idx = integer
         if sparsity:
             self.sparse_idx = self._validate_indices(sparsity)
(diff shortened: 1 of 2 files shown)


Transurgeon (approved):
Looks good! I think the less than a quarter of entries being non-zero check seems reasonable. But maybe we should also advertise more clearly somewhere that the ``value_sparse`` field exists.

PR #2846, opened 2025-06-28
Fix docstring in Constraint.violation: remove mention of squaring

napinoco:
## Description
This PR fixes a docstring typo in `cvxpy/constraints/constraint.py`.
The description incorrectly stated that the residual is squared, but in fact it is not.

For example, the `violation` methods in other constraint classes such as [`SecondOrderCone`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/constraints/second_order.py#L105) and [`NonPos`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/constraints/nonpos.py#L95) compute the norm via `np.linalg.norm(x, ord=2)` without squaring the result.

## Type of change
- [ ] Bug fix

- [X] Write unittests.

### DIFF
--- a/cvxpy/constraints/constraint.py
+++ b/cvxpy/constraints/constraint.py
@@ -149,7 +149,7 @@ class Constraint(u.Canonical):
 
         .. math::
 
-            ||\\Pi(v) - v||_2^2
+            ||\\Pi(v) - v||_2
 
         where :math:`v` is the value of the constrained expression and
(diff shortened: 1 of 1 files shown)


PTNobel:
Zero and Equality are also not squared

PTNobel (approved):

PTNobel:
I checked a few more cones and none seem to be squared

PR #2847, opened 2025-06-30
Doc improvements for some atoms, plus an Expression.cast function.

### LINKED ISSUE #2644: Documentation needed for quantum_cond_entr
https://github.com/cvxpy/cvxpy/blob/14124e110fc1efe0e1cfac1be8414ad9109969ff/cvxpy/atoms/quantum_cond_entr.py#L12-L20

@aryamanjeendgar this atom is not properly documented. It also fails silently if sys is neither 0 nor 1. Please fix.

rileyjmurray:
This started just as a plan to resolve #2644. I ended up making documentation changes in quite a few atoms related to quantum information and quadrature-based approximations of relative entropy cones.

More material changes:
* von_neumann_entr now checks that the input is Hermitian, rather than merely square.
* quantum_rel_entr's .value function now references an EVAL_TOL static class variable to determine how to handle small negative eigenvalues.
* adds a new Expression.cast function. This is just a wrapper around Expression.cast_to_const, which we use all over the place in ways that don't make sense given the function's name. I think it'd be good to replace Expression.cast_to_const with Expression.cast systematically, but that's better left to another PR.

I noticed that quantum relative entropy tests take _forever_ to run when MOSEK is installed, since in those cases we let ourselves try and run larger tests. It might be wise for us to disable those tests unless specifically requested. If someone has an idea of how we'd do that, I'd appreciate it if you could open an issue proposing that change.

I'm also seeing weird test failures about error messages not having the exact form the tests expect. Those failures are unrelated to my changes, so I haven't tried to fix them.

### DIFF
--- a/cvxpy/atoms/quantum_rel_entr.py
+++ b/cvxpy/atoms/quantum_rel_entr.py
@@ -19,22 +19,40 @@ import numpy as np
 from scipy import linalg as LA
 from scipy.stats import entropy
 
+from cvxpy import settings
 from cvxpy.atoms.atom import Atom
 from cvxpy.constraints.constraint import Constraint
 
 
 class quantum_rel_entr(Atom):
     """
+    An approximation of the quantum relative entropy between systems with (possibly un-normalized) 
+    density matrices :math:`X` and :math`Y:`
+     
+    .. math::
+        \\operatorname{tr}\\left( X ( \\log X - \\log Y ) \\right).
+      
+    The approximation uses a quadrature scheme described in https://arxiv.org/abs/1705.00812.
+
     Parameters
     ----------
     X : Expression or numeric
         A PSD matrix
     Y : Expression or numeric
         A PSD matrix
+    quad_approx : Tuple[int, int]
+        quad_approx[0] is the number of quadrature nodes and quad_approx[1] is the number of scaling
+        points in the quadrature scheme from https://arxiv.org/abs/1705.00812.
+
+    Notes
+    -----
+    This function does not assume :math:`\\operatorname{tr}(X)=\\operatorname{tr}(Y)=1,` which
+    would be required for most uses of this function in the context of quantum information.
     """
 
+    EVAL_TOL = min(settings.ATOM_EVAL_TOL, 1e-6)
+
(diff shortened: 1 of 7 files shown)


cvxpy/atoms/quantum_rel_entr.py
```diff
@@ -19,22 +19,40 @@
...
+    Notes
+    -----
+    This function does not assume :math:`\\operatorname{tr}(X)=\\operatorname{tr}(Y)=1,` which
+    would be required for most uses of this function in the context of quantum information.
     """
 
+    EVAL_TOL = min(settings.ATOM_EVAL_TOL, 1e-6)
+
     def __init__(self, X, Y, quad_approx: Tuple[int, int] = (3, 3)) -> None:
-        # TODO: add a check that N is symmetric/Hermitian.
```
PTNobel, line 37: Where is this TODO implemented?

PTNobel (approved):
I'm not very knowledgeable about the relevant math, but the code seems seems fine from a style perspective

cvxpy/atoms/quantum_rel_entr.py
rileyjmurray: A call to ``quantum_rel_entr.validate_arguments`` gets triggered by ``super(quantum_rel_entr, self).__init__(X, Y)`` in quantum_rel_entr's ``__init__``. This had already been the case, so the TODO was stale.

cvxpy/atoms/quantum_rel_entr.py
```diff
@@ -49,12 +67,11 @@ def numeric(self, values):
...
-            assert np.all(x >= 0)
-            x_sum = x.sum()
-            val = -entropy(x)
-            un_normalized = (x_sum * val + np.log(x_sum)*x_sum)
-            return un_normalized
+        if np.any(w1 < - self.EVAL_TOL) or np.any(w2 < -self.EVAL_TOL):
+            return np.inf
+        else:
+            w1[w1 < 0] = 0
+            w2[w2 < 0] = 0
```
Transurgeon, line 74: I don't understand the changes here. Are these two implementations equivalent?
Also, where was func being called previously?

Transurgeon:
I think we should merge this, my review comment was more for my understanding. But I trust that Riley knows what he is doing.

PR #2825, opened 2025-06-08
Replace custom axis handling with numpy.array_utils.normalize_axis_*

7astro7:
Addresses issue #2816 

- Replaced manual axis checks with:
  - `normalize_axis_index` in `partial_trace.py` and `partial_transpose.py`
  - `normalize_axis_index` and `normalize_axis_tuple` in `axis_atom.py`
- Removed redundant `ValueError` checks for out-of-bounds axes
- Removed tests that manually validated axis bounds (since NumPy now handles this)

## Description
Please include a short summary of the change.
Issue link (if applicable):
Please see above. 

## Type of change
- [x] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [] Run the benchmarks to make sure your change doesn’t introduce a regression.

FYI I did not see the benchmark test.

PR #2865, opened 2025-07-16
BUG: raise ValueError for invalid axis argument to norm

Transurgeon:
## Description
Straightforward from title.
However, I am thinking maybe the norm class is supposed to sub-class from AxisAtom? Then the validation could be done there? Not sure if its really important though. (edit: NVM, the norm file is just a factory class that creates pnorm atoms which do subclass from AxisAtom)

Issue link (if applicable): someone posted this on Ed for the 364a summer class. Thanks for pointing this @dance858!

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/atoms/norm.py
+++ b/cvxpy/atoms/norm.py
@@ -67,6 +67,8 @@ def norm(x, p: Union[int, str] = 2, axis=None, keepdims: bool = False):
         else:
             raise RuntimeError('Unsupported matrix norm.')
     else:
+        if not isinstance(axis, int):
+            raise ValueError(f'Axis must be an integer, not {type(axis)}.')
         if p == 1 or x.is_scalar():
             return norm1(x, axis=axis, keepdims=keepdims)
         elif str(p).lower() == "inf":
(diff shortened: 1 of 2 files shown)


PTNobel:
Shouldn't we support `axis` being a tuple of ints? What is the bug here?

PTNobel (changes requested):
Blocking merging of this for now.

Transurgeon:
> Shouldn't we support `axis` being a tuple of ints? What is the bug here?

oh right definitely. I think the bug is that we are currently able to define a norm2 with a variable as the axis argument.

PTNobel:
So notably it doesn't work if Variable isn't a scalar...

The bug is on this line: https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/axis_atom.py#L79

We could do the following: check if axis is a CVXPY expression and error saying "please don't" and we should also add the error NumPy has on empty axes

PTNobel:
#2825 seems to be a better way to fix this.

Transurgeon:
> #2825 seems to be a better way to fix this.

yeah agreed, let me close this PR.

Issue #2876, opened 2025-07-27
Abstract method `cite` in `Solver` class breaks `semver` guarantees for CVXPY 1.7


enzbus:
Hello dear `CVXPY` developers. Just noting that enforcing that new `cite` method is abstract, in CVXPY 1.7, breaks API in a way that is incompatible with semantic versioning.

I'm referring to this line: https://github.com/cvxpy/cvxpy/blob/e6d60294a0051c20bf1f0ad1032b3632559c8054/cvxpy/reductions/solvers/solver.py#L75

The API that breaks is a user-defined custom solver class, which I take to be part of the public API, [since it is documented here](https://www.cvxpy.org/tutorial/solvers/index.html#custom-solvers).

**Suggested solution**
Remove the `@abc.abstractmethod` decorator, using the default implementation, which raises `NotImplementedError` on usage of the `cite` method in custom solver classes. This ensures that custom solver classes that worked with CVXPY 1.6 don't break with 1.7.

Transurgeon:
let's close this issue per the comments in #2877

PR #2877, opened 2025-07-31
Fix `cite()` semver issue

govindchari:
## Description
Fix semver issue described here: https://github.com/cvxpy/cvxpy/issues/2876

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/solver.py
+++ b/cvxpy/reductions/solvers/solver.py
@@ -72,7 +72,6 @@ class Solver(Reduction):
         """
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def cite(self, data):
         """Return bibtex citation for the solver.
         """


enzbus:
Thank you! In any case technically this is a regression/bug fix :) (EDIT: Ok, minor.)

SteveDiamond:
Thanks @govindchari! The maintainers have been discussing this issue, and our position is that semantic versioning only covers the user facing API (constructing variables, expressions, problems, etc.). The developer API for custom solvers is not covered by semantic versioning. We are happy to discuss more in the discord with you or @enzbus

enzbus:
Thank you. I don't have any arguments against your position, as it's your choice where to draw the line between public and private APIs. I found out about the `Solver` base class(es) only recently, from the documentation. Before reading that I was assuming it was private. Thank you anyways for accommodating my request.

PR #2872, opened 2025-07-24
Fix status mapping issues

govindchari:
## Description
Fix various deprecation warnings in Xpress. Currently fixed the status mapping warnings. Still getting the following warnings:

`/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:231: DeprecationWarning: Deprecated in Xpress 9.5: use problem.attributes.solvestatus and problem.attributes.solstatus instead
  results_dict['status'] = self.prob_.getProbStatus()
/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:232: DeprecationWarning: Deprecated in Xpress 9.5: use problem.attributes.solvestatus and problem.attributes.solstatus instead
  results_dict['getProbStatusString'] = self.prob_.getProbStatusString()
/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:233: DeprecationWarning: Deprecated in Xpress 9.5: use problem.attributes.objval instead
  results_dict['obj_value'] = self.prob_.getObjVal()
/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:249: DeprecationWarning: Deprecated in Xpress 9.5: use problem.attributes.solvestatus and problem.attributes.solstatus instead
  results_dict['getProbStatusString'] = self.prob_.getProbStatusString()
/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:252: DeprecationWarning: Deprecated in Xpress 9.5: use problem.attributes.objval instead
  results_dict['getObjVal'] = self.prob_.getObjVal()
/home/govind/Desktop/git/cvxpy/cvxpy/reductions/solvers/qp_solvers/xpress_qpif.py:256: DeprecationWarning: Deprecated in Xpress 9.5: use problem.getDuals instead
  results_dict['getDual'] = self.prob_.getDual()
`
Would like some input on the cleanest way of fixing these. The obvious way is to do a version check before calling the getXXX() functions but that seems sloppy.

Issue link (if applicable):

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
[...]

PR #2878, opened 2025-07-31
CI: update macOS runners to ``macos-14``

Transurgeon:
## Description
This [official github blog](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down) states that the ``macos-13`` action runners are being deprecated on September 1st, 2025. However, they are the last runners that are on x86 intel machines, see this [discussion](https://github.com/astropy/astropy/issues/18456). 

Issue link (if applicable): 
Per discussion in #2872

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, macos-13, windows-2022 ]
+        os: [ ubuntu-22.04, macos-14, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12", "3.13" ]
         include:
           # These are intended to just add their extra parameter to existing matrix combinations;
@@ -83,7 +83,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, macos-13, windows-2022 ]
+        os: [ ubuntu-22.04, macos-14, windows-2022 ]
         python-version: [ 3.9, "3.10", "3.11", "3.12", "3.13" ]
         build-cvxpy-base: [ true, false ]  # whether to build cvxpy-base (true) or regular cvxpy (false)
         include:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
             python-version: 3.12
             openmp: "True"
             single_action_config: false
-          - os: macos-13
+          - os: macos-14
             python-version: 3.12
             single_action_config: true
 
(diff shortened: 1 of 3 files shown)


yemine0x01:
Is possible also to have this in the `test_optional_solvers` CI? and also to have an ARM architecture?

Transurgeon:
> Is possible also to have this in the `test_optional_solvers` CI? 

yes. Will update the PR now.

> and also to have an ARM architecture?

the macos-14 option already adds an ARM runner. See the table of runners [here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images). 

but first there is some debugging to do with regards to installing the ``mkl`` package (which can only be installed on intel machines).

Transurgeon:
@usamamuneeb sorry for the ping.. but had a quick question for you.

The changes in this PR led to a failing SDPA test for macos-ARM machines (the macos-14 configuration for the runner defaults to ARM). I first ran into an issue that the ``mkl`` library couldn't be installed on the ARM machines (which makes sense), but after skipping the installation I get this failing SDPA test. 

My assumption is that SDPA is using ``mkl``'s linear solver and not installing it leads to a regression? Could you look into this please? Is there a way to make this test pass by installing an alternative to ``mkl`` for macos ARM?

usamamuneeb:
Hello @Transurgeon, great to hear from you!

The problem as usual is the fragile LP5 test which has always been very sensitive for `sdpa-python`, despite parameter tuning. This is the reason in the past we would run LP5 test only on `sdpa-multiprecision`, as that would usually handle it in all environments. We used to skip LP5 for `sdpa-python` (as it passed in some, failed in others).

I doubt the MKL dependency issue you saw was from SDPA. The `sdpa-python` wheels (which you are using) are themselves built natively on the ARM based `macos-14` runners (unlike `sdpa-multiprecision` which are cross compiled on the Intel based `macos-13` runners). That being said, none of SDPA wheels should have an MKL dependency.

Transurgeon:
> I doubt the MKL dependency issue you saw was from SDPA. The `sdpa-python` wheels (which you are using) are themselves built natively on the ARM based `macos-14` runners (unlike `sdpa-multiprecision` which are cross compiled on the Intel based `macos-13` runners). That being said, none of SDPA wheels should have an MKL dependency.

ok great to hear.. I guess we can merge this PR then. But it is still weird that this particular test is consistently failing.

usamamuneeb:
If you'd also like to silence it, you can perhaps wrap this test with (1) a check that you are using `sdpa-multiprecision` package, or (2) if `sdpap.sdpacall.sdpacall.get_backend_info()["gmp"]` returns True, whichever is convenient.

I couldn't quite find a set of parameters that will make LP5 for SDPA pass across all environments (and there seem to be factors other than just the BLAS implementation that introduce inconsistencies).

Issue #2290, opened 2023-11-19
Adding cp.squeeze


Transurgeon:
**Is your feature request related to a problem? Please describe.**
Missing squeeze operation for full compatibility with NumPy. This currently has the same functionality as flatten since cvxpy expressions are limited to 2-dimensions. However with the addition of nd-expressions, the squeeze operation could be helpful to ensure shape consistency. 

**Describe the solution you'd like**
Implement an equivalent operation to [np.squeeze](https://numpy.org/doc/stable/reference/generated/numpy.squeeze.html) in cvxpy.

yemine0x01:
Is this still of interest? I'd like to give it a try.

SteveDiamond:
It's still of interest. Contributions are very helpful.

yemine0x01:
I’ve been thinking about this and was considering implementing a `Squeeze` class as a subclass of `AffAtom`, similar to how the `reshape` class is structured. Does this approach make sense, or am I heading in the wrong direction?

Transurgeon:
@eminyouskn yes I think your approach makes sense! In fact, I think squeeze can even just be a wrapper around reshape? But you would need to compute the output shape by iterating over all dimensions and removing the ones.

yemine0x01:
@Transurgeon I implement it as you suggested. Thanks.

yemine0x01:
I think we can close this issue.

Transurgeon:
thanks @eminyouskn nice work!

PR #2887, opened 2025-08-08
Squeeze

eminyouskn:
## Description
Implement `cp.squeeze` #2290 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/atoms/affine/squeeze.py
```diff
@@ -0,0 +1,68 @@
...
+limitations under the License.
+"""
+from __future__ import annotations
+
+import numbers
+from typing import Tuple
+
+from cvxpy.atoms.affine.reshape import reshape
+
+
```
PTNobel, line 23: Thoughts on implementing this as a function instead of a class? I can write an example in an hour or two when I get to a keyboard.

cvxpy/atoms/affine/squeeze.py
```diff
@@ -0,0 +1,68 @@
+"""
+Copyright 2013 Steven Diamond
```
PTNobel: Update year and put The CVXPY Developers

yemine0x01: Will fix it. I just copy the reshape file.

cvxpy/atoms/affine/squeeze.py
yemine0x01: Are you saying that the reshape need to be a function or the squeeze is just a function wrapper around reshape class?

PTNobel: I think squeeze can just be a function wrapper around reshape!

PTNobel (approved):
Looks good to merge once CI finishes! Thanks for doing this!

PR #2888, opened 2025-08-08
Enable Support for Advanced HIGHS Options

ghackebeil:
This pull requests enables support for advanced HIGHS solver options that are not accessible through the HighsOptions class. If we just call `solver.setOptionValue(key, value)`, we are able to set the same options on the HighsOptions class (e.g., threads) as well as additional options that are not included there (e.g., presolve_rule_off).

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


ghackebeil:
Looks like I need to do some additional error checking and raise on bad option names/values.

ghackebeil:
failure seems to be from an unrelated flaky test

Transurgeon (approved):
LGTM, thanks @ghackebeil !

PR #2897, opened 2025-08-14
Update inv_prod.py

PTNobel:
## Description
Eliminates the double call to power in inv_prod, so it is canonicalized as a single power(geo_mean) instead of power(power(geo_mean).

## Type of change
- [x] Bug fix

- [ ] Write unittests.


Transurgeon (approved):
LGTM and nice catch!

PR #2898, opened 2025-08-14
QPALM interface

tttapa:
## Description
Adds an interface for the QPALM solver (https://github.com/kul-optec/QPALM).

## Type of change
- [ ] Bug fix

- [x] Write unittests.

One thing that's unclear to me is how to add QPALM to the GitHub Actions CI. IIUC, only the optional conic solvers are tested in the CI, but not the optional QP solvers.


cvxpy/reductions/solvers/qp_solvers/qpalm_qpif.py
```diff
@@ -0,0 +1,138 @@
...
+
+        qp_data = qpalm.Data(n_var, n_con)
+        qp_data.Q = sp.triu(P).tocsc()
+        qp_data.q = q
+        qp_data.A = A
+        qp_data.bmin = b_min
+        qp_data.bmax = b_max
+
+        settings = qpalm.Settings()
+        settings.max_iter = 1000
```
PTNobel: Do all these setting have to be specified by us? Does qpalm not set defaults? I'd prefer not to have to maintain the solver options.

cvxpy/tests/test_constant_atoms.py
```diff
@@ -30,13 +30,14 @@
...
+from cvxpy.settings import CLARABEL, CVXOPT, MOSEK, OSQP, QPALM, ROBUST_KKTSOLVER, SCS
 
 ROBUST_CVXOPT = "robust_cvxopt"
 SOLVER_TO_TOL = {SCS: 1e-2,
                  CLARABEL: 1e-7,
-                 OSQP: 1e-1}
-SOLVERS_TO_TRY = [CLARABEL, SCS, OSQP]
+                 OSQP: 1e-1,
+                 QPALM: 1e-7}
+SOLVERS_TO_TRY = [CLARABEL, SCS, OSQP, QPALM]
```
PTNobel: This structure is reserved for CVXPY default solvers only. Is there a reason we need to run these tests?

Even if there is a reason, the solver should be conditionally added to SOLVERS_TO_TRY based on if it is installed.

cvxpy/tests/test_qp_solvers.py
```diff
@@ -459,6 +459,29 @@ def test_warm_start(self) -> None:
         result2 = prob.solve(solver="OSQP", warm_start=False)
         self.assertAlmostEqual(result, result2)
 
+    def test_qpalm_warmstart(self) -> None:
```
PTNobel, line 462: Please add more tests, comparable to the other QP solvers.

PTNobel (changes requested):
Mostly fine, a few small nits and some questions.

PTNobel:
aditionally, please add installing qpalm to the install_optional_solvers script

tttapa:
Thanks! I've pushed the requested changes.

cvxpy/reductions/solvers/qp_solvers/qpalm_qpif.py
tttapa: QPALM's default tolerances are quite lax. As a result, some of the solutions it produces using those tolerances are not accurate enough for all tests to pass. The tolerance values here are chosen to match PIQP, which is also a second-order solver. I'll add this as a comment, and remove the `max_iter` value, for which the default should suffice. Note that OSQP also requires tweaking the tolerances: 
https://github.com/cvxpy/cvxpy/blob/9b1f4628d78f4a29c667d0698ad1e29ea4c3982d/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L88-L91

cvxpy/tests/test_constant_atoms.py
tttapa: No particular reason. I've reverted this change.

cvxpy/tests/test_qp_solvers.py
tttapa: Could you clarify which tests you mean specifically? I'm grepping for PIQP and PROXQP, but the only solver-specific test seems to be the warm-starting ones. The normal QPALM solver test should be covered by `cvxpy/tests/test_qp_solvers.py::TestQp::test_all_solvers`.

PTNobel: You're totally right! I forgot how we handled QP tests. Sorry for the confusing comment.

PTNobel (approved):
Looks good

Issue #2851, opened 2025-07-06
OSQP 1.0 `PendingDeprecationWarning` on `solve` method `raise_error` default parameter


enzbus:
**Describe the bug**
OSQP 1.0 `solve` method throws a `PendingDeprecationWarning`, which is hidden by default (but is caught by warning filters, like what I had) on the `raise_error` parameter which will be required in the future.

Here's the OSQP code that throws the warning on missing default value: https://github.com/osqp/osqp-python/blob/a8b8eaefc3bb3bccaa992e2b2e456bd6cc8fcfaa/src/osqp/interface.py#L403

Here's where that method is invoked in CVXPY, with no default value passed: https://github.com/cvxpy/cvxpy/blob/2cb7206b4cf635f42308d75dffa6640dc5408c7b/cvxpy/reductions/solvers/qp_solvers/osqp_qpif.py#L132

**To Reproduce**
``` python
import warnings
import cvxpy as cp
warnings.simplefilter('error')
cp.Problem(cp.Minimize(0.), [cp.Variable() == 0]).solve(solver='OSQP')
```

Prints (end of stack trace):

``` text
    warnings.warn(
    ~~~~~~~~~~~~~^
        'The default value of raise_error will change to True in the future.',
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        PendingDeprecationWarning,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
PendingDeprecationWarning: The default value of raise_error will change to True in the future.
```

**Expected behavior**
OSQP interface in CVXPY should pass `raise_error=False` (current behavior) or manage the exception(s) thrown.

**Version**
 - OSQP Version: 1.0
 - CVXPY Version: latest

**Additional context**

Transurgeon:
this has been fixed and should be available in the 1.7.3 release!

PR #2901, opened 2025-08-16
Pass in raise_error=False to osqp solve method

govindchari:
## Description
Pass in `raise_error=False` to osqp solve method.

In order to keep existing behavior as mentioned in the issue linked below, we must pass in `raise_error=False`.

Note I bumped the required version of OSQP to >=1.0.0, since it is the first version of OSQP that has the argument `raise_error`.

If anyone feels strongly about this version bump, we can keep the version the same, but have a version check before the `.solve()` call. However, if we go down this path, the minimum version of OSQP should still be bumped to `>=0.6.3` since osqp version 0.6.2 (which is currently the minimum version supported in `pyproject.toml`) throws the error `AttributeError: module 'osqp' has no attribute '__version__'` when its version is queried.

Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/2851

## Type of change
- [x] Bug fix

- [ ] Write unittests.


PTNobel (approved):
The version bump is fine by me! I'll let someone else weigh in before merging

SteveDiamond (approved):
Bumping the version is fine with me too

Issue #2533, opened 2024-08-16
Max Iter seems to be handled differently across solvers


vhartman:
Hey,
First, thanks for the work!

**Description:**
Reaching the maximum number of iterations is treated as error in the OSQP-interface (and in most other solvers).
In e.g. CLARABEL, gurobi, it is not treated as an error, but as USER_LIMIT.

Is this intended behavior, or is s.USER_LIMIT simply a newer addition, and adapting previous solvers was not done yet, or is this an intended design choice?

**Background:** 
In model predictive control, I might want to do only N iterations of the solver, and I do not expect the solver to converge at all, so I would not treat this as an error. When using e.g. OSQP directly, this is possible, since the status is not reported as error. In cvxpy, the max_iter_reached is mapped to Error, which raises an exception, and makes the previously specified behavior impossible. (I am very much aware that doing real time iterations in python is likely not very sensible)

Thanks!

SteveDiamond:
As you guessed, USER_LIMIT is a more recent addition to the status codes. It would be reasonable for the solvers to be more consistent about using the USER_LIMIT status.

vhartman:
Thanks for getting back so quick. I'll have a look if I can get set up for contributing and make a PR.

PR #2906, opened 2025-08-19
If solver hits max iters it should return s.USER_LIMIT

govindchari:
## Description
If solver hits max iters it should return `s.USER_LIMIT`.

Another question is if the user interrupts the solver with ctrl+C, should the solver return `s.SOLVER_ERROR` or `s.USER_LIMIT`. Some interfaces do one, others do the other. 

Issue link (if applicable):
https://github.com/cvxpy/cvxpy/issues/2533

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


govindchari:
Tests are failing since there is a test that expects the solver to raise an error. This error is raised by setting max_iters to 1 in OSQP and solving a problem. However, by changing max iters to a user limit error rather than a solver error, this no longer raises an error so the test fails. Any ideas on what to do?

PTNobel:
For both of them you can use the problem: `cp.Problem(cp.Minimize(cp.quad_form(cp.Variable(1) + 1, np.array([[-1]]), True))).solve('OSQP')`

PR #2908, opened 2025-08-21
Grad bilinear

langestefan:
## Description

This change makes it possible to compute gradients for bilinear expressions. This gradient can then be used to linearize expressions that contain products of variables, such as in DCCP problems. 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond (approved):
Nice!

PTNobel (approved):

PR #2920, opened 2025-08-28
Implement Trace(AB) using vdot for performance efficiency

thisisRMak:
## Description
Implement Trace(AB) using vdot for performance efficiency. Move trace to trace_internal and use trace as a wrapper function that determines when to use vdot for a faster computation `vdot(A,B)` versus fall back to traditional trace `trace_internal`
Reference: https://github.com/cvxpy/cvxpy/discussions/2359
In experimental testing we see 
- 7x improvement on canonicalization of the following problem reducing execution time from 7.46 seconds to 1.05 seconds
- We also see 10x improvement in ``semidefinite_programming.SemidefiniteProgramming.time_compile_problem``
```py
import numpy as np
import cvxpy as cp
import time

# --- Step 1: Generate synthetic data ---
np.random.seed(0)
n = 500     # dimension (number of variables)
N = 100    # number of samples

# Create a random positive definite covariance matrix
A = np.random.randn(n, n)
true_cov = A @ A.T + n * np.eye(n)  # SPD matrix
X = np.random.multivariate_normal(mean=np.zeros(n), cov=true_cov, size=N)

# Sample covariance
S = np.cov(X, rowvar=False)

# --- Step 2: Define Graphical Lasso problem ---
lambda_reg = 0.1
K = cp.Variable((n, n), symmetric=True)

# Objective: -log det(K) + trace(SK) + λ * ||K||_1
log_det_term = -cp.log_det(K)
trace_term   = cp.trace(S @ K)
l1_penalty   = lambda_reg * cp.norm1(K)

objective = cp.Minimize(log_det_term + trace_term + l1_penalty)

constraints = [K >> 0]  # enforce K ≽ 0

start_time = time.perf_counter()
data, _, _ = prob.get_problem_data(solver=cp.CLARABEL)
end_time = time.perf_counter()
elapsed_time = end_time - start_time
print(f"Function execution time: {elapsed_time:.6f} seconds")
```

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/__init__.py
+++ b/cvxpy/atoms/__init__.py
@@ -35,7 +35,7 @@ from cvxpy.atoms.affine.reshape import deep_flatten, reshape
 from cvxpy.atoms.affine.squeeze import squeeze
 from cvxpy.atoms.affine.concatenate import concatenate
 from cvxpy.atoms.affine.sum import sum
-from cvxpy.atoms.affine.trace import trace, trace_internal
+from cvxpy.atoms.affine.trace import trace, Trace
 from cvxpy.atoms.affine.transpose import (transpose, permute_dims, swapaxes, moveaxis)
 from cvxpy.atoms.affine.upper_tri import upper_tri, vec_to_upper_tri
 from cvxpy.atoms.affine.vec import vec
--- a/cvxpy/atoms/__init__.py
+++ b/cvxpy/atoms/__init__.py
@@ -35,7 +35,7 @@ from cvxpy.atoms.affine.reshape import deep_flatten, reshape
 from cvxpy.atoms.affine.squeeze import squeeze
 from cvxpy.atoms.affine.concatenate import concatenate
 from cvxpy.atoms.affine.sum import sum
-from cvxpy.atoms.affine.trace import trace
+from cvxpy.atoms.affine.trace import trace, trace_internal
 from cvxpy.atoms.affine.transpose import (transpose, permute_dims, swapaxes, moveaxis)
 from cvxpy.atoms.affine.upper_tri import upper_tri, vec_to_upper_tri
 from cvxpy.atoms.affine.vec import vec
--- a/cvxpy/atoms/affine/trace.py
+++ b/cvxpy/atoms/affine/trace.py
@@ -24,7 +24,22 @@ from cvxpy.atoms.affine.binary_operators import MulExpression
 from cvxpy.constraints.constraint import Constraint
 
 
-class trace(AffAtom):
+def trace(expr):
+    """
+    TLDR: Use alternate formulation for Trace(A@B) for more efficient computation.
+    Trace(A@B) normally is O(n^3) because of the A@B operation. 
+    However, Trace(A@B) only requires diagonal entries of A@B, which can be
+    computed by taking the sum of element-wise product of A.T * B in O(n^2) time.
+    In fact, vdot does this operation more robustly, using conj(A) instead of transpose. 
+    """
+    if isinstance(expr, MulExpression): 
+        from cvxpy.atoms.affine.binary_operators import vdot
+        return vdot(expr.args[0], expr.args[1])
+    else:
+        return trace_internal(expr)
+
+
+class trace_internal(AffAtom):
     """The sum of the diagonal entries of a matrix.
 
     Parameters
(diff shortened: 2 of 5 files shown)


PTNobel:
Thanks for the contribution! We synced synchronously; I'll review once you have the new design.

thisisRMak:
> Thanks for the contribution! We synced synchronously; I'll review once you have the new design.

Thanks Parth. PR is updated per our discussion. Fewer changes than expected. 

Please let William and me know if you have suggestions to improve or edit further?

PTNobel (approved):

PTNobel:
I think DGP's canonicalizers need an update too, but I'll check tonight

PTNobel:
I think this usage also needs to be fixed: https://github.com/cvxpy/cvxpy/blob/9e51b84e3130d653a16efa5c8b80cd82a035eb01/cvxpy/reductions/dgp2dcp/canonicalizers/__init__.py#L7

Can we also test some complex matrices? I'm a bit worried that the vdot conj is incorrect

Transurgeon:
> I think this usage also needs to be fixed:
> 
> https://github.com/cvxpy/cvxpy/blob/9e51b84e3130d653a16efa5c8b80cd82a035eb01/cvxpy/reductions/dgp2dcp/canonicalizers/__init__.py#L7
> 
> Can we also test some complex matrices? I'm a bit worried that the vdot conj is incorrect

It is testing complex matrices already. The tests pass for TestComplex and also conic solver tests have a complex test case.
We noticed this because there were 4 failing tests without the change to the complex2real reduction.

thisisRMak:
Thanks Parth and William. I've made the change for dgp2dcp's canonicalizer. Looks like no other reduction's canonicalizer requires this change. No change in ruff and pytest locally. 

1. If the code change looks correct, could one of you please approve the workflows to run for the latest checkin?
2. No tests failed for `dgp2dcp/canonicalizers/__init__.py` earlier. That's why we missed it earlier. Are there tests we should be introducing to catch this? If so, could use some guidance on test cases? Or if that's not necessary for now, happy to come back and review/add tests at a later date also.
3. Like William said, 4 tests were failing for complex2real earlier. That's how we knew our changes in `complex2real/canonicalizers/__init__.py` were ok to checkin. Is that satisfactory for complex2real for our change? Or might there be something still missing in the complex2real test cases as they are now?

jonathanberthias:
Naming wise, what would you think of renaming `trace_internal` to `Trace` (with a capital T)? It's a pattern that is already used a little in the code base. Off the top of my head, there's `sum`/`Sum` and `hstack`/`Hstack`, probably a few others.

Transurgeon:
> Naming wise, what would you think of renaming `trace_internal` to `Trace` (with a capital T)? It's a pattern that is already used a little in the code base. Off the top of my head, there's `sum`/`Sum` and `hstack`/`Hstack`, probably a few others.

ohhh yeah really nice suggestion, I definitely like it!

about the dgp trace tests, I personally think its fine to not add one for this PR @PTNobel , I don't have the bandwith to come up with a test and we are almost certain the behavior will be the same as before. What do you think?

PTNobel:
> about the dgp trace tests, I personally think its fine to not add one for this PR @PTNobel , I don't have the bandwith to come up with a test and we are almost certain the behavior will be the same as before. What do you think?

I'd like a test; we can just have min cp.trace(X), subject to X == 2 and solve with dgp=True.

thisisRMak:
Thanks @PTNobel @Transurgeon @jonathanberthias . 
- [x] trace_internal is now renamed to Trace
- [x] added trace_test_dgp2dcp in test_atoms.py
  - Does it match what you had in mind @PTNobel ?
  - Do we want `trace_test_dgp2dcp` placed in `test_dgp.py` instead of `test_atoms.py`?
- [ ] Do we want a similar test added for `test_trace_complex2real`? 
  - I've verified this test case fails without our complex2real fix, but passes with it. I've pushed this change also so we can see if passing workflows, in case the group wants to include this. Plmk if we prefer it to be placed in `test_complex.py` instead of `test_atoms.py` as I've implemented it now?
  - Like @Transurgeon mentioned earlier, trace under complex2real functionality was already getting exercised under pytest, since these 4 tests failed earlier without the change to complex2real's canonicalizer. Now `test_trace_complex2real` makes it 5 test failures, but maybe its helpful if it explicitly pins the blame on trace when these tests fail together.

If changes in cvxpy/reductions/complex2real/canonicalizers/__init__.py are skipped

BEFORE adding `test_trace_complex2real`
```
FAILED cvxpy/tests/test_complex.py::TestComplex::test_log_det - AssertionError
FAILED cvxpy/tests/test_complex.py::TestComplex::test_special_idx - AssertionError
FAILED cvxpy/tests/test_conic_solvers.py::TestSCS::test_complex_matrices - AssertionError
FAILED cvxpy/tests/test_von_neumann_entr.py::Test_von_neumann_entr::test_1_complex - AssertionError
```

AFTER adding `test_trace_complex2real`
```
FAILED cvxpy/tests/test_atoms.py::TestAtoms::test_trace_complex2real - AssertionError
FAILED cvxpy/tests/test_complex.py::TestComplex::test_log_det - AssertionError
FAILED cvxpy/tests/test_complex.py::TestComplex::test_special_idx - AssertionError
FAILED cvxpy/tests/test_conic_solvers.py::TestSCS::test_complex_matrices - AssertionError
FAILED cvxpy/tests/test_von_neumann_entr.py::Test_von_neumann_entr::test_1_complex - AssertionError
```

thisisRMak:
There was a bug in test_trace_complex2real. Minimize needs a "Real" target. Added a norm like some other tests in test_complex.py seem to be doing. Wasn't a scientific decision, just poking around to see something that works. 

Suggestions welcome to either drop test_trace_complex2real completely or to do a different test than what I've hacked together here.

Issue #2806, opened 2025-05-09
signs not correct in str(expr)


sebastiendementen:
**Describe the bug**
The equation expression is not showing correct sign when using -

**To Reproduce**
```python
import cvxpy

a, b = cvxpy.Variable(1, name="a"), cvxpy.Variable(1, name="b")
c = a - b
print(c)
print(-c)

```

**Expected behavior**
Displaying
```
a + -b
-a + b
```
or 
```
a + -b
-(a + -b)
```

**Output**
The current output of the expression -c is misleading
```
a + -b
-a + -b
```

**Version**
 - OS: windows
 - CVXPY Version:1.6.4

Transurgeon:
ah right, thanks for catching this. The issue is that it's difficult to produce the right outputs in a general sense. I suppose what's happening here is that the neg_expr class is prepending a ``-`` in front of the original expression (and doesn't detect that it is a substraction). 
I suppose we could add brackets (corresponding to your second example).

rileyjmurray:
Formatting text representations of expressions is a recurring issue for us.
It would be valuable to make reusable infrastructure for text-based
expression parsing and reformatting.

PR #2924, opened 2025-09-01
Correct string representation of NegExpression

thisisRMak:
## Description
Correct string representation of NegExpression
Issue link (if applicable): #2806

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -875,15 +875,19 @@ class TestExpressions(BaseTest):
         assert exp.is_affine()
         self.assertEqual(exp.sign, s.UNKNOWN)
         assert not exp.is_nonneg()
-        # self.assertEqual(exp.canonical_form[0].shape, (2, 1))
-        # self.assertEqual(exp.canonical_form[1], [])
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_expressions.py
```diff
@@ -875,16 +875,20 @@ def test_neg_expression(self) -> None:
         assert exp.is_affine()
         self.assertEqual(exp.sign, s.UNKNOWN)
         assert not exp.is_nonneg()
-        # self.assertEqual(exp.canonical_form[0].shape, (2, 1))
-        # self.assertEqual(exp.canonical_form[1], [])
-        # self.assertEqual(exp.name(), "-%s" % self.x.name())
```
SteveDiamond, line 880: There should be an assertion checking the string representation of -self.x

thisisRMak: Thanks! Added back the test for -x. 

Not sure what the other two commented out tests were doing. Plmk if we want to retain those comments or bring back those tests somehow?

SteveDiamond: No need, this is good.

Issue #2619, opened 2024-11-09
Further improvements to string representations of expressions


Transurgeon:
The string representation of expressions and constraints is invaluable when a user wants to know why a problem isn't DCP for example. However, as problems get more complicated and the expressions contain a lot of nested atoms, it can sometime become almost incomprehensible. This issue will serve as a tracking issue since there are many things that can be improved upon. I will comment with some examples I have encountered so far (and I will continue to add more sub-examples as I continue using cvxpy).

jonathanberthias:
One thing that I think can help a lot when mentioning complex constraints and expression is the ability to give them a name, and use that instead of the current format. Currently only the variables can be given names, but most of the time the expressions also have a domain specific meaning.

This was discussed many years ago in #322 but it was never implemented.

Transurgeon:
Thank you for pointing us to this issue @jonathanberthias , I really like the idea that was proposed and will think more about it. @SteveDiamond should we reopen the issue that was mentioned above (the proposed solution can cover a lot of the improvements we were thinking about)?

SteveDiamond:
Sure, reopened.

DresdenGman:
Hi! I reviewed the label support merged in #2935 and found what appears to be a remaining recursive-formatting gap in the two transform expression types that define custom `name()` implementations: `indicator` and `PartialProblem`.

For example, a labeled constraint inside an indicator is currently not surfaced:

```python
x = cp.Variable(name="x")
constraint = (x >= 0).set_label("nonnegative")
expr = cp.transforms.indicator([constraint])

expr.format_labeled()
# Current: "Indicator([Inequality(Constant(...))])"
# The nested "nonnegative" label is not included.
```

The same occurs for a labeled objective expression inside a `PartialProblem`: `format_labeled()` falls back to `PartialProblem.name()` and prints the unlabeled inner problem.

I’m considering a focused follow-up with this scope:

- implement recursive `format_labeled()` behavior for `indicator` and `PartialProblem`;
- preserve their existing `name()` and ordinary `str()` output;
- cover nested labels, top-level labels, and unlabeled parity with focused tests;
- make no public API changes.

I also noticed the renderer de-duplication direction discussed after #2935. My inclination is to keep that broader refactor separate so this fix remains reviewable, unless maintainers would prefer the shared-renderer work to be addressed together.

Would this focused follow-up be welcome under #2619, or would you prefer it tracked separately?

Issue #322, opened 2017-02-03
Associating Metadata with Expressions and Constraints


ssanderson:
**TL;DR** I'm interested in adding the ability to customize the name of non-variable expressions and/or tag expressions with extra metadata.  I'm curious if you think that's a reasonable potential enhancement for CVXPY.

Hi!  As mentioned over in https://github.com/cvxgrp/cvxpy/issues/313, I'm working on adding a DSL for portfolio optimization to Quantopian/Zipline using CVXPY under the hood.  In general, the structure of my library is that it provides high-level `Objective` and `Constraint` classes like `MaximizeAlpha` or `MaxGrossLeverage`, which expand out into one or many CVXPY Objective/Constraint instances.  For example, here's what a [pair trade](https://en.wikipedia.org/wiki/Pairs_trade) constraint looks like currently:

<details>

```python

class Pair(Constraint):
    """
    A constraint representing a pair of inverse-weighted stocks.

    Parameters
    ----------
    long : Asset
        The asset to long.
    short : Asset
        The asset to short.
    hedge_ratio : float, optional
        The ratio between the respective absolute values of the long and short
        weights. Required to be greater than 0. Default is 1.0, signifying
        equal weighting.
    tolerance : float, optional
        The amount by which the hedge ratio of the calculated weights is
        allowed to differ from the given hedge ratio, in either direction.
        Required to be greater than or equal to 0. Default is 0.0.
    """

    @normalize_positive_number('Pair', ['tolerance'])
    @normalize_strictly_positive_number('Pair', ['hedge_ratio'])
    def __init__(self, long, short, hedge_ratio=1.0, tolerance=0.0):
        self._long = long
        self._short = short
        self._hedge_ratio = hedge_ratio
        self._tolerance = tolerance

    def to_cvxpy(self, old_weights, target_weights):
        # old_weights is a pandas.Series indexed by Asset
        # new_weights is the cvxpy.Variable over which we're optimizing
        universe = old_weights.index

        long_ix, short_ix = universe.get_indexer([self._long, self._short])

        long_weight = target_weights[long_ix]
        short_weight = target_weights[short_ix]

        # The hedge ratio constraint is better understood logically as
        #       hedge_ratio = long_weight / -short_weight,
        #
        # to ensure that hedge_ratio is positive, but rearranged as
        #       hedge_ratio * -short_weight = long_weight,
        #
        # so that we cannot divide by 0. Then we add tolerance to make the
        # upper and lower bound constraints

[...]

PR #2928, opened 2025-09-05
Add custom display names for expressions and constraints

ClayCampaigne:
## Description
Adds set_display_name() and display_name() methods to expressions and constraints, allowing users to assign custom labels for better debugging and model interpretation.

Key features:
- Method chaining: set_display_name() returns self to allow (x >= 0).set_display_name("bounds")
- Hybrid behavior: constraints show display names in str(), expressions don't (prevents component expression display name labels from leaking into compound expression names)
- No breaking changes: all existing code works exactly as before
- Comprehensive testing: 18 tests covering all functionality and edge cases
- Documentation: added to intro tutorial with practical examples

🎉 Generated with [Claude Code](https://claude.ai/code)
Issue link (if applicable): https://github.com/cvxpy/cvxpy/issues/322, https://github.com/cvxpy/cvxpy/issues/2619

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/constraints/constraint.py
+++ b/cvxpy/constraints/constraint.py
@@ -52,13 +52,45 @@ class Constraint(u.Canonical):
     def __str__(self):
         """Returns a string showing the mathematical constraint.
         """
-        return self.name()
+        return self.display_name()
 
     def __repr__(self) -> str:
         """Returns a string with information about the constraint.
         """
         return "%s(%s)" % (self.__class__.__name__,
                            repr(self.args[0]))
+
+    def set_display_name(self, display_name: str):
+        """Set a custom display name for this constraint.
+        
+        Parameters
+        ----------
+        display_name : str
+            Custom name to display before the mathematical constraint.
+            
+        Returns
+        -------
+        Constraint
+            Returns self to allow method chaining.
+            
+        Examples
+        --------
+        >>> x = cp.Variable(3)
+        >>> con = (x >= 0).set_display_name("non_negative")
+        >>> str(con)
+        'non_negative: 0.0 <= x'
+        """
+        self._display_name = str(display_name) if display_name is not None else None
+        return self
+
+    def display_name(self) -> str:
+        """Get the display name of this constraint.
+        
+        Returns custom display name if set, otherwise mathematical form.
+        """
+        if hasattr(self, '_display_name') and self._display_name is not None:
+            return f"{self._display_name}: {self.name()}"
+        return self.name()
 
     def _construct_dual_variables(self, args) -> None:
         self.dual_variables = [cvxtypes.variable()(arg.shape) for arg in args]
--- a/cvxpy/constraints/zero.py
+++ b/cvxpy/constraints/zero.py
@@ -34,7 +34,7 @@ class Zero(Constraint):
     def __str__(self):
         """Returns a string showing the mathematical constraint.
         """
-        return self.name()
+        return self.display_name()
 
     def __repr__(self) -> str:
         """Returns a string with information about the constraint.
@@ -106,7 +106,7 @@ class Equality(Constraint):
     def __str__(self):
         """Returns a string showing the mathematical constraint.
         """
-        return self.name()
+        return self.display_name()
 
     def __repr__(self) -> str:
         """Returns a string with information about the constraint.
--- a/cvxpy/expressions/expression.py
+++ b/cvxpy/expressions/expression.py
@@ -175,6 +175,38 @@ class Expression(u.Canonical):
         """str : The string representation of the expression.
         """
         raise NotImplementedError()
+    
+    def set_display_name(self, display_name: str):
+        """Set a custom display name for this expression.
+        
+        Parameters
+        ----------
+        display_name : str
+            Custom name to display before the mathematical expression.
+            
+        Returns
+        -------
+        Expression
+            Returns self to allow method chaining.
+            
+        Examples
+        --------
+        >>> x = cp.Variable(3)
+        >>> expr = cp.sum(x).set_display_name("total")
+        >>> expr.display_name()
+        'total: Sum(x, None, False)'
+        """
+        self._display_name = display_name
+        return self
+
+    def display_name(self) -> str:
+        """Get the display name of this expression.
+        
+        Returns custom display name if set, otherwise mathematical form.
+        """
+        if hasattr(self, '_display_name') and self._display_name is not None:
+            return f"{self._display_name}: {self.name()}"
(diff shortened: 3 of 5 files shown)


PTNobel:
This looks really cool!

ClayCampaigne:
Thanks for the review! Lmk if you want me to pare back some of the new tests. 18 new tests, 333 new lines, seems like overkill for this feature now that I look back at it.

cvxpy/constraints/constraint.py
```diff
@@ -52,14 +52,46 @@ def __init__(self, args, constr_id=None) -> None:
...
+        """
+        self._display_name = str(display_name) if display_name is not None else None
+        return self
+
+    def display_name(self) -> str:
+        """Get the display name of this constraint.
+        
+        Returns custom display name if set, otherwise mathematical form.
+        """
+        if hasattr(self, '_display_name') and self._display_name is not None:
```
jonathanberthias, line 91: Could we initialize the attribute to `None` in `__init__` to avoid having to check whether it exists every time? Or maybe the default should be an empty string?

cvxpy/constraints/constraint.py
```diff
@@ -52,14 +52,46 @@ def __init__(self, args, constr_id=None) -> None:
...
+        --------
+        >>> x = cp.Variable(3)
+        >>> con = (x >= 0).set_display_name("non_negative")
+        >>> str(con)
+        'non_negative: 0.0 <= x'
+        """
+        self._display_name = str(display_name) if display_name is not None else None
+        return self
+
+    def display_name(self) -> str:
```
jonathanberthias, line 86: At first, I expected this method to return the set display name, not the name and the stringified constraint. I guess it's due to the infortunate `name` used for the `str` implementation, but maybe we can do better now?
I would suggest having this method return just the display name, whether it has been set or the default, and then put the logic of how to display the constraint inside `__str__` directly. What do you think?

cvxpy/constraints/constraint.py
```diff
@@ -52,14 +52,46 @@ def __init__(self, args, constr_id=None) -> None:
...
-        return self.name()
+        return self.display_name()
 
     def __repr__(self) -> str:
         """Returns a string with information about the constraint.
         """
         return "%s(%s)" % (self.__class__.__name__,
                            repr(self.args[0]))
 
+    def set_display_name(self, display_name: str):
```
jonathanberthias, line 63: I'm afraid this method's name is too long given how frequently I would use it. For instance, a typical use-case could be:
```py
def build_problem(data):
    weights = cp.Variable(len(data))
    constraints = [
        (weights @ data.values <= data.upper_bound).set_display_name("no over-allocation")
    ]
```
but the constraint line is already too long and `black` or `ruff` formatters would split it which make things very weird:
```py
def build_problem(data):
    weights = cp.Variable(len(data))
    constraints = [
        (weights @ data.values <= data.upper_bound).set_display_name(
            "no over-allocation"
        )
    ]
```

So the simple solution would be to reduce the length of the name. How about `with_name`? Or maybe even just `named`?

cvxpy/constraints/zero.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, expr, constr_id=None) -> None:
     def __str__(self):
         """Returns a string showing the mathematical constraint.
         """
-        return self.name()
+        return self.display_name()
```
jonathanberthias, line 37: I'm pretty sure this override can be removed entirely

cvxpy/constraints/zero.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, lhs, rhs, constr_id=None) -> None:
...
```
jonathanberthias, line 109: Same here

cvxpy/expressions/expression.py
```diff
@@ -175,6 +175,38 @@ def name(self) -> str:
...
+        --------
+        >>> x = cp.Variable(3)
+        >>> expr = cp.sum(x).set_display_name("total")
+        >>> expr.display_name()
+        'total: Sum(x, None, False)'
+        """
+        self._display_name = display_name
+        return self
+
+    def display_name(self) -> str:
```
jonathanberthias, line 202: I have the same doubt as for the constraint, even though it's less important to me for expressions since I would have to explicitly call this method anyway.
Still, I wouldn't expect `display_name` to show me more than the name I set

jonathanberthias (commented):
Thanks a lot for tackling this! I'm not a maintainer here, but I can definitely see myself using this feature quite a bit, hence my remarks and limitations. It's all up for discussion of course!

cvxpy/constraints/constraint.py
PTNobel: Agreed that this would be a better approach

cvxpy/constraints/constraint.py
PTNobel: This makes sense to me

cvxpy/constraints/constraint.py
ClayCampaigne: another option I considered is `set_label`, which makes clearer that the display name has two parts, the label and the original mathematical formulation of the constraint, which in this design is preserved as a part of the display name.
I'm open to `with_name` or `named` as well. Maybe someone else can weigh in.

cvxpy/constraints/constraint.py
ClayCampaigne: @jonathanberthias maybe `label` works well as the method name for your suggested approach? Then the idea of my current proposal would be that the constraint displays with either just the original "stringified constraints," and if we have set it, prefixed by the label. 
I kind of like the option to have both a user-set label and the current stringified constraint, but I'm not super strongly attached to it.

cvxpy/constraints/zero.py
ClayCampaigne: Good catch. There is no reason for this child class to implement the same method.

ClayCampaigne:
Thanks for the feedback! I may come back with some design questions on discord before addressing these comments @jonathanberthias @PTNobel

cvxpy/constraints/constraint.py
jonathanberthias: I really like label 💯 it is much clearer IMO than display name

ClayCampaigne:
I'm going to close this one and replace it with #2935 which is more full-featured. Thanks for the feedback!

cvxpy/constraints/constraint.py
PTNobel: I like label!

Issue #2896, opened 2025-08-13
`cp.harmonic_mean` gradient computation fails


langestefan:
**Describe the bug**
Computing the gradient of `cp.harmonic_mean` will throw an exception in `scipy.sparse`.

**To Reproduce**
```python
>>> x = cp.Variable(2, name="x", value=[1.0,1.0])
>>> expr = cp.harmonic_mean(x)
>>> expr.grad[x]
```

**Output**
```python
>>> x = cp.Variable(2, name="x", value=[1.0,1.0])
>>> expr = cp.harmonic_mean(x)
>>> expr.grad[x]
Traceback (most recent call last):
  File "<python-input-33>", line 1, in <module>
    expr.grad[x]
    ^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/atoms/atom.py", line 408, in grad
    grad_arg = arg.grad
               ^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/atoms/atom.py", line 402, in grad
    grad_self = self._grad(arg_values)
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/atoms/pnorm.py", line 234, in _grad
    return self._axis_grad(values)
           ~~~~~~~~~~~~~~~^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/atoms/axis_atom.py", line 103, in _axis_grad
    D = sp.csc_array(D)
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/scipy/sparse/_compressed.py", line 95, in __init__
    coo = self._coo_container(arg1, dtype=dtype)
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/scipy/sparse/_coo.py", line 95, in __init__
    self.data = getdata(M[coords], copy=copy, dtype=dtype)
                ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2933, opened 2025-09-09
Fix issue #2896: harmonic mean gradient computation dtype error

ClayCampaigne:
## Description
Fixes issue #2896.
The gradient computation for `cp.harmonic_mean` was failing with `ValueError: scipy.sparse does not support dtype object`.

The issue occurred when computing gradients with `p=-1` (used by harmonic_mean), where `self.p` is a `Fraction` object. Computing `np.power(value, self.p - 1)` with Fraction exponents creates object dtype arrays that scipy.sparse rejects.

### Changes made:
- Cast Fraction exponents to float in `_column_grad` to prevent dtype=object arrays
- Ensure consistent vector semantics with `ravel(order='F')` 
- Fix sign handling for `p>1` with negative values to prevent incorrect gradients

### Tests added:
- Regression test using exact reproduction from issue #2896
- Test for negative p values including harmonic mean case (p=-1)
- Test for gradient sign correctness with p>1 and negative values  
- Test for column-shaped variables to verify vector semantics
- Analytic gradient value checks for correctness

## Type of change
- [x] Bug fix

- [x] Write unittests.
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. [didn't do: benchmarks are now a separate github workflow]

### DIFF
--- a/cvxpy/atoms/pnorm.py
+++ b/cvxpy/atoms/pnorm.py
@@ -249,15 +249,21 @@ class Pnorm(AxisAtom):
         if self.p < 1 and np.any(value <= 0):
             return None
         D_null = sp.csc_array((rows, 1), dtype='float64')
+        # Ensure vector semantics and consistent column-major vectorization.
+        value = np.asarray(value).ravel(order='F')
         denominator = np.linalg.norm(value, float(self.p))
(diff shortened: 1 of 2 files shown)


PTNobel:
Thanks for doing this!

PR #2955, opened 2025-10-03
Update Python version in devcontainer configuration

Ni2002ka:
## Description
Previously, the devcontainer was using python 3.10, while the pytests require python 3.11. 

## Type of change
- Bug fix: Update the python image in the devcontainer to 3.12.


Transurgeon (approved):
LGTM! can confirm in person that this change fixes testing on codespaces.

Issue #2752, opened 2025-03-12
Support something like `np.stack` for stacking to higher dimension


yasirroni:
**Is your feature request related to a problem? Please describe.**
Since v1.6.0., cvxpy support higher dimension. But, we still don't have support to stacking variable to higher dimension.

```python
import cvxpy as cp
import numpy as np

# Create a 3D variable of shape (2, 3, 4)
x = cp.Variable((2, 3, 4))

# Define the constraints and objective
constraints = [x <= 0.5]
objective = cp.Maximize(cp.sum(x))

# Formulate the problem and solve
problem = cp.Problem(objective, constraints)
problem.solve()

print("Optimal value for cp.Variable(2,3,4):", problem.value)
print("Optimal solution for cp.Variable(2,3,4):", x.value)
```

Relying in numpy will not work because numpy will treat it as single dimension

```python
# Create two 2D variables (shape 3x4)
x1 = cp.Variable((3, 4))
x2 = cp.Variable((3, 4))

# Use np.stack to combine them into a 3D array
stacked_x = np.stack([x1, x2], axis=0)

# Define the constraints and objective
constraints = [x1 <= 0.5, x2 <= 0.5]
objective = cp.Maximize(cp.sum(stacked_x))

# Formulate the problem and solve
problem = cp.Problem(objective, constraints)
problem.solve()

print("\nOptimal value for np.stack:", problem.value)
print("Optimal solution for stacked_x:", stacked_x)
```

**Describe the solution you'd like**
Add `cp.stack` to cvxpy

```python
# Create two 2D variables (shape 3x4)
x1 = cp.Variable((3, 4))
x2 = cp.Variable((3, 4))

# Use np.stack to combine them into a 3D array
stacked_x = cp.stack([x1, x2], axis=0)

# Define the constraints and objective
constraints = [x1 <= 0.5, x2 <= 0.5]
objective = cp.Maximize(cp.sum(stacked_x))

# Formulate the problem and solve
problem = cp.Problem(objective, constraints)
problem.solve()

print("\nOptimal value for np.stack:", problem.value)
print("Optimal solution for stacked_x:", stacked_x)
```

**Describe alternatives you've considered**
Back to old way, using loops over dimensions.

**Additional context**
Related to https://github.com/cvxpy/cvxpy/issues/198#event-15391664856

Transurgeon:
we support cp.concatenate which should cover what you expect from np.stack? Although I am not sure of the exact API.
[...]

PR #2956, opened 2025-10-03
Add cp.stack atom

ClayCampaigne:
## Description
Add a NumPy-style `cp.stack` atom that inserts a new axis via reshape/concatenate, export it, cover it with unit tests, and document it in the affine atom reference.
  
Issue: https://github.com/cvxpy/cvxpy/issues/2752

## Type of change
- [ ] Bug fix

- [x] Write unittests.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression. [n/a]

### DIFF
--- a/cvxpy/atoms/affine/stack.py
+++ b/cvxpy/atoms/affine/stack.py
@@ -1,4 +1,19 @@
-"""NumPy-style stacking that inserts a new axis for CVXPY Expressions.
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+NumPy-style stacking that inserts a new axis for CVXPY Expressions.
 
 This mirrors the semantics of ``np.stack``, where they make sense, for symbolic
 expressions: every input must share an identical shape, and the result gains a
@@ -18,8 +33,6 @@ Examples
 >>> z.shape
 (3, 2)
 """
-
-from __future__ import annotations
 
 from typing import Iterable, Sequence
 
--- a/cvxpy/atoms/affine/stack.py
+++ b/cvxpy/atoms/affine/stack.py
@@ -1,0 +1,94 @@
+"""NumPy-style stacking that inserts a new axis for CVXPY Expressions.
+
+This mirrors the semantics of ``np.stack``, where they make sense, for symbolic
+expressions: every input must share an identical shape, and the result gains a
+new axis of length ``len(arrays)`` at the requested position. Options that are
+specific to NumPy's ndarray type system (for example ``out`` or ``dtype``) are
+intentionally unsupported.
+
+Examples
+--------
+>>> import cvxpy as cp
+>>> a = cp.Parameter((3,))
+>>> b = cp.Parameter((3,))
+>>> y = cp.stack([a, b], axis=0)
+>>> y.shape
+(2, 3)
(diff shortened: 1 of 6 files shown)


ClayCampaigne:
I think during the test jobs, pytest is picking up an older installed copy of CVXPY in the runner’s virtualenv (.venv/lib/.../cvxpy/tests/*). It looks like the cached environment needs to be cleaned. Could someone with access rerun the workflow in a fresh venv (e.g., drop the cache or remove the preinstalled wheel before pytest)?

PTNobel:
Rerunning tests!

cvxpy/tests/atoms/affine/test_stack.py
PTNobel, line 1: Can we add a test that canonicalization works correctly?

Something like:
```
x = cp.Variable()
y = cp.Variable()
z = cp.Variable(2)
z_tilde = cp.stack([x, y])
cp.Problem(cp.Minimize(0), [z_tilde == z, x == 1, y == 2])
assert all(z.value == np.array([1, 2])
```

cvxpy/atoms/affine/stack.py
```diff
@@ -0,0 +1,94 @@
...
+>>> b = cp.Parameter((3,))
+>>> y = cp.stack([a, b], axis=0)
+>>> y.shape
+(2, 3)
+>>> z = cp.stack([a, b], axis=-1)
+>>> z.shape
+(3, 2)
+"""
+
+from __future__ import annotations
```
Transurgeon: could it be this causing the test issues? I don't see ``annotations`` being used anywhere else. Is it necessary to have this?

PTNobel: Can we also remove this independent of the test issue?

ClayCampaigne: whoops, sorry I let that sneak in there! Thanks for catching that @Transurgeon. I don't think that explains test failures though.

ClayCampaigne:
Not sure about these CI failures. This is what my chatgpt says:

> • That pytest error isn’t caused by anything lingering in your tree—it’s coming from the way the CI job is set up. The workflow does a non‑editable install (uv pip install .), which copies the package (and our test suite) into .venv/lib/python3.12/site-packages/cvxpy. A few steps later the job runs pytest cvxpy/tests, so pytest tries to collect the tests out of the checkout, but Python has already imported cvxpy from site‑packages. When pytest imports cvxpy.tests.test_expression_methods, it gets the
>   site‑packages copy, while the collector is holding the file path from the repo, and pytest aborts with the “imported module … is not the same file we want to collect” mismatch.
> 
>   You can’t fix that on your branch; the options are all on the CI side:
>   - Change the install step to editable (uv pip install -e .), so the installed package points back at the working tree and pytest sees the same file in both places.
>   - Or, keep the non‑editable install but run tests from the installed copy instead: pytest --pyargs cvxpy.tests.
>   - Alternatively, remove the site‑packages copy of cvxpy (or its __pycache__) before running pytest, but that defeats the purpose of installing it in the first place.
> 
>   So the branch is clean; nothing’s wrong locally. To unblock the pipeline the maintainers need to tweak the CI script (most  simply switching to an editable install or changing the pytest invocation).

Transurgeon:
@ClayCampaigne I think we can try what chatgpt recommends.
can you do the following diff and see if it would work?
```
diff --git a/continuous_integration/test_script.sh b/continuous_integration/test_script.sh
index bec16e0f..4bafcf11 100644
--- a/continuous_integration/test_script.sh
+++ b/continuous_integration/test_script.sh
@@ -20,7 +20,7 @@ if [ $USE_OPENMP == "True" ] && [ $RUNNER_OS == "Linux" ]; then
     export OMP_NUM_THREADS=4
 else
     uv pip list
-    uv pip install .
+    uv pip install -e .
```
Thanks! I tried committing to your branch but got some permission issues.

ClayCampaigne:
Thanks for the suggestion, @Transurgeon . I am iterating on this issue along the lines you have suggested.

Transurgeon:
@ClayCampaigne thanks for looking into this, it seems to be fixed now!
don't worry about the windows tests failing, we might skip that one for now and find a fix later.

Issue #2981, opened 2025-11-01
Raise an informative error when variable bounds are passed as a tuple


rileyjmurray:
CVXPY ends up erroring when variable bounds are passed as a tuple. The current error message is hard to decipher:

<img width="876" height="621" alt="Image" src="https://github.com/user-attachments/assets/2ba7e8ca-1a6d-4037-97dd-2f952c327d99" />

Luckily I had Parth sitting next to me when this happened. He told me that this something of a design choice (don't want users thinking bounds are exclusive, after all). That being the case, we should have an informative error message that tells users to pass bounds as a list.

PTNobel:
We fixed this in some of the bounds reworking.

PR #2982, opened 2025-11-01
adds proper error message for specifying tuple bounds

Transurgeon:
## Description
This PR raises an error when the user specifies a tuple for variable bounds. One other option could be to convert it to a list before processing it further, but maybe we should stick to this for now. 
Issue link (if applicable): #2981 

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel:
Should we just add tuple support for bounds instead?

Transurgeon:
> Should we just add tuple support for bounds instead?

I don't see why not.

Transurgeon:
I'll close this for now, I think we have other things that are higher priority at the moment.

PR #3018, opened 2025-11-27
POC Rust backend

SteveDiamond:
## Description
This is a first stab at a complete rust backend. The rust backend here fully matches the functionality of the scipy backend, and is fast enough to replace cvxcore as the default backend.

I suspect there are issues with this backend that I missed, but it is promising enough that we can start migrating away from cvxcore in the 1.8 release IMO. Maybe we would keep cvxcore but deprecate it in 1.8 and drop it in 1.9?

## Type of change
- [ ] Bug fix

- [x] Write unittests.


SteveDiamond:
Main open items:
- CI doesn't work, but builds locally
- Compatibility issues with python 3.14, requires PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 in front of every command
- Slow down relative to CPP and SCIPY backends for well structured problems like the Lasso example in `rust_benchmarks`

Transurgeon:
> Main open items:
> 
> * CI doesn't work, but builds locally

This is fixed by using rust-toolchain instead

> * Compatibility issues with python 3.14, requires PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 in front of every command

This was fixed by updating pyo3 to 0.27.0

> * Slow down relative to CPP and SCIPY backends for well structured problems like the Lasso example in `rust_benchmarks`

this was fixed by using Arc to avoid copying data.. need to investigate more to understand it well.

Transurgeon:
I think the backend changed quite a lot and we probably won't have plans to merge this anytime soon.
I will close the PR for now, but we will definitely remember the ``vibe-rust`` branch and revisit it when needed.

Issue #2485, opened 2024-06-22
Support for solving a batch of problems


chrisyeh96:
## Background, motivation, and inspiration

I would love for cvxpy to support the ability to solve a batch of similar problems. Specifically, I am interested in the setting where the problem structure is fixed, and there is a batch of different problem parameters.

The motivation and inspiration for this feature request comes from cvxpylayers (which internally depends on diffcp), which supports solving a batch of cvxpy problems using a ThreadPool. See code here:
https://github.com/cvxgrp/diffcp/blob/cdef024c6b656814f92e3ca10f47ae4feaace1fb/diffcp/cone_program.py#L90-L94.

cvxpylayers allows users to define a problem in cvxpy, then pass in parameters as PyTorch `torch.Tensor` objects, which are allowed to have a batch dimension. In the following example code, a standard cvxpy problem is defined. All the `cp.Variable` and `cp.Parameter` objects are written for a single problem (no batch dimension). However, the parameters in PyTorch have a `batch_size` dimension, and the problem solution is returned with a batch dimension as well:

```python
import cvxpy as cp
from cvxpylayers.torch import CvxpyLayer
import torch

n, m = 2, 3
x = cp.Variable(n)
A = cp.Parameter((m, n))
b = cp.Parameter(m)
constraints = [x >= 0]
objective = cp.Minimize(0.5 * cp.pnorm(A @ x - b, p=1))
problem = cp.Problem(objective, constraints)
assert problem.is_dpp()

cvxpylayer = CvxpyLayer(problem, parameters=[A, b], variables=[x])

batch_size = 5
A_tch = torch.randn(batch_size, m, n, requires_grad=True)
b_tch = torch.randn(batch_size, m, requires_grad=True)

# solve the problem
solution, = cvxpylayer(A_tch, b_tch)

print(solution.shape)
```

Output:
```
torch.Size([5, 2])
```

More generally, I am motivated by the problem of wanting to solve a number of optimization problems as part of the forward inference pass of a neural network. cvxpylayers (via diffcp) always calculates the gradient as well, but at inference time, I do not need the gradient information. Therefore, I hope that cvxpy can directly solve a batch of problems.

## Possible API and design considerations

I see several possible APIs for supporting a batch of problem parameters, and the following ideas are by no means a comprehensive list.

**1) `cp.Parameter.batch_value`, `cp.Variable.batch_value`, and `prob.solve(batch=True)`**

[...]

PR #3039, opened 2025-12-16
Exploration of batch support

SteveDiamond:
## Description
This is an exploration of one possible way to add batch support. I followed approach 2 from #2485:

```
var = cp.Variable(n)
param = cp.Parameter(m, n)
param.set_value(np.ones([batch_size, m, n]), batch=True)
...
prob.solve()

# solution is stored in:
var.value  # shape [batch_size, n]

# optimal value is stored as:
prob.value  # shape [batch_size]
```

I considered a lot of different ways to add batch support and this approach seemed the most consistent with the existing CVXPY API. But it would be helpful to discuss alternatives. Some topics that came up while exploring this change:

1. Do we allow only one batch dimension or an arbitrary number? I felt that adding an arbitrary number of batch dimensions was more consistent with NumPy and a similar amount of work to implement, but I'm not sure about this.
2. Do we only allow solvers that support batching to work in batch mode? Or do we implement some kind of loop or parallelization over solvers without batch support? My perspective is that the implementation of batching is a solver concern, and that CVXPY should only allow batch mode with solvers that support batching natively.
3. Do we make expression.value work fully with batch mode? This is a major code change but very compatible with NumPy so maybe ok.
4. How do we handle batched sparse parameters?

Issue link (if applicable): #2485 

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_batched.py
+++ b/cvxpy/tests/test_batched.py
@@ -627,3 +627,138 @@ class TestBatchedSolveEndToEnd:
         # Compare
         np.testing.assert_allclose(prob.value, individual_values, rtol=1e-4)
         np.testing.assert_allclose(x.value, np.array(individual_x), rtol=1e-4)
+
+
+class TestBatchedValueAccess:
+    """Tests for .value behavior with batched variables/parameters."""
+
+    def test_variable_value_batched_shape(self):
+        """Variable.value returns batched array after batched solve."""
+        n = 3
+        x = cp.Variable(n)
+        c = cp.Parameter(n)
+
+        prob = cp.Problem(cp.Minimize(c @ x), [x >= 0, cp.sum(x) == 1])
+
+        batch_size = 5
+        c_vals = np.random.randn(batch_size, n)
+        c.set_value(c_vals, batch=True)
+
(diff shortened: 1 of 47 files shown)


cvxpy/tests/test_batched.py
```diff
@@ -0,0 +1,1389 @@
...
+        )
+
+        # Check shapes
+        assert q_batch.shape == (3, 4, param_cone_prog.x.size)
+        assert d_batch.shape == (3, 4)
+        assert A_batch.shape == (3, 4, param_cone_prog.constr_size, param_cone_prog.x.size)
+        assert b_batch.shape == (3, 4, param_cone_prog.constr_size)
+
+
+class TestBatchedSolveEndToEnd:
```
SteveDiamond, line 533: You can see examples of end to end tests with batching in this class.

Transurgeon:
@SteveDiamond is this PR fine to close? I think batch support is really cool but definitely unclear how we would go about it. Maybe we can revisit this later?

SteveDiamond:
Yes the amount of changes required for batch support as evidenced by this PR make it way out of scope.

PR #3070, opened 2026-01-13
Replace Eigen sparse Cholesky with QDLDL (and add optimizations for PSD matrices)

Transurgeon:
Replace the Eigen-based C++ sparse Cholesky extension with QDLDL, which is already an indirect dependency via OSQP and is Apache 2.0 licensed.

Changes:
- Update sparse_cholesky() in linalg.py to use qdldl.Solver()
- Convert LDL factors to Cholesky form: L_chol = (I + L) @ sqrt(D)
- Add qdldl >= 0.1.7 as explicit dependency in pyproject.toml
- Remove sparsecholesky C++ extension from build system
- Delete cvxpy/utilities/cpp/sparsecholesky/ directory
- Update tests to remove skipif decorators

## Description
Issue link (if applicable): from contributing list.

I was just playing around with Claude to see if sparse cholesky could be removed. My motivation for this is once we get rid of the CPP backend (CVXCORE), it wouldn't make much sense to rely on (and import) the gigantic ``Eigen`` library anymore just for computing the sparse cholesky factorization.
QDLDL seems to work well on the existing tests.
Would love to hear your thoughts on this, and if its something we should do moving forward.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


Transurgeon:
@claude could you review this once more? Does it make sense to get rid of ``sparsecholesky`` from Eigen or should we keep it?

cvxpy/utilities/linalg.py
```diff
@@ -234,31 +236,44 @@ def sparse_cholesky(A, sym_tol=settings.CHOL_SYM_TOL, assume_posdef=False):
...
+        L_chol = spar.csr_array(L_chol)
+
+        # QDLDL's permutation p defines P where P[i, p[i]] = 1, giving:
+        #   A = P @ L_chol @ L_chol.T @ P.T
+        # The caller expects L_out[p_out, :] @ L_out[p_out, :].T == A
+        # Since L[p, :] = P @ L, we need L[p_inv, :] where p_inv = argsort(p)
+        # so that L[p_inv, :] @ L[p_inv, :].T = A (verified empirically)
+        p_inv = np.argsort(p)
+
+        return 1.0, L_chol, p_inv
```
PTNobel: What is the 1.0

cvxpy/utilities/linalg.py
```diff
@@ -234,31 +236,44 @@ def sparse_cholesky(A, sym_tol=settings.CHOL_SYM_TOL, assume_posdef=False):
...
-    outpivs = np.array(outpivs)
-    L = spar.csr_array((outvals, (outrows, outcols)), shape=(n, n))
-    return 1.0, L, outpivs
+        solver = qdldl.Solver(A_upper, upper=True)
+        L_unit, D, p = solver.factors()
+
+        # QDLDL returns: A == P @ (I + L) @ diag(D) @ (I + L).T @ P.T
+        # For positive definite: all D > 0
+        # Convert to standard Cholesky: L_chol = (I + L) @ diag(sqrt(D))
+        if not np.all(D > 0):
```
PTNobel: Why do we need D>0 rather than just D>=0?

cvxpy/tests/test_linalg_utils.py
```diff
@@ -74,20 +65,32 @@ def test_generic(self, use_expression=False):
...
         self.check_gram(L[p, :], A)
 
-    @pytest.mark.skipif(missing_extension, reason="requires sparse_cholesky")
     def test_expression(self):
         self.test_generic(use_expression=True)
     
-    @pytest.mark.skipif(missing_extension, reason="requires sparse_cholesky")
     def test_singular(self):
-        # error on singular PSD matrix
+        # error on singular PSD matrix (rank deficient)
```
PTNobel: I don't understand why we error on singular matrices

cvxpy/utilities/linalg.py
```diff
@@ -193,25 +193,27 @@ class SparseCholeskyMessages:
...
-    We raise a ValueError if Eigen's Cholesky fails or if we certify indefiniteness
-    before calling Eigen. While checking for indefiniteness, we also check that
+    We raise a ValueError if QDLDL's factorization fails or if we certify indefiniteness
+    before calling QDLDL. While checking for indefiniteness, we also check that
      ||A - A'||_Fro / sqrt(n) <= sym_tol, where n is the order of the matrix.
     """
-    import cvxpy.utilities.cpp.sparsecholesky as spchol  # noqa: I001
-    import cvxpy.expressions.cvxtypes as cvxtypes #noqa: I001
-    
+    import qdldl
```
PTNobel: Why is this not at the top of the file?

rileyjmurray:
LMK if my input is needed on this. I originally wrote this module.

Transurgeon:
thanks for your review @PTNobel , let's come back to this once the osqp team releases python3.14 builds for qdldl.

PTNobel:
QDLDL has 3.14t wheels now!

cvxpy/tests/test_linalg_utils.py
Transurgeon: it is because we require the matrix to be full rank.

cvxpy/utilities/linalg.py
Transurgeon: fixed.

cvxpy/utilities/linalg.py
Transurgeon: claude added a comment for it, seems to be the sign of the eigenvalues.

cvxpy/utilities/linalg.py
Transurgeon: claude added a comment for it, its for diagonal entries to be strictly positive that way we can take their square roots, or something like that.

Transurgeon (commented):
@rileyjmurray @PTNobel I synced this PR with master and added some clarifying comments. Please have a look! It seems like a nice cleanup :).

PTNobel (approved):

Transurgeon:
@rileyjmurray pinging again in case you have time to look at it.
@PTNobel added some cleanups for PSD matrices to keep using sparse ldl factorization instead of converting to dense.

rileyjmurray:
No need to wait for my approval if Parth already did a pass. He and I
synced on a related issue recently.

PTNobel:
@rileyjmurray I rewrote the core of this PR; it would be wise to get a second set of eyes on it

cvxpy/tests/test_linalg_utils.py
```diff
@@ -74,26 +65,83 @@ def test_generic(self, use_expression=False):
...
+        # Verify permutation is non-trivial (not identity or simple reversal)
+        p_inv = np.argsort(p)
+        self.assertFalse(np.array_equal(p, p_inv),
+                         "Test requires non-symmetric permutation")
+
+    def test_indefinite_with_zero_diagonal(self):
+        # [[1, 2], [2, 0]] has non-negative diagonal but is indefinite.
+        # The zero-diagonal row is not all-zero, so we must reject it.
+        A = sp.csc_array(np.array([[1.0, 2.0], [2.0, 0.0]]))
+        with self.assertRaises(ValueError):
```
rileyjmurray: check for ``msg=lau.SparseCholeskyMessages.INDEFINITE``, in line with ``test_nonsingular_indefinite``.

cvxpy/utilities/linalg.py
```diff
@@ -184,81 +185,124 @@ class SparseCholeskyMessages:
...
+        L_expanded = sp.lil_array((n, k))
+        L_expanded[nonzero_idx, :] = L_sub[p_sub, :]
+        L_expanded = sp.csr_array(L_expanded)
+        return sign, L_expanded, np.arange(n)
+
+    # QDLDL expects upper triangular CSC format
+    A_upper = sp.triu(A, format='csc')
+
+    # Tolerance for near-zero pivots, matching decomp_quad's dense path.
+    tol = 1e6 * np.finfo(float).eps  # ≈ 2.2e-10
```
rileyjmurray: This is a strange way to define a tolerance.

cvxpy/utilities/linalg.py
```diff
@@ -184,81 +185,124 @@ class SparseCholeskyMessages:
...
+        p_inv = np.argsort(p)
+
+        return sign, L_chol, p_inv
+
+    except ValueError:
+        raise
+    except Exception as e:
+        raise ValueError(
+            f"{SparseCholeskyMessages.FACTORIZATION_FAILED}: {e}"
+        ) from e
```
rileyjmurray, line 307: Strange exception handling

rileyjmurray (commented):
See below for minor comments. (Leaving as "comment" rather than "request changes" so you don't have to wait for me to approve later.)

cvxpy/utilities/linalg.py
PTNobel: Wait really? This is like the way to add a prefix to an Exception. Seemed totally normal to me

cvxpy/utilities/linalg.py
PTNobel: I don't remember how this came to be; it might have just been progressively increasing the multiplier till I was happy

rileyjmurray: I'm okay with a magic number. I'd like it to reference a variable that's settable at runtime, like we have for various projection tolerances.

cvxpy/utilities/linalg.py
rileyjmurray: The prefix isn't what's weird to me. What's weird is catching and then trivially re-raising a ValueError.

rileyjmurray:
ping @PTNobel responded to your comments on my comments

cvxpy/utilities/linalg.py
PTNobel: Yeah that's how you wrap an error?

rileyjmurray: Why not the following pattern?
```python
>>> def divzero():
...     return 1/0
... 
>>> def wrapped_divzero():
...     try:
...         divzero()
...     except Exception as e:
...         raise ValueError('Hello!') from e
... 
>>> 
>>> wrapped_divzero()
  File "<stdin>", line 3, in wrapped_divzero
  File "<stdin>", line 2, in divzero
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "<stdin>", line 5, in wrapped_divzero
ValueError: Hello!

```

PTNobel: That's a totally different message! That suggests we failed to successfully handle the error we were trying to handle, this is just adding context and normalizing the type of an error.

rileyjmurray: Got it. You can mark this as resolved, then.

PR #3002, opened 2025-11-19
New feature: Disciplined Nonlinear Programming

Transurgeon:
## Description
Please include a short summary of the change.
This PR introduces DNLP.
Issue link (if applicable):

## Type of change
- [x] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #3108, opened 2026-02-01
Disciplined nonlinear programming (DNLP)

Transurgeon:
# Add Disciplined Nonlinear Programming (DNLP) extension to CVXPY

## Summary

This PR adds **Disciplined Nonlinear Programming (DNLP)** support to CVXPY, enabling users to formulate and solve general nonlinear programs by freely mixing smooth functions with nonsmooth convex and concave functions under disciplined composition rules.

For theoretical background, see: [Disciplined Nonlinear Programming](https://web.stanford.edu/~boyd/papers/dnlp.html)

### Usage

```python
import cvxpy as cp
import numpy as np

x = cp.Variable(n)
x.value = np.ones(n)  # initial point required
prob = cp.Problem(cp.Minimize(objective), constraints)
prob.solve(nlp=True, solver=cp.IPOPT)
```

### DNLP expression system and composition rules
- New expression properties: `is_smooth()`, `is_linearizable_convex()`, `is_linearizable_concave()` on `Expression`, `Atom`, `Leaf`, `Objective`, and constraint classes
- `is_dnlp()` validation on `Problem` to check DNLP compliance
- `is_atom_smooth()` method on atoms to classify smooth vs. non-smooth atoms

### New atoms
- **Trigonometric:** `sin`, `cos`, `tan` (`cvxpy/atoms/elementwise/trig.py`)
- **Hyperbolic:** `sinh`, `tanh`, `asinh`, `atanh` (`cvxpy/atoms/elementwise/hyperbolic.py`)

### Dnlp2Smooth reduction
- `Dnlp2Smooth` reduction (`cvxpy/reductions/dnlp2smooth/`) that rewrites non-smooth atoms into smooth equivalents using auxiliary variables
- 18 smooth canonicalizers for: `abs`, `max`, `maximum`, `norm_inf`, `sum_largest`, `div`, `entr`, `exp`, `geo_mean`, `huber`, `kl_div`, `log`, `log_sum_exp`, `logistic`, `multiply`, `pnorm`, `power`, `prod`, `quad_over_lin`, `rel_entr`, and trig/hyperbolic atoms

### NLP solver infrastructure
- **Base `NLPsolver` class** (`cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py`) with:
  - `Bounds`: extracts variable and constraint bounds from the reduced problem
  - `Oracles`: provides function/derivative oracles (objective, gradient, constraints, Jacobian, Hessian of the Lagrangian)
- **NLP solving chain** (`cvxpy/reductions/solvers/nlp_solving_chain.py`): orchestrates `FlipObjective → CvxAttr2Constr → Dnlp2Smooth → NLPSolver`, handles initial point construction (`_set_nlp_initial_point`), and `best_of` multi-start logic
- **Solver variant mapping** (`NLP_SOLVER_VARIANTS`) for algorithm selection (e.g., `"knitro_sqp"`)

### Differentiation engine integration
- `C_problem` wrapper (`cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py`) around the [SparseDiffPy](https://github.com/SparseDifferentiation/SparseDiffPy) C library, providing `objective_forward()`, `gradient()`, `jacobian()`, `hessian()` oracles
- AST-to-C converters (`converters.py`) with `ATOM_CONVERTERS` dict mapping ~40 atom types to C constructors, including sparse parameter matmul fusion
- `Parameter` support: parameters are registered with the C engine and values can be updated without rebuilding the expression tree

### Four NLP solver interfaces
| Solver | File | License |
|--------|------|---------|
| [IPOPT](https://github.com/coin-or/Ipopt) | `ipopt_nlpif.py` | EPL-2.0 (open source) |
| [Knitro](https://www.artelys.com/solvers/knitro/) | `knitro_nlpif.py` | Commercial |
| [COPT](https://www.copt.de/) | `copt_nlpif.py` | Commercial |
| [Uno](https://github.com/cuter-testing/uno) | `uno_nlpif.py` | Open source |

### CI
- New `test_nlp_solvers` GitHub Actions workflow (`.github/workflows/test_nlp_solvers.yml`) that installs system IPOPT via `uv` and runs NLP solver tests

### Documentation
- DNLP tutorial (`doc/source/tutorial/dnlp/index.rst`)
- CLAUDE.md for developer onboarding

### Comprehensive test suite (29 test files, ~3,800 lines)
- **Derivative verification** via `DerivativeChecker` (finite-difference checks of gradients, Jacobians, and Hessians)
- **Atom-level tests:** abs, entropy/KL-div/rel-entropy, huber/sum-largest, hyperbolic, log-sum-exp, matmul, prod, sum, broadcast, power-flow
- **Problem-level tests:** scalar and matrix problems, DNLP validation, initialization, best-of multi-start, quasi-Newton methods
- **Solver interface tests:** cross-solver comparison (IPOPT, Knitro, COPT, Uno)
- **Application tests:** risk parity, Sharpe ratio, ML Gaussian stress test
- **Diff engine stress tests:** affine matrix/vector atoms, sparse matmul, sparse multiply, quad_form

## Diff stats

- **108 files changed**, ~8,300 insertions, ~170 deletions
- ~3,300 lines of core NLP infrastructure (solvers, reductions, diff engine)
- ~3,900 lines of tests
- ~500 lines of atom additions/modifications

## Test plan

- [ ] All existing CVXPY tests pass (no regressions to DCP/DGP/DQCP)
- [ ] `pytest cvxpy/tests/nlp_tests/` passes with IPOPT installed
- [ ] NLP solver CI workflow (`test_nlp_solvers`) passes
- [ ] Pre-commit checks (ruff, actionlint, jsonschema) pass
- [ ] Cross-platform build (Ubuntu/macOS/Windows) succeeds


Transurgeon:
@claude could you review this? I made a new PR branching from #3002, which resets the README and CLAUDE.md to be from CVXPY master, and only appended a small section about DNLP to them.

Transurgeon:
@SteveDiamond @PTNobel here's an updated PR for the DNLP integration.
@dance858 and I have been working on a differentiation engine which computes derivatives in C, we were thus able to get rid of all the ``_jacobian_`` and ``_hess_vec`` oracles per atom. Albeit, we now have many bindings (one for each atom) which serve to convert expression trees into their C equivalent.

There is a weird failure only on windows and py311, I will investigate it at some point.. is it blocking for merging?
There's also documentation that need to be added, but we plan on doing that at a later stage (in another PR) if that's okay.

.github/workflows/test_nlp_solvers.yml
```diff
@@ -0,0 +1,46 @@
...
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+      - uses: actions/checkout@v5
+        with:
+          submodules: recursive
+      - name: Set up Conda
```
PTNobel: Can we use the same uv setup we use in all the other CI?

cvxpy/atoms/__init__.py
```diff
@@ -71,6 +71,8 @@
 from cvxpy.atoms.elementwise.sqrt import sqrt
 from cvxpy.atoms.elementwise.square import square
 from cvxpy.atoms.elementwise.xexp import xexp
+from cvxpy.atoms.elementwise.trig import sin, cos, tan
+from cvxpy.atoms.elementwise.hyperbolic import sinh, asinh, tanh, atanh
```
PTNobel, line 76: I am opposed to totally non-convex/non-concave atoms being in the top-namespace of cvxpy. I'd like to put them in cp.nlp like we have cp.logic

cvxpy/atoms/atom.py
```diff
@@ -182,12 +182,24 @@ def is_atom_concave(self) -> bool:
         """Is the atom concave?
         """
         raise NotImplementedError()
-
+
```
PTNobel, line 185: 😭

cvxpy/atoms/pnorm.py
```diff
@@ -176,6 +176,16 @@ def is_atom_concave(self) -> bool:
         """
         return self.p < 1
 
+    def is_atom_esr(self) -> bool:
+        """Is the atom esr?
+        """
+        return self.p > 1
```
PTNobel: Shouldn't this be >= ?

cvxpy/expressions/expression.py
```diff
@@ -338,6 +344,18 @@ def is_concave(self) -> bool:
         """Is the expression concave?
         """
         raise NotImplementedError()
+    
+    #@abc.abstractmethod
```
PTNobel: How much trouble would it be to enable this at this point?

cvxpy/problems/problem.py
```diff
@@ -1216,6 +1230,93 @@ def _solve(self,
                 self.unpack(chain.retrieve(soln))
                 return self.value
 
+        if nlp and self.is_dnlp():
```
PTNobel, line 1076: I think this is way too much logic for this file; can we put it in a file called nlp_solving_chain.py perhaps?

cvxpy/problems/problem.py
```diff
@@ -1524,7 +1625,7 @@ def unpack(self, solution) -> None:
     def unpack_results(self, solution, chain: SolvingChain, inverse_data) -> None:
         """Updates the problem state given the solver results.
 
-        Updates problem.status, problem.value and value of
+        Updates problem.status, problem.value and value ofro
```
PTNobel: ?

cvxpy/problems/problem.py
```diff
@@ -1561,6 +1662,103 @@ def unpack_results(self, solution, chain: SolvingChain, inverse_data) -> None:
...
+        initial value is specified, look at the bounds. If both lb and ub are 
+        specified, we initialize the variables to be their midpoints. If only
+        one of them is specified, we initialize the variable one unit from 
+        the bound. If none of them is specified, we initialize it to zero.
+        """
+        for var in self.variables():
+            if var.value is not None:
+                continue
+
+            bounds = var.bounds
```
PTNobel: Should we use `get_bounds`?

cvxpy/problems/problem.py
```diff
@@ -1561,6 +1662,103 @@ def unpack_results(self, solution, chain: SolvingChain, inverse_data) -> None:
...
+                # Midpoint if both finite, one from bound if only one finite, zero if none
+                init = (lb_finite * ub_finite * 0.5 * (lb0 + ub0) +
+                        lb_finite * (~ub_finite) * (lb0 + 1.0) +
+                        (~lb_finite) * ub_finite * (ub0 - 1.0))
+                # Broadcast to variable shape (handles scalar bounds)
+                init = np.broadcast_to(init, var.shape).copy()
+                var.save_value(init)
+
+
+    def set_random_NLP_initial_point(self, run) -> dict:
```
PTNobel: I wonder if we should move this logic to both methods to a helper method? On that note why are these public methods?

cvxpy/reductions/dnlp2smooth/canonicalizers/entr_canon.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import numpy as np
+
+from cvxpy.expressions.variable import Variable
+
+MIN_INIT = 1e-4
+
+def entr_canon(expr, args):
+    t = Variable(args[0].shape, bounds=[0, None])
```
PTNobel: Should we use the `expr.get_bounds` infrastructure?

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
```diff
@@ -0,0 +1,50 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from cvxpy.expressions.variable import Variable
+
+
+def sinh_canon(expr, args):
```
PTNobel: Can we deduplicate the canonicalization functions? A lot of them are identical...

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
```diff
@@ -0,0 +1,111 @@
...
+limitations under the License.
+"""
+
+from typing import Tuple
+
+from cvxpy import problems
+from cvxpy.expressions.expression import Expression
+from cvxpy.problems.objective import Minimize
+from cvxpy.reductions.canonicalization import Canonicalization
+from cvxpy.reductions.dnlp2smooth.canonicalizers import SMOOTH_CANON_METHODS as smooth_canon_methods
```
PTNobel, line 23: Why not keep this in all caps?

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    Reduce a disciplined nonlinear program to an equivalent smooth program
+
+    This reduction takes as input (minimization) expressions and converts
+    them into smooth expressions.
+    """
+    def __init__(self, problem=None) -> None:
+        super(Canonicalization, self).__init__(problem=problem)
+        self.smooth_canon_methods = smooth_canon_methods
+
```
PTNobel, line 37: Why is this stored as part of the class?

cvxpy/reductions/solvers/nlp_solvers/ipopt_nlpif.py
```diff
@@ -0,0 +1,203 @@
...
+        hessian_approx = 'exact'
+        if solver_opts:
+            hessian_approx = solver_opts.get('hessian_approximation', 'exact')
+        use_hessian = (hessian_approx == 'exact')
+
+        oracles = Oracles(bounds.new_problem, bounds.x0, len(bounds.cl),
+                          verbose=verbose, use_hessian=use_hessian)
+
+        nlp = cyipopt.Problem(
+        n=len(data["x0"]),
```
PTNobel: Please tab this in...

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,499 @@
...
+                lower.extend([0.0] * constraint.size)
+                upper.extend([np.inf] * constraint.size)
+                new_constr.append(lower_ineq_to_nonneg(constraint))
+            elif isinstance(constraint, NonPos):
+                lower.extend([0.0] * constraint.size)
+                upper.extend([np.inf] * constraint.size)
+                new_constr.append(nonpos2nonneg(constraint))
+        canonicalized_prob = self.problem.copy([self.problem.objective, new_constr])
+        self.new_problem = canonicalized_prob
+        self.cl = np.array(lower)
```
PTNobel, line 125: Not a huge fan of the list approach. Couldn't we precompute the size we need and then modify it in place?

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,499 @@
...
+                upper.extend([0.0] * constraint.size)
+                new_constr.append(lower_equality(constraint))
+            elif isinstance(constraint, Inequality):
+                lower.extend([0.0] * constraint.size)
+                upper.extend([np.inf] * constraint.size)
+                new_constr.append(lower_ineq_to_nonneg(constraint))
+            elif isinstance(constraint, NonPos):
+                lower.extend([0.0] * constraint.size)
+                upper.extend([np.inf] * constraint.size)
+                new_constr.append(nonpos2nonneg(constraint))
```
PTNobel, line 122: Shouldn't nonpos2nonneg be part of the canonicalization in the solving_chain?

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,499 @@
...
+        canonicalized_prob = self.problem.copy([self.problem.objective, new_constr])
+        self.new_problem = canonicalized_prob
+        self.cl = np.array(lower)
+        self.cu = np.array(upper)
+
+    def get_variable_bounds(self):
```
PTNobel: Can't we use extract_lower_bounds/extract_upper_bounds?

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,499 @@
...
+
+        self._jac_structure = (jac_coo.row.astype(np.int32),
+                               jac_coo.col.astype(np.int32))
+        return self._jac_structure
+
+    def hessian(self, x, duals, obj_factor):
+        """Returns the lower triangular Hessian values in COO format. """
+        if not self.use_hessian:
+            # Shouldn't be called when using quasi-Newton, but return empty array
+            return np.array([])
```
PTNobel: Shouldn't we error?

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,499 @@
...
+        # Extract lower triangular values
+        mask = hess_coo.row >= hess_coo.col
+
+        return hess_coo.data[mask]
+
+    def hessianstructure(self):
+        """Returns the sparsity structure of the lower triangular Hessian."""
+        if not self.use_hessian:
+            # Return empty structure when using quasi-Newton approximation
+            return (np.array([], dtype=np.int32), np.array([], dtype=np.int32))
```
PTNobel: Error here?

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
```diff
@@ -0,0 +1,342 @@
...
+                return 1
+
+        # Define objective gradient callback
+        # Signature: gradient(number_variables, x, gradient, user_data) -> int
+        # Must write result to gradient array and return 0 on success
+        def gradient_callback(number_variables, x, gradient, user_data):
+            try:
+                x_arr = to_numpy(x, number_variables)
+                grad = oracles.gradient(x_arr)
+                for i in range(n):
```
PTNobel, line 190: This looks like it wouldn't scale? Is there no vectorized way to do this? Can we ask Uno to add a way to do this?

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
```diff
@@ -0,0 +1,342 @@
...
+        # Set constraints if there are any
+        if m > 0:
+            # Define constraints callback
+            # Signature: constraints(n, m, x, constraint_values, user_data) -> int
+            def constraints_callback(number_variables, number_constraints, x,
+                                     constraint_values, user_data):
+                try:
+                    x_arr = to_numpy(x, number_variables)
+                    cons = oracles.constraints(x_arr)
+                    for i in range(m):
```
PTNobel, line 208: Also seems suboptimal

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
```diff
@@ -0,0 +1,342 @@
...
+            nnz_jacobian = len(jac_rows)
+
+            # Define Jacobian callback
+            # Signature: jacobian(n, nnz, x, jacobian_values, user_data) -> int
+            def jacobian_callback(number_variables, number_jacobian_nonzeros, x,
+                                  jacobian_values, user_data):
+                try:
+                    x_arr = to_numpy(x, number_variables)
+                    jac_vals = oracles.jacobian(x_arr)
+                    # Flatten in case it's returned as a 2D memoryview
```
PTNobel, line 225: Sorry what is this protecting against?

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
```diff
@@ -0,0 +1,342 @@
...
+                    jac_vals_arr = np.asarray(jac_vals).flatten()
+                    for i in range(nnz_jacobian):
```
PTNobel, line 227: Same concern.

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
```diff
@@ -0,0 +1,342 @@
...
+        # This matches our oracles.hessian convention
+        def hessian_callback(number_variables, number_constraints, number_hessian_nonzeros,
+                             x, objective_multiplier, multipliers, hessian_values, user_data):
+            try:
+                x_arr = to_numpy(x, number_variables)
+                mult_arr = to_numpy(multipliers, number_constraints) if m > 0 else np.array([])
+                hess_vals = oracles.hessian(x_arr, mult_arr, objective_multiplier)
+                # Flatten in case it's returned as a 2D array
+                hess_vals_arr = np.asarray(hess_vals).flatten()
+                for i in range(nnz_hessian):
```
PTNobel, line 265: Same concern.

cvxpy/settings.py
```diff
@@ -202,7 +204,7 @@
 # Numerical tolerances
 EIGVAL_TOL = 1e-10
 PSD_NSD_PROJECTION_TOL = 1e-8
-GENERAL_PROJECTION_TOL = 1e-10
```
PTNobel, line 206: Can you explain this change?

pyproject.toml
```diff
@@ -93,6 +93,7 @@ SCS = []
 XPRESS = ["xpress>=9.5"]
 DAQP = ["daqp"]
 KNITRO = ["knitro"]
+# IPOPT = ["cyipopt"]
```
PTNobel: What's going on here?

README.md
PTNobel, line 1: What's going on with these changes?

setup.py
PTNobel, line 1: I thought this file was deprecated; are we still using it?

PTNobel (commented):
Reviewed everything except the bindings for the diffengine and the tests.

pyproject.toml
Transurgeon: this had some issues because it only worked well with conda.
for some reason installation through pip would only give the interface ``cyipopt`` not the actual solver ``Ipopt``.

.github/workflows/test_nlp_solvers.yml
Transurgeon: I think ``uv`` doesn't install ``Ipopt`` automatically.. only the interface, this was a good way for us to test things (since for some reason the conda installation is able to install the solver source code automatically).

cvxpy/atoms/__init__.py
Transurgeon: okay makes sense, does this mean that users will now have to write ``cp.nlp.sin`` in their models?

cvxpy/expressions/expression.py
Transurgeon: hmmm that's a good question, I think we haven't covered many of the matrix atoms, like log_det or lambda_sum_largest, I don't know if they make much sense for DNLP. Maybe we would just need to set them all to false for these atoms?

cvxpy/settings.py
Transurgeon: It was because of the derivative checker in Ipopt, it perturbs by around 1e-6 and it would violate bounds (which use this tolerance). Then CVXPY would complain and crash the entire program. Is this fine to change or should we keep it to 1e-10?
It isn't a big blocker, but definitely a QoL improvement if somebody wants to check derivatives.

README.md
Transurgeon: sorry, I tried to revert the README to be the same as CVXPY using claude, but maybe it went back to an older version. Will fix.

cvxpy/atoms/elementwise/power.py
```diff
@@ -212,6 +212,16 @@ def is_atom_concave(self) -> bool:
         # p == 0 is affine here.
         return _is_const(self.p) and 0 <= self.p.value <= 1
 
+    def is_atom_esr(self) -> bool:
+        """Is the atom esr?
+        """
+        return _is_const(self.p)
+
+    def is_atom_hsr(self) -> bool:
+        """Is the atom hsr?
```
SteveDiamond: could you change these comments and the is atom esr one to spell out the acronym or maybe link to a definition?

cvxpy/atoms/max.py
```diff
@@ -110,6 +120,10 @@ def is_atom_log_log_concave(self) -> bool:
         """
         return False
 
+    def is_smooth(self):
```
SteveDiamond: I don't see the is_smooth in as many atoms. is it true by default?

cvxpy/reductions/eliminate_pwl/canonicalizers/norm_inf_canon.py
```diff
@@ -37,4 +37,9 @@ def norm_inf_canon(expr, args, solver_context: SolverInfo | None = None):
     else:  # shape = (m, 1)
         promoted_t = reshape(t, (x.shape[0], 1), order='F') @ Constant(np.ones((1, x.shape[1])))
 
+    # for DNLP we must initialize the new variable (DNLP guarantees that 
+    # x.value will be set when this function is called)
+    if expr.value is not None:
+        t.value = expr.value
```
SteveDiamond, line 43: That's interesting. Propagating variable values like this could be useful for warm starting as well.

cvxpy/atoms/max.py
PTNobel: It is ESR and HSR by default. I'm not sure why it is special cased here tbh.

cvxpy/expressions/expression.py
PTNobel: I don't see why the matrix atoms wouldn't make sense for DNLP... It would be nice if we have a clear story for the docs of what's not implemented.

cvxpy/atoms/__init__.py
PTNobel: That's my idea!

.github/workflows/test_nlp_solvers.yml
PTNobel: Hmmm, I am not a conda fan; so I would prefer to not depend on it. Could we use that solver that starts with an A for the tests so we can use uv?

cvxpy/settings.py
PTNobel: Two thoughts:
1. I'm very concerned about loosening the projection tolerance, but I don't know where all it is used. So I'll have to do some looking before I say yes or no.
2. Would our users ever check derivatives? They'd never be manually implementing a derivative so why would they want to use the derivative checker?

cvxpy/atoms/pnorm.py
Transurgeon: fixed.

cvxpy/problems/problem.py
Transurgeon: fixed.

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
Transurgeon: deduplicated through a ``_common.py`` canonicalization file.

cvxpy/reductions/solvers/nlp_solvers/ipopt_nlpif.py
Transurgeon: done.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
Transurgeon: yes done.

cvxpy/atoms/elementwise/power.py
Transurgeon: would it be better to use lconvex (linearizable to convex)? This is the definition we end up using in the paper.

Transurgeon: maybe we would have it like ``Lconvex`` with a capital L.

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
Transurgeon: I followed the convention in dcp2cone, maybe it is not necessary, should I remove it?

.github/workflows/test_nlp_solvers.yml
Transurgeon: I am not sure which one you are talking about.. we tried to use uno by default but there are some issues with it.

PTNobel: Yeah I meant uno... Hmmm okay... That's difficult

PTNobel: I'll poke around at this.

cvxpy/reductions/eliminate_pwl/canonicalizers/norm_inf_canon.py
PTNobel: Implemented in #3123

cvxpy/atoms/elementwise/power.py
PTNobel: Any reason we can't say "is_atom_linearizable_convex"? That's not that much longer than is_atom_log_log_convex

cvxpy/expressions/expression.py
rileyjmurray, line 1: New functions in the Expression class need clear documentation. Docstrings written here will be informative for all downstream classes. IMO, the ideal situation is that contributors/developers should not have to refer to the DNLP paper to understand what these functions do. If that's not possible then these docstrings need to refer to extremely specific parts of the paper (e.g., "see Definition 3.1 ...",  "Per Equation (4.9), ...", "By Theorem 5 ...")

cvxpy/expressions/variable.py
```diff
@@ -52,6 +52,8 @@ def __init__(
         self._value = None
         self.delta = None
         self.gradient = None
+        # bounds for sampling initial points in DNLP problems
+        self.sample_bounds = None
```
rileyjmurray: Clear documentation needed.

cvxpy/expressions/expression.py
rileyjmurray: @Transurgeon is the issue that some spectral functions can't be made to satisfy the smoothness requirements of DNLP? (I haven't read the paper.)

cvxpy/reductions/eliminate_pwl/canonicalizers/abs_canon.py
```diff
@@ -24,4 +25,10 @@ def abs_canon(expr, args, solver_context: SolverInfo | None = None):
     bounds = get_expr_bounds_if_supported(expr, solver_context)
     t = Variable(expr.shape, bounds=bounds)
     constraints = [t >= x, t >= -x]
+
+    # for DNLP we must initialize the new variable (DNLP guarantees that 
+    # x.value will be set when this function is called)
+    if expr.value is not None:
+        t.value = expr.value
+
```
rileyjmurray, line 33: These changes can be applied without referencing DNLP. @PTNobel do your new variable-value propagation PRs handle this?

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,517 @@
...
+        self.constraints_forward_passed = False
+
+
+# TODO: maybe add a cchecker like this to the diff-engine? Or rather do a checker that
+# uses cvxpy expressions to evaluate values. It will be slower, but will better test 
+# consistency with cvxpy.
+class DerivativeChecker:
+    """
+    A utility class to verify derivative computations by comparing
+    C-based diff engine results against Python-based evaluations.
```
rileyjmurray: "against central-difference approximations implemented in Python."

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,517 @@
...
+            x_minus = x.copy()
+            x_plus[j] += epsilon
+            x_minus[j] -= epsilon
+            
+            c_plus = constraint_func(x_plus)
+            c_minus = constraint_func(x_minus)
+            
+            numerical_jac[:, j] = (c_plus - c_minus) / (2 * epsilon)
+        
+        match = np.allclose(c_jac_dense, numerical_jac, rtol=1e-4, atol=1e-5)
```
rileyjmurray: Please justify these values of magic constants or make it possible to change them at runtime. Better yet, have a function that just returns the pair ``c_jac_dense, numerical_jac``, and then have a wrapper function that includes a tolerance-based check.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
```diff
@@ -0,0 +1,517 @@
...
+        """Prints information at every Ipopt iteration."""
+        self.iterations = iter_count
+        self.objective_forward_passed = False
+        self.constraints_forward_passed = False
+
+
+# TODO: maybe add a cchecker like this to the diff-engine? Or rather do a checker that
+# uses cvxpy expressions to evaluate values. It will be slower, but will better test 
+# consistency with cvxpy.
+class DerivativeChecker:
```
rileyjmurray: For any function in this class that goes through some effort to compute a C-engine quantity and a corresponding Python-implemented quantity, I suggest the function simply return both quantities. If we need wrappers that perform specific tolerance-based checks for those quantities then they can be just that --- wrappers.

^I've left a comment to that effect for check_jacobian but it applies to many functions in this class.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
rileyjmurray, line 1: I'd like files that define new base classes to include thorough type annotations.

cvxpy/settings.py
rileyjmurray: I suggest we change this value at runtime as needed before calling IPOPT. Then after IPOPT returns but before we return control to the calling function we can restore the old value (here, 1e-10).

cvxpy/expressions/expression.py
Transurgeon: yes precisely, atleast that's my intuition. I am not 100% sure if that's the case.. but it's just something we haven't explored more.

cvxpy/reductions/eliminate_pwl/canonicalizers/abs_canon.py
PTNobel: Yep; they've added this

cvxpy/expressions/expression.py
rileyjmurray: I've thought about this a little more. In order to compute derivatives of spectral functions (e.g., $\lambda_ {\text{max}}(X)$ or the spectral norm) you'll need a subroutine for symmetric eigendecomposition.

It would be a huge lift to implement a reliable symmetric eigendecomposition from scratch. It would be foolish to try and get good speed out of that implementation.

The more reasonable approch would be to somehow haul in eigendecomposition that we know will be shipped with numpy/scipy.

rileyjmurray: > hmmm that's a good question, I think we haven't covered many of the matrix atoms, like log_det or lambda_sum_largest

The gradient of log-det isn't actually that hard to compute. It's just $\nabla( \log\det X ) = -X^{-1}$ . Many other spectral functions require eigendecomposition, though, per my comment above.

cvxpy/atoms/elementwise/power.py
Transurgeon: I fixed this on DNLP master to be more like the paper, will probably make another PR

cvxpy/atoms/max.py
Transurgeon: nice catch! will remove this and other instances of it.

cvxpy/expressions/expression.py
Transurgeon: I fixed the docstring on DNLP master with a more descriptive annotation, will make another PR after some other cleanups.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
PTNobel: Not finite difference?

cvxpy/expressions/expression.py
Transurgeon: we enabled this on DNLP master. For now I think we won't be supporting matrix atoms, they are a bit out of scope and we can leave them for future work. I should definitely try to add documentation for how to add atoms and whatnot.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
Transurgeon: maybe we should make a new reduction called ``normalized_constraints``? I followed the ConeMatrixStuffing code which uses ``nonpos2nonneg``.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
Transurgeon: I moved these to the tests, they shouldn't be exposed to users nor run when solving an NLP.

cvxpy/problems/problem.py
Transurgeon: moved in https://github.com/cvxgrp/DNLP/pull/156

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
Transurgeon: let's leave this as a TODO.. I poked around and they didn't seem to have this (I think maybe because it is a C interface)?

cvxpy/settings.py
Transurgeon: Since we have our own derivative checker we can probably just revert it back to 1e-10. Will do that.

cvxpy/expressions/variable.py
Transurgeon: added in https://github.com/cvxgrp/DNLP/pull/158

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
Transurgeon: added in https://github.com/cvxgrp/DNLP/pull/158, documents the bounds class as well as NLP solver a bit better.

cvxpy/problems/problem.py
Transurgeon: made a PR for this in DNLP master, thanks for the great suggestion!

cvxpy/problems/problem.py
Transurgeon: yes I think I refactored that for DNLP master.

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
Transurgeon: I think I cleaned this up as well on master, I'll make another PR to cvxpy master soon.

cvxpy/reductions/solvers/nlp_solvers/uno_nlpif.py
Transurgeon: nothing it seems like, will fix at some point.

.github/workflows/test_nlp_solvers.yml
Transurgeon: We fixed the issue by automatically installing ipopt through system commands.

Transurgeon:
@PTNobel @SteveDiamond @rileyjmurray could you have another look at this PR please?
The summary of changes are below:
- we added the sparsediffpy package as a dependency to compute derivatives.
- documentation for the NLP solver and dnlp2smooth reductions, sample_bounds attribute for variables and many more. Thanks @rileyjmurray for the remainder on this end.
- one documentation page in the user guide, which follows the structure of the other disciplines.
- we moved the derivative checker to the tests.. it shouldn't be used by users.
- a lot of cleanups to the solving chain and other small things. Thanks again to @PTNobel for pointing them out.
- caching the oracles class for multiple solves in best_of. Thanks @SteveDiamond for the suggestion/reminder.

@dance858 and I also did a thorough scan of the changes and we are quite happy about the state of the PR/fork.
Apart from test changes, and new canonicalization methods, there are very little changes to the existing CVXPY infrastructure.

One big question is moving the non-convex atoms to a new namespace. Would it be fine to do that in another PR? It would be nice if we could discuss this as well more. I agree that if we create many new atoms it would be good to put them under a new namespace, but maybe we could just keep the trigonometric ones for now?

The code is certainly not perfect, and we will definitely keep making many improvements over time. But I think we were able to integrate most of the big necessary changes for the project. It is also quite tedious to keep the fork synced (especially with all the recent PRs we have been doing). So to summarize, I think it is ready to be merged into CVXPY master!

cvxpy/atoms/affine/broadcast_to.py
```diff
@@ -30,6 +30,7 @@ class broadcast_to(AffAtom):
     def __init__(self, expr, shape) -> None:
         self.broadcast_shape = shape
         self._shape = expr.shape
+        self.broadcast_type = None
```
SteveDiamond: What is this for? can you add a comment?

CLAUDE.md
```diff
@@ -1,29 +1,92 @@
-# CVXPY Development Guide
```
SteveDiamond, line 1: Can you comment on the changes made to this file and the reasons for them? I see the discussion of DNLP but the changes are much broader than that.

SteveDiamond:
Thanks for all your work on this PR! I agree we should focus on the code as is and defer further design decisions like the nlp namespace. There will be many rounds of revision of all new features in 1.9 before we release so we should take an iterative development approach.

cvxpy/atoms/affine/broadcast_to.py
Transurgeon: it was to compute the derivatives.. but that is now done elsewhere. I'll remove it.

CLAUDE.md
Transurgeon: I haven't been keeping track of the changes of the CLAUDE.
Maybe I'll just revert the changes for now?

cvxpy/atoms/__init__.py
```diff
@@ -161,6 +163,12 @@
     ptp
 ]
 
+NON_SMOOTH_ATOMS = [
```
PTNobel: This list feels incomplete; where are the norms etc?

cvxpy/atoms/affine/transpose.py
```diff
@@ -116,6 +116,7 @@ def graph_implementation(
         """
         return (lu.transpose(arg_objs[0], self.axes), [])
 
+
```
PTNobel: Do we need this new line

cvxpy/atoms/elementwise/hyperbolic.py
```diff
@@ -0,0 +1,203 @@
...
+    def numeric(self, values):
+        """Returns the elementwise sinh of x.
+        """
+        return np.sinh(values[0])
+    
+    def sign_from_args(self) -> Tuple[bool, bool]:
+        """Returns sign (is positive, is negative) of the expression.
+        """
+        # Always unknown.
+        raise NotImplementedError("sign_from_args not implemented for sinh.")
```
PTNobel: Shouldn't this just be False, False?

cvxpy/atoms/elementwise/kl_div.py
```diff
@@ -94,3 +98,6 @@ def _domain(self) -> List[Constraint]:
         """Returns constraints describing the domain of the node.
         """
         return [self.args[0] >= 0, self.args[1] >= 0]
+    
+    def point_in_domain(self, argument=0):
```
PTNobel: What's the contract on point in domain? It has a different signature in different places and I can't find a base implementation

cvxpy/atoms/elementwise/trig.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def is_decr(self, idx) -> bool:
+        """Is the composition non-increasing in argument idx?
+        """
+        return False
+
+    def _domain(self) -> List[Constraint]:
+        """Returns constraints describing the domain of the node.
+        """
+        return [-np.pi/2 < self.args[0], self.args[0] < np.pi/2]
```
PTNobel: This is pretty restrictive no?

PTNobel: Also wouldn't this error with the no strict inequalities message

cvxpy/constraints/zero.py
```diff
@@ -133,6 +137,10 @@ def is_dcp(self, dpp: bool = False) -> bool:
                 return self.expr.is_affine()
         return self.expr.is_affine()
 
+    def is_dnlp(self) -> bool:
+        """A zero constraint is DNLP if its argument is smooth representable."""
```
PTNobel:
```suggestion
        """An equality constraint is DNLP if its argument is smooth representable."""
```

cvxpy/expressions/expression.py
```diff
@@ -327,6 +327,14 @@ def is_affine(self) -> bool:
         """Is the expression affine?
         """
         return self.is_constant() or (self.is_convex() and self.is_concave())
+    
+    @perf.compute_once
+    def is_smooth(self) -> bool:
+        """Is the expression smooth?
+        """
+        return self.is_constant() or (
```
PTNobel, line 335: Do we need the first case

cvxpy/expressions/variable.py
```diff
@@ -31,7 +31,16 @@
 
 
 class Variable(Leaf):
-    """The optimization variables in a problem."""
+    """The optimization variables in a problem.
+
+    Attributes
+    ----------
+    sample_bounds : tuple[np.ndarray, np.ndarray] | None
```
PTNobel, line 38: I'm not a giant fan of only documenting this one

cvxpy/problems/problem.py
```diff
@@ -290,6 +290,13 @@ def is_dcp(self, dpp: bool = False) -> bool:
         """
         return all(
           expr.is_dcp(dpp) for expr in self.constraints + [self.objective])
+    
+    @perf.compute_once
```
PTNobel, line 294: We're pretty inconsistent about where we put these; which is fine since we don't need them when they're faster to compute, but I just want to confirm you were thoughtful about this

cvxpy/reductions/dnlp2smooth/canonicalizers/div_canon.py
```diff
@@ -0,0 +1,55 @@
...
+
+import numpy as np
+
+from cvxpy.atoms.affine.binary_operators import multiply
+from cvxpy.expressions.variable import Variable
+
+MIN_INIT = 1e-4
+
+# We canonicalize div(f(x), g(x)) as z * y = f(x), y = g(x), y >= 0.
+# In other words, it assumes that the denominator is nonnegative.
```
PTNobel, line 25: Is this documented anywhere a user would see this

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
PTNobel: They seem to still be duplicated

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
```diff
@@ -0,0 +1,50 @@
...
+def asinh_canon(expr, args):
+    if isinstance(args[0], Variable):
+        return expr, []
+    else:
+        t = Variable(args[0].shape)
+        if args[0].value is not None:
+            t.value = args[0].value
+        return expr.copy([t]), [t == args[0]]
+
+def atanh_canon(expr, args):
```
PTNobel, line 19: Why does this not check if it's variable first

PTNobel, line 19: Why does this not check if arg[0] is a variable first?

cvxpy/reductions/dnlp2smooth/canonicalizers/log_canon.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import numpy as np
+
+from cvxpy.expressions.variable import Variable
+
+MIN_INIT = 1e-4
+
+def log_canon(expr, args):
+    t = Variable(args[0].shape, bounds=[0, None])
```
PTNobel: Why is this nonnegative

PTNobel (changes requested):
GitHub is breaking on me so I'm only halfway done.

I'm really confused why we're not using the bounds propagation infrastructure; I think it's perfect for this

cvxpy/reductions/dnlp2smooth/canonicalizers/multiply_canon.py
```diff
@@ -0,0 +1,63 @@
...
+# If a user insert x * x where x is a variable it gets canonicalized to 
+# square(x) before this function is called.
+def multiply_canon(expr, args):
+    t1 = args[0]
+    t2 = args[1]
+    constraints = []
+
+    # if either is constant, no canonicalization needed
+    if t1.is_constant() or t2.is_constant():
+        return expr.copy([t1, t2]), []
```
PTNobel, line 30: Why return a copy with the same args

cvxpy/reductions/dnlp2smooth/canonicalizers/power_canon.py
```diff
@@ -0,0 +1,55 @@
...
+    if p == 0:
+        return ones, []
+    elif p == 1:
+        return x, []
+    elif isinstance(p, int) and p > 1:
+        if isinstance(x, Variable):
+            return expr.copy(args), []
+
+        t = Variable(shape)
+        if x.value is not None:
```
PTNobel, line 39: I thought at this point in the DNLP process all expressions had values

PTNobel: We claim that in the pnorm canon

cvxpy/reductions/dnlp2smooth/canonicalizers/prod_canon.py
```diff
@@ -0,0 +1,39 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import numpy as np
+
+from cvxpy.expressions.variable import Variable
+
+
+def prod_canon(expr, args):
```
PTNobel: I'm surprised we're not using a log canonicalization like for geo_mean; just want to confirm that's intentional

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
PTNobel: Fine if we're consistent

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
```diff
@@ -0,0 +1,111 @@
...
+        # smoothen constraints
+        for constraint in problem.constraints:
+            # canon_constr is the constraint re-expressed in terms of
+            # its canonicalized arguments, and aux_constr are the constraints
+            # generated while canonicalizing the arguments of the original
+            # constraint
+            canon_constr, aux_constr = self.canonicalize_tree(
+                constraint, False)
+            canon_constraints += aux_constr + [canon_constr]
```
PTNobel, line 60: We can save some list allocations by using extend and append, but I don't think it matters

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
```diff
@@ -0,0 +1,110 @@
...
+limitations under the License.
+"""
+
+import numpy as np
+
+import cvxpy as cp
+
+# Import the low-level C bindings
+try:
+    from sparsediffpy import _sparsediffengine as _diffengine
```
PTNobel: I don't like erroring on import for a missing dependency. We don't do that anywhere else in the library. Can we not move this to the init or something?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
```diff
@@ -0,0 +1,427 @@
+
+"""Converters from CVXPY expressions to C diff engine expressions.
```
PTNobel: I think we generally put the licence text about the description of the file.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
```diff
@@ -0,0 +1,110 @@
+"""Wrapper around C problem struct for CVXPY problems.
```
PTNobel: I generally prefer license text above the description of the file

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
```diff
@@ -0,0 +1,427 @@
...
+"""
+
+import numpy as np
+from scipy import sparse
+
+import cvxpy as cp
+from cvxpy.reductions.inverse_data import InverseData
+
+# Import the low-level C bindings
+try:
```
PTNobel: I really don't like doing this in a bunch of places: could we add it to cvxpy_types or something?

cvxpy/reductions/solvers/nlp_solvers/ipopt_nlpif.py
PTNobel: Why is this at the same indentation as the nlp =

cvxpy/reductions/solvers/nlp_solvers/nlp_solver.py
PTNobel: That's fine with me for now then; I think this could go in ExactCone2Cone in the future

PTNobel (changes requested):
All tests seem to be missing copyright headers

PTNobel:
Regarding your comments, I started my review before I saw it.

I don't think this PR is ready to merge. Notably, many files are messing copyright headers. I'm pretty sure the domain property of tan always errors when accessed.

Also, I think some edits made in response to my previous review never made it onto this branch possibly? There's a handful of comments you replied to saying something was done that wasn't. 

I think for all of my code quality changes (of which there are 25 small changes or so), as long as they're listed on an open issue somewhere (could be here with a check list; could be on cvxgrp/DNLP) I'll be satisfied. I'm worried if they're only on this PR it'll be very hard to track them after merge. 

Regarding my bigger concerns, (1. using the get_bounds infrastructure to narrow bounds, 2. deduplicating canon functions and having consistent implementation for the different smooth canons) I'm happy to just have an issue filed for each and for them to be addressed later.

cvxpy/atoms/__init__.py
PTNobel: Ahhh okay we only support the 2-norm, but I still think it should be here. Same with sum_largest

PTNobel: Isn't Huber no smooth under our definition

dance858: Nice catch. I think this is a fossil since pre-DNLP (before we enforced a discipline). We don't use so I removed it.

cvxpy/atoms/affine/transpose.py
dance858: Done.

cvxpy/atoms/elementwise/hyperbolic.py
dance858: Nice catch. I fixed this and the other similar errors in the same file.

cvxpy/atoms/elementwise/kl_div.py
dance858: We were never actually using the point in domains, so I removed them.

We used to use them, but when we changed so we set the initial values of the original variables before the canonicalization rather than after, we don't use point in domain. Nice catch!

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
dance858: Because atanh does not have full domain, so we should always introduce a new variable.

Also, I fixed the duplication. Thanks for pointing this out.

dance858: Because atanh does not have full domain, so we should always introduce a new variable.

cvxpy/reductions/dnlp2smooth/canonicalizers/log_canon.py
dance858: Because the argument to log is always nonnegative.

cvxpy/reductions/dnlp2smooth/canonicalizers/multiply_canon.py
dance858: I think it is to guarantee that the rest of the expression is canonicalized? If we don't copy, it does not seem to canonicalize an expression like (1/f(x)) * 3 correctly. @Transurgeon, do you remember?

cvxpy/reductions/dnlp2smooth/canonicalizers/power_canon.py
dance858: Yes, in principle all expressions should have values, so the checks of the form if value is not None would be unnecessary. However, if we accidentally fail to propagate a value to some expression, removing these checks would result in an obscure crash. Instead, we keep the explicit checks. After canonicalization, we verify that all values have been initialized. If any value is missing, we raise a clear error, something like
"This variable does not have a value. This is a bug and should be reported."

dance858:
> All tests seem to be missing copyright headers

Done

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
dance858: Agreed. Fixed.

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
dance858: Here we followed the structure in dcp2cone.py

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
dance858: Agreed. Fixed.

cvxpy/reductions/dnlp2smooth/canonicalizers/prod_canon.py
dance858: For prod we don't necessarily know that all variables are positive, so we don't do the same trick as in geo-mean trick.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
dance858: @Transurgeon can you fix this comment?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
dance858: @Transurgeon Can you fix this comment?

cvxpy/atoms/elementwise/trig.py
dance858: We must restrict the domain of tan to one interval so the domain is connected. The interval we chose is typically the interval you think about what you say tan. So I think this domain makes sense and is not restrictive.

We don't use the domain anywhere, so we didn't think carefully about strict versus constrict inequality. When I check the domain of log I see that it is x >= 0, so I'll change the domain of tan to non-strict to be consistent with that.

cvxpy/reductions/dnlp2smooth/canonicalizers/div_canon.py
dance858: I added it to the documentation. I also raise a warning if the denominator is not nonnegative.

cvxpy/expressions/expression.py
dance858: Yes we need it. Alternatively, we could define is_linearizable_convex and is_linearizable_concave in the constant class. Let me know what you prefer

cvxpy/problems/problem.py
dance858: @Transurgeon Can you look into this?

dance858:
I fixed most of the comments but tagged @Transurgeon in a few that he'll fix later today. Then we'll update the PR. I also implemented a common canonicalizer for smooth atoms with full domain. Great suggestion. Many thanks for the review. Please check if my answers above make sense.

cvxpy/expressions/variable.py
dance858: @Transurgeon Can you look into this?

cvxpy/reductions/dnlp2smooth/canonicalizers/multiply_canon.py
Transurgeon: I explored this a bit more and think I have an answer.
The arguments that are passed to the canonicalizer are themselves canonicalized.. hence they might differ from the original arguments of the expression. Thus we must point the expression to the new arguments (as a copy). This is actually existing behavior in both [dcp2cone](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/dcp2cone/dcp2cone.py#L148) and dnlp2smooth as well.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
Transurgeon: thanks for pointing this out! I ended up just removing the try catch entirely.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
Transurgeon: It seems like it can just be imported directly without the try-catch.. without any other issues.

cvxpy/expressions/expression.py
PTNobel: We do define is_linearizable_convex andis_linearizable_concave on the Constant class since it inherits from Leaf.

Regardless, not an issue and there might be some special cases about convex functions of constants that are worth supporting so let's keep it!

cvxpy/reductions/dnlp2smooth/canonicalizers/hyperbolic_canon.py
PTNobel: Ahh of course. Thanks!

cvxpy/reductions/dnlp2smooth/canonicalizers/log_canon.py
PTNobel: LOL I swear I knew that

cvxpy/reductions/dnlp2smooth/canonicalizers/multiply_canon.py
PTNobel: Huh; today I learned! Thanks William.

cvxpy/reductions/dnlp2smooth/canonicalizers/power_canon.py
PTNobel: Ahhh makes sense. Thank you Daniel.

cvxpy/reductions/dnlp2smooth/canonicalizers/prod_canon.py
PTNobel: Of course. Thanks.

PTNobel:
Thanks Daniel! I looked at the PR in the other repo and it looks great. I think there's maybe 3 or 4 things to file issues about, merge your PR there, and port all the changes from master to cvxpy-merge on the other repo and then we'll be all set!

Really excited to get this all merged. It is super exciting.

Transurgeon:
> Thanks Daniel! I looked at the PR in the other repo and it looks great. I think there's maybe 3 or 4 things to file issues about, merge your PR there, and port all the changes from master to cvxpy-merge on the other repo and then we'll be all set!
> 
> Really excited to get this all merged. It is super exciting.

Thanks Parth! I'll go ahead and update this branch then. I'll then move over the issues to cvxpy master (there must be a way to do this somehow?). Then you guys could have a final look.. pinging @SteveDiamond and @rileyjmurray for approval as well.

Transurgeon:
@PTNobel sorry for the ping.. but I just moved all issues from DNLP to cvxpy (with a new tag for them).
Out of the things you mention above, it seems like only the ``get_bounds`` infrastructure change is missing?
I think @dance858 has worked on deduplicating the common canon files.
It would be nice if you could have a final look and approve this if it looks good!

PTNobel:
I think get_bounds is the main thing missing. I don't see an issue for it yet?

Transurgeon:
> I think get_bounds is the main thing missing. I don't see an issue for it yet?

Just opened one now.. feel free to add more info if needed.

PTNobel (approved):

PR #3123, opened 2026-02-05
Add value propagation for warm-startable solvers

PTNobel:
## Description
Add "value propagation" to CVXPY, mirroring the existing bound propagation pattern. When a solver supports warm starting from initial values, propagate variable `.value` attributes through the expression tree to set values on auxiliary variables created during canonicalization. This gives solvers a complete warm start (not just original variable values, but also the derived auxiliary variable values).

Adds a `WARM_STARTABLE` flag to the `Solver` base class, wires it through `SolverInfo`/`SolvingChain`, creates a `get_expr_value_if_supported` helper, and uses it in 8 canonicalizers (exp, log, pnorm, power, abs, max, maximum, norm_inf). Marks Gurobi and Xpress (conic + QP) as `WARM_STARTABLE = True`.

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel:
@claude Rereview this PR. Note that the benchmarks are for measuring canonicalization, so they won't test anything interesting here. The end-to-end tests would be nice, but we don't have those commercial solvers in CI. So we need to wait for an open-source solver to add better warm starting support.

PTNobel:
@claude can you fix all the failing tests by putting guards in all the value implementations that are erroring so they return None if there is no value set?

PTNobel:
@Transurgeon @dance858 

FYI, I made a few changes to DNLP's solving chain and reductions here. It more closely matches the DCP chain now where we have a solver context object everywhere that is used to make decisions about canonicalizatoin.

Right now, all DNLP solvers use the same context object, but this let's us have consistent behavior in DCP about value propagation and makes the DNLP solving chain more extensible.

Open to thoughts and further discussion.

PTNobel:
@dance858 @Transurgeon can I get reviews on this PR? It obviously isn't directly DNLP related, but I think we'll want to share infra between DNLP and this branch eventually... (for example in eliminate_pwl, we already have the same code)

nej1gotnochill:
Hello @PTNobel  I’ve started looking into this PR and the current codebase to understand the value propagation flow and how it interacts with warm startable solvers. The main idea quite makes sense to me is to propagate the values of user provided variables through canonicalization so that auxiliary variables also receive appropriate initial values before being passed to the solver. I’ll first go through the existing implementation and the merge conflicts carefully, and then work on updating the PR against the current codebase. Have a nice day !

nej1gotnochill:
I've gone through the PR and have compared it with the current master.There are quite aa few conflicts because the PR itself is fairly old like it was posted in February and ever since then some parts of the current codebase have changed significantly. 

Meanwhile through my investigation I have identified a few points I'd like to clarify before I start resolving everything: 

**1) Should the DNLP related changes be kept in the PR or it would be better to make them deferred to a followup?**
**2) Secondly the master currently has some unconditional value propogation in the PWL canonicalizers(like the `abs` , `max `,etc.). Also should the PR's solver context dependent behaviour replace that?**
**3) I saw that the PR also added the `solver_context ` parameters across many DGP2DCP canonicalizers,  although they don't appear to use them. Should I preserve those changes or avoid the additional churn?**
**4) Lastly , should SCS still be considered warm startable for this feature?**

I will be more than happy to proceed with the core DCP value propogation changes first if that's the preferred scope. I just wanted to confirm these points before resolving the conflicts so that I maintain the direction to which this PR is directed , Thank you , Have a nice day!

nej1gotnochill:
After systematic investigation across multiple solver paths and problem structures I don't think there's a convincing use case for auxiliary value propagation that justifies the additional implementation complexity.

The key findings in my investigation are as follows:
1. The HiGHS interface does not consume `init_value` on the first solve, subsequent warm starts uses the cached solution which already contains the auxiliary values. 
2. For the formulations tested, auxiliary values are determined by the original variables at the solution. 
3. SCS and HiGHS benchmarks showed no meaningful performance improvement from propogation.
4. The remaining potentially relevant path appears to be Gurobi's handling of initial variable values but I did not investigate / benchmark that path because it wasn't part of the requested investigation.

Given the evidence so far and the additional changes that would be required to make the HiGHS path relevant I think that closing this PR is the appropriate conclusion. Thank you , have a nice day !

Issue #3059, opened 2026-01-03
CuClarabel always solves verbosely


govindchari:
**Describe the bug**
When using CuClarabel in CVXPY, the CuClarabel solver logs are always printed even when the `verbose=False` is passed to `problem.solve()`. 

**Expected behavior**
No solver output should be printed to screen.

**Version**
 - OS: Ubuntu
 - CVXPY Version: Building off master and using Clarabel.jl v0.11.0

**Additional context**

PR #3132, opened 2026-02-09
Fix CUCLARABEL ignoring verbose flag

prasu-baran:
## Description
This PR ensures that the verbose flag passed to prob.solve() is correctly respected by the CUCLARABEL solver.
Earlier , **CUCLARABEL**  used Clarabel’s default verbosity, which could result in solver output being printed even when **_verbose=False._** My proposed change explicitly forwards the verbosity setting to Clarabel via its Settings object.

Issue link : #3059 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import contextlib
+import io
 import math
 import os
 import re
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_conic_solvers.py
```diff
@@ -595,7 +596,21 @@ def update_parameters(P, A, b, q):
         # consecutive solves, no data update
         result1 = prob.solve(solver=cp.CLARABEL, warm_start=False)
         self.assertAlmostEqual(result1, result2)
-
+
```
PTNobel, line 599: No blank lines with just spaces

PTNobel (commented):
Overall looks great! Thanks for doing this. One minor formatting thing, and then this is good to merge.

cvxpy/tests/test_conic_solvers.py
prasu-baran: Before every new function there is a blank line existed everywhere and i have done with like that only

prasu-baran:
Hi @PTNobel !,
All the functions are placed like that only have seen in the same file as well in other files too

bsauvage1:
It would be great to get this fixed. As a temp workaround, here's an unelegant way to do it, with a context manager around solve() method if verbose=False:
```
@contextlib.contextmanager
def suppress_julia_output():
    """Redirect Julia's stdout/stderr to devnull at the Julia runtime level.
    """
    from juliacall import Main as jl

    jl.seval(
        "global _cuc_old_out = redirect_stdout(devnull); global _cuc_old_err = redirect_stderr(devnull)"
    )
    try:
        yield
    finally:
        jl.seval("redirect_stdout(_cuc_old_out); redirect_stderr(_cuc_old_err)")

```

PR #3121, opened 2026-02-04
DRAFT Make quad_form DPP compatible for QP + conic quadratic solvers

SteveDiamond:
## Description
This is a draft MR to make quad_form DPP compatible for QP + conic quadratic solvers. Adding a new scope seems like the best way to enable this feature, and then the solving chain needs to be reorganized to first choose a solver and then reason about DPP. I will sync with Parth about the solving chain refactor.

## Type of change
- [x] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [x] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #3142, opened 2026-02-14
Add DPP support for quad_form(x, P) with parametric P

PTNobel:
Rebase of #3121 onto master.

## Description
Enable DPP (Disciplined Parametrized Programming) for `quad_form(x, P)` when `P` is a `Parameter(PSD=True)` and the solver supports quadratic objectives.

Key changes:
- Add `quad_form_dpp_scope` context manager in `scopes.py` that relaxes `QuadForm.is_atom_convex/concave()` to accept param-affine P matrices
- Enter `quad_form_dpp_scope` in the solving chain when the solver supports `quad_obj`, so that the DPP check recognizes parametric P as valid
- Handle parametric P in `coeff_extractor.py` quadratic coefficient extraction
- Add comprehensive tests in `test_quad_dpp.py`

Problems like `min x'Px` where `P` is a PSD Parameter are now DPP when using solvers that support quadratic objectives (CLARABEL, SCS, OSQP, etc.), enabling cached re-solves.

## Type of change

- [x] Write unittests.


cvxpy/utilities/coeff_extractor.py
```diff
@@ -130,22 +131,79 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                    # Only need shape info, not value.
+                    n = P_expr.shape[0]
+                    P_param_offset = self.param_id_map[P_expr.id]
                 else:
-                    P = sp.coo_matrix(P.value)
+                    # For constant P, use the numeric value.
+                    assert (
+                        P_expr.value is not None
+                    ), "P matrix must be instantiated before calling extract_quadratic_coeffs."
+                    P_val = P_expr.value
```
PTNobel, line 154: So this always casts to a `np.ndarray` right? We have to use `.sparse_value` to get the sparse version?

PTNobel (commented):
So does this only support Parameters in cp.quad_form? No support for expressions of Parameters? I would presume a user would have to psd_wrap any expression of Parameters, but still feels like that should exist?

cvxpy/utilities/coeff_extractor.py
PTNobel: Yes this always makes a dense matrix; since P isn't a parameter, if the user had a sparse constant, this does fill in.

PTNobel:
@maxschaller uhhh another breakage of your package :sweat_smile: 

This one looks to be caused by changes in the caching of canonicalization depending on scopes? I'm not really sure. Happy to sync and explore on Saturday if you're coming to the hackathon.

maxschaller:
> @maxschaller uhhh another breakage of your package 😅
> 
> This one looks to be caused by changes in the caching of canonicalization depending on scopes? I'm not really sure. Happy to sync and explore on Saturday if you're coming to the hackathon.

@PTNobel thanks for the heads-up. Could you please rebase onto master / merge master back? Compatibility with CVXPYgen might be (partially) fixed by #3149 already.

cvxpy/utilities/coeff_extractor.py
SteveDiamond: Good question! I traced this through and `Constant.value` actually does preserve sparsity — `Constant.__init__` checks `intf.is_sparse(value)` and stores it via `DEFAULT_SPARSE_INTF` when true. So `P_expr.value` returns a sparse matrix when the user passed one in.

Verified empirically by instrumenting `extract_quadratic_coeffs`:

```
qf_atom.P type:        Constant
P_expr.value type:     scipy.sparse._csc.csc_array
P_expr.value is sparse: True
```

Interestingly, the old code on master had a latent bug here — it checked `sp.issparse(P)` on the Constant object itself (always False), making the sparse-to-COO branch dead code. The new code correctly checks `P_val` (the actual value), so the sparse path works now.

SteveDiamond (approved):

SteveDiamond:
@Transurgeon did you want to look at this?

PTNobel:
@claude review the changes

PTNobel:
@claude Please rereview this PR.

cvxpy/problems/problem.py
```diff
@@ -361,14 +362,39 @@ def is_dpp(self, context: str = 'dcp') -> bool:
...
+                    obj_dpp = self.objective.is_dcp(dpp=True)
+                # Check constraints: with scope for QCQP, without for QP.
+                if quad_form_dpp == 'qcqp':
+                    with scopes.quad_form_dpp_scope():
+                        constrs_dpp = all(
+                            c.is_dcp(dpp=True) for c in self.constraints)
+                else:
+                    constrs_dpp = all(
+                        c.is_dcp(dpp=True) for c in self.constraints)
+                expr_dpp = obj_dpp and constrs_dpp
```
Transurgeon: do we need the ``qcqp`` context? I don't see it being used and am wondering if its even necessary in the first place?

cvxpy/utilities/coeff_extractor.py
Transurgeon, line 1: the changes in this file are a bit complicated, I don't quite get it.. could add some explanations or documentation maybe? It creates a new code path that is very different than what was there before.
Also, it seems like the parametric case doesn't handle ``SymbolicQuadForm`` for non-scalar quadforms.. does this mean we only support parameters appearing in explicit QuadForms?

Transurgeon (commented):
I am not sure I would approve this just yet.. sorry for the pushback.
The changes are quite complicated and I would like to understand better.
I'll do another pass for the changes in coeff_extractor, the other scope changes make sense though.

cvxpy/problems/problem.py
PTNobel: No it's future proofing for adding QCQP support; we can drop it

cvxpy/utilities/coeff_extractor.py
```diff
@@ -130,22 +130,93 @@ def extract_quadratic_coeffs(self, affine_expr, quad_forms):
...
+                        # Any parametric coefficient would make this quadratic in parameters.
+                        non_const_mask = c_nonzero_idxs != (num_params - 1)
+                        if np.any(non_const_mask):
+                            raise ValueError(
+                                "DPP quad_form requires x to be parameter-free. "
+                                "Found parameter dependence in x, which would make "
+                                "the expression quadratic in parameters."
+                            )
+
+                        coef_val = c_nonzero_vals[0]
```
PTNobel: @Transurgeon can you clean up this logic? ^^^

Transurgeon: claude suggested something like this:
```
  coef_val = c_part[0, -1]                                                                                                                                                                                                                                            
  if coef_val == 0:                                     
      P_tup = TensorRepresentation.empty_with_shape((n, n))                                                                                                                                                                                                           
  else:
      assert not np.any(c_part[0, :-1]), (                                                                                                                                                                                                                            
          "Bug: parametric quad_form has parameter-dependent affine head"
      )                                                                                                                                                                                                                                                               
      # ... proceed with coef_val and P_coeffs extraction
```

Transurgeon: I made the changes in the newest commit to the branch.. tests all seem to pass

Transurgeon (approved):
Looks good to me after discussing offline with @PTNobel .
Feel free to merge this!

PR #3147, opened 2026-02-17
Remove shared variable/parameter IDs from reductions

PTNobel:
## Description

Stop sharing IDs between original and transformed leaves across reductions (`CvxAttr2Constr`, `Dgp2Dcp`, `Complex2Real`). Each reduction now creates fresh variable/parameter IDs and explicitly tracks mappings, enabling correct chain-rule composition for `backward()` and `derivative()`.

### Bug fix

`derivative()` with complex variables silently dropped imaginary deltas. When `Complex2Real` splits a complex variable into real and imaginary parts, the imaginary variable gets a fresh ID. On CVXPY 1.8.1, `split_solution` only extracted outer variable IDs, so the imaginary delta was lost and `z.delta` was always real-valued.

### Key changes

- **Dict-based chain rule API**: Change `var_forward`/`var_backward`/`param_forward`/`param_backward` from per-variable `(var, value) -> value` signatures to dict-based `{id: value} -> {id: value}` signatures on the `Reduction` base class. This naturally handles 1-to-many mappings (e.g. complex -> real + imag).
- **`var_id_map` / `param_id_map` properties**: Each reduction exposes `{orig_id: [new_id, ...]}` mappings. `Chain.compose_var_id_map()` / `compose_param_id_map()` compose these across the full chain.
- **`CvxAttr2Constr`**: Fresh IDs for reduced variables/parameters; dict-based differentiation methods.
- **`Dgp2Dcp`**: Fresh IDs; dict-based API; correctly handles parameters that appear both as `log(p)` and directly (e.g. power atom exponents).
- **`Complex2Real`**: Dict-based `var_forward`/`var_backward`/`param_forward`/`param_backward`; proper complex <-> real/imag chain rule composition.
- **`Problem.backward` / `derivative`**: Pipe dicts through reduction chains instead of per-variable loops.
- **Remove provenance infrastructure**: Delete `leaf_of_provenance`, `set_leaf_of_provenance`, `attributes_were_lowered`, and provenance handling in `cone_matrix_stuffing`.
- **`perspective_canon`**: Use `var_id_map` to look up reduced variable columns instead of assuming shared IDs.

## Type of change
- [x] Bug fix

- [x] Write unittests.


PTNobel:
So this breaks gcsopt, but I think it is still worth doing:

1. I don't like that there are multiple possible answers to "what is the shape of the variable with id x"
2. I think we want to move downstream users aways from depending on this quirk of variable IDs if we're serious about providing concrete matrices in the future
3. I think the amount of logic via variable provence was hidden in matrix_stuffing and param_prog was bad, and I think it is good to make people use the chain to get that info
4. I added a proper API to track variable replacements in the reduction chain that users can use.
5. I think these changes to having more stuff at the chain level and allowing variable substitution this way, gives us way more flexibility around real long-term goals like automatic problem rescaling etc

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -246,8 +247,7 @@ def apply(self, problem):
                 for key in reduction_attributes:
                     if new_attr[key]:
                         new_attr[key] = None if key == 'bounds' else False
-                reduced_param = Parameter(n, id=param.id, **new_attr)
-                reduced_param.set_leaf_of_provenance(param)
+                reduced_param = Parameter(n, **new_attr)
```
maxschaller: Instantiating the reduced parameter with `name=param.name()` helps name parameters during code generation, which stores parameters in their reduced form. For pure cvxpy users, this should make no difference, as they would not access the reduced parameter directly.

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -263,26 +263,55 @@ def apply(self, problem):
...
         return cvxtypes.problem()(obj, constr), inverse_data
 
+    @property
+    def var_id_map(self):
+        return {orig.id: new.id for orig, new in self._variables.items()}
+
     def update_parameters(self, problem) -> None:
         """Update reduced parameter values from original parameters."""
         for param, reduced_param in self._parameters.items():
             reduced_param.value = lower_value(param)
```
maxschaller: We should check `if param.value is not None`, since `update_parameters` might be called before parameter values are assigned (for example during code generation).

PTNobel: Fixed

cvxpy/reductions/cvx_attr2constr.py
PTNobel: Fixed

cvxpy/reductions/cvx_attr2constr.py
```diff
@@ -263,27 +263,56 @@ def apply(self, problem):
...
-            return None
-        return recover_value_for_leaf(param, dparams[reduced_param.id])
-
-    def param_forward(self, param, delta):
-        """Transform full-size delta to reduced-size delta."""
-        if param not in self._parameters:
-            return None
-        reduced_param = self._parameters[param]
-        return {reduced_param.id: lower_value(param, delta)}
+    def var_forward(self, dvars):
```
maxschaller: Purely cosmetic: Keep `var_backward` before `var_forward`?

PTNobel:
I'm completely convinced this change is worth it: Complex2Real decomposes one Variable to two, but only one variable inherits the original ID of the Variable. So anything depending on the variable id mapping will mess up on complex problems I think

PTNobel:
I was able to write 3 tests that fail on master and pass here, so this is a bug fix.

PTNobel:
We're not running test_derivative in CI because we don't have diffcp installed.

cvxpy/problems/problem.py
```diff
@@ -1219,20 +1219,23 @@ def backward(self) -> None:
         backward_cache = self._solver_cache[s.DIFFCP]
         DT = backward_cache["DT"]
         zeros = np.zeros(backward_cache["s"].shape)
-        # del_vars: dictionary of variable gradients (delta/∂ with respect to variables)
-        # Maps variable IDs to their gradient arrays for the backward pass
-        del_vars = {}
+        # broadcast_to creates O(1) read-only views; safe because chain rule
+        # ops in var_backward/param_backward create new arrays, never mutate.
+        _ONE = np.float64(1.0)
+        _ZERO = np.float64(0.0)
```
Transurgeon: I don't understand this notation.. it seems very different than what we do usually. and also are these variables necessary? Also I don't fully understand the comment, why are we talking about mutation?

cvxpy/problems/problem.py
```diff
@@ -1313,41 +1306,42 @@ def derivative(self) -> None:
         backward_cache = self._solver_cache[s.DIFFCP]
         param_prog = self._cache.param_prog
         D = backward_cache["D"]
-        param_deltas = {}
+        # broadcast_to creates O(1) read-only views; safe because chain rule
+        # ops in param_forward/var_forward create new arrays, never mutate.
+        _ZERO = np.float64(0.0)
```
Transurgeon: same comment as above.

cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
```diff
@@ -78,10 +78,15 @@ def perspective_canon(expr, args, solver_context: SolverInfo | None = None):
...
+    # Build mapping from original var IDs to reduced var IDs.
+    # Reductions like CvxAttr2Constr may replace variables with new IDs.
+    var_id_map = chain.compose_var_id_map()
+
     end_inds = sorted(prob_canon.var_id_to_col.values()) + [x_canon.shape[0]]
 
     for var in expr.f.variables():
-        start_ind = prob_canon.var_id_to_col[var.id]
+        reduced_id = var_id_map.get(var.id, [var.id])[0]
+        start_ind = prob_canon.var_id_to_col[reduced_id]
```
Transurgeon, line 93: can you explain the semantics of this reduced_id?
Why are we using ``get`` with a var.id inside a list then indexing at zero?

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -331,14 +322,6 @@ def split_adjoint(self, del_vars=None):
...
             var = self.id_to_var[var_id]
             col = self.var_id_to_col[var_id]
-            if var.attributes_were_lowered():
-                orig_var = var.leaf_of_provenance()
-                if cvx_attr2constr.attributes_present(
-                        [orig_var], cvx_attr2constr.SYMMETRIC_ATTRIBUTES):
-                    delta = delta + np.swapaxes(delta, -2, -1)
-                    di = np.arange(delta.shape[-1])
-                    delta[..., di, di] /= 2
-                delta = cvx_attr2constr.lower_value(orig_var, delta)
```
Transurgeon, line 341: we should document somewhere that there is no more ``var.attributes_were_lowered`` or something similar. And this code won't be needed any more. With that being said, where do we ensure this behavior still happens in this refactor?

cvxpy/reductions/dgp2dcp/dgp2dcp.py
```diff
@@ -56,13 +56,28 @@ class Dgp2Dcp(Canonicalization):
...
 
     def accepts(self, problem):
         """A problem is accepted if it is DGP.
         """
         return problem.is_dgp()
 
+    @property
+    def var_id_map(self):
+        if self.canon_methods is None:
+            return {}
```
Transurgeon, line 73: where do we populate self.canon_methods? Maybe I missed something, but this was not quite clear to me when I looked at this file.

cvxpy/reductions/dgp2dcp/dgp2dcp.py
```diff
@@ -76,53 +91,84 @@ def update_parameters(self, problem) -> None:
             if param in self.canon_methods._parameters:
                 self.canon_methods._parameters[param].value = np.log(param.value)
 
-    def param_backward(self, param, dparams):
+    def var_backward(self, del_vars):
+        """Apply chain rule for exp transformation in backward diff.
+
+        For DGP, x_gp = exp(x_cone), so dx_gp/dx_cone = exp(x_cone) = x_gp.
+        Transforms from outer (original) var IDs to inner (log-space) var IDs.
+        """
```
Transurgeon, line 104: I don't understand this comment and the one below, why are we talking about exponential cones when we are actually in the ``dgp2dcp`` file?

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
```diff
@@ -91,8 +91,6 @@ def set_column_names_from_variables(lp, variables):
     """
     column_names = []
     for variable in variables:
-        # leaf_of_provenance() handles auto-generated vars (nonneg=True, etc.)
-        variable = variable.leaf_of_provenance() or variable
```
Transurgeon, line 95: why was this ``variable.leaf_of_provenance()`` call even here in the first place?
and how do we handle this more natively to ensure this still works?
This is more a question than a comment.

cvxpy/reductions/chain.py
```diff
@@ -2,6 +2,30 @@
...
+    Each step map comes from a reduction's ``var_id_map`` or
+    ``param_id_map``.  The result is a single mapping from the
+    outermost original IDs to lists of the innermost final IDs.
+    """
+    result = {}
+    for step_map in step_maps:
+        # Update existing mappings whose current targets were renamed.
+        for orig_id, cur_ids in result.items():
+            new_ids = []
+            for cur_id in cur_ids:
```
Transurgeon: that's a lot of for loops, is there any way to clean this up potentially?

cvxpy/tests/test_derivative.py
```diff
@@ -584,9 +658,11 @@ def test_param_used_in_exponent_and_elsewhere(self) -> None:
         alpha.delta = 1e-5
         problem.solve(solver=cp.DIFFCP, gp=True, requires_grad=True, eps=1e-5)
         self.assertAlmostEqual(x.value, 1 - base**(0.5) - 0.5**2)
+        # Check backward first — on CVXPY 1.8.1 this gave wrong gradient
         problem.backward()
-        problem.derivative()
         self.assertAlmostEqual(alpha.gradient, -np.log(base)*base**(0.5) - 2*0.5)
+        # Check derivative — on CVXPY 1.8.1 this raised KeyError
+        problem.derivative()
```
Transurgeon, line 665: these comments say that it raised an error but don't actually change the test assertions?
Is it because the order of calling derivative and the assertion swapped and now it passes?

Transurgeon (commented):
@PTNobel I finally got to reviewing this.
Overall I like the changes, there's a lot of nice cleanups. But there are many questions and things I don't understand, which I left as comments.
I think we can merge this once they are addressed.

cvxpy/reductions/dgp2dcp/dgp2dcp.py
PTNobel: We're not talking about exponential cones; `x_cone` is the variable in the cone program rather than the one in the gp

cvxpy/tests/test_derivative.py
PTNobel: We never ran the test_derivative in CI before, so the test used to fail

SteveDiamond (approved):
This is a nice cleanup. LGTM. Potentially worth adding some comments around the items will asked about, but not blocking.

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
PTNobel: We needed the variable names for constructing the LP files correctly. The variable name is now always propogated.

cvxpy/reductions/dgp2dcp/dgp2dcp.py
PTNobel: I am doing some new names which will hopefully be clearer...

cvxpy/reductions/dgp2dcp/dgp2dcp.py
PTNobel: In .apply; I'll add a comment

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
PTNobel: There are a ton of tests of this part of the codebase. Every PSD program tests it. I didn't think we needed more tests. Who are we documenting this for? We already talked to Tobia and Max, and I don't think this was a part of the public API, since you can't reach the affected variable object right?

I'm fine putting it in the release notes.

cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
PTNobel: perspective doesn't support complex variables, so we are guaranteed that there is only new id. Hence, we do the [0].

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
PTNobel: To clarify provenance-based recovery in split_solution/split_adjoint is now done by CvxAttr2Constr.var_forward/var_backward/invert via recover_value_for_leaf and the symmetric reshape. It is now in the reduction chain instead of floating over here.

cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
PTNobel: var_id_map maps var.id to a list of variables because some reductions introduce multiple variables for one original problem variable. (Notably complex).

PR #3150, opened 2026-02-19
Fix quad_form_canon silently dropping M1 term for indefinite P

goyalpalak18:
I found a pretty critical bug in `quad_form_canon` where it silently drops the positive-eigenvalue component of an indefinite matrix. This results in the solver optimizing a completely wrong objective function but still returning an `OPTIMAL` status. 

### The Problem
When a matrix is indefinite, `decomp_quad` correctly returns both `M1` (the positive-eigenvalue component) and `M2` (the negative-eigenvalue component). In `quad_form_canon.py` (lines 33-38), the code assigns the `M1` expression correctly, but then unconditionally overwrites it with the `M2` expression right below it. 

Because the `M1` term never reaches `quad_over_lin_canon`, the solver effectively minimizes only the negative-eigenvalue part ($-scale * \Vert M_2^T x \Vert^2$), which flips the geometry of the problem and returns bogus optimal values without throwing any exceptions. 

This is especially dangerous for pipelines using `cp.quad_form(x, P, assume_PSD=True)` with matrices built via floating-point math (e.g., $A^T A$ or covariance matrices) that might have tiny negative eigenvalues (~1e-14). 

### What I Changed
I made a couple of minor logic updates in `quad_form_canon.py` (lines 33-45) to make the branches mutually exclusive:
1. **Explicit Guard:** Added an `if M1.size > 0 and M2.size > 0:` check to catch indefinite matrices immediately. Instead of proceeding, it now raises a clear `ValueError` advising the user to validate their matrix construction.
2. **Proper Branching:** Changed the second unconditional `if M2.size > 0` to an `else` for the NSD-only case. 

### Impact
Now, instead of failing silently and giving users a mathematically incorrect `OPTIMAL` status, the code fails fast with an actionable error. The PSD and NSD paths remain completely unaffected.

I've verified the fix locally and all 21 relevant tests are passing cleanly. Let me know if you need anything else to get this merged!


goyalpalak18:
Hey @PTNobel ,would you mind taking a look at this when you get a chance? Thanks!

PTNobel:
This is super interesting! Thanks for catching it. Could you add some tests that fail on master but pass on this branch?

And also add some tests that show what the error we expect is?

goyalpalak18:
Thanks @PTNobel for the review!

Added `test_indefinite_assume_psd_raises` to cover this. It correctly fails on `master` (where it was silently returning an incorrect `OPTIMAL`) and passes here. I also made sure it verifies the `ValueError` message includes "indefinite" and "PSD".

Makes sense that this only crops up with `assume_PSD=True`, since the DCP check catches it otherwise. 

Let me know if it looks good!

cvxpy/reductions/dcp2cone/canonicalizers/quad_form_canon.py
```diff
@@ -30,9 +30,17 @@ def quad_form_canon(expr, args, solver_context: SolverInfo | None = None):
...
+            "(e.g., use P = (P + P.T) / 2 and clip negative eigenvalues)."
+        )
+
     if M1.size > 0:
         expr = sum_squares(Constant(M1.T) @ args[0])
-    if M2.size > 0:
+    else:
         scale = -scale
         expr = sum_squares(Constant(M2.T) @ args[0])
     obj, constr = quad_over_lin_canon(expr, expr.args)
```
PTNobel: Shouldn't this pass solver_context as well

cvxpy/tests/test_quad_form.py
```diff
@@ -216,3 +216,19 @@ def test_assume_psd(self) -> None:
         # Transform to a SolverError.
         with pytest.raises(cp.SolverError):
             prob.solve(solver=cp.OSQP)
+
+    def test_indefinite_assume_psd_raises(self) -> None:
+        """quad_form_canon must raise ValueError for indefinite P with assume_PSD=True.
+
+        On master, the M1 (positive-eigenvalue) term was silently dropped.
```
PTNobel: Let's say in CVXPY 1.8.1 and earlier instead of master since this will hopefully be part of master soon!

PTNobel (commented):
A couple minor comments.

One thing that's really confusing me is why aren't we doing an LDL decomposition to check for indefiniteness before we do a full eigenvalue decomposition? This is in the decompose_quad.

That would be both faster, and improve the sparsity pattern which would give higher performance. I think @dance858 once showed me data on this and using a triangular block for an SOC constraint can be a big improvement. 

No worries if you don't want to look into this, but it hit me as I was looking at this PR.

PTNobel:
Huh, seems the new test isn't passing in CI. Let me know if you need help investigating that.

PTNobel:
Forget everything I said about LDL and eigendecomposition. I'm going to deal with it myself, if you can fix the tests and address the 2 comments that would be great!

goyalpalak18:
@PTNobel I have pushed the changes . Please take a look at this !

PTNobel (approved):

SteveDiamond (approved):

PR #3156, opened 2026-02-22
Rewrite MOSEK interface to use ACC API (no Dualize/Slacks)

PTNobel:
## Description

Rewrite the MOSEK solver interface to use MOSEK's ACC (Affine Conic Constraints) API instead of the old Dualize/Slacks approach. This eliminates unnecessary problem transformations, simplifies the code (~150 fewer lines), and adds new capabilities.

**Key changes:**
- **Require MOSEK 10+** (ACC API support)
- **Direct primal formulation** — no more Dualize (continuous) or Slacks (MIP) reductions
- **PSD via `appendsvecpsdconedomain`** — no bar variables, uses √2-scaled svec format (same as SCS)
- **Native PowConeND support** — previously unsupported
- **MISDP support** — MOSEK 10 handles mixed-integer SDP via ACC
- **Efficient linear constraints** — zero/nonneg cones use MOSEK linear constraints; SOC/PSD/exp/pow use ACC
- **Dual extraction** — `gety` for linear constraints + `getaccdoty` for ACCs, with proper exp cone permutation and PSD svec→full expansion

**Architecture (before → after):**
- Before: Problem → format_constraints → Dualize/Slacks → build_task (old cone API) → solve → Dualize.invert/Slacks.invert
- After: Problem → `_prepare_data_and_inv_data` → `apply_parameters` → `_build_task` (ACC API) → solve → direct extraction

## Type of change
- [ ] Bug fix

**Breaking:** Requires MOSEK 10+. MOSEK 9 is no longer supported.

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
@@ -177,9 +177,6 @@ class MOSEK(ConicSolver):
         data[s.BOOL_IDX] = [int(t[0]) for t in problem.x.boolean_idx]
         data[s.INT_IDX] = [int(t[0]) for t in problem.x.integer_idx]
         data['psd_variable_info'] = problem.psd_variable_info
-        inv_data['num_psd_var_linking'] = sum(
-            rs for _, rs, _, _ in problem.psd_variable_info
-        )
         return data, inv_data
 
     def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts,
@@ -210,17 +207,25 @@ class MOSEK(ConicSolver):
         return {'task': task, 'solver_options': solver_opts}
 
     @staticmethod
-    def _add_afe_acc(task, A, b, row, dim, domain_idx):
-        """Add an ACC for rows A[row:row+dim, :] x + b[row:row+dim] in domain."""
-        afe_base = task.getnumafe()
-        task.appendafes(dim)
-        A_block = A[row:row + dim, :]
+    def _sparse_coo(A_block):
+        """Extract (row, col, val) COO entries from a dense or sparse matrix."""
         if sp.sparse.issparse(A_block):
             A_coo = A_block.tocoo()
-            rows_i, cols_j, vals = A_coo.row, A_coo.col, A_coo.data
+            return A_coo.row, A_coo.col, A_coo.data
         else:
-            rows_i, cols_j = np.nonzero(A_block)
-            vals = np.asarray(A_block[rows_i, cols_j]).ravel()
+            rows, cols = np.nonzero(A_block)
+            vals = np.asarray(A_block[rows, cols]).ravel()
+            return rows, cols, vals
+
+    @staticmethod
+    def _add_afe_acc(task, A, b, row, dim, domain_idx):
+        """Add an ACC for rows A[row:row+dim, :] x + b[row:row+dim] in domain.
+
+        Returns the AFE base index.
+        """
+        afe_base = task.getnumafe()
+        task.appendafes(dim)
+        rows_i, cols_j, vals = MOSEK._sparse_coo(A[row:row + dim, :])
         if len(vals) > 0:
             task.putafefentrylist(
                 (rows_i + afe_base).tolist(),
@@ -229,6 +234,7 @@ class MOSEK(ConicSolver):
             )
         task.putafegslice(afe_base, afe_base + dim, b[row:row + dim].tolist())
         task.appendacc(domain_idx, list(range(afe_base, afe_base + dim)), None)
+        return afe_base
 
     @staticmethod
     def _build_task(task, data):
@@ -236,6 +242,10 @@ class MOSEK(ConicSolver):
 
         Linear (zero/nonneg) constraints use efficient MOSEK linear constraints.
         All other cones (SOC, PSD, exp, pow3d, powND) use ACC.
+
+        PSD variables are handled without linking constraints: barvar coefficients
+        are injected directly into the objective, linear constraints, and AFE rows
+        via putbarcblocktriplet / putbarablocktriplet / putafebarfblocktriplet.
         """
         import mosek
 
@@ -247,6 +257,19 @@ class MOSEK(ConicSolver):
 
         # Variables
         task.appendvars(n)
+
+        # Create barvars for PSD variables early.
+        psd_var_info = data.get('psd_variable_info', [])
+        psd_var_barvars = []
+        for var_offset, reduced_size, psd_dim, is_nsd in psd_var_info:
+            barvar_idx = task.getnumbarvar()
+            task.appendbarvars([psd_dim])
+            tri_rows, tri_cols = np.tril_indices(psd_dim)
+            order = np.lexsort((tri_rows, tri_cols))
+            psd_var_barvars.append((
+                var_offset, reduced_size, psd_dim, is_nsd,
+                barvar_idx, tri_rows[order], tri_cols[order],
+            ))
 
         # Variable bounds
         if data[s.LOWER_BOUNDS] is not None and data[s.UPPER_BOUNDS] is not None:
@@ -266,9 +289,22 @@ class MOSEK(ConicSolver):
             task.putvarboundlist(np.arange(n, dtype=np.int32),
                                 [mosek.boundkey.fr] * n, o, o)
 
+        # Fix PSD variable scalar entries to zero.
+        # Their A-matrix coefficients contribute nothing; barvar coefficients
+        # (added later) provide the actual contribution.
+        for var_offset, reduced_size, *_ in psd_var_barvars:
+            idx = np.arange(var_offset, var_offset + reduced_size,
+                            dtype=np.int32)
+            task.putvarboundlist(idx, [mosek.boundkey.fx] * reduced_size,
+                                [0.0] * reduced_size, [0.0] * reduced_size)
+
         # Objective: min c'x
         task.putclist(np.arange(n, dtype=np.int32), c)
         task.putobjsense(mosek.objsense.minimize)
+
+        # Track A-row to MOSEK-index mappings for PSD variable post-processing.
+        linear_maps = []  # (A_row_start, A_row_end, mosek_con_offset)
+        afe_maps = []     # (A_row_start, A_row_end, mosek_afe_base)
 
         row = 0
 
@@ -278,13 +314,7 @@ class MOSEK(ConicSolver):
         if num_eq > 0:
             con_offset = task.getnumcon()
             task.appendcons(num_eq)
-            A_eq = A[row:row + num_eq, :]
-            if sp.sparse.issparse(A_eq):
-                A_coo = A_eq.tocoo()
-                eq_rows, eq_cols, eq_vals = A_coo.row, A_coo.col, A_coo.data
-            else:
-                eq_rows, eq_cols = np.nonzero(A_eq)
-                eq_vals = np.asarray(A_eq[eq_rows, eq_cols]).ravel()
+            eq_rows, eq_cols, eq_vals = MOSEK._sparse_coo(A[row:row + num_eq, :])
             if len(eq_vals) > 0:
                 task.putaijlist(
                     (eq_rows + con_offset).tolist(),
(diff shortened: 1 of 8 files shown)


PTNobel:
@aszekMosek can you look at this change? We're considering switching to ACC so that we have Ax + b in K instead of using our Dualize reduction to have the conic constraint direclty on x.

Riley thought this might have a performance hit, but I was under the impression this was preferred.

PTNobel:
@claude can you fix the CI so the Mosek tests run in test additional solvers

PTNobel:
@claude, that didn't work.

PTNobel:
@claude can you fix the bug in this PR? It seems that dual variable recovery is broken.

aszekMosek:
Hi @PTNobel , and thanks for working on it.

I will test it, but it may take some time. It's also been a while since the last time I understood cvxpy internals.

Just some initial comments:

* rewriting cones to ACC, and removing dualization are unrelated. I'm not sure why but there seems to be some confusion that the format is related to dualization. Whether the problem is input as ``Ax+b-slack = 0, slack in cone`` or ``Ax+b in conic_domain`` is the same thing, and whether cvxpy dualizes prior to that is a completely separate topic. 
* rewriting cones to ACC is a good idea because the old cones will one day be removed.
* whether skipping dualization is a good idea is hard to say. If the cvxpy team one day decided that it is better to dualize then just changing from cones to ACC itself does invalidate that decision from the point of view of efficiency in Mosek. On the other hand likely it is good on its own from the point of view of cvxpy code, and because Mosek gets what the user sees and not transposed.
* MISDP is never supported no matter how it is input. If it "worked" then we have a bug somewhere in Mosek by not rejecting it.
* Inputting all SDPs via the vectorized domain and not ``barvar`` is probably a uniform performance killer for everybody solving SDP due to https://docs.mosek.com/latest/faq/faq.html#why-is-my-sdp-lmi-slow-or-running-out-of-memory, while the current solution was good for sparse LMIs, but let me check.

PTNobel:
>     rewriting cones to ACC, and removing dualization are unrelated. I'm not sure why but there seems to be some confusion that the format is related to dualization. Whether the problem is input as Ax+b-slack = 0, slack in cone or Ax+b in conic_domain is the same thing, and whether cvxpy dualizes prior to that is a completely separate topic.

So my understanding was the old Mosek API was min. c^T x s.t. Ax = b, x in K and now we're doing min. c^T x s.t. Ax + b in K. The latter is the standard form every other solver uses, so we were hoping to bring the Mosek interface closer to the other solvers.

>    rewriting cones to ACC is a good idea because the old cones will one day be removed.

Awesome!

>    whether skipping dualization is a good idea is hard to say. If the cvxpy team one day decided that it is better to dualize then just changing from cones to ACC itself does invalidate that decision from the point of view of efficiency in Mosek. On the other hand likely it is good on its own from the point of view of cvxpy code, and because Mosek gets what the user sees and not transposed.

We were working on extending our built-in IIS support and passing user bounds directly to Mosek, and I think the dualization was making both impossible.

>    MISDP is never supported no matter how it is input. If it "worked" then we have a bug somewhere in Mosek by not rejecting it.

Let me fix that.

>    Inputting all SDPs via the vectorized domain and not barvar is probably a uniform performance killer for everybody solving SDP due to https://docs.mosek.com/latest/faq/faq.html#why-is-my-sdp-lmi-slow-or-running-out-of-memory, while the current solution was good for sparse LMIs, but let me check.

I will look into this. Thank you!

SteveDiamond:
@parthb83 you should run this interface and the current interface against all the CBLIB problems https://github.com/cvxpy/solver-benchmarks/issues/1

rileyjmurray:
I ran a test problem in my quantum information work that had a (real-valued) LMI of order 144. Here's a glimpse at the results. It shows a 15x performance hit.

CVXPY 1.7.5 (latest on conda-forge) has the dualization codepath. It takes 1 second to solve on my M2 Max machine.
```
(CVXPY) Feb 23 11:08:40 PM: Optimizer  - Constraints            : 2664            
(CVXPY) Feb 23 11:08:40 PM: Optimizer  - Cones                  : 1               
(CVXPY) Feb 23 11:08:40 PM: Optimizer  - Scalar variables       : 3                 conic                  : 3               
(CVXPY) Feb 23 11:08:40 PM: Optimizer  - Semi-definite variables: 3                 scalarized             : 10596   
...
(CVXPY) Feb 23 11:08:41 PM: 11  1.1e-08  3.8e-09  3.7e-15  1.00e+00   -1.251204093e-01  -1.251204092e-01  8.2e-10  0.94  
(CVXPY) Feb 23 11:08:41 PM: Optimizer terminated. Time: 0.94  
```

This branch does not have the dualization codepath. It takes 15 seconds to solve on my M2 Max machine.
```
(CVXPY) Feb 23 11:15:59 PM: Optimizer  - Constraints            : 10598           
(CVXPY) Feb 23 11:15:59 PM: Optimizer  - Cones                  : 1               
(CVXPY) Feb 23 11:15:59 PM: Optimizer  - Scalar variables       : 2665              conic                  : 2665            
(CVXPY) Feb 23 11:15:59 PM: Optimizer  - Semi-definite variables: 3                 scalarized             : 10596           
...
(CVXPY) Feb 23 11:16:10 PM: 10  3.9e-09  3.3e-09  2.4e-14  9.95e-01   -1.251203974e-01  -1.251203980e-01  3.3e-09  15.00 
(CVXPY) Feb 23 11:16:10 PM: Optimizer terminated. Time: 15.01   
```

PTNobel:
I'm still not a huge fan of Mosek having a different codepath than everything else; I wonder if we could lift dualization to not just be Mosek specific.

Also, how come the constraints and variable count went up in this setting?

cvxpy/reductions/solvers/conic_solvers/mosek_conif.py
```diff
@@ -313,339 +205,370 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
+            tri_rows = tri_rows[order]
+            tri_cols = tri_cols[order]
+
+            # Create PSD matrix variable
+            barvar_idx = task.getnumbarvar()
+            task.appendbarvars([psd_dim])
+
+            # Add vec_len equality constraints: A_psd[k,:] @ x - <M_k, X> = -b[k]
+            con_offset = task.getnumcon()
+            task.appendcons(vec_len)
```
rileyjmurray: @PTNobel you're adding ``psd_dim * (psd_dim + 1) // 2`` equality constraints here. These are inevitable in the slack variables approach, and they're the price of not dualizing.

rileyjmurray: We could reduce the advantage of the dualize reduction if we tracked the PSD=True Variable attribute like we track bound constraints now. In fact, if we made that change then there would be situations where dualizing makes things worse!

We'd need to intelligently choose between dualizing or not. My suggestion is that we should dualize if the problem has an LMI (i.e., not as PSD=True constraint, but an algebraic LMI expression) that's bigger than all PSD cones stemming from PSD=True Variable attributes.

No other solver has this problem because no other solver has a method of specifying PSD=True as though it were a bound constraint.

PTNobel: Ahhh okay

PTNobel:
@rileyjmurray can you run that runtime experiment again? I reworked the support so that:

1. It adds no constraints for Variable(PSD=True)
2. It does add both primal and bar variables, but the primal variables are bound to always be 0, and I'm assuming the MOSEK presolver will clean that up without any trouble...

rileyjmurray:
@PTNobel 

> @ rileyjmurray can you run that runtime experiment again? I reworked the support so that:
> 
> 1. It adds no constraints for Variable(PSD=True)
> 2. It does add both primal and bar variables, but the primal variables are bound to always be 0, and I'm assuming the MOSEK presolver will clean that up without any trouble...

I got basically the same result. Presolve log shown for good measure. (I don't understand what you mean by "the primal variables are bound to always be 0.")
```
...
(CVXPY) Feb 24 08:52:17 PM: Presolve started.
(CVXPY) Feb 24 08:52:17 PM: Linear dependency checker started.
(CVXPY) Feb 24 08:52:17 PM: Linear dependency checker terminated.
(CVXPY) Feb 24 08:52:17 PM: Eliminator started.
(CVXPY) Feb 24 08:52:17 PM: Freed constraints in eliminator : 0
(CVXPY) Feb 24 08:52:17 PM: Eliminator terminated.
(CVXPY) Feb 24 08:52:17 PM: Eliminator started.
(CVXPY) Feb 24 08:52:17 PM: Freed constraints in eliminator : 0
(CVXPY) Feb 24 08:52:17 PM: Eliminator terminated.
(CVXPY) Feb 24 08:52:17 PM: Eliminator - tries                  : 2                 time                   : 0.00            
(CVXPY) Feb 24 08:52:17 PM: Lin. dep.  - tries                  : 1                 time                   : 0.00            
(CVXPY) Feb 24 08:52:17 PM: Lin. dep.  - primal attempts        : 1                 successes              : 1               
(CVXPY) Feb 24 08:52:17 PM: Lin. dep.  - dual attempts          : 0                 successes              : 0               
(CVXPY) Feb 24 08:52:17 PM: Lin. dep.  - primal deps.           : 0                 dual deps.             : 0               
(CVXPY) Feb 24 08:52:17 PM: Presolve terminated. Time: 0.00    
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - threads                : 12              
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - solved problem         : the primal      
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - Constraints            : 10598           
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - Cones                  : 1               
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - Scalar variables       : 2665              conic                  : 2665            
(CVXPY) Feb 24 08:52:21 PM: Optimizer  - Semi-definite variables: 3                 scalarized             : 10596           
...
(CVXPY) Feb 24 08:52:32 PM: 10  8.1e-09  7.9e-09  1.3e-13  9.87e-01   -1.251203726e-01  -1.251203738e-01  7.9e-09  14.91 
(CVXPY) Feb 24 08:52:32 PM: Optimizer terminated. Time: 14.91   
```
This does not surprise me. The largest PSD cone in this problem comes from a proper LMI, not from a PSD=True Variable attribute.

I'll try some examples where PSD=True Variable attribute causes the largest PSD cone.

rileyjmurray:
@PTNobel the proposed changes also cause a major increase in complexity for test_quantum_rel_entr.py. That file defines three test problems.
 * The first problem is easy to solve for MOSEK and Clarabel on both branches (master and proposed changes here).
 * The second and third problems are too expensive to Clarabel.
     * If we use MOSEK interface on master then my M2 Max machine solves them in 3 seconds each, peak RAM of 2.7GB.
     * If we use the proposed MOSEK interface ...
        *  test 2 needs 15GB RAM, 47 seconds to reach the first iteration, total solve time of 142 seconds. This is a 47x slowdown.
        * test 3 needs 13.7 GB RAM, 71 seconds to reach the first iteration, total solve time of 252 seconds. This is an 84x slowdown.

Still working finding you a problem where the new approach (not dualizing) is useful.

rileyjmurray:
Here's a toy script that defines a correlation matrix completion problem: 
[correlation-completion-example.py](https://github.com/user-attachments/files/25538686/correlation-completion-example.py)

Docstrings of the "log_proposed_n120_r20" and "log_current_n120_r20" functions show output when running on my laptop with a matrix of order 120. "Proposed" means this PR and "current" means master. There's a 6.7x slowdown for proposed (52.50 seconds, all solver) relative to current (7.844 seconds, 2.556 of which spent dualizing).

This problem has a structure that should make dualization not-strictly-necessary. The only PSD constraint is from a Variable attribute.

Note: the example is adapted from LLM output. It wrote the constraint in a somewhat weird way: ``[cp.multiply(mask.astype(float), X - S) == 0]``. It's simpler to write it as `` [X[mask] == S[mask]]``.

PTNobel:
I think #3275 is a more reasonable path forward.

Transurgeon:
let's close this PR and try to make progress with #3275 then.

PR #3159, opened 2026-02-23
Fix backward() setting param.gradient to float 0.0 instead of zero nd…

goyalpalak18:
## Summary
I ran into a bug in `Problem.backward()` where parameter gradients are being initialized as a bare Python float (`0.0`) instead of a properly shaped numpy array (`np.zeros(param.shape)`). 

If a parameter gets transformed by a reduction (like a log-transform in `Dgp2Dcp`), its raw ID might not show up in `dparams`. When the chain rule fallback returns `None` in these cases, the unhandled parameter's gradient silently defaults to that initial `0.0`.

Since downstream code expects `param.gradient` to be an `ndarray`, this `float` initialization causes a couple of frustrating problems:
* **Unexpected Crashes:** Anything trying to call `.shape`, `.reshape()`, or `np.linalg.norm()` on the gradient immediately throws an `AttributeError`.
* **Silent Failures:** Gradient update operations like `param.value -= lr * 0.0` silently execute but broadcast incorrectly on matrix parameters, resulting in a no-op step with completely wrong semantics.

## The Fix
I updated the initialization in `cvxpy/problems/problem.py` (around line 1227):

**From:** `grad = 0.0`
**To:** `grad = np.zeros(param.shape)`

This single line ensures that `param.gradient` is always a correctly shaped `ndarray` after `.backward()` finishes, even if no reduction explicitly handles the parameter. 

## Impact
* Eliminates the silent scalar type corruption in ML+convex hybrid pipelines (like differentiable DGP).
* Stops the unexpected `AttributeError` crashes in post-backward code.
* Completely safe with zero risk of regression, since `np.zeros(param.shape) + ndarray` behaves numerically identical to `0.0 + ndarray`.


goyalpalak18:
Hi @PTNobel , could you please have a look at this when you have a moment? Thanks!

cvxpy/problems/problem.py
```diff
@@ -1224,7 +1224,7 @@ def backward(self) -> None:
         # reductions that transform parameters (e.g., DGP log transform,
         # Complex2Real split into real/imag).
         for param in self.parameters():
-            grad = 0.0
+            grad = np.zeros(param.shape)
```
PTNobel: Should we use broadcast_to to save the allocation?

PTNobel (approved):
Looks pretty good, one mild comment

cvxpy/problems/problem.py
goyalpalak18: Thank you for the review and suggestion. Using np.broadcast_to makes perfect sense to avoid the allocation overhead. I have pushed the fix.

PR #3162, opened 2026-02-23
Fix backward() assigning read-only broadcast_to view to param.gradient

goyalpalak18:
### Description

I ran into an issue where `Problem.backward()` is assigning a read-only numpy view to `param.gradient`, which ends up crashing in-place gradient operations downstream.

Currently in `cvxpy/problems/problem.py:1227`, the gradient is initialized using `np.broadcast_to(0.0, param.shape)`. This returns an array with `flags.writeable == False`. When a parameter isn't present in `dparams`, this read-only view gets assigned directly to `param.gradient`. 

While standard updates like `param.value -= lr * param.gradient` create a new array and mask the bug, any in-place processing—like gradient clipping, normalization, or momentum accumulation—fails entirely with `ValueError: output array is read-only`. 

It looks like this was introduced in `c056ce863` as a memory micro-optimization. However, it breaks the contract that `param.gradient` should be a writable `ndarray`. The `derivative()` method actually still uses `np.zeros` for its zero-initialization step, so this brings `backward()` back into alignment with that.

### The Fix

I reverted `np.broadcast_to(0.0, param.shape)` back to `np.zeros(param.shape)` in `Problem.backward()`. The memory savings from `broadcast_to` aren't worth breaking the API contract and causing non-deterministic crashes in differentiable ML pipelines.


goyalpalak18:
Hi @PTNobel , please have a look at this when you get a chance. Thanks!

SteveDiamond (approved):
Thanks for catching this!

PR #3164, opened 2026-02-24
Fix Dgp2Dcp.param_backward KeyError when log-param not in dparams

goyalpalak18:
## Description

I ran into a `KeyError` crash in `problem.backward()` when a DGP parameter ends up being constant-folded. This PR adds the missing dictionary guard to `Dgp2Dcp.param_backward` to fix it.

### The Problem

When `dparams` is built by `apply_param_jac`, it only includes parameters whose IDs actually appear in `param_id_to_col`. If a DGP log-parameter gets folded into the constant offset during canonicalization, it never makes it into the parametric tensor. As a result, its ID is completely absent from `dparams`.

Right now, `Dgp2Dcp.param_backward` tries to access `dparams[new_param.id]` blindly. This naked access throws a `KeyError` and crashes the backward pass with no real error message. 

As a side note, this can also cause silent failures: if `param.value` is near zero, the unguarded `1.0 / param.value` division can produce `inf` and silently poison the gradient accumulation with NaNs. Every other reduction pass (like `CvxAttr2Constr` and `Complex2Real`) safely guards their parameter dictionary accesses, but `Dgp2Dcp` was missing it.

### The Solution

I added a single standard guard line to `Dgp2Dcp.param_backward`:
`if new_param.id not in dparams: return None`

This brings it in line with how the other reduction passes handle it. Now, if the parameter's ID is missing from `dparams`, we safely return `None`. The gradient accumulator in `problem.backward()` correctly stays at `np.zeros(param.shape)`, reflecting the true zero gradient, and we completely avoid the silent NaN path.


goyalpalak18:
Hi @PTNobel @SteveDiamond , could you please have a look at this when you have a moment? Thanks!

PTNobel:
#3147 also fixes this bug; I am fine with merging this for now since it is a small change, but I want to flag the redesign.

goyalpalak18:
Got it, thanks @PTNobel for pointing out #3147. Makes sense to get this in as a small patch for now.

SteveDiamond:
@goyalpalak18 please add a test that passes with this fix but fails without it.

goyalpalak18:
@SteveDiamond , I just added the test in cvxpy/tests/test_derivative.py. I set it up to call dgp.param_backward(p, {}) with an empty dparams dict right after apply(). Without the guard, it throws the KeyError; with the fix, it safely returns None. I also put it in a class that doesn't depend on diffcp so it runs unconditionally.

SteveDiamond (approved):
Thanks!

PR #3167, opened 2026-02-26
Add hypothesis fuzz tests for reshape N-D inference and round-trip

goyalpalak18:
## Description

Following up on the discussion on my previous PR (#3165), I put together a property-based fuzzing suite for `cp.reshape` using the `hypothesis` library. 

I've added these fuzzing tests in a new dedicated file. To keep CI runtimes bounded and avoid timing out the test runner, I set `deadline=None` and capped `max_examples` at 150–200. I also intentionally scoped these test cases to dimensions and properties not already covered by `test_reshape_negative_one` in `test_atoms.py`, specifically targeting N=4, 5, and 6 since the existing regression test only covers N=3.

## Summary of Changes

I created `cvxpy/tests/test_fuzz_reshape.py` to test dynamically generated N-dimensional shapes across both `order='C'` and `order='F'`. I focused on verifying four core properties:

* **N-D Inference Parity:** I ensure that `cp.reshape` inference perfectly matches `numpy.reshape` for any valid N-D shape containing exactly one `-1`.
* **Round-Trip Consistency:** I check that consecutive reshapes recover the original array without data corruption, silent reordering, or transpose errors.
* **DCP Metadata Preservation:** I assert that `is_affine()`, `is_nonneg()`, and `is_nonpos()` attributes properly survive the reshape node for arbitrarily generated dimensions.
* **Guard Rail Consistency:** I confirm that any shape containing multiple `-1` dimensions consistently raises the correct `AssertionError`, guarding against future refactors inadvertently softening this hard constraint.


goyalpalak18:
Hey @SteveDiamond @PTNobel , here are those hypothesis tests we talked about in #3165. Let me know what you think!

PTNobel (approved):

SteveDiamond:
Nice! Please add @settings(derandomize=True) so the tests are deterministic.

goyalpalak18:
Done, added derandomize=True to all four @settings calls.

SteveDiamond (approved):

PR #3169, opened 2026-02-27
Fix DGP pnorm scalar canonicalization promoting p instead of x

goyalpalak18:
## Description
I came across a pretty critical silent failure in `dgp2dcp/pnorm_canon` where scalar DGP `pnorm` objectives and constraints were evaluating incorrectly. 

Basically, the canonicalizer was accidentally promoting the numeric exponent `p` instead of the log-domain variable `x`. Because of this, the objective function became a constant with respect to the optimization variable. The solver was receiving a problem with absolutely no variable dependence and just returning an arbitrary feasible point instead of the true optimum. The worst part is that it would silently report an "optimal" status, giving the illusion that everything worked perfectly.

## Root Cause
While digging into `cvxpy/reductions/dgp2dcp/canonicalizers/pnorm_canon.py`, I noticed the issue happens on line 12. Inside the `if x.shape == tuple():` guard, the code was mistakenly executing `promote(p, (1,))`. 

This cast the Python float exponent (e.g., `2.0`) to a constant vector, which completely discarded the original variable expression. The intent was clearly to reshape the scalar expression into a 1-element vector so the downstream `hstack` works, but the wrong object was being promoted. 

## Changes Made
* **Fixed the typo:** It's literally a one-character fix! I changed `x = promote(p, (1,))` to `x = promote(x, (1,))` so the actual log-domain expression is the thing being promoted to a 1-element vector.
* **Added tests:** I added `test_pnorm_scalar` to `cvxpy/tests/test_dgp.py` as a regression test to ensure this stays fixed, covering scalar p-norms for $p=2$ and $p=3$.

## Impact
* Scalar DGP `pnorm` objectives and constraints will now actually compute the correct values.
* DGP problems with scalar variables will return the true optimum rather than just a random feasible point.
* Most importantly, this eliminates a really dangerous silent failure mode.
* Since this modified logic is strictly contained within the scalar check branch, there are absolutely no behavioral changes or risks to non-scalar inputs.


goyalpalak18:
Hi @PTNobel @SteveDiamond , could you please have a look at this when you have a moment? Thanks!

SteveDiamond (approved):
Thanks for fixing this! I checked and the bug has been there for 7 years :(

goyalpalak18:
Thanks @SteveDiamond ! Glad we could finally get it fixed.

cvxpy/tests/test_dgp.py
```diff
@@ -298,3 +298,19 @@ def test_sparse_variable_not_dgp(self) -> None:
...
             cvxpy.Variable(3, diag=True, neg=True)
+
+    def test_pnorm_scalar(self) -> None:
+        """Regression test: scalar DGP pnorm must canonicalize x, not the exponent p."""
+        # pnorm of a single positive scalar equals that scalar,
+        # so the minimum subject to x >= lb must be lb.
+        x = cvxpy.Variable(pos=True)
+        prob = cvxpy.Problem(cvxpy.Minimize(cvxpy.pnorm(x, p=2)), [x >= 2.5])
+        prob.solve(gp=True)
+        self.assertEqual(prob.status, "optimal")
```
SteveDiamond: you should use cvxpy.OPTIMAL here not the string

goyalpalak18: Done. Thanks for catching this!

PR #3174, opened 2026-02-28
Fix decomp_quad sparse NSD using column permutation instead of row pe…

goyalpalak18:
## Description
I ran into a nasty silent bug where `cp.quad_form(x, P)` produces incorrect optimization results when `P` is a sparse Negative Semidefinite (NSD) matrix. I tracked the issue down to `decomp_quad` applying the wrong permutation array. 

## The Bug
When decomposing a sparse NSD matrix, the code relies on `sparse_cholesky`. The documented mathematical contract for this function guarantees that `L[p, :] @ L[p, :].T == |A|`. 

However, the existing code was returning `L[:, p]` (a column permutation) instead of `L[p, :]` (a row permutation). Because the AMD ordering is almost never an identity permutation for sparse matrices, this resulted in a completely wrong factorization. The worst part is that it doesn't throw any errors—it just silently computes the wrong objective value. For instance, in a 6x6 tridiagonal test case I ran, the objective was off by about 1.54.

## The Fix
In `cvxpy/atoms/quad_form.py`, I simply swapped out the column permutation for the correct row permutation in the sparse NSD branch:
- **Before:** `return 1.0, np.empty((0, 0)), L[:, p]`
- **After:** `return 1.0, np.empty((0, 0)), L[p, :]`

## Testing
I added a new regression test (`test_sparse_nsd` in `test_quad_form.py::TestDecompQuad`) to make sure this doesn't happen again. 
* It uses a sparse negative discrete Laplacian (a tridiagonal NSD matrix) to intentionally trigger a non-identity AMD ordering (specifically `[4,5,3,2,1,0]`).
* It confirms the factorization identity `P == scale * (M1@M1.T - M2@M2.T)`.
* With this patch, the row permutation yields an error of `~4.4e-16` (correct), whereas the old buggy column permutation yields an error of `1.0`.


goyalpalak18:

PTNobel (approved):
Thanks so much for catching this!

PR #3183, opened 2026-03-01
Add Huber Scale Parameter

WilliamZhang20:
## Description
Issue link (if applicable): #1377 

Adds a scale parameter to the Huber atom. 

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/cone2cone/affine2direct.py
+++ b/cvxpy/reductions/cone2cone/affine2direct.py
@@ -21,7 +21,7 @@ from cvxpy import settings as s
 from cvxpy.constraints.exponential import ExpCone as ExpCone_obj
 from cvxpy.constraints.nonpos import NonNeg as NonNeg_obj
 from cvxpy.constraints.power import PowCone3D as PowCone_obj
-from cvxpy.constraints.psd import PSD as PSD_obj
+from cvxpy.constraints.psd import SvecPSD as SvecPSD_obj
 from cvxpy.constraints.second_order import SOC as SOC_obj
 from cvxpy.constraints.zero import Zero as Zero_obj
 from cvxpy.reductions.solution import Solution
@@ -154,9 +154,12 @@ class Dualize:
         to the second-order-cone under the CVXPY standard ({ z : z[0] >= || z[1:] || }).
         We map these variables back to dual variables for SOC constraints in (P-Opt).
 
-        solution.primal_vars[PSD] is a list of symmetric positive semidefinite matrices
-        which result by lifting the vectorized PSD blocks of "y" back into matrix form.
-        We assign these as dual variables to PSD constraints appearing in (P-Opt).
+        solution.primal_vars[PSD] is a list of triangular (svec) vectors: the
+        column-major lower-triangular entries of the PSD blocks of "y", one vector
+        per cone. PSD constraints reach a dualizing solver in this triangular form
+        via the PSDToSvecPSD reduction (see #3268), so we concatenate the cones
+        belonging to each SvecPSD constraint and assign the result as that
+        constraint's dual variable in (P-Opt).
 
         solution.primal_vars[DUAL_EXP] is a vector of concatenated length-3 slices of y, where
         each constituent length-3 slice belongs to dual exponential cone as implied by the CVXPY
@@ -190,9 +193,17 @@ class Dualize:
                 dv = np.concatenate(direct_prims[SOC][i:i + block_len])
                 dual_vars[con.id] = dv
                 i += block_len
-            for i, con in enumerate(constr_map[PSD_obj]):
-                dv = direct_prims[PSD][i]
-                dual_vars[con.id] = dv
+            # direct_prims[PSD] holds one triangular (svec) vector per cone,
+            # ordered to match ConeDims.psd; a batched SvecPSD constraint spans
+            # con.num_cones() consecutive cones. Concatenating those cones gives
+            # the constraint's dual in the svec form PSDToSvecPSD.recover_dual
+            # expects.
+            psd_idx = 0
+            for con in constr_map[SvecPSD_obj]:
+                k = con.num_cones()
+                blocks = direct_prims[PSD][psd_idx:psd_idx + k]
+                dual_vars[con.id] = np.concatenate(blocks)
+                psd_idx += k
             i = 0
             for con in constr_map[ExpCone_obj]:
                 dv = direct_prims[DUAL_EXP][i:i + con.size]
--- a/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
+++ b/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
@@ -469,11 +469,13 @@ class ConeMatrixStuffing(MatrixStuffing):
                 shape = con_obj.shape
                 dual_value = solution.dual_vars.get(new_con)
                 # TODO rationalize Exponential.
-                if dual_value is not None:
-                    if shape == () or isinstance(con_obj, (ExpCone, SOC)):
-                        dual_vars[old_con] = dual_value
-                    else:
-                        dual_vars[old_con] = np.reshape(dual_value, shape, order="F")
+                if dual_value is None:
+                    continue
+                if shape == () or isinstance(con_obj, (ExpCone, SOC)) or \
+                        (isinstance(con_obj, PSD) and con_obj.num_cones() > 1):
+                    dual_vars[old_con] = dual_value
+                else:
+                    dual_vars[old_con] = np.reshape(dual_value, shape, order="F")
 
         primal_vars = {}
         if solution.status not in s.SOLUTION_PRESENT:
@@ -485,22 +487,6 @@ class ConeMatrixStuffing(MatrixStuffing):
             shape = inverse_data.var_shapes[var_id]
             size = np.prod(shape, dtype=int)
             primal_vars[var_id] = np.reshape(x_opt[offset: offset + size], shape, order="F")
-
-        # Remap dual variables if dual exists (problem is convex).
-        if solution.dual_vars is not None:
-            for old_con, new_con in con_map.items():
-                con_obj = inverse_data.id2cons[old_con]
-                shape = con_obj.shape
-                # TODO rationalize Exponential.
-                if shape == () or isinstance(con_obj, (ExpCone, SOC)) or \
-                        (isinstance(con_obj, PSD) and con_obj.num_cones() > 1):
-                    dual_vars[old_con] = solution.dual_vars[new_con]
-                else:
-                    dual_vars[old_con] = np.reshape(
-                        solution.dual_vars[new_con],
-                        shape,
-                        order='F'
-                    )
 
         return Solution(solution.status, opt_val, primal_vars, dual_vars,
                         solution.attr)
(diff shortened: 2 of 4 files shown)


PTNobel:
Thanks for working on this! So one of the goals is that t can be a variable and not just a constant. I don't think you added support for that?

cvxpy/atoms/elementwise/huber.py
```diff
@@ -19,13 +19,47 @@
...
+        Positive scalar halfwidth. Defaults to 1.
+    t : Expression, Variable, Constant, or None
+        Optional positive scale parameter. When provided as a Variable,
+        enables concomitant scale estimation. Must be concave (or affine).
+        When None or omitted, the standard two-argument huber is returned.
+    """
+    if t is None:
+        return _HuberAtom(x, M)
+    else:
+        return _HuberPerspectiveAtom(x, t, M)
```
Transurgeon: I think its fine to not have underscores for the names of the atoms.
Last time we had something similar was for Trace and we called it TraceAtom.
I feel like underscores should be reserved for private functions.

cvxpy/atoms/elementwise/huber.py
```diff
@@ -112,4 +149,150 @@ def _grad(self, values):
...
+        else:
+            # non-increasing in t
+            return True
+
+    def is_quadratic(self) -> bool:
+        """Not quadratic in general when t is a variable."""
+        return False
+
+    def has_quadratic_term(self) -> bool:
+        return False
```
Transurgeon: I think these should be true under some conditions.. like when t is constant and x affine?
Not sure about this, but there must be some cases where we can take the quadratic path.

cvxpy/reductions/dcp2cone/canonicalizers/quad/__init__.py
```diff
@@ -29,6 +29,6 @@
     quad_over_lin: quad_over_lin_canon,
     Power: power_canon,
     PowerApprox: power_canon,
-    huber: huber_canon,
+    _HuberAtom: huber_canon,
```
Transurgeon: @PTNobel should we have a HuberPerspective somewhere here as well?
It is not necessarily quadratic but can be when ``t`` is constant.. so not sure how we should approach this.

doc/source/functions/functions_table.rst
```diff
@@ -408,15 +408,19 @@
...
+   * - :ref:`huber(x, M=1, t=None) <huber>`
+   
+       :math:`M \geq 0,\ t > 0`
+     - :math:`\begin{aligned}
+         & t = \text{None}: \\
+         &\quad \text{if } |x| \leq M\colon \\&\quad x^2 \\
+         &\quad \text{if } |x| > M\colon \\&\quad 2M|x| - M^2 \\
+         & \\
+         & t > 0\colon \\&\quad t\,\mathrm{huber}(x/t, M)
+       \end{aligned}`
```
Transurgeon: let's maybe make two entries for both huber and huber perspective?
you can just keep the previous one so that we have both in the table.

cvxpy/reductions/dcp2cone/canonicalizers/huber_canon.py
```diff
@@ -42,3 +44,48 @@ def huber_canon(expr, args, solver_context: SolverInfo | None = None):
...
+    t = args[1]    # canonicalized t, scalar, concave/affine
+    shape = x.shape
+
+    # Reparametrized auxiliaries: ñ = t*n,  s̃ = t*s
+    n_tilde = Variable(shape, name="huber_persp_n")
+    s_tilde = Variable(shape, name="huber_persp_s")
+
+    # Canonicalize quad_over_lin(n_tilde, t) inline to SOC constraints
+    qol_expr = quad_over_lin(n_tilde, t)
+    qol_obj, constr_qol = quad_over_lin_canon(qol_expr, qol_expr.args)
```
Transurgeon: I think there could be a potential bug here.
quad_over_lin returns a scalar since it sums over all inputs.. so when x is a vector it will sum all squared entries of n_tilde. 
I think we want something like ``x**2/t`` elementwise.
You can look into passing an axis argument to quad_over_lin to see if that would fix this issue.

Transurgeon: a simple test that claude suggested is the following:
```
cp.huber(np.array([1.0, 2.0]), M=1, t=2.0).value
```
it should return a vector of length 2 rather than a scalar.
(note I haven't tried this.. just speculating that it might not work).

cvxpy/tests/test_atoms.py
```diff
@@ -1056,6 +1062,117 @@ def test_huber(self) -> None:
...
+        
+        M_param.value = 1.0
+        t_param.value = 1.0
+        result = problem_t.solve(solver=cp.CLARABEL)
+        x_val1 = x.value
+        
+        # Change t and resolve
+        t_param.value = 2.0
+        result = problem_t.solve(solver=cp.CLARABEL)
+        x_val2 = x.value
```
Transurgeon: I am not sure we handle this properly in the canonicalization.. we should also test things like dpp scope and expression.is_dpp().
I feel like having t as a parameter is quite complicated.. so maybe we can just not support that for now?

cvxpy/tests/test_atoms.py
```diff
@@ -1056,6 +1062,117 @@ def test_huber(self) -> None:
...
+        
+        # With larger t, the huber term is more heavily weighted
+        # so x should be closer to minimizing huber term (x closer to 1.5)
+        self.assertIsNotNone(x_val1)
+        self.assertIsNotNone(x_val2)
```
Transurgeon: I personally prefer to use pytest semantics.
so this would be:
```
assert x_val1 is not None
```

cvxpy/tests/test_atoms.py
```diff
@@ -1056,6 +1062,117 @@ def test_huber(self) -> None:
...
+        )
+        prob.solve(solver=cp.CLARABEL)
+        self.assertEqual(prob.status, cp.OPTIMAL)
+        np.testing.assert_allclose(x.value, target, atol=0.5)
+        self.assertGreater(t.value, 0)
+
+        # --- DCP check: huber(affine x, M, t=Variable) is convex ---
+        x = cp.Variable(3)
+        t = cp.Variable()
+        self.assertTrue(cp.huber(x, M=1, t=t).is_convex())
```
Transurgeon: this is a bit all over the place.. maybe we can write a separate test for this?
and we could include the dpp stuff there if you wanna include it.

cvxpy/tests/test_atoms.py
```diff
@@ -1056,6 +1062,117 @@ def test_huber(self) -> None:
...
+        self.assertAlmostEqual(cp.huber(0.5, M=1, t=2).value, 0.125)  # 0.5^2/2
+        self.assertAlmostEqual(cp.huber(0.5, M=1, t=3).value, 1.0/12)  # 0.5^2/3
+        
+        # When |x/t| > M: result = t * (2*M*|x/t| - M^2)
+        self.assertAlmostEqual(cp.huber(2, M=1, t=2).value, 2.0)  # 2*(2*1*1 - 1)
+        self.assertAlmostEqual(cp.huber(2, M=1, t=3).value, 4.0/3)  # 3*(2*1*(2/3) - 1)
+        
+        # Test with negative x (should give same result due to abs)
+        self.assertAlmostEqual(cp.huber(-0.5, M=1, t=2).value, 0.125)
+        self.assertAlmostEqual(cp.huber(-2, M=1, t=2).value, 2.0)
```
Transurgeon: let's test more than just scalar inputs for ``x``.
Also it would be really cool if we could find a small and realistic example of the concomitant estimation or whatever to show the usage of this huber perspective atom (but maybe that's for another PR).

cvxpy/atoms/elementwise/huber.py
```diff
@@ -112,4 +149,150 @@ def _grad(self, values):
...
+
+    Parameters
+    ----------
+    x : Expression
+        Affine expression (the function is nonmonotonic in x).
+    M : int, float, Constant, or Parameter
+        Non-negative scalar halfwidth constant.
+    t : Expression
+        Concave (or affine) positive scalar expression. May be a Variable
+        to enable concomitant scale estimation.
```
Transurgeon: do we really allow ``t`` to be a general expression? It seems quite complicated.. and I am not sure this is handled in the canonicalization naturally.
Maybe we should just stick to ``Variable`` or ``Constant``.. what do you think?

Transurgeon (commented):
really nice work @WilliamZhang20 I left quite a few review comments.
There is some work left but really nice job so far.. also @PTNobel please double check my review and let me know if I said anything incorrect.

cvxpy/atoms/elementwise/huber.py
PTNobel: Well we can assume t is nonconstant, so I don't think this is the case?

cvxpy/reductions/dcp2cone/canonicalizers/quad/__init__.py
PTNobel: I think we should assume t nonconstant since if t is constant we should have a Huber atom

cvxpy/tests/test_atoms.py
PTNobel: I don't think t as a parameter is any harder?

cvxpy/atoms/elementwise/huber.py
PTNobel: I think we should definitely support arbitrary expression; the canonicalization process should handle it fine

cvxpy/atoms/elementwise/huber.py
```diff
@@ -19,13 +19,47 @@
...
+        Positive scalar halfwidth. Defaults to 1.
+    t : Expression, Variable, Constant, or None
+        Optional positive scale parameter. When provided as a Variable,
+        enables concomitant scale estimation. Must be concave (or affine).
+        When None or omitted, the standard two-argument huber is returned.
+    """
+    if t is None:
+        return HuberAtom(x, M)
+    else:
+        return HuberPerspectiveAtom(x, t, M)
```
Transurgeon: @WilliamZhang20 maybe we should change this to the following (based on Parth's comment that when ``t`` is constant we should just call huber):
```
    if t is None:
        return HuberAtom(x, M)
    else:
        if t.is_constant():
           return t * HuberAtom(x, M)
        else:
        return HuberPerspectiveAtom(x, t, M)
```

Transurgeon: then we are able to not have HuberPerspectiveAtom under quad.. and we can also set its quadratic attribute to false.

doc/source/functions/functions_table.rst
WilliamZhang20: @Transurgeon Do you want the new entry to be called HuberPersepectiveAtom and the old entry still called Huber instead of HuberAtom?

Transurgeon:
@WilliamZhang20 the changes in this PR also fail some tests for DNLP.
```
=========================== short test summary info ============================
FAILED cvxpy/tests/nlp_tests/test_huber_sum_largest.py::TestNonsmoothNontrivial::test_huber - NotImplementedError: Atom 'HuberAtom' not supported
============ 1 failed, 174 passed, 74 skipped, 1 warning in 13.19s =============
```
could you look into that? I think it just comes down to changing the canon mapping inside ``dnlp2smooth/init.py``.

WilliamZhang20:
Will reopen another one...

PR #3184, opened 2026-03-01
Fix MulExpression missing is_hermitian() for Hermitian self-product

goyalpalak18:
## Description

I ran into a critical issue where minimizing the squared Frobenius norm of a Hermitian matrix—`Minimize(trace(X @ X))`—crashes with a `ValueError: The 'minimize' objective must be real valued.` 

This looks like a direct follow-on to the recent fix in commit `69b1ecdff` for the elementwise multiply (`*`) operator. The `MulExpression` (`@` matrix multiply) operator was missing an `is_hermitian()` override. Because of this, even though the self-product `X @ X` of a Hermitian matrix is mathematically guaranteed to be Hermitian `(X @ X)† = X† @ X† = X @ X`, CVXPY falls back to `Expression.is_hermitian()`, which evaluates to `False`. 

This false negative cascades into `trace()`, making `trace(X @ X).is_real()` return `False`, which completely blocks standard SDP, nuclear norm, and Lyapunov stability workflows. Using `cp.real()` as a workaround just triggers a separate DCP violation.

## The Fix

I've implemented a minimal, conservative fix across two files:

1. **`cvxpy/atoms/affine/binary_operators.py`**: Added an `is_hermitian()` method to `MulExpression`. It specifically checks for the self-product case (`self.args[0] is self.args[1]`) where the argument is strictly Hermitian. This has zero false positives for a general `A @ B` product.
2. **`cvxpy/atoms/affine/trace.py`**: Currently, `trace(A @ B)` fast-paths to `vdot`, which silently drops Hermitian property information. I updated the logic so that if the `MulExpression` is Hermitian, it bypasses the `vdot` fast-path and falls through to the `Trace` class so `Trace.is_real()` evaluates correctly.

## Impact

* Unblocks `trace(X @ X)` and `cp.sum(X @ X)` in objectives for complex Hermitian variables.
* No behavior change or performance hit for non-Hermitian or standard real-valued matrix multiplications.

### DIFF
--- a/cvxpy/atoms/affine/binary_operators.py
+++ b/cvxpy/atoms/affine/binary_operators.py
@@ -247,6 +247,18 @@ class MulExpression(BinaryOperator):
         # For matrix multiplication, use matmul bounds
         return bounds_utils.matmul_bounds(lb1, ub1, lb2, ub2)
 
+    def is_hermitian(self) -> bool:
+        """Is the expression Hermitian?
+
+        A @ A is Hermitian whenever A is Hermitian, because
+        (A @ A)† = A† @ A† = A @ A.  Other products are not
+        Hermitian in general (they require commutativity), so
+        fall back to the real-and-symmetric check.
+        """
+        if self.args[0] is self.args[1] and self.args[0].is_hermitian():
+            return True
+        return self.is_real() and self.is_symmetric()
+
     def is_incr(self, idx) -> bool:
         """Is the composition non-decreasing in argument idx?
         """
--- a/cvxpy/atoms/affine/trace.py
+++ b/cvxpy/atoms/affine/trace.py
@@ -27,12 +27,16 @@ from cvxpy.constraints.constraint import Constraint
 def trace(expr):
     """
     TLDR: Use alternate formulation for trace(A@B) for more efficient computation.
-    trace(A@B) normally is O(n^3) because of the A@B operation. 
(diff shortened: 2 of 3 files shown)


PTNobel:
Dropping the fast path is a huge performance hit; I'd love to figure out a way to keep it.

PTNobel:
Otherwise this looks great!

goyalpalak18:
Thank you for the review @PTNobel . I've pushed an update to address this.

I reverted the Trace routing so the $O(n^2)$ vdot fast path is always preserved. For the Hermitian case, the code now wraps the vdot result with cp.real(). This acts as a zero-cost wrapper that correctly propagates is_real() == True upward without having to materialize the full matrix product. 

Please let me know if this looks good to you, or if any other changes are needed!

SteveDiamond:
I don't think this PR is a good idea. We are already stretching DCP by allowing X @ X. Why not use the Frobenius norm atom?

goyalpalak18:
Thanks for explaining that, @SteveDiamond . I agree that cp.sum_squares(X) is the correct way to handle this under DCP rules.

My main reason for opening the PR was just that the current ValueError about the objective needing to be real-valued is pretty confusing for a user to debug if they accidentally write trace(X @ X).

Would you prefer I just close this PR, or should I update it to catch this specific case and return a more helpful error message pointing to sum_squares?

PR #3196, opened 2026-03-06
Fix silent wrong objective in `cp.trace(A @ B)` for non-symmetric matrices

goyalpalak18:
## Description
I'm submitting this PR to fix a critical regression introduced in PR #2920. Currently, `cp.trace(A @ B)` silently computes trace(Aᴴ @ B) instead of trace(A @ B) for non-Hermitian/non-symmetric matrices. 

If a user writes `cp.minimize(cp.trace(C @ X))` where `C` is asymmetric, the solver returns a solution that is optimal for the completely wrong problem, with no warnings or errors thrown.

## Root Cause
The previous optimization in PR #2920 introduced a fast-path shortcut using `vdot(A, B)`. However, there was a mix-up between conjugation and transposition. 

In CVXPY, `vdot(A, B)` computes ∑ᵢⱼ conj(Aᵢⱼ) · Bᵢⱼ, which is exactly trace(Aᴴ @ B). The correct mathematical identity for trace(A @ B) is ∑ᵢⱼ Aᵢⱼ · Bⱼᵢ, which translates to sum(A ⊙ Bᵀ). 

This slipped through the test suite because existing tests primarily evaluated PSD/symmetric variables where Xᵢⱼ = Xⱼᵢ, effectively masking the bug.

## The Fix
1. **Updated the fast-path math:** In `cvxpy/atoms/affine/trace.py`, I replaced `vdot(args[0], args[1])` with `sum(multiply(args[0], args[1].T))`. This correctly evaluates to sum(A ⊙ Bᵀ) for all matrices (real, complex, symmetric, or asymmetric) while preserving the O(n²) performance motivation of the original PR by avoiding the full matrix multiplication.
2. **Updated the test suite:** I modified `test_trace_AB` to use non-symmetric matrices. It now explicitly checks for numerical correctness rather than relying on fragile structural assertions (like checking for a `conj` node, which was an implementation detail of the flawed `vdot` path).


goyalpalak18:

cvxpy/atoms/affine/trace.py
```diff
@@ -27,14 +27,30 @@
...
+
+    When the MulExpression is Hermitian (e.g. X @ X^H for Hermitian X), the
+    result is provably real; it is wrapped with real() so that is_real()
+    correctly returns True without routing to the O(n^3) Trace(expr) path.
     """
-    if isinstance(expr, MulExpression): 
-        from cvxpy.atoms.affine.binary_operators import vdot
-        return vdot(expr.args[0], expr.args[1])
+    if isinstance(expr, MulExpression):
+        from cvxpy.atoms.affine.binary_operators import multiply
```
SteveDiamond: imports at the top

cvxpy/tests/test_atoms.py
SteveDiamond: please install the pre-commit

cvxpy/tests/test_atoms.py
```diff
@@ -882,20 +882,28 @@ def test_Trace(self) -> None:
...
-        assert len(t.args) == 1
-        assert isinstance(t.args[0], multiply)
-        assert len(t.args[0].args) == 2
-        assert isinstance(t.args[0].args[0], conj)
-        assert len(t.args[0].args[0].args) == 1
-        assert isinstance(t.args[0].args[0].args[0], reshape)
-        assert isinstance(t.args[0].args[1], reshape)
+
+        # Structural check: trace(A@B) is a scalar Sum of an element-wise multiply.
+        from cvxpy.atoms.affine.sum import Sum
```
SteveDiamond: imports at the top

cvxpy/tests/test_atoms.py
```diff
@@ -882,20 +882,28 @@ def test_Trace(self) -> None:
         assert isinstance(t, cp.Trace)
 
     def test_trace_AB(self) -> None:
-        """Test the trace(AB) gets canonicalized to vdot(A,B)
+        """Test that trace(A @ B) = sum(A * B.T), correct for non-symmetric matrices.
```
SteveDiamond, line 883: can you add a couple more test cases in this function for more thorough coverage?

cvxpy/atoms/affine/trace.py
goyalpalak18: Moved these up to the top. Verified it doesn't cause any circular imports.

cvxpy/tests/test_atoms.py
goyalpalak18: My bad! Fixed the import sorting. I'll set up pre-commit locally so I don't miss this again.

cvxpy/tests/test_atoms.py
goyalpalak18: Fixed, moved to the top.

cvxpy/tests/test_atoms.py
goyalpalak18: Added a few more cases: non-square factors (3x4 @ 4x3), complex non-Hermitian matrices, and a solve-level check with an asymmetric cost matrix.

SteveDiamond (approved):

PR #3214, opened 2026-03-10
Fix targets_and_priorities ignoring flipped tar/lim for negative prio…

goyalpalak18:
### Description

While working with multi-objective scalarizations, I stumbled across a subtle stale-variable bug in `cvxpy.transforms.targets_and_priorities()` that causes problems with negative priorities to silently solve against the wrong reference points.

I noticed that when `priorities[i] < 0`, the function correctly computes the negated `obj`, `tar`, and `lim` around line 91. However, on lines 95 and 99, the code still references the original `targets[i]` and `limits[i]` from the outer loop instead of using the locally updated `tar` and `lim`. 

Because Python doesn't warn us about unused local variables or stale outer references, this typo easily slipped through. The docstring correctly notes that negative priorities should flip the objective sense, but the code was effectively ignoring the flipped targets and limits later in the loop.

### Why this is critical

* **Silent Failure:** The penalty `delta` and hard-limit indicator compute against the original (un-negated) target/limit. The problem still successfully solves to `OPTIMAL` status, but it drives the optimizer toward the completely wrong reference point, resulting in an incorrect primal solution. Users currently get absolutely no error or warning that their scalarization is numerically wrong.
* **Test Blindspot:** I checked the existing tests in `test_scalarize.py` and saw they do use negative priorities, but they only assert `is_convex()` / `is_concave()`. Because the expressions maintain the correct curvature, our tests were passing even though the underlying numerical values were completely off.

### The Fix

My fix is super minimal—literally just two one-word substitutions on lines 95 and 99:
* Changed `targets[i]` to `tar`
* Changed `limits[i]` to `lim`

This ensures both the penalty delta and the hard-limit indicator consistently use the locally-flipped values we computed in the `if priorities[i] < 0` block. I've run the test suite locally and all tests continue to pass cleanly.


goyalpalak18:

PTNobel (approved):

PTNobel:
Is this part of the code tested at all? I fear not...

goyalpalak18:
I looked at `test_scalarize.py` and the existing tests for negative priorities only check `is_convex()` and `is_concave()`. They don't actually solve anything or check the numerical objective values, which is why this bug was completely invisible to the test suite.

I would be happy to add a quick numerical regression test to `test_scalarize.py` in this PR to make sure it stays fixed, if you'd like?

SteveDiamond:
> I looked at `test_scalarize.py` and the existing tests for negative priorities only check `is_convex()` and `is_concave()`. They don't actually solve anything or check the numerical objective values, which is why this bug was completely invisible to the test suite.
> 
> I would be happy to add a quick numerical regression test to `test_scalarize.py` in this PR to make sure it stays fixed, if you'd like?

Yes, whenever you fix a bug please add a test that would fail without the bugfix.

goyalpalak18:
@SteveDiamond Just pushed the regression test (`test_negative_priority_regression` in `test_scalarize.py`). 

I set up a small problem that explicitly fails on `master` without this fix. Under the old code, the stale variables end up creating an impossible indicator constraint like `(x-1)^2 <= -0.5`, which makes the problem evaluate to `INFEASIBLE`. 

With the fix, the bounds are flipped correctly, the constraints are valid, and it optimally solves to `x = 0.5`. The test now asserts both the `OPTIMAL` status and the exact numerical solution, so it will reliably catch this bug.

SteveDiamond (approved):

PR #3218, opened 2026-03-12
feat: add get_attr_dual() API for PSD/NSD dual variable recovery

dhru189:
## Summary

Extends dual variable recovery to ALL attribute constraints, addressing #3200.

## Problem

When variables are declared with attributes like PSD=True, nonneg=True etc., implicit constraints are added internally.
The dual variables for these constraints were never recovered  or exposed after solving.

## Solution

Introduces Problem.get_attr_dual(var, attr=None) — a unified  method that returns dual values for ANY attribute constraint.

## Usage

# PSD variable
X = cp.Variable((3,3), PSD=True)
prob.solve()
prob.get_attr_dual(X)          # {'PSD': array([[...]]}
prob.get_attr_dual(X, 'PSD')   # array([[...]])

# nonneg variable  
x = cp.Variable(3, nonneg=True)
prob.solve()     prob.get_attr_dual(x)          # {'nonneg': 1.0}
prob.get_attr_dual(x, 'nonneg') # 1.0

## Note

All attribute constraints (PSD, NSD, nonneg, nonpos, bounds)  are now supported via a unified API as suggested in review.


PTNobel:
Thanks for starting to think about this! I think one design axiom I have is that all attributes should have the option to have dual variables; certainly all the bounds do

dhru189:
> Thanks for starting to think about this! I think one design axiom I have is that all attributes should have the option to have dual variables; certainly all the bounds do

Thanks for the feedback! That makes sense as a design principle. So the API should expose dual variables for all constraint-generating attributes (PSD, NSD, nonneg, nonpos, bounds, etc.) through a unified interface rather than just PSD/NSD.

Would something like prob.get_attr_dual(var) returning a dict of {attr_name: dual_value} work, or do you have a preferred API shape in mind?

SteveDiamond:
Thanks for looking at this! I just wanted to flag that API changes like this require two reviewers to approve.

PTNobel:
> Would something like prob.get_attr_dual(var) returning a dict of {attr_name: dual_value} work, or do you have a preferred API shape in mind?

I have no idea; maybe the user should specify which attribute they want the duals of? 

I'd love to see a design doc with different proposals and recommendations

dhru189:
> > Would something like prob.get_attr_dual(var) returning a dict of {attr_name: dual_value} work, or do you have a preferred API shape in mind?
> 
> I have no idea; maybe the user should specify which attribute they want the duals of?
> 
> I'd love to see a design doc with different proposals and recommendations

Here are a few designs I was thinking about:

Option A (what I have now):
prob.get_attr_dual(var)          # returns all as dict   ,prob.get_attr_dual(var, 'PSD')   # specific one

Option B (on the variable itself):
var.attr_dual('PSD')

Option C (like constraint.dual_value):
prob.dual_value(var, attr='PSD')

I'm leaning toward Option A since it feels closest to how constraint.dual_value already works in CVXPY. But happy to go with whatever direction makes most sense to you!

PTNobel:
Consider putting together a longer doc: https://github.com/PTNobel/PTNobel/blob/main/cvxpy_design_docs/sparsity-api.md

It would be good to check what CVX and Convex.jl do as well. Does JuMP even support attributes?

dhru189:
Here's the design doc as requested:
https://gist.github.com/dhru189/19920d5d7c952cd49ab44f5d2d8da68e

Open to feedback  :

PTNobel:
Thanks for sending it! I am acknowledging receipt, but I am pretty busy and probably won't get to really sit with it for a bit.

dhru189:
> Thanks for sending it! I am acknowledging receipt, but I am pretty busy and probably won't get to really sit with it for a bit.

"Still in progress — waiting for design doc feedback before updating the implementation."

PTNobel:
Hmmm?

dhru189:
> Hmmm?

Sorry for the confusion! GitHub auto-prompted  "Still in progress?" so I just confirmed the status. No rush on the design doc review at all!

dhru189:
Resolved merge conflict with upstream master. 
PR is now up to date and ready for review!

dhru189:
Thinking about how this fits into post-solver verification — with attribute dual recovery, we can now include PSD/nonneg/bound constraints in KKT checks on the original problem, not just the canonicalized form.

This should complement DNLP dual recovery and move us toward complete dual coverage.

Does this align with the intended direction, or would you prefer any design changes here?

PTNobel:
I agree we'd need to access dual variables for attributes to do KKT verification on the original problem.

dhru189:
> I agree we'd need to access dual variables for attributes to do KKT verification on the original problem.
Based on our discussion, I’m planning to proceed with:
prob.get_attr_dual(var, attr=None)
- returns dict if attr=None
- returns specific dual if attr is provided
This keeps it consistent with constraint.dual_value while allowing flexibility.

I’ll refine the implementation and add more tests for all attribute types.Happy to adjust based on further feedback!

PR #3220, opened 2026-03-12
Fix sum_largest_canon DNLP warm-start using wrong threshold

goyalpalak18:
## Description
I noticed a bug in `sum_largest_canon.py` where the DNLP warm-start produces an infeasible initialization, causing NLP solvers to start from a constraint-violating state. 

The LP epigraph form of `sum_k_largest(x)` has the constraint `x_i <= t_i + q`. However, the current code computes the threshold `q` using the k-th smallest element instead of the k-th largest. Additionally, it only initializes `t` for the top-k elements. This means any "middle" elements (where `x_i > q`) violate the feasibility constraint. 

When passed to an NLP solver (like IPOPT or Knitro), this infeasible warm start forces a costly restoration phase or, worse, silently converges to a spurious local minimum.

## Changes Proposed
I implemented a straightforward fix to exactly match the LP-optimal primal solution:
* Removed `idx_of_smallest` as it was computing the incorrect threshold.
* Updated `q.value` to `np.min(x.value[idx_of_largest])` so it correctly targets the k-th largest element.
* Updated `t.value` to `np.maximum(0, x.value - q.value)` to ensure all elements are properly initialized, not just the top-k.

## Impact
With this fix, the DNLP warm-start for `sum_largest` (and `lambda_sum_largest`) will always begin from a feasible primal point. This allows NLP solvers to skip the restoration phase entirely, heavily reducing solve time, and prevents silent, incorrect sub-optimal returns in sequential convex approximation pipelines.

## Testing
I included regression tests to verify that `q` and `t` are computed correctly and that the constraints `x_i <= t_i + q` and `t_i >= 0` hold true for all elements, including the middle ones.


goyalpalak18:
Hi @PTNobel @SteveDiamond ! Could you please have a look at this when you have a moment? Thanks!

Transurgeon (commented):
nice catch! could add some tests that were failing previously? (if that's too difficult, could you just add some simple tests?)

goyalpalak18:
@Transurgeon I've added a regression test in e9cf9f4. It reproduces the exact scenario that caused the issue—specifically the middle elements between q and the top-k—to ensure that q, t, and both constraints are properly satisfied after the warm-start initialization.

Transurgeon (commented):
@dance858 could you have a double check at this PR please?
It looks good to me.. but weird that we didn't catch it or run into this error before.

cvxpy/tests/nlp_tests/test_huber_sum_largest.py
```diff
@@ -18,10 +18,38 @@
...
 
 import cvxpy as cp
+from cvxpy.atoms.sum_largest import sum_largest
+from cvxpy.expressions.variable import Variable
+from cvxpy.reductions.eliminate_pwl.canonicalizers.sum_largest_canon import sum_largest_canon
 from cvxpy.reductions.solvers.defines import INSTALLED_SOLVERS
 from cvxpy.tests.nlp_tests.derivative_checker import DerivativeChecker
 
 
+class TestSumLargestCanonWarmStart:
```
Transurgeon, line 28: maybe we need to add something like this as well? ``@pytest.mark.skipif('IPOPT' not in INSTALLED_SOLVERS, reason='IPOPT is not installed.')``

dance858:
Hi @goyalpalak18,

Thank you for raising this issue. There is indeed an issue with the code. The current line 

`idx_of_smallest = sorted_indices[:k]`

 should be 

`idx_of_smallest = sorted_indices[:-k]` 

The LP in the canonicalization has one solution $q^\star = x_{[k+1]}$ and $t_i = 0$ for $i = k + 1, ..., n$ and $t_i = x_i - q^\star$ for $i = 1, ..., k$. If you make the change above, then the code implements this solution. Many thanks for finding this.

cvxpy/tests/nlp_tests/test_huber_sum_largest.py
PTNobel: I don't think so? It doesn't actually IPOPT, just the canon file

Transurgeon: yes, sorry.

goyalpalak18:
Thanks @dance858 , you're totally right. I've updated the fix to use `sorted_indices[:-k]` like you suggested since it's definitely cleaner. I also updated the regression test to check against the exact LP-optimal values.

Transurgeon (approved):

PTNobel (approved):

PR #3222, opened 2026-03-12
Add cp.nlp submodule namespace for NLP atoms

Nikitaa104:
Closes #3198
Creates a dedicated cvxpy.nlp namespace for atoms that require an NLP solver (nlp=True). Users can now write:
(python)
import cvxpy as cp
x = cp.Variable()
prob = cp.Problem(cp.Minimize(cp.nlp.sin(x)), [x >= 0])
prob.solve(nlp=True)

Atoms available under cp.nlp: sin, cos, tan, sinh, tanh, asinh, atanh
Note: cp.sin etc. still work as before. cp.nlp provides a dedicated namespace making it clear these atoms require NLP solvers.
AI assistance was used for codebase exploration, as per CVXPY's AI disclosure policy.

### DIFF
--- a/cvxpy/tests/test_nlp_namespace.py
+++ b/cvxpy/tests/test_nlp_namespace.py
@@ -1,7 +1,7 @@
 """Tests for cp.nlp namespace."""
 import cvxpy as cp
-from cvxpy.atoms.elementwise.trig import sin, cos, tan
-from cvxpy.atoms.elementwise.hyperbolic import sinh, tanh, asinh, atanh
+from cvxpy.atoms.elementwise.hyperbolic import tanh
+from cvxpy.atoms.elementwise.trig import cos, sin
 
 
 class TestNLPNamespace:
--- a/cvxpy/tests/test_nlp_namespace.py
+++ b/cvxpy/tests/test_nlp_namespace.py
@@ -1,5 +1,7 @@
 """Tests for cp.nlp namespace."""
 import cvxpy as cp
+from cvxpy.atoms.elementwise.trig import sin, cos, tan
+from cvxpy.atoms.elementwise.hyperbolic import sinh, tanh, asinh, atanh
 
 
 class TestNLPNamespace:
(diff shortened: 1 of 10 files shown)


PTNobel:
See https://github.com/cvxpy/cvxpy/pull/3208#issuecomment-4045988678

Also, I think if we add the NLP namespace we should remove the NLP atoms from the top-level interface.

Nikitaa104:
Done. @PTNobel, NLP atoms are now removed from the top-level cp namespace and only accessible via cp.nlp. Updated the tests accordingly: removed test_atoms_same_as_direct (which asserted cp.sin is cp.nlp.sin, no longer valid since cp.sin doesn't exist) and added test_atoms_not_in_top_level to enforce the policy. All 2002 tests pass locally.

PTNobel:
Looks like the NLP tests aren't passing yet...

Nikitaa104:
Updated: the nlp_tests were still using cp.sin, etc., directly, which broke after removing them from the top-level namespace. All five affected test files now use cp.nlp.*.

cvxpy/tests/test_nlp_namespace.py
```diff
@@ -0,0 +1,39 @@
+"""Tests for cp.nlp namespace."""
+import cvxpy as cp
+from cvxpy.atoms.elementwise.trig import sin, cos, tan
+from cvxpy.atoms.elementwise.hyperbolic import sinh, tanh, asinh, atanh
+
+
+class TestNLPNamespace:
```
SteveDiamond, line 7: please install the pre-commit

SteveDiamond:
@Transurgeon I think this namespace is the right approach to make it more clear that the NLP solver is a different animal than the convex solvers and much more exploratory.

cvxpy/tests/test_nlp_namespace.py
Nikitaa104: Removed unused imports (tan, sinh, asinh, atanh) and sorted the import block to satisfy ruff. All 5 tests pass.

SteveDiamond (approved):
LGTM. I updated the DNLP tutorial docs to mention the namespace and also commented out sigmoid from the list of special DNLP atoms, since it is not implemented.

PR #3246, opened 2026-03-20
Fix sigma_max gradient crash on non-square matrices

goyalpalak18:
## Description                                                                                                                                                 
  Fix a `ValueError` crash when calling `.grad` on `sigma_max` with non-square matrices.                                                                         
                                                                                                                                                                 
  The root cause is in `LA.svd(values[0])` within `sigma_max._grad()`, which uses `full_matrices=True` by default. For rectangular matrices where dimensions     
  differ, this produces full unitary matrices with misaligned inner dimensions during gradient computation, preventing access to gradients for rectangular weight
   matrices commonly used in machine learning and control theory.                                                                                                
                                                                                                                                                               
  The fix is to use `LA.svd(values[0], full_matrices=False)` for economy-size SVD. This is the same fix that was applied to `normNuc` in PR #2426 (April 2024)   
  but was never applied to `sigma_max`.
                                                                                                                                                                 
  Verified that the analytical gradient now perfectly matches the numerical gradient to machine precision for non-square matrices.                               
   
  ## Type of change                                                                                                                                              
  - [x] Bug fix                                                                                                                                                  
                                                                                                                                                                 
  - [x] Write unittests.                                                                                                                                         


PTNobel:
Please use our PR template when opening PRs

cvxpy/atoms/sigma_max.py
```diff
@@ -55,7 +55,7 @@ def _grad(self, values):
             A list of SciPy CSC sparse matrices or None.
         """
         # Grad: U diag(e_1) V.T
-        U, s, V = LA.svd(values[0])
+        U, s, V = LA.svd(values[0], full_matrices=False)
         ds = np.zeros(len(s))
         ds[0] = 1
         D = U.dot(np.diag(ds)).dot(V)
```
PTNobel: Could we avoid the dense matrix in the middle with some broadcasting and multiplies? Would be nice while we're cleaning this up

PTNobel (approved):
Changes look great!

goyalpalak18:
> Please use our PR template when opening PRs

Thanks @PTNobel  for pointing that out! I've updated the description to follow the PR template!

cvxpy/atoms/sigma_max.py
goyalpalak18: Good call, replaced it with U[:, 0:1] @ V[0:1, :]. No dense diagonal needed.

PTNobel: I think you forgot to push

goyalpalak18: Sorry about that, pushed the fix!

PR #3247, opened 2026-03-21
Fix MatrixFrac._grad wrong vectorization order for matrix X

goyalpalak18:
## Description                                                                                                                                                 
                                                                                                                                                                                           
 `MatrixFrac._grad` (used by `Atom.grad`) computes the gradient of `matrix_frac(X, P) = tr(X^T P^{-1} X)` with respect to `X` and `P`. The gradient w.r.t. `X` is `(P^{-1} + P^{-T})X`, which must be vectorized in Fortran (column-major) order per CVXPY convention.
                                                                                                                                                                 
  **Bug:** Line 77 used `DX.T.ravel(order='F')`, which is equivalent to `DX.ravel(order='C')` (row-major order). This produces **silently wrong gradients** whenever `X` has more than one column (i.e., matrix `X` with shape `(n, p)` where `p > 1`). For vector `X` (single column), the transpose is a no-op and the bug is masked.                                                                                                                                                 
                  
  **Impact:** `Atom.grad` returns incorrect values for `matrix_frac` with matrix `X`, with errors up to ~0.97 (catastrophic). This does **not** affect `problem.backward()` or `problem.derivative()`, which use `diffcp` internally.
                                                                                                                                                                 
  **Fix:** Remove the `.T` so the line reads `DX.ravel(order='F')`, matching CVXPY's column-major vectorization convention. The gradient w.r.t. `P` (lines 80–84) is correct because its double-transpose cancels out.
                                                                                                                                                                 
  **Verification:** Finite-difference check confirms the fix produces correct gradients (max error ~1.85e-08 vs ~0.97 before the fix).                           
   
  ## Type of change                                                                                                                                              
  - [x] Bug fix                                                                                                                                                  
                                                                                                                                                                 
  - [x] Write unittests.                                                                                                                                         


goyalpalak18:
Hi @PTNobel ! Could you please have a look at this whenever you have a moment? Thanks!

PTNobel:
Please do not tag me when you open PRs. I check PRs very frequently and I get to many GitHub notifications already.

PTNobel (approved):

PR #3252, opened 2026-03-23
Fix targets_and_priorities wrong sign for affine Maximize objectives

goyalpalak18:
## Description                                                                                                                                                 
                                                                                                                                                                 
Currently, when passing a `Maximize` objective with an affine argument (like a plain `cp.Variable()`) to `targets_and_priorities`, the optimizer silently pushes the variable in the completely wrong direction—toward the lower bound instead of the target/upper bound. It doesn't throw an error; it just yields a valid but completely incorrect Pareto solution.
                                                                                                                                                                 
  **The Root Cause:**
  On line 93 of `cvxpy/transforms/scalarize.py`, the code uses `obj.args[0].is_convex()` to figure out the sign. Because affine expressions are simultaneously convex and concave in CVXPY, `is_convex()` evaluates to `True`. This forces the code down the `Minimize` path (setting `sign = 1`) regardless of what the actual objective type is.               

  **The Fix:**                                                                                                                                                   
  I updated line 93 to determine the sign based on the actual objective type rather than the argument's curvature.
  - **Before:** `sign = 1 if obj.args[0].is_convex() else -1`                                                                                                    
  - **After:** `sign = 1 if isinstance(obj, Minimize) else -1`
                                              
  This ensures that linear/affine `Maximize` objectives (and `Minimize` objectives with negative priorities) retain their correct optimization directions. I've also added two regression tests in `test_scalarize.py` to ensure this behavior is locked down. All 9 tests in the suite now pass.
                                                                                                                                                                 
  ## Type of change                                                                                                                                              
  - [x] Bug fix
                                          
  - [x] Write unittests.                                                                                                                                         


PTNobel (approved):

PTNobel:
Wow this part of the code seems really undertested 😅

PTNobel:
Great catch!

PR #3262, opened 2026-03-26
Allow array denominator in quad_over_lin

### LINKED ISSUE #1197: Vectorized `quad-over-lin` or similar?
I needed to work around the following issue: the intuitive cost function I would like to express is as below in pseudo code

```python
x = cp.Variable(n, nonneg=True)
y = <array of n positive values>

objective = cp.Minimize(cp.sum(cp.square(x - y) / (x + y)))
```

The canonical form is an SOCP, but cvxpy resolves this (incorrectly) to quasiconvex, so an alternative is needed. See below for the two alternatives I could get to work.

It would be nice if e.g., `quad_over_lin` also accepted `y` same shape as `X` since it's pretty clear that

```python
cp.sum([cp.quad_over_lin(xx - yy, xx + yy) for (xx, yy) in zip(x, y)])
```

is just a sum of special cases of `quad_over_lin`.

**Describe alternatives you've considered**

1. quick abuse of `quad_over_lin`, but it's not vectorized and very slow with n>1e3:
```python
cp.sum([cp.quad_over_lin(xx - yy, xx + yy) for (xx, yy) in zip(x, y)])
```

2. just bypass the automatic reduction step for quad-over-lin and write the canonical form by hand:

```python
t = cp.Variable(n, nonneg=True)
objective = cp.Minimize(cp.sum(t))
constraints += SOC(t=x + y + t, X=vstack([x + y - t, 2 * (x - y)]), axis=0,)
```

This is super fast, but kind of defeats the purpose of higher order expressions and makes the problem harder to read and experiment with.

**Additional context**
I spent 20 minutes trying to implement a variation of `quad_over_lin` but realized there's a fair bit of non-trivial logic in a lot of reductions...

gugar20:
## Description

Closes #1197.

- Allow `quad_over_lin(x, y)` where `x` and `y` are broadcast-compatible.
- Broadcasts, computes element-wise `x_ij^2 / y_ij`, then reduces along `axis`.
- Follow the `AxisAtom` convention: `axis=None` (default) sums all elements to a scalar. `axis=()` gives element-wise output with no reduction. `axis=k` reduces along axis `k`.
- Update all four canonicalizers: SOC (dcp2cone), QP (quad), DGP (dgp2dcp), and DNLP (dnlp2smooth).

## Type of change
- [ ] Bug fix

- [x] Write unittests.


PTNobel:
Please use our PR template

PTNobel:
Can you explain why we wouldn't include broadcasting between x and y? What do you mean by reduce?

PTNobel:
Please install and run the pre commit

PTNobel:
My first thought in my review is that the tests seem way more complicated than for any comparable atoms. Can you compress them?

gugar20:
I will look into broadcasting, by reduce I meant sum

I thought I had set up pre-commit properly, everything is green locally, I'll resolve.

I will simplify the tests to match other similar atoms.

gugar20:
I implemented broadcasting, cut down on the tests and checked my setup of pre-commit (but didn't find any failures).

cvxpy/reductions/dcp2cone/canonicalizers/quad/quad_over_lin_canon.py
```diff
@@ -47,7 +59,7 @@ def quad_over_lin_canon(expr, args, solver_context=None):
             axes = (normalize_axis_index(axis, ndim),)
         else:
             axes = normalize_axis_tuple(axis, ndim)
-
+
```
PTNobel: I had thought this would be prohibited my the pre-commit... I need to debug why it isn't firing

gugar20: Claude flagged that as rule W293; I tried applying it but got a lot of changes so left it out and removed the whitespace

PTNobel: Okay we just merged W293 in as part of #3264, thanks for ctching it wasn't enforced!

cvxpy/atoms/affine/broadcast_to.py
```diff
@@ -24,6 +24,10 @@
 from cvxpy.constraints.constraint import Constraint
 
 
+def lazy_broadcast_to(expr, shape):
+    """Broadcast expr to shape only if necessary."""
+    return expr if expr.shape == shape else broadcast_to(expr, shape)
```
PTNobel: broadcast_to is a No-Op in many cases, I'm surprised this is doing anything. What inspired you to add it?

cvxpy/atoms/quad_over_lin.py
```diff
@@ -152,12 +177,47 @@ def is_decr(self, idx) -> bool:
...
+            raise ValueError(
+                "The shapes of the arguments to quad_over_lin are incompatible."
+            )
+        # Validate axis against broadcast ndim
+        if self.axis is not None:
+            axes = [self.axis] if isinstance(self.axis, int) else self.axis
+            normalize_axis_tuple(axes, len(bshape))
+        Atom.validate_arguments(self)
+
+    def shape_from_args(self) -> Tuple[int, ...]:
```
PTNobel: Shouldn't AxisAtom do all this for us?

cvxpy/atoms/quad_over_lin.py
```diff
@@ -173,3 +233,30 @@ def is_qpwa(self) -> bool:
         """Quadratic of piecewise affine if x is PWL and y is constant.
         """
         return self.args[0].is_pwl() and self.args[1].is_constant()
+
+def _unbroadcast(
```
PTNobel, line 236: Is this the transpose of broadcasting?

cvxpy/atoms/affine/broadcast_to.py
gugar20: I initially added it because I broke 2 tests* and thought it caused a benchmark regression. I just reran the affected benchmark a few times with/without and it actually looked like noise, so I removed it (and updated affected tests). I should mention that since this PR would always call `broadcast_to` in SOC and QP canonicalizers and it [doesn't support cpp](https://github.com/cvxpy/cvxpy/pull/3262/changes#diff-197344a5ec808555da576f2e6540d2e0f6797896259f8870cd1156aee4fb440dL34), these problems will be canonicalized with Scipy. Using `lazy_broadcast_to` prevented that (returned the variable itself, so didn't add a cpp blocker to `quad_over_lin`).

Another notable difference is that even when `broadcast_to` is a no-op, the QP canonicalizer adds an extra expression [here](https://github.com/cvxpy/cvxpy/pull/3262/changes#diff-ff35bcf9c651e91bd75633bef59d06c85080b2943348b76a41f00c1e46f178bdL53), so e.g. users with array x, scalar y (the existing supported behavior) would see different problem_data. This is why I changed the tests which used problem data to adjust for the larger P. I don't know if it affects performance though.

cvxpy/atoms/quad_over_lin.py
gugar20: AxisAtom infers the shape from the first argument, but with broadcasting in quad_over_lin we need both x and y to determine the full broadcast shape, and apply the axis reductions to this shape. For example, if x.shape = (3,), y.shape = (5,4,3); broadcast_shape = (5,4,3) and axis=0 gives (4,3). We'd get () from AxisAtom (scalar result).

Edit: Maybe I misunderstood, did you mean that this behavior should be handled by AxisAtom? I _think_ quad_over_lin is the only multi-argument subclass to AxisAtom

cvxpy/atoms/quad_over_lin.py
gugar20: It sums on the extra axes between `arr` and the target shape, the `unbroadcast` name doesn't exactly convey this but it's used for lack of a better name. (And because it applies the logic of `np.broadcast` backwards to determine the axes to sum on)

PR #3263, opened 2026-03-27
Fix targets_and_priorities wrong direction when priority <= off_target

goyalpalak18:
## Description

I found a critical bug in `targets_and_priorities` where setting an objective's priority less than or equal to `off_target` causes the solver to silently reverse the optimization direction.

Currently, when a `Maximize` objective has `priority <= off_target`, the `pos(delta)` coefficient becomes zero or negative. This makes the scalarized expression affine rather than strictly concave. Since the existing logic determines the output type purely from curvature analysis, and `is_convex()` evaluates to true for affine expressions, it wraps the problem in a `Minimize` objective.

This results in the solver returning an `OPTIMAL` status while driving the variable in the exact opposite direction of what was intended, with no warnings or errors. This naturally occurs if someone tries to disable an objective by setting `priority=0`.

## Fix

I updated the logic in `cvxpy/transforms/scalarize.py` to track the intended direction directly from the objective types rather than relying purely on curvature.

- When all objectives agree (e.g., all `Minimize` or all `Maximize` after evaluating priority signs), the code now explicitly enforces that direction.
- If the resulting curvature is inconsistent with the enforced direction, it raises a clear error instead of silently failing.
- Mixed objective problems fall back to the original curvature-based logic.

## Testing

- Added two new regression tests in `cvxpy/tests/test_scalarize.py` to verify the correct optimization direction is maintained when `priority <= off_target`, for both `Maximize` and `Minimize` symmetric cases.
- Passed local test suite.

## Type of change
- [x] Bug fix

## Contribution checklist
- [x] Write unittests.


SteveDiamond (approved):
Thanks for fixing!

PR #3280, opened 2026-04-05
fix: raise ValueError when cp.Parameter receives a sparse value (#2890)

### LINKED ISSUE #2890: Naive usage of sparse `cp.Parameter` can lead to uncaught errors
**Describe the bug**

A naive user (like me) might think you can initialize a parameter with a sparse matrix like:
```python
grad = y.grad[x]
<Compressed Sparse Column sparse array of dtype 'float64'
        with 4 stored elements and shape (4, 4)>
  Coords        Values
  (0, 0)        1.0
  (1, 1)        1.0
  (2, 2)        1.0
  (3, 3)        1.0

p = cp.Parameter((4, 4), value=grad, name="p")
```

This is obviously wrong but will not be caught until canonicalization occurs. 

**To Reproduce**

```python
import cvxpy as cp

x = cp.Variable((2, 2), value=[[1, 0], [0, 1]])

# simple expression
y = x + 1

# naive approach
grad = y.grad[x]
p = cp.Parameter((4, 4), value=grad, name="p")

# use the proper sparse parameter approach
# grad = y.grad[x].tocoo()
# sparsity_pattern = (grad.row, grad.col)
# p = cp.Parameter((4, 4), sparsity=sparsity_pattern, name="p")
# p.value_sparse = coo_array((grad.data, sparsity_pattern), shape=(4, 4))

obj = cp.Minimize(cp.sum(cp.square(cp.vec(x, order="F").T @ p)))
prob = cp.Problem(obj, [x >= 0])
print(f"Is DCP compliant: {prob.is_dcp()}")
prob.solve(solver=cp.SCS, verbose=True)
```

**Expected behavior**

I think some form of validation on the `value` field of `cp.Parameter` needs to happen to exit early and provide good feedback. 

**Output**

<details Trace>
  <summary><i>Trace</i></summary>

```python
Is DCP compliant: True
===============================================================================
                                     CVXPY                                     
                                     v1.7.1                                    
===============================================================================
(CVXPY) Aug 09 09:00:57 PM: Your problem has 4 variables, 4 constraints, and 16 parameters.
(CVXPY) Aug 09 09:00:57 PM: It is compliant with the following grammars: DCP, DQCP
(CVXPY) Aug 09 09:00:57 PM: CVXPY will first compile your problem; then, it will invoke a numerical solver to obtain a solution.
(CVXPY) Aug 09 09:00:57 PM: Your problem is compiled with the CPP canonicalization backend.
                                  Compilation                                  
(CVXPY) Aug 09 09:00:57 PM: Compiling problem (target solver=SCS).
(CVXPY) Aug 09 09:00:57 PM: Reduction chain: Dcp2Cone -> CvxAttr2Constr -> ConeMatrixStuffing -> SCS
(CVXPY) Aug 09 09:00:57 PM: Applying reduction Dcp2Cone
(CVXPY) Aug 09 09:00:57 PM: Applying reduction CvxAttr2Constr
(CVXPY) Aug 09 09:00:57 PM: Applying reduction ConeMatrixStuffing
(CVXPY) Aug 09 09:00:57 PM: Applying reduction SCS
TypeError: float() argument must be a string or a real number, not 'csc_array'

The above exception was the direct cause of the following exception:

  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/sparsity_issue.py", line 21, in <module>
    prob.solve(solver=cp.SCS, verbose=True)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/problems/problem.py", line 609, in solve
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/problems/problem.py", line 1191, in _solve
    data, solving_chain, inverse_data = self.get_problem_data(
                                        ~~~~~~~~~~~~~~~~~~~~~^
        solver, gp, enforce_dpp, ignore_dpp, verbose, canon_backend, kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/problems/problem.py", line 805, in get_problem_data
                         ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/reductions/chain.py", line 76, in apply
                   ~~~~~~~^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/reductions/solvers/conic_solvers/scs_conif.py", line 215, in apply
    return super(SCS, self).apply(problem)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/reductions/solvers/conic_solvers/conic_solver.py", line 362, in apply
    P, c, d, A, b = problem.apply_parameters(quad_obj=True)
                    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py", line 208, in apply_parameters
    param_vec = canonInterface.get_parameter_vector(
        self.total_param_size,
    ...<2 lines>...
        param_value,
        zero_offset=zero_offset)
  File "/mnt/c/Users/lange/OneDrive/Projects/2025/cvx/dccp/.venv/lib/python3.13/site-packages/cvxpy/cvxcore/python/canonInterface.py", line 61, in get_parameter_vector
    param_vec[col:col + size] = value
    ~~~~~~~~~^^^^^^^^^^^^^^^^
ValueError: setting an array element with a sequence.
```

</details>

**Version**
 - OS: W11 + WSL2 ubuntu 22.04 LTS
 - CVXPY Version: 1.7.1

Vivaan-Atharva:
Fixes #2890

## Problem
`cp.Parameter` silently accepted a scipy sparse matrix as its `value`,
storing it without error. The failure only surfaced later during
canonicalization with a confusing traceback.

## Reproduction
```python
import cvxpy as cp
x = cp.Variable((2, 2), value=[[1, 0], [0, 1]])
grad = (x + 1).grad[x]  # sparse matrix
p = cp.Parameter((4, 4), value=grad)  # should error here, didn't
```

## Fix
Override the `value` setter in `Parameter` to check `scipy.sparse.issparse()`
before delegating to `Leaf.value.fset`. Catches the error at assignment time
with a clear message pointing to `cp.Constant` as the alternative.

## Tests
- Rejects CSC, CSR, COO sparse arrays at construction time
- Rejects sparse via `.value =` setter after construction  
- Reproduces the exact case from the issue (grad sparse matrix)
- Verifies dense, None, scalar still accepted

### DIFF
--- a/cvxpy/expressions/constants/parameter.py
+++ b/cvxpy/expressions/constants/parameter.py
@@ -15,8 +15,6 @@ limitations under the License.
 """
 from __future__ import annotations
 
-import scipy.sparse as sp
-
 import cvxpy.lin_ops.lin_utils as lu
 from cvxpy import settings as s
 from cvxpy.expressions.expression import Expression
@@ -76,6 +74,7 @@ class Parameter(Leaf):
 
     @Leaf.value.setter
     def value(self, val) -> None:
+        import scipy.sparse as sp
         if sp.issparse(val):
             raise ValueError(
                 f"Cannot set cp.Parameter.value to a sparse matrix "
--- a/cvxpy/expressions/constants/parameter.py
+++ b/cvxpy/expressions/constants/parameter.py
@@ -83,7 +83,7 @@ class Parameter(Leaf):
                 "attribute and assign via `.value_sparse`. "
                 "To use a sparse constant, use cp.Constant() instead."
             )
-        Leaf.value.fset(self, val)
+        Leaf.value.fset(self, val)  # type: ignore[misc]
 
     def get_data(self):
         """Returns info needed to reconstruct the expression besides the args.
--- a/cvxpy/expressions/constants/parameter.py
+++ b/cvxpy/expressions/constants/parameter.py
@@ -15,6 +15,8 @@ limitations under the License.
 """
 from __future__ import annotations
 
+import scipy.sparse as sp
+
 import cvxpy.lin_ops.lin_utils as lu
 from cvxpy import settings as s
 from cvxpy.expressions.expression import Expression
@@ -71,6 +73,18 @@ class Parameter(Leaf):
         self.gradient = None
         super(Parameter, self).__init__(shape, value, **kwargs)
         self._is_constant = True
+
+    @Leaf.value.setter
+    def value(self, val) -> None:
+        if sp.issparse(val):
+            raise ValueError(
+                f"Cannot set cp.Parameter.value to a sparse matrix "
+                f"(type: {type(val).__name__}). "
+                "To use a sparse parameter, construct it with the `sparsity` "
+                "attribute and assign via `.value_sparse`. "
+                "To use a sparse constant, use cp.Constant() instead."
+            )
+        Leaf.value.fset(self, val)
 
     def get_data(self):
         """Returns info needed to reconstruct the expression besides the args.
--- a/cvxpy/tests/test_parameter_sparse.py
+++ b/cvxpy/tests/test_parameter_sparse.py
@@ -1,0 +1,71 @@
+"""
+Tests for Issue #2890: cp.Parameter should reject sparse matrix values
+at assignment time with a clear ValueError, not silently accept them
+and fail later during canonicalization.
+"""
+
+import numpy as np
+import pytest
+import scipy.sparse as sp
+
+import cvxpy as cp
+
+
+class TestParameterSparseValueRejection:
+
+    @staticmethod
+    def _make_sparse_csc(n: int = 4) -> sp.csc_array:
+        return sp.eye(n, format="csc")
+
+    @staticmethod
+    def _make_sparse_from_grad() -> sp.spmatrix:
+        x = cp.Variable((2, 2), value=[[1, 0], [0, 1]])
+        y = x + 1
+        return y.grad[x]
+
+    def test_rejects_sparse_csc_at_construction(self):
+        with pytest.raises(ValueError, match="sparse"):
+            cp.Parameter((4, 4), value=self._make_sparse_csc(4))
+
+    def test_rejects_sparse_csr_at_construction(self):
+        with pytest.raises(ValueError, match="sparse"):
+            cp.Parameter((3, 3), value=sp.eye(3, format="csr"))
+
+    def test_rejects_sparse_coo_at_construction(self):
+        with pytest.raises(ValueError, match="sparse"):
+            cp.Parameter((2, 2), value=sp.eye(2, format="coo"))
+
+    def test_rejects_grad_sparse_at_construction(self):
+        grad = self._make_sparse_from_grad()
+        with pytest.raises(ValueError, match="sparse"):
+            cp.Parameter((4, 4), value=grad, name="p")
+
+    def test_rejects_sparse_via_setter(self):
+        p = cp.Parameter((4, 4), name="p")
+        with pytest.raises(ValueError, match="sparse"):
+            p.value = self._make_sparse_csc(4)
+
+    def test_rejects_grad_sparse_via_setter(self):
+        p = cp.Parameter((4, 4), name="p")
+        with pytest.raises(ValueError, match="sparse"):
+            p.value = self._make_sparse_from_grad()
+
+    def test_accepts_dense_ndarray(self):
+        p = cp.Parameter((3, 3))
+        p.value = np.eye(3)
+        assert p.value is not None
+
+    def test_accepts_none(self):
+        p = cp.Parameter((3, 3))
+        p.value = None
+        assert p.value is None
+
+    def test_accepts_scalar(self):
+        p = cp.Parameter()
+        p.value = 3.14
+        assert abs(p.value - 3.14) < 1e-10
+
+    def test_error_mentions_constant_alternative(self):
+        p = cp.Parameter((4, 4))
+        with pytest.raises(ValueError, match="cp.Constant"):
+            p.value = self._make_sparse_csc(4)
--- a/cvxpy/tests/test_expressions.py
+++ b/cvxpy/tests/test_expressions.py
@@ -367,8 +367,8 @@ class TestExpressions(BaseTest):
 
         # Test valid diagonal parameter.
         p = Parameter((2, 2), diag=True)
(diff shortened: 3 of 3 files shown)


cvxpy/expressions/constants/parameter.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import cvxpy.lin_ops.lin_utils as lu
+import scipy.sparse as sp
```
SteveDiamond: please install the pre-commit

cvxpy/expressions/constants/parameter.py
```diff
@@ -72,6 +73,19 @@ def __init__(
         super(Parameter, self).__init__(shape, value, **kwargs)
         self._is_constant = True
 
+    @Leaf.value.setter
+    def value(self, val) -> None:
+        if sp.issparse(val):
+            raise ValueError(
+                "Setting a cp.Parameter value to a sparse matrix or array "
```
SteveDiamond: This error message is not quite right. We have a distinction now between .value and .value_sparse. I think now that Scipy has ND sparse arrays we can pass in sparse inputs via .value.

cvxpy/expressions/constants/parameter.py
Vivaan-Atharva: Donee, installed pre commit and ran it, ruff fixed the import ordering.

cvxpy/expressions/constants/parameter.py
Vivaan-Atharva: Updated the message, it no longer claims Parameters are dense only. It now points to the sparsity attribute + .value_sparse for sparse parameters, and cp.Constant() for the constant use case.

cvxpy/expressions/constants/parameter.py
```diff
@@ -83,7 +83,7 @@ def value(self, val) -> None:
                 "attribute and assign via `.value_sparse`. "
                 "To use a sparse constant, use cp.Constant() instead."
             )
-        Leaf.value.fset(self, val)
+        Leaf.value.fset(self, val)  # type: ignore[misc]
```
PTNobel, line 86: We generally avoid just masking pyright errors; can you explain why we can't fix the underlying issue? I think the problem is we don't have an annotation on val.

cvxpy/tests/test_parameter_sparse.py
PTNobel, line 1: Why is this a new file? If we keep it as a new file it needs the copyright header etc.

cvxpy/tests/test_parameter_sparse.py
```diff
@@ -0,0 +1,71 @@
...
+        p = cp.Parameter((3, 3))
+        p.value = np.eye(3)
+        assert p.value is not None
+
+    def test_accepts_none(self):
+        p = cp.Parameter((3, 3))
+        p.value = None
+        assert p.value is None
+
+    def test_accepts_scalar(self):
```
PTNobel, line 63: Surely we already have this test somewhere?

cvxpy/tests/test_parameter_sparse.py
```diff
@@ -0,0 +1,71 @@
...
+import scipy.sparse as sp
+
+import cvxpy as cp
+
+
+class TestParameterSparseValueRejection:
+
+    @staticmethod
+    def _make_sparse_csc(n: int = 4) -> sp.csc_array:
+        return sp.eye(n, format="csc")
```
PTNobel, line 18:
```suggestion
        return sp.eye_array(n, format="csc")
```

cvxpy/tests/test_parameter_sparse.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def _make_sparse_from_grad() -> sp.spmatrix:
+        x = cp.Variable((2, 2), value=[[1, 0], [0, 1]])
+        y = x + 1
+        return y.grad[x]
+
+    def test_rejects_sparse_csc_at_construction(self):
+        with pytest.raises(ValueError, match="sparse"):
+            cp.Parameter((4, 4), value=self._make_sparse_csc(4))
```
PTNobel, line 28: Why are we using a helper method here but not elsewhere?

cvxpy/tests/test_parameter_sparse.py
PTNobel: Overall I think this is way too many tests

Vivaan-Atharva: > Why is this a new file? If we keep it as a new file it needs the copyright header etc.

moved everything into test_parameters_failures, deleted the new file

cvxpy/tests/test_parameter_sparse.py
Vivaan-Atharva: yeah removed those, they're already in test_parameters_successes

cvxpy/tests/test_parameter_sparse.py
Vivaan-Atharva: removed the helper, just inlined sp.eye_array directly

cvxpy/expressions/constants/parameter.py
Vivaan-Atharva: ah got it, added an assert to narrow the type instead

PR #3288, opened 2026-04-13
Fix backward/derivative crash for Hermitian parameters

goyalpalak18:
## Description
I ran into a bug where `backward()` and `derivative()` blow up with a `ValueError` when you have a Hermitian `cp.Parameter` and `requires_grad=True`.

What's going on is `Complex2Real` and `CvxAttr2Constr` both touch the same intermediate real parameter but disagree on its shape. `CvxAttr2Constr` compacts the real gradient into upper-triangle form on the way back, but `Complex2Real` doesn't know that — it just tries to add the compact `(3,)` array to the full `(2,2)` imaginary gradient and you get a shape mismatch. Same thing in reverse during the forward pass.

The fix is straightforward: in `param_backward`, I use `recover_value_for_leaf` to expand the compact real gradient back to a full matrix before combining. In `param_forward`, I use `lower_value` to pack the real delta into compact form before passing it along. Both functions already existed in `cvx_attr2constr` — I just wasn't calling them.

Added a regression test that hits both code paths with a 2×2 Hermitian parameter.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/complex2real/complex2real.py
+++ b/cvxpy/reductions/complex2real/complex2real.py
@@ -36,6 +36,12 @@ from cvxpy.reductions.complex2real.canonicalizers import (
 from cvxpy.reductions.complex2real.canonicalizers import (
     Complex2RealCanonMethods,
 )
+from cvxpy.reductions.cvx_attr2constr import (
+    SYMMETRIC_ATTRIBUTES,
+    attributes_present,
+    lower_value,
+    recover_value_for_leaf,
+)
 from cvxpy.reductions.reduction import Reduction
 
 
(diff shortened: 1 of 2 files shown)


PTNobel:
Does #3147 also fix this? I feel like this is spooky action at a distance which I'm generally scared of

goyalpalak18:
Ah yeah, #3147 would definitely fix this since it gets rid of the shared ID thing entirely. My PR is just a smaller stopgap for the current code. Happy to close this if you'd rather just land #3147.

PTNobel:
I'll block this PR for now on #3147, if we start prepping a release and we haven't reached consensus on it, then I'll withdraw my objections and we can merge this PR.

PTNobel:
#3147 is merged! This bug is fixed...

Issue #1355, opened 2021-04-29
warm_start not working


pizquierdo2:
**Describe the bug**
Hello everybody. First of all, thank you very much for this incredible tool. I'm trying to solve a problem that seems to be very difficult: even in very little instances the solvers struggle to find a solution (and the give a warning saying that is inaccurate). Now, because of the nature of the problem, the solution might be very close to the zero vector. That's why I tried to give the solver a hand and use ```warm_start``` having instantiated previously the zero vector to the variables of the problem. However, the solver gives a solution that is way worse than the one initialized with warm start.

**To Reproduce**
```
import cvxpy as cp
import numpy as np

C_sound = 0.43633231299858233
G_sound = np.asarray([[0.4156921938165309 - 0.23999999999999963j, 0.29692299558323665 + 0.1714285714285704j],
                      [-0.665569940421644 + 0.009673545578146104j, 0.14151808351117373 + 0.21848875925703226j],
                      [-4.656133759306296e-16 - 0.8j, 2.287657973439295e-16 + 0.26666666666666666j],
                      [-0.665569940421644 + 0.009673545578146104j, 0.14151808351117298 + 0.21848875925703276j],
                      [-2.078460969082654 - 1.1999999999999944j, -0.18895099718933187 + 0.10909090909090943j]])
u0_sound = np.asarray([0.009906430846671255 - 0.0010830386455277078j, 0.008187941456002363 + 0.00178927838865346j,
                       0.005509237543204766 - 0.007985358398429381j, -0.006763413493560299 - 0.009293348678396056j,
                       -0.007994080650317923 - 0.004615384615384564j])
T_w_sound = np.asarray(
    [285855.5924994196, 383076.31134604354, 299134.1010521829, 224575.94674223074, 325376.06626492966])

C_silence = 43633231.29985823
G_silence = np.asarray([[-0.18895099718933187 + 0.10909090909090943j, -2.078460969082654 - 1.1999999999999944j],
                        [0.14151808351117373 + 0.21848875925703226j, -0.6655699404216439 + 0.009673545578147284j],
                        [2.287657973439295e-16 + 0.26666666666666666j, -4.656133759306296e-16 - 0.8j],
                        [0.14151808351117373 + 0.21848875925703226j, -0.6655699404216441 + 0.009673545578144925j],
                        [0.29692299558323665 + 0.1714285714285704j, 0.4156921938165309 - 0.23999999999999963j]])
[...]

PR #3237, opened 2026-03-17
Update docs to clarify warm start/initial guess support (#2912)

nikhil6393:
Updated "Warm start" documentation to clarify that providing an initial guess via the [value](cci:1://file:///d:/downloads/all%20projects/new4/cvxpy/cvxpy/expressions/expression.py:131:4-136:35) field is not supported by many solvers.

Issue link: #2912

## Type of change
- [ ] Bug fix
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [x] Other (Documentation)

## [Contribution checklist]
- [x] Check that your code adheres to our coding style.
- [x] Run the unittests and check that they’re passing (N/A for docs).

PR #3293, opened 2026-04-15
docs(solvers): clarify warm start does not use variable values as initial guess

### LINKED ISSUE #2912: DOC: providing initial guess is not supported
The documentation [suggests](https://www.cvxpy.org/tutorial/solvers/index.html#warm-start):

> Warm start can also be used to provide an initial guess the first time a problem is solved. The initial guess is constructed from the `value` field of the problem variables.

It seems to be incorrect, as per https://github.com/cvxpy/cvxpy/issues/1355, e.g. (modified from [here](https://stackoverflow.com/questions/52314581/initial-guess-warm-start-in-cvxpy-give-a-hint-of-the-solution)):

```
import cvxpy as cvx

# Examples: linear programming
# Create two scalar optimization variables.
x = cvx.Variable(value=0.5)
y = cvx.Variable(value=0.5)
# Create 4 constraints.
constraints = [x >= 0, y >= 0, x + y >= 1, 2 * x + y >= 1]

# Form objective.
obj = cvx.Minimize(x + y)

# Form and solve problem.
prob = cvx.Problem(obj, constraints)
prob.solve(warm_start=True)  # Returns the optimal value.
print("status:", prob.status)
print("optimal value", prob.value)
print("optimal var", x.value, y.value)

# status: optimal
# optimal value 1.0000000021945776
# optimal var 0.4971185775945734 0.5028814246000043

```

Expected to obtain a solution `(0.5, 0.5)` with smaller objective value (`1.0`) if `(x.value, y.value)` is used as an initial guess.

So, the documentation seems to be misleading and needs to be updated.

I believe the documentation should explicitly mention that `warm_start=True` cannot be used to provide the initial guess by the user, up until this support is implemented.

armorbreak001:
Fixes #2912

The warm start documentation incorrectly stated that `warm_start=True` uses
the `value` field of problem variables as an initial guess on first solve.
This is misleading — issue #1355 confirmed that warm start does not use
variable values as an initial guess.

**Changes:**
- Replaced the incorrect statement with a `.. note::` block that explicitly
  clarifies warm start is only for reusing cached solver state across multiple
  solves of the same problem instance (e.g., parameter sweeps in a loop).

**Verification:**
- Read the updated section at `doc/source/tutorial/solvers/index.rst`
  under the "Warm start" heading
- Confirm the new note clearly states the limitation

### DIFF
--- a/doc/source/tutorial/solvers/index.rst
+++ b/doc/source/tutorial/solvers/index.rst
@@ -289,10 +289,11 @@ The speed up in this case comes from caching the KKT matrix factorization.
 If ``A`` were a parameter, factorization caching would not be possible and the benefit of
 warm start would only be a good initial point.
 
-Warm start can also be used to provide an initial guess the first time a problem is solved.
-The initial guess is constructed from the ``value`` field of the problem variables.
-If the same problem is solved a second time, the initial guess is constructed from the
-cached previous solution as described above (rather than from the ``value`` field).
+.. note::
+   Warm start does **not** use the ``value`` field of problem variables as an initial guess
+   on the first solve. It only provides benefits when solving the **same** problem instance
+   multiple times (e.g., in a loop with changing parameters), by reusing cached solver
+   state such as factorizations or the previous solution point.
 
 .. _solveropts:
 


rileyjmurray:
I’m under the impression that warm start semantics have changed since that
issue was filed five years ago. I think the semantics now differ from one
solver to another. Can you double check? Sorry I’m writing this from my
phone.

armorbreak001:
@rileyjmurray Thanks for flagging this — good catch from your phone! 🙏

You're right that I should verify whether the semantics have diverged across solvers since #1355 was filed. Let me check:

1. **OSQP / ECOS / SCS** — These are the most commonly used solvers with `warm_start`. I'll test each to confirm whether `Variable(value=x)` actually propagates as an initial guess on first solve.

2. **The original issue (#2912) references #1355** which showed that at least for some solvers, the `value` field was NOT being used. But if solver interfaces have been updated since then, the doc may now be accurate again (making this PR unnecessary or partially wrong).

I'll run a quick verification script against the latest main and report back here. If the behavior has indeed changed, I'll update or close this PR accordingly.

Thanks for the review!

Transurgeon:
this is a duplicate from #3237

armorbreak001:
Thanks for the heads-up, @Transurgeon! I wasn'\''t aware of #3237 when I opened this. Closing in favor of the earlier PR — no need to review duplicate work. 🙏

armorbreak001:
Closing as duplicate of #3237 (per maintainer guidance)

PR #3294, opened 2026-04-16
Fix constant bmat Hermitian metadata

magic-alt:
## Description

Fixes `cp.bmat(...)` for constant-only block matrices so the result preserves symmetric/Hermitian metadata.

Previously, `bmat` always assembled blocks through `hstack`/`vstack`, which produced a `Vstack` expression even when every block was constant. That dropped the constant matrix metadata used by `quad_form`, so symmetric/Hermitian block matrices built with `cp.bmat` were rejected with:

```text
ValueError: Quadratic form matrices must be symmetric/Hermitian.
```

This change casts blocks up front and constant-folds all-constant block matrices via `np.block(...)`, returning a `Constant` instead of a stacked affine expression. That preserves the correct symmetric/Hermitian properties.

Issue link (if applicable): Fixes #2786

## Type of change
- [x] Bug fix

- [x] Add our license to new files. No new files were added.
- [x] Write unittests.

## Tests
- `python -m pytest cvxpy/tests/test_atoms.py`
- `python -m pytest cvxpy/tests/test_complex.py -k quad_form`

### DIFF
--- a/cvxpy/atoms/affine/bmat.py
+++ b/cvxpy/atoms/affine/bmat.py
@@ -14,8 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import numpy as np
+
+from cvxpy.atoms.affine.affine_atom import AffAtom
 from cvxpy.atoms.affine.hstack import hstack
 from cvxpy.atoms.affine.vstack import vstack
+from cvxpy.expressions.constants import Constant
 
 
 def bmat(block_lists):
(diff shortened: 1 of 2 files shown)


magic-alt:
Follow-up update: I addressed the automated review concerns in the latest push (`2590ec0`).

Changes included:
- Restrict constant-folding in `cp.bmat(...)` to parameter-free blocks so `Parameter`-based block matrices remain parameterized.
- Preserve sparsity for sparse constant blocks by assembling them with `scipy.sparse.bmat(...)` instead of `np.block(...)`.
- Add regression tests covering sparse constant Hermitian structure and parameterized `bmat` behavior.

Local verification:
- `python -m pytest cvxpy/tests/test_atoms.py`
- `python -m pytest cvxpy/tests/test_complex.py -k quad_form`

CLA is now signed as well.

PR #3295, opened 2026-04-16
Map cuOpt UnboundedOrInfeasible status to INFEASIBLE_OR_UNBOUNDED

ramakrishnap-nv:
## Summary

cuOpt added a new termination status \`UnboundedOrInfeasible\` (enum value 11) for both LP and MILP results, returned by its PSLP presolver when the two cases cannot be disambiguated. The existing \`STATUS_MAP_LP\` / \`STATUS_MAP_MIP\` in the CUOPT adapter do not handle it, producing \`KeyError: <LPTerminationStatus.UnboundedOrInfeasible: 11>\` on any unbounded-or-infeasible LP.

Adds the mapping to \`s.INFEASIBLE_OR_UNBOUNDED\` in both tables (matches how HiGHS handles \`kUnboundedOrInfeasible\`).

Tracked upstream in NVIDIA/cuopt#1114.


SteveDiamond (approved):
Thanks for fixing!

Issue #3301, opened 2026-04-19
Use of square Hessian when calling HiGHS results in asymmetry error and native heap corruption


raphaelsaavedra:
**Describe the bug**

On the HiGHS side, this bug is thoroughly described in https://github.com/ERGO-Code/HiGHS/issues/2979.

Note that this happens because cvxpy passes the Hessian in square format to HiGHS. This seems to happen here: 
https://github.com/cvxpy/cvxpy/blob/6e22a84fc19cfcb33b4a6c2803837e9d22ce2b1c/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py#L196

As @odow noted in the issue referenced above, while this can (and possibly should) be solved on HiGHS, it is common practice to use the triangle Hessian to avoid issues like this. I'm wondering if there's good reason to use the square Hessian or if it should be replaced with triangle.

**To Reproduce**

A minimal python reproduction is included in the issue I linked above.

**Expected behavior**

If triangle Hessian is used, the floating-point-level asymmetries don't make it to HiGHS and no error happens.

**Output**

Also reported in the above issue.

**Version**
-  
 - OS: Reproduced on both macOS arm64 (local) and Linux x86_64 (containers).
 - CVXPY Version: 1.7.5, but looking at the code I believe this behavior persists in the latest versions.

odow:
Closed by #3302?

PR #3302, opened 2026-04-19
Use triangular Hessian format when passing QP to HiGHS

raphaelsaavedra:
## Description

Switches the Hessian format used in the HiGHS QP interface from square to triangle. Currently, cvxpy's canonicalization can produce square Hessians where upper and lower triangles have floating-point asymmetries. HiGHS >= 1.14.0 added a strict symmetry check (https://github.com/ERGO-Code/HiGHS/pull/2857) that rejects such matrices, and the subsequent error-handling path causes native heap corruption (SIGABRT/SIGSEGV), killing the Python process. See https://github.com/ERGO-Code/HiGHS/issues/2979 for the full report.

While the heap corruption is ultimately a HiGHS bug, enforcing square Hessian symmetry is not strictly a bug, and using a triangle Hessian is common behavior, so it seems appropriate to make this change.

Issue link (if applicable): #3301 

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
@@ -193,10 +193,15 @@ class HIGHS(QpSolver):
             hessian = model.hessian_
             hessian.dim_ = model.lp_.num_col_
             assert P.format == "csc"
-            hessian.format_ = hp.HessianFormat.kSquare
-            hessian.start_ = P.indptr
-            hessian.index_ = P.indices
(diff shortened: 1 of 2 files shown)


SteveDiamond:
This change is welcome. We primarily rely on the solver maintainers and other contributors to keep their interfaces in good shape.

SteveDiamond (approved):

PTNobel:
I think this PR broke QPs for all HiGHS versions before 1.14.0 since they required lower triangular Hessians in my experiments.

PR #3308, opened 2026-04-29
Fall back to dense eigh in sparse_cholesky for rank-deficient PSD inputs

Transurgeon:
## Description
`sparse_cholesky` (added in #3070, partially patched in #3270) raises                    
`ValueError: Cholesky factorization failed` on simple rank-deficient PSD
inputs such as `[[1, 1], [1, 1]]`. The root cause is that QDLDL has no                   
pivoting and bails out as soon as a zero pivot appears mid-elimination —                 
which the existing rank-deficient handling in #3270 doesn't cover (it                    
only special-cases all-zero rows and zero pivots that land last in                       
QDLDL's elimination order).                                                              
                                                                                         
This adds a dense `numpy.linalg.eigh` fallback inside                                    
`sparse_cholesky`'s existing `except Exception` branch. When QDLDL fails,                
we eigendecompose `A`, classify it as PSD/NSD/indefinite using the same                  
tolerance logic as the QDLDL branch, and return an `n × rank(A)` sparse                  
factor built from the nonzero eigenpairs. The function's documented                      
contract (`L[p,:] @ L[p,:].T == sign * A`, `k = rank(A)`) now holds for                  
all PSD/NSD inputs. 

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_linalg_utils.py
+++ b/cvxpy/tests/test_linalg_utils.py
@@ -136,6 +136,22 @@ class TestSparseCholesky(BaseTest):
         with self.assertRaises(ValueError, msg=lau.SparseCholeskyMessages.INDEFINITE):
             lau.sparse_cholesky(A, 0.0)
 
+    def test_rank_deficient_full_diagonal(self):
+        # [[1,1],[1,1]] is rank-1 PSD with no zero diagonal entries; QDLDL
+        # alone cannot factor it, so this exercises the eigh fallback.
+        A = sp.csc_array(np.array([[1.0, 1.0], [1.0, 1.0]]))
+        sign, L, p = lau.sparse_cholesky(A)
+        self.assertEqual(sign, 1.0)
+        self.assertEqual(L.shape, (2, 1))
+        self.check_gram(L[p, :], A)
+
+    def test_nsd_rank_deficient_full_diagonal(self):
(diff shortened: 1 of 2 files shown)


odow:
This is the same saga as https://github.com/jump-dev/MathOptInterface.jl/issues/2973 ultimately "fixed" https://github.com/jump-dev/MathOptInterface.jl/pull/2967. Even then I don't like it. There's no magic fix. Every option has trade-offs. 

We discussed doing this https://github.com/jump-dev/MathOptInterface.jl/pull/2931 but decided not to merge because of the performance issues for larger models.

PTNobel:
Okay, so I looked at our implementation more; we're using LDL for low-rank factorization in the dense case via Bunch-Kaufman pivoting.

It should be possible to do the same in the sparse case, especially since we don't need symbolic analysis reuse or anything. 

Let me try some experiments and see what we can do...

PTNobel:
Here's a very messy implementation that preserves sparsity and handles low-rank 

https://github.com/cvxpy/cvxpy/pull/3311

PTNobel:
Here's an easy fix: #3313.

I would like to---at some point---use a Bunch-Kaufman-based sparse-LDL factorization so that we can get low-rank support, but maybe that's a different problem.

PTNobel:
#3313 fixes this better.

PR #3313, opened 2026-05-04
Fix sparse_cholesky fallback for rank-deficient PSD/NSD inputs

PTNobel:
## Description
QDLDL is designed for quasi-definite matrices and fails on rank-deficient PSD/NSD inputs in two ways:

1. **Throws "not quasi-definite"** outright — happens for ~18% of random low-rank PSD seeds (e.g. `n=6, k=2`, seeds 3, 16, 22, 24, …) and for benign cases like `[[1,1,0],[1,1,0],[0,0,1]]` (eigenvalues `[0, 1, 2]`). The previous code re-raised this as `Cholesky factorization failed` with no fallback.
2. **Silently returns a wrong factorization** — for some inputs (e.g. seed 128, `n=6, k=2`), QDLDL returns tiny pivots like `D[j]=-3.6e-30` paired with huge L entries (`|L|~7.5e13`) from internal division by those pivots. The previous "low-rank" path masked out columns where `|D[j]| < tol*scale` — but that drops a `D[j]·L[:,j]L[:,j]^T` contribution of magnitude `~0.02`, producing a wrong factorization without warning.

Both modes now fall back to a dense pivoted LDL via `scipy.linalg.ldl` (Bunch-Kaufman). The LDL+2x2-block-resolution logic was already implemented inline in `decomp_quad`; this PR extracts it into a shared `dense_ldl_decomp` helper in `cvxpy.utilities.linalg` that both call sites use.

After QDLDL succeeds, we additionally validate the residual `‖sign·A − L[p,:]L[p,:]^T‖_F` whenever any columns were dropped (rank-deficient case); full-rank PD inputs skip the residual check since the QDLDL identity is exact.

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/quad_form.py
+++ b/cvxpy/atoms/quad_form.py
@@ -17,7 +17,6 @@ limitations under the License.
 
 import numpy as np
 import scipy.sparse as sp
-from scipy import linalg as LA
 
 from cvxpy.atoms.affine.wraps import psd_wrap
 from cvxpy.atoms.atom import Atom
@@ -25,7 +24,7 @@ from cvxpy.expressions.constants.parameter import is_param_affine, is_param_free
 from cvxpy.expressions.expression import Expression
 from cvxpy.interface.matrix_utilities import is_sparse
 from cvxpy.utilities import scopes
-from cvxpy.utilities.linalg import sparse_cholesky
+from cvxpy.utilities.linalg import dense_ldl_decomp, sparse_cholesky
 from cvxpy.utilities.warn import warn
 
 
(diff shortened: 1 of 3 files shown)


rileyjmurray:
The saga never ends 🫠

PTNobel:
@rileyjmurray , this is just a workaround. The ultimate solution is to use a dynamically-pivoted (like Bunch-Kauffman) sparse LDL library that handles low-rank correctly.

My cursory survey suggests that are none with Python bindings and an acceptable license, but that faer does implement it (at least only pivoting within supernodes, which is hopefully good enough) so we could make a very small rust library for it pretty easily...

SteveDiamond (approved):
lgtm

PR #3314, opened 2026-05-07
Extensions CI

maxschaller:
## Description
CVXPYgen now installs CVXPYlayers from sources in its CI.

## Type of change
- [ ] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [x] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #3315, opened 2026-05-07
Add compose_constr_id_map

maxschaller:
## Description
Along with the new ID assignment logic, `Chain.compose_var_id_map` and `Chain.compose_param_id_map` compose the ID mappings for variables and parameters, respectively, end-to-end. However, the composition of constraint IDs was missing, which is needed by, _e.g._, CVXPYlayers. This PR introduces `Chain.compose_constr_id_map`. Test coverage via #3314.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/chain.py
+++ b/cvxpy/reductions/chain.py
@@ -23,6 +23,17 @@ def _compose_id_map(step_maps):
     return result
 
 
+def _extract_constr_step_map(r, inv):
+    """Return the cons_id_map dict for one (reduction, inverse_data) pair."""
+    # CvxAttr2Constr stores (id2new_var, id2old_var, cons_id_map); no-op case returns ().
+    from cvxpy.reductions.cvx_attr2constr import CvxAttr2Constr
+    if isinstance(r, CvxAttr2Constr) and len(inv) == 3:
+        return inv[2]
+    if hasattr(inv, "cons_id_map") and isinstance(inv.cons_id_map, dict):
+        return inv.cons_id_map
+    return {}
+
+
 class Chain(Reduction):
     """A logical grouping of multiple reductions into a single reduction.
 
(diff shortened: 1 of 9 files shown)


cvxpy/reductions/chain.py
```diff
@@ -23,6 +23,17 @@ def _compose_id_map(step_maps):
     return result
 
 
+def _extract_constr_step_map(r, inv):
+    """Return the cons_id_map dict for one (reduction, inverse_data) pair."""
+    # CvxAttr2Constr stores (id2new_var, id2old_var, cons_id_map); no-op case returns ().
+    from cvxpy.reductions.cvx_attr2constr import CvxAttr2Constr
```
PTNobel: Do we need the inline import?

cvxpy/reductions/chain.py
```diff
@@ -23,6 +23,17 @@ def _compose_id_map(step_maps):
...
+    if isinstance(r, CvxAttr2Constr) and len(inv) == 3:
```
PTNobel: Should we just change CA2C to store it on the reduction object? It seems simpler

cvxpy/reductions/chain.py
```diff
@@ -23,6 +23,17 @@ def _compose_id_map(step_maps):
...
+        return inv[2]
+    if hasattr(inv, "cons_id_map") and isinstance(inv.cons_id_map, dict):
```
PTNobel: I'm happy to require reductions to have the map. What do we do for param and var maps?

cvxpy/reductions/chain.py
maxschaller: Obsolete after recent commit.

cvxpy/reductions/chain.py
maxschaller: Agreed!

cvxpy/reductions/chain.py
maxschaller: We require `var_id_map` and `param_id_map` as properties of `Reduction`. The latest commit adds `cons_id_map` as property as well.

PTNobel (approved):

PR #3316, opened 2026-05-07
Robustify scope managers

maxschaller:
## Description
Previously, when using `dpp_scope` or `quad_form_dpp_scope`, any exception under the `with` block would leave the `dpp_scope_active` or `quad_form_dpp_scope_active` flag, respectively, permanently changed. This PR avoids this by wrapping the `yield` statement in `try:` and `finally:`. It also adds `suspend_quad_form_dpp_scope` for suspending the `quad_form_dpp_scope`. This is useful when, _e.g._, using `quad_form_dpp_scope` for checking the objective, but not for the constraints.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/utilities/scopes.py
+++ b/cvxpy/utilities/scopes.py
@@ -40,8 +40,10 @@ def dpp_scope() -> Generator[None, None, None]:
     """
     prev_state = getattr(_thread_local, 'dpp_scope_active', False)
     _thread_local.dpp_scope_active = True
-    yield
-    _thread_local.dpp_scope_active = prev_state
+    try:
+        yield
+    finally:
+        _thread_local.dpp_scope_active = prev_state
 
 
 def dpp_scope_active() -> bool:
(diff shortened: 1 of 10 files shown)


cvxpy/utilities/scopes.py
```diff
@@ -62,10 +64,37 @@ def quad_form_dpp_scope() -> Generator[None, None, None]:
...
+        _thread_local.quad_form_dpp_scope_active = prev
 
 
 def quad_form_dpp_scope_active() -> bool:
     """Returns True if a `quad_form_dpp_scope` is active."""
     return getattr(_thread_local, 'quad_form_dpp_scope_active', False)
+
+
+@contextlib.contextmanager
+def suspend_quad_form_dpp_scope() -> Generator[None, None, None]:
```
PTNobel: What is this going to be used for? Partial minimize problems?

PTNobel:
Maybe a test or two that raise errors would be ideal?

cvxpy/utilities/scopes.py
maxschaller: Removed as per simplifications in https://github.com/cvxpy/cvxpylayers/pull/238.

PTNobel (approved):

PR #3323, opened 2026-05-12
Docs: DNLP to front page

maxschaller:
## Description
Adding info about DNLP to the front page.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -62,13 +62,13 @@ For example, the following code solves a least-squares problem with box constrai
     print(constraints[0].dual_value)
 
 This short script is a basic example of what CVXPY can do. In addition
-to convex programming, CVXPY also supports parametrized programming,
-geometric programming, mixed-integer convex programs, quasiconvex programs,
-and nonlinear programs.
+to convex programming, CVXPY also supports :doc:`parametrized programming </tutorial/dpp/index>`,
+:doc:`geometric programming </tutorial/dgp/index>`, :ref:`mixed-integer convex programs <mip>`,
+:doc:`quasiconvex programming </tutorial/dqcp/index>`,
+and :doc:`nonlinear programming </tutorial/dnlp/index>`.
 
 For a guided tour of CVXPY, check out the :doc:`tutorial
-</tutorial/index>` (including a tutorial on :doc:`nonlinear programming </tutorial/dnlp/index>`).
-For applications to machine learning, control, finance, and
+</tutorial/index>`. For applications to machine learning, control, finance, and
 more, browse the :doc:`library of examples </examples/index>`.
 For applications that involve nonlinear programming, visit the
 `library of DNLP examples <https://github.com/cvxgrp/dnlp-examples/>`_.
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -65,10 +65,27 @@ This short script is a basic example of what CVXPY can do. In addition
 to convex programming, CVXPY also supports a generalization of geometric
 programming, mixed-integer convex programs, and quasiconvex programs.
 
+CVXPY's :doc:`Disciplined Nonlinear Programming </tutorial/dnlp/index>` (DNLP)
+system extends CVXPY to general (nonconvex) nonlinear programs. DNLP lets
+you mix smooth functions (such as :math:`\cos`, products of variables,
+and quadratic forms) with nonsmooth convex and concave atoms, under clear
+:ref:`composition rules <dnlp-atoms-and-expressions>`.
+CVXPY then interfaces with a nonlinear solver such as
+IPOPT or KNITRO. For example, DNLP allows you to have
+``objective = cp.Maximize(cp.sum_squares(A @ x - b))`` or
+``constraints = [cp.multiply(x, (1 - x)) <= 0]`` in the above problem,
+which are not convex, but adhere to DNLP.
+Pass ``nlp=True`` to ``solve()`` to invoke DNLP. Note that, unlike convex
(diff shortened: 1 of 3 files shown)


maxschaller:
@dance858 @Transurgeon Should we add IPOPT and UNO to [install/index.rst](https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst)? Should we mention SparseDiffPy/SparseDiffEngine somewhere?

doc/source/index.rst
```diff
@@ -65,10 +65,27 @@ This short script is a basic example of what CVXPY can do. In addition
 to convex programming, CVXPY also supports a generalization of geometric
 programming, mixed-integer convex programs, and quasiconvex programs.
 
+CVXPY's :doc:`Disciplined Nonlinear Programming </tutorial/dnlp/index>` (DNLP)
```
SteveDiamond: This is more prominently than we normally highlight new features. I'm not sure what would be better exactly but it feels a bit off somehow.

doc/source/index.rst
```diff
@@ -65,10 +65,27 @@ This short script is a basic example of what CVXPY can do. In addition
 to convex programming, CVXPY also supports a generalization of geometric
```
SteveDiamond: This description of cvxpy problem classes needs to change. Maybe mention DNLP here.

doc/source/index.rst
maxschaller: See latest commit.

doc/source/index.rst
maxschaller: See latest commit. How about we just link to every of the classes? We have a tutorial section for each except mixed-integer, I think.

SteveDiamond: Great idea!

doc/source/index.rst
SteveDiamond: Looks good.

doc/source/index.rst
maxschaller: See the latest commit.

dance858:
> @dance858 @Transurgeon Should we add IPOPT and UNO to [install/index.rst](https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst)? Should we mention SparseDiffPy/SparseDiffEngine somewhere?

Good idea! 

Somewhere we should mention what NLP solvers we support (Ipopt, Uno, COPT, and several different algorithms from Knitro).  I'm not sure what the convention is. Do we mention this in the DNLP tutorial?  

I can also add the installation instructions for the NLP solvers to index.rst if that's the right place.

SteveDiamond:
> > @dance858 @Transurgeon Should we add IPOPT and UNO to [install/index.rst](https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst)? Should we mention SparseDiffPy/SparseDiffEngine somewhere?
> 
> Good idea!
> 
> Somewhere we should mention what NLP solvers we support (Ipopt, Uno, COPT, and several different algorithms from Knitro). I'm not sure what the convention is. Do we mention this in the DNLP tutorial?
> 
> I can also add the installation instructions for the NLP solvers to index.rst if that's the right place.

The NLP solvers would go in the big table of solvers https://www.cvxpy.org/tutorial/solvers/index.html#choosing-a-solver
It may be too big with another column though.

Definitely the new solvers should be added to this section https://www.cvxpy.org/install/index.html

maxschaller:
> > > @dance858 @Transurgeon Should we add IPOPT and UNO to [install/index.rst](https://github.com/cvxpy/cvxpy/blob/master/doc/source/install/index.rst)? Should we mention SparseDiffPy/SparseDiffEngine somewhere?
> > 
> > 
> > Good idea!
> > Somewhere we should mention what NLP solvers we support (Ipopt, Uno, COPT, and several different algorithms from Knitro). I'm not sure what the convention is. Do we mention this in the DNLP tutorial?
> > I can also add the installation instructions for the NLP solvers to index.rst if that's the right place.
> 
> The NLP solvers would go in the big table of solvers https://www.cvxpy.org/tutorial/solvers/index.html#choosing-a-solver It may be too big with another column though.
> 
> Definitely the new solvers should be added to this section https://www.cvxpy.org/install/index.html

I added IPOPT and UNO to install/index.rst. It seems that the solver table will become too big with another column. Also, it would be ambiguous whether pure NLP solvers handle, for example, QPs. Technically they do, but it might be misread as solving them with the same guarantees as the convex optimization solvers.

maxschaller:
How about we add a paragraph before or after the table, that just lists the available NLP solvers?

SteveDiamond:
> How about we add a paragraph before or after the table, that just lists the available NLP solvers?

We could add a solver table in the new DNLP section doc/source/tutorial/dnlp/index.rst instead. The solver section is very confusing so I don't want to add more complex content.

SteveDiamond:
Discussed offline with Parth. The discussion of NLP solvers should go in doc/source/tutorial/dnlp/index.rst but a table is not the right format.

maxschaller:
Added a paragraph to doc/source/tutorial/dnlp/index.rst, which lists the supported NLP solvers, with a link to doc/source/install/index.rst for installation instructions.

SteveDiamond (approved):

SteveDiamond:
Thanks for all the fixes!

PTNobel:
> Technically they do, but it might be misread as solving them with the same guarantees as the convex optimization solvers.

Conceptual question, doesn't ipopt guarantee a KKT point if it returns solved? Wouldn't that necessarily have the same guarantees as a convex solver?

PR #3324, opened 2026-05-13
Make solver_path fall back on non-optimal statuses

### LINKED ISSUE #3307: Add option to make solver_path behavior more strict
Currently, when a `solver_path` is provided, the subsequent solvers are tried only if a `SolverError` is encountered.  However, often times a particular solver can fail due to numerical issues.  In these cases, a `SolverError` is not raised.  Rather, the problem status is something other than `OPTIMAL`.

It would be nice if there was an option to make the behavior of `solver_path` more strict so that only solutions with an `OPTIMAL` status are accepted.  If a non-`OPTIMAL` status or `SolverError` is encountered, then we move on to try the next solver.

curious7-web:
## Description
This PR updates `solver_path` fallback behavior so that a solver is only accepted when the resulting problem status is `OPTIMAL`. If a solver returns a non-optimal status, such as `user_limit`, the next solver in the path is tried, similar to the existing fallback behavior for `SolverError`.

Closes #3307.

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -547,7 +547,10 @@ class Problem(u.Canonical):
                             self, *args, solver=solver_name, **solver_kwargs, **kwargs)
                 else:
                     raise ValueError(ENTRY_ERROR_MSG)
-                s.LOGGER.info("Solver %s succeeds", solver_name)
+                if self.status == s.OPTIMAL:
+                    s.LOGGER.info("Solver %s succeeds", solver_name)
+                else:
+                    continue
                 return solution
             except error.SolverError as e:
                 s.LOGGER.info("Solver %s failed: %s", solver_name, e)
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -550,6 +550,10 @@ class Problem(u.Canonical):
                 if self.status == s.OPTIMAL:
                     s.LOGGER.info("Solver %s succeeds", solver_name)
                 else:
+                    s.LOGGER.info("Solver %s returned non-optimal status %s",
+                                  solver_name,
+                                  self.status
+                                  )
                     continue
                 return solution
             except error.SolverError as e:
@@ -1051,7 +1055,7 @@ class Problem(u.Canonical):
                     if bibtex:
                         print(_CITATION_STR)
                         print(CITATION_DICT["CVXPY"])
-                        print(CITATION_DICT["DQCP"])                             
+                        print(CITATION_DICT["DQCP"])
                 reductions = [dqcp2dcp.Dqcp2Dcp()]
                 start = time.time()
                 if type(self.objective) == Maximize:
@@ -1089,9 +1093,9 @@ class Problem(u.Canonical):
 
             # Cite problem grammar.
             if self.is_dcp():
-                print(CITATION_DICT["DCP"]) 
+                print(CITATION_DICT["DCP"])
             if gp:
-                print(CITATION_DICT["DGP"]) 
+                print(CITATION_DICT["DGP"])
 
             # Cite solver.
             print(solving_chain.reductions[-1].cite(data))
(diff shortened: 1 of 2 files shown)


cvxpy/problems/problem.py
```diff
@@ -575,7 +547,10 @@ def _solve_solver_path(self, solve_func, solvers:list[tuple[str, dict] | str],
                             self, *args, solver=solver_name, **solver_kwargs, **kwargs)
                 else:
                     raise ValueError(ENTRY_ERROR_MSG)
-                s.LOGGER.info("Solver %s succeeds", solver_name)
+                if self.status == s.OPTIMAL:
+                    s.LOGGER.info("Solver %s succeeds", solver_name)
+                else:
+                    continue
```
warwickmm, line 557: Would it make sense to add a log entry here with the solver and status?

cvxpy/tests/test_problem.py
```diff
@@ -1542,15 +1540,17 @@ def test_solve_solver_path(self) -> None:
             self.assertIsNotNone(Problem(cp.Minimize(
                 cp.sum_squares(cp.matmul(A, cp.Variable(40)) - b))).solve(
                 solver_path=solvers))
-
+        # valid input, non-optimal first solver falls back to next solver
+        problem = Problem(cp.Minimize(
+            cp.sum_squares(cp.matmul(A, cp.Variable(40)) - b)))
+        problem.solve(solver_path=[(s.OSQP, {'max_iter': 1}), s.CLARABEL])
+        self.assertEqual(problem.status, s.OPTIMAL)
```
warwickmm, line 1548: Should we also check something like
```
self.assertEqual(problem.solver_stats.solver_name, s.CLARABEL)
```

curious7-web:
Thanks, updated. I added a log entry for non-optimal statuses before falling back, and strengthened the test to assert that the fallback reaches CLARABEL.

PTNobel (approved):
Thanks for this!

cvxpy/problems/problem.py
```diff
@@ -575,7 +547,14 @@ def _solve_solver_path(self, solve_func, solvers:list[tuple[str, dict] | str],
                             self, *args, solver=solver_name, **solver_kwargs, **kwargs)
                 else:
                     raise ValueError(ENTRY_ERROR_MSG)
-                s.LOGGER.info("Solver %s succeeds", solver_name)
+                if self.status == s.OPTIMAL:
```
SteveDiamond, line 550: I don't love this condition. Shouldn't it terminate after INFEASIBLE or UNBOUNDED? And it's not clear that it should continue after OPTIMAL_INACCURATE or the other INACCURATE statuses.

curious7-web: Ah yeah, good catch. No point in trying another solver if the problem is already INFEASIBLE or UNBOUNDED. I'll open a follow-up PR to terminate early on those

warwickmm: So, the focus on a strictly OPTIMAL status in #3307 was intentional.  I've experienced cases where a solver would yield an INFEASIBLE status, but the same solver with different options (not directly related to any tolerances) would yield an OPTIMAL status.  Or, larger more complex problems would yield an INACCURATE or INFEASIBLE status when using a free solver (e.g., CLARABEL or OSQP), but result in OPTIMAL solutions when using commercial solvers like CPLEX, GUROBI, or MOSEK.  If all the solvers claim the problem is infeasible, then that status will still be retained.

A user can implement this loop themselves, but I thought that since solver_path was already supported, at least an option to make it strict could be desirable.

SteveDiamond: I see, that's interesting. We can leave it as is and see if we get more feedback.

curious7-web: Yeah, that makes total sense. I hadn't considered the free vs. commercial solver discrepancy on INFEASIBLE flags. I'll hold off on the follow-up PR. Thanks for the context @warwickmm!

PR #3331, opened 2026-05-19
Simplifies version tracking and release procedures

PTNobel:
## Description

I was reworking the CVXPYlayers release procedures, and realized we could generate version numbers from the version control system via `setuptools-scm`; I went ahead and had Claude prototype the change. I think it looks nice and seems like an improvement and figured I'd check what others think.

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


SteveDiamond:
This is a nice idea!

PTNobel:
## ⚠️ Manual step required after merge: create the `dummy-v1.9.0` anchor tag

The CI failures on this PR (`test_derivative`, `test_backends`, `pyright`, etc. — all failing at `uv sync`) come from a tag-reachability problem with the new setuptools-scm setup, and **one manual `git` step is required after this PR merges** for `master` to compute a correct version.

### The problem

`semver-pep440-release-branch` derives the version from the nearest `git describe` tag. But cvxpy's release tags (`v1.7.x`–`v1.9.0`) live on `release/*` branches and are **not ancestors of `master`**. So `git describe` from `master` skips them and falls back to the ancient `v1.1.18`:

| describe config | `master` version |
|---|---|
| `--match v*` (this PR, before fix) | `1.2.0.dev805+g…` ❌ |
| `--match v* --match dummy-v*` (this PR, after fix) + anchor tag | `1.10.0.dev1+g…` ✅ |

A wrong base version (`1.2.0`) also fails `diffcp`'s `cvxpy>=1.6.3` requirement during resolution, which is the proximate cause of the red CI.

### What this PR now does

- `describe_command` also matches `dummy-v*` — an **anchor tag** on `master` that makes the most recent minor release reachable. This is additive: it does **not** move or touch the published `v1.9.0` tag.
- `build.yml` tag builds + the `DEPLOY` gate are restricted to `v*`, so `dummy-v*` tags can never trigger a PyPI deploy.

### Required manual step after merge

Create the anchor tag on the `master` commit `release/1.9.x` was branched from (`eb1fea237`, the parent of `cd95b324c`):

```
git tag -a dummy-v1.9.0 eb1fea237 -m "setuptools-scm anchor: master 1.9 branch point"
git push origin dummy-v1.9.0
```

This adds **no commits** to `master` — it is a ref only. Verified locally: with this tag, `master` HEAD resolves to `1.10.0.dev1+gcd95b324c`, and `release/1.9.x` still resolves to the real `v1.9.0` (it is closer). Going forward, the procedure in `PROCEDURES.md` tags `v1.10.0` directly on `master`, so once 1.10.0 ships no further anchor tags are needed.

PTNobel:
@SteveDiamond can we merge this PR?

rileyjmurray (approved):
LGTM!

(G = *gorgeous*.)

PR #3334, opened 2026-05-20
Update xpress.getversion() to getVersion() for Xpress 9.9

brunoficoxpress:
## Description
`xpress.getversion()` was deprecated in Xpress 9.9 in favour of `xpress.getVersion()`.
Updates `import_solver` to use the new API with a fallback for users on Xpress 9.8 and earlier.

## Type of change
- [x] Bug fix

## Contribution checklist
- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
@@ -64,7 +64,10 @@ class XPRESS(ConicSolver):
         """Imports the solver.
         """
         import xpress
-        self.version = xpress.getversion()
+        try:
+            self.version = xpress.getVersion()
+        except AttributeError:
+            self.version = xpress.getversion()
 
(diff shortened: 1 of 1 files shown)


PTNobel:
When was getVersion added? If it is old enough, it probably isn't worth bothering with the backup.

Also, please always use our PR template.

brunoficoxpress:
> When was getVersion added? If it is old enough, it probably isn't worth bothering with the backup.
> 
> Also, please always use our PR template.

Thanks @PTNobel for the feedback. `getVersion()` was introduced in Xpress 9.9 (just released), so users on 9.8 and earlier will hit AttributeError with the fallback. We'll likely be able to drop `.getversion()` once 9.9 is widely adopted, this is just so users on 9.9 do not get the deprecation warning.

I updated the PR description with the template, and will always use it from now on regardless of how small the change is.

PTNobel (approved):

PR #3335, opened 2026-05-20
Fix MOSEK conic interface: SDP dual KeyError + xfail unsupported bounds test

PTNobel:
## Description
Fixes the `TestMosek` failures that surfaced once #3332 repaired `is_mosek_available()` and un-skipped the suite. All three issues are pre-existing latent bugs/omissions that the skipped suite had been hiding — none is a regression introduced by a recent PR's intent, but two are incomplete follow-through from earlier PRs.

**1. SDP dual recovery dropped for `SvecPSD` constraints (`test_mosek_sdp_1/2/batched/power`).**
#3268 routed PSD constraints through the new `SvecPSD` cone for solvers that request the triangular form — MOSEK does (`SvecPSD` is in `MOSEK.SUPPORTED_CONSTRAINTS`). It updated `ConeDims` to count both `PSD` and `SvecPSD` constraints, but missed `Dualize.invert` in `affine2direct.py`, which mapped PSD-constraint duals from `constr_map[PSD]` only. A dualizing solver receiving `SvecPSD` constraints therefore recovered no PSD dual at all, and `con.dual_value` came back `None`. `Dualize.invert` now iterates `PSD` and `SvecPSD` constraints together (in `ConeDims.psd` order, consuming `num_cones()` barvar blocks per constraint) and returns `SvecPSD` duals in triangular svec form — the raw symmetric entries `PSDToSvecPSD.recover_dual` feeds to `tri_to_full`.

**2. `KeyError` in `ConeMatrixStuffing.invert` when a solver omits a dual.**
#3127 added a second dual-remapping loop after the `SOLUTION_PRESENT` early return that indexed `solution.dual_vars[new_con]` directly, overriding the safe `.get()` loop above it. The two loops are merged into one safe pass: `.get()` + `None`-skip from the original, plus the multi-cone PSD branch from #3127. (Necessary alongside fix 1 so a genuinely-absent dual no longer crashes `invert`.)

**3. `test_mosek_lp_bound_attr` (`KeyError: 'lower_bounds'`).**
`test_lp_bound_attr` asserts a solver consumes variable bounds *natively* (`LOWER_BOUNDS` populated, `dims.nonneg == 0`). MOSEK has `BOUNDED_VARIABLES = False`: its solving chain desugars bounds into nonneg constraints, and its custom `apply()` never populates `LOWER_BOUNDS`. The test was added by #2637, which only wired the MIP/`Slacks` path — never the LP/`Dualize` path, never `BOUNDED_VARIABLES`. Marked `xfail(strict=True)` so it flips to a failure — a loud signal to remove the marker — once MOSEK gains native bounds support.

Result: the full `TestMosek` suite passes (`test_optional_solvers` CI), with `test_mosek_lp_bound_attr` as the single tracked `xfail`.

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.

> Note: MOSEK is not installed/licensed in the dev environment; the `TestMosek` suite is verified by the `test_optional_solvers` CI job, which carries the MOSEK CI license. It passes on both ubuntu-22.04 and macos-14.

### DIFF
--- a/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
+++ b/cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
@@ -469,11 +469,13 @@ class ConeMatrixStuffing(MatrixStuffing):
                 shape = con_obj.shape
                 dual_value = solution.dual_vars.get(new_con)
                 # TODO rationalize Exponential.
-                if dual_value is not None:
-                    if shape == () or isinstance(con_obj, (ExpCone, SOC)):
-                        dual_vars[old_con] = dual_value
-                    else:
-                        dual_vars[old_con] = np.reshape(dual_value, shape, order="F")
+                if dual_value is None:
+                    continue
+                if shape == () or isinstance(con_obj, (ExpCone, SOC)) or \
+                        (isinstance(con_obj, PSD) and con_obj.num_cones() > 1):
+                    dual_vars[old_con] = dual_value
+                else:
+                    dual_vars[old_con] = np.reshape(dual_value, shape, order="F")
 
         primal_vars = {}
         if solution.status not in s.SOLUTION_PRESENT:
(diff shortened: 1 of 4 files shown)


cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -469,11 +469,13 @@ def invert(self, solution, inverse_data):
...
                 # TODO rationalize Exponential.
-                if dual_value is not None:
-                    if shape == () or isinstance(con_obj, (ExpCone, SOC)):
-                        dual_vars[old_con] = dual_value
-                    else:
-                        dual_vars[old_con] = np.reshape(dual_value, shape, order="F")
+                if dual_value is None:
+                    continue
+                if shape == () or isinstance(con_obj, (ExpCone, SOC)) or \
+                        (isinstance(con_obj, PSD) and con_obj.num_cones() > 1):
```
SteveDiamond, line 475: What about SvecPSD objects?

SteveDiamond (approved):
Looks good other than one minor question.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
PTNobel: SvecPSD's dual is canonicalized 1D, it doesn't need to be reshaped unlike batched expcone and soc which need to be reshaped since they're part of the public API.

PR #3338, opened 2026-05-22
Fix COPT PSD bounds with unbounded variables and NLP interface 

PTNobel:
## Description

We currently dualize PSD problems before passing it to COPT, which is a problem when not all variables are bounded because we end up with infinite values in the problem we pass COPT. This is both unsupported on their side, and bad style on ours.

This fixes that. I don't have a COPT license right now so @terryweng03 can you run the tests on this branch?

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/test_nlp_solvers.yml
+++ b/.github/workflows/test_nlp_solvers.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install COPT solver
         run: uv pip install -e ".[COPT]"
       - name: Install Uno solver
-        run: uv pip install unopy
+        run: uv pip install -e ".[UNO]"
       - name: Verify Uno import
         run: uv run python -c "import unopy; print('unopy imported successfully')"
       - name: Install KNITRO solver
--- a/.github/workflows/test_nlp_solvers.yml
+++ b/.github/workflows/test_nlp_solvers.yml
@@ -29,8 +29,10 @@ jobs:
       - name: Install dependencies
         run: |
           uv venv
-          uv pip install -e ".[testing,COPT]"
+          uv pip install -e ".[testing]"
           uv pip install cyipopt
+      - name: Install COPT solver
+        run: uv pip install -e ".[COPT]"
       - name: Install Uno solver
         run: uv pip install unopy
       - name: Verify Uno import
--- a/.github/workflows/test_nlp_solvers.yml
+++ b/.github/workflows/test_nlp_solvers.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv venv
-          uv pip install -e ".[testing]"
+          uv pip install -e ".[testing,COPT]"
           uv pip install cyipopt
       - name: Install Uno solver
         run: uv pip install unopy
--- a/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
@@ -87,8 +87,10 @@ class COPT(ConicSolver):
                   7: s.OPTIMAL_INACCURATE,  # imprecise
                   8: s.USER_LIMIT,          # time out
                   9: s.SOLVER_ERROR,        # unfinished
-                 10: s.USER_LIMIT,          # interrupted
-                 11: s.USER_LIMIT           # iteration limit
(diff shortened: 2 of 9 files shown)


rileyjmurray:
> we end up with infinite values in the problem we pass COPT.

@PTNobel can you elaborate a bit on what happened?

PTNobel:
@rileyjmurray we were adding Ax<= ub (and the corresponding lower bound) and then dualizing. The issue is ub contained infinite values, so after dualization we ended up with an objective function containing infinite coefficients.

rileyjmurray:
Ohhhh yikes. Got it. Glad they were able to reach out and you were able to fix quickly.

PTNobel:
I reenabled COPT in CI!

PTNobel:
> model.loadConeMatrix and model.loadMatrix already supports scipy.sparse.csc_array as the input type for argument A.

From Terry in the other thread. I'll fix this before merge

.github/workflows/test_nlp_solvers.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv venv
-          uv pip install -e ".[testing]"
+          uv pip install -e ".[testing,COPT]"
```
rileyjmurray: seems like an awkward hack

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
```diff
@@ -16,6 +16,46 @@
...
+        finite_lb = np.isfinite(lb)
+        n_lb = int(np.count_nonzero(finite_lb))
+        if n_lb:
+            extra_rows.append(_bound_selector(n_lb, n, np.flatnonzero(finite_lb), -1.0))
+            extra_b.append(-lb[finite_lb])
+
+    if extra_rows:
+        insert_at = dims[s.EQ_DIM] + dims[s.LEQ_DIM]
+        bound_block = sp.vstack(extra_rows, format='csc')
+        A = sp.vstack([A[:insert_at], bound_block, A[insert_at:]], format='csc')
```
rileyjmurray: If we're stacking vertically then this will almost certainly be converted to CSR internally before a subsequent conversion to CSC. `bound_block` should be CSR.

rileyjmurray (commented):
Minor comments. Technically these are requested changes but I don't want github to have to wait for me after you'd change things.

.github/workflows/test_nlp_solvers.yml
PTNobel: Moved COPT into its own install step (3a1864afd), matching the cyipopt and Uno steps.

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
PTNobel: Good catch — `bound_block` (and `_bound_selector`) now build CSR in 3a1864afd, so the outer vstack no longer pays a CSC->CSR->CSC round trip.

cvxpy/reductions/solvers/nlp_solvers/copt_nlpif.py
```diff
@@ -256,6 +258,15 @@ def EvalHess(self, xdata, sigma, lambdata, outdata):
         for key, value in solver_opts.items():
             model.setParam(key, value)
 
+        # COPT may evaluate the gradient/Jacobian/Hessian callbacks at the
```
dance858, line 259: looks great!

SteveDiamond (commented):
This looks good in terms of code changes, but the test_qp_solvers part of test_optional_solvers isn't working anymore after this PR.

PTNobel:
@terryweng03 it looks like AVX-512 instructions are being run on CPUs without AVX-512 in CI and we're getting a SIGILL.

PTNobel:
I can't reproduce the AVX-512 issue with Qemu. I'll make a real minimizer and report back.

PTNobel:
I'm giving up on reproducgin the SIGILL. It seems to be very rare.

PTNobel:
> This looks good in terms of code changes, but the test_qp_solvers part of test_optional_solvers isn't working anymore after this PR.

Seems to just be a mac issue, I'll investigate.

rileyjmurray (approved):

PR #3340, opened 2026-05-22
Add KNITRO to the nlp solvers CI job

PTNobel:
## Description
Adds KNITRO to the `test_nlp_solvers` CI job so the KNITRO-gated tests under `cvxpy/tests/nlp_tests/` run instead of being skipped, and fixes the issues that surfaced from doing so.

### CI
- Install the `knitro` package and write the CI license from the existing `KNITRO_LICENSE` secret (guarded so forks without the secret skip cleanly).
- Drop `-s` from the nlp `pytest` invocation. `-s` disabled output capture, flooding the log with ~35k lines per OS leg of IPOPT/diff-engine output; without it pytest only replays output for failing tests.
- Run KNITRO nlp tests in a **separate pytest process** (`-m knitro`) from the rest (`-m "not knitro"`).

### Native-library fix
`import cvxpy` eagerly built `INSTALLED_SOLVERS` by importing every solver, which loaded the native KNITRO runtime — and on macOS its bundled `libomp.dylib` — into every process. With a second OpenMP runtime present, an IPOPT solve segfaults on macOS.

- `KNITRO` (conic/qp/nlp) and `IPOPT` solver classes now override `is_installed()` to detect their package with `importlib.util.find_spec()` instead of importing it. The native library loads only on an actual solve.
- `is_knitro_available()` in `test_conic_solvers.py` now detects KNITRO via env vars instead of `import knitro`, so collecting the nlp suite no longer loads the native library. `test_interfaces.py`'s duplicate definition is removed in favour of the shared one.
- `test_optional_solvers.yml` exports `ARTELYS_LICENSE` when the license secret is present, since `is_knitro_available()` now gates on it.

Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/test_nlp_solvers.yml
+++ b/.github/workflows/test_nlp_solvers.yml
@@ -35,7 +35,17 @@ jobs:
         run: uv pip install unopy
       - name: Verify Uno import
         run: uv run python -c "import unopy; print('unopy imported successfully')"
+      - name: Install KNITRO solver
+        run: uv pip install knitro
+      - name: Set up KNITRO license
+        if: env.KNITRO_LICENSE != ''
+        run: |
+          echo "$KNITRO_LICENSE" > ~/artelys_lic.txt
+          echo "ARTELYS_LICENSE=$HOME" >> $GITHUB_ENV
       - name: Print installed solvers
         run: uv run python -c "import cvxpy; print(cvxpy.installed_solvers())"
       - name: Run nlp tests
         run: uv run pytest -s -v cvxpy/tests/nlp_tests/
+
+    env:
+      KNITRO_LICENSE: ${{ secrets.KNITRO_LICENSE }}
--- a/.github/workflows/test_nlp_solvers.yml
+++ b/.github/workflows/test_nlp_solvers.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Print installed solvers
         run: uv run python -c "import cvxpy; print(cvxpy.installed_solvers())"
       - name: Run nlp tests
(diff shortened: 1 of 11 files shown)


PTNobel:
There is way too much output in the NLP tests; I'm going to try to disable verbose mode?

PTNobel:
Hmm, seems to be a segfault on macOS

PTNobel:
@dance858 any idea why we have a segfault here

PTNobel:
Hmm there are duplicate openmp symbols

PTNobel:
That worked!

dance858:
Did everything work out or should I take a look at something?

On a different note, the change you made in the commit “Scope the UNO skips to the UNO parametrization” was fixed on the `cvxpy-sync-1.9` branch in DNLP. At first I was a bit confused since those changes didn’t seem to be present here.

Edit: never mind about my comment. The fix was added on the `cvxpy-sync-1.9` branch, but it looks like it was reverted before the merge or during a later change. In any case, the code is there now which is what matters. Thanks!

PTNobel:
I figured it out; it looks to be a libomp symbol duplication between Knitro and Ipopt. Can you do a code review of these changes?

dance858:
> I figured it out; it looks to be a libomp symbol duplication between Knitro and Ipopt. Can you do a code review of these changes?

Great!

I'm not an expert on environment variables so unfortunately, I can't comment on those changes. The "more normal DNLP changes" (bounds in nlp_solver.py and Uno guards in the tests) look good though. Thanks for fixing this.

SteveDiamond (approved):
LGTM

PR #3341, opened 2026-05-23
Isolate KNITRO conic tests and warn on OpenMP-runtime conflicts

PTNobel:
## Description
Fixes the `test_optional_solvers (macos-14)` job which began segfaulting after #3340. Also adds a runtime warning when two OpenMP-bundling solvers are about to coexist in one process.

### Root cause
Among the ~20 wheels installed by `uv sync --all-extras --dev`, exactly two ship an OpenMP runtime in their macOS arm64 wheel:

| Wheel | OpenMP runtime | Where |
|---|---|---|
| `knitro` | LLVM `libomp.dylib` | `knitro/lib/libomp.dylib` |
| `cvxopt` | GNU `libgomp.1.dylib` | `cvxopt/.dylibs/libgomp.1.dylib` (pulled in by `libopenblasp-r0.3.24.dylib`, which every cvxopt extension links) |

LLVM libomp and GNU libgomp were never designed to coexist in one process.

Pre-#3340, `TestKNITRO`'s `@unittest.skipUnless(is_knitro_available(), ...)` decorator did `import knitro; KN_new(); KN_free()` at pytest **collection time**, so KNITRO's libomp was the first OpenMP loaded. CVXOPT/GLPK tests then loaded libgomp second, and the suite finished.

After #3340 the order reverses: `is_knitro_available()` now only reads env vars and runs `find_spec`, so cvxopt loads libgomp first (early in the test run), and when `TestKNITRO::test_knitro_lp_0` finally executes at ~76% of the suite, KNITRO loads libomp second and `KN_solve` segfaults on its first parallel section.

### Changes
1. **CI split.** `TestKNITRO` in `test_conic_solvers.py` gets `@pytest.mark.knitro`, and `test_optional_solvers.yml` runs `-m "not knitro"` and `-m knitro` as separate pytest processes for `test_conic_solvers.py`. Mirrors the existing pattern in `test_nlp_solvers.yml`.
2. **Runtime warning.** New `cvxpy/reductions/solvers/openmp_conflict.py` exposes `warn_if_omp_conflict(importing)`. Called from `import_solver()` on the KNITRO (conic / QP / NLP), IPOPT, and CVXOPT solver classes. Detects the conflict via `sys.modules` (does not import anything itself) and emits a `RuntimeWarning` naming the conflicting packages and runtimes.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/.github/workflows/test_optional_solvers.yml
+++ b/.github/workflows/test_optional_solvers.yml
@@ -50,8 +50,15 @@ jobs:
         run: uv pip install moreau --extra-index-url https://${FURY_TOKEN}:@pypi.fury.io/optimalintellect/
       - name: Print installed solvers
         run: uv run python -c "import cvxpy; print(cvxpy.installed_solvers())"
+      # KNITRO bundles LLVM libomp; CVXOPT bundles GNU libgomp via its
+      # OpenBLAS dependency. Loading both in one process can crash inside
+      # KNITRO's KN_solve on macOS, so run KNITRO tests in a separate
+      # process (mirrors test_nlp_solvers.yml).
       - name: Run test_conic_solvers
-        run: uv run pytest -rs cvxpy/tests/test_conic_solvers.py
+        run: uv run pytest -rs -m "not knitro" cvxpy/tests/test_conic_solvers.py
+      - name: Run test_conic_solvers (KNITRO)
+        if: always()
+        run: uv run pytest -rs -m knitro cvxpy/tests/test_conic_solvers.py
       - name: Run test_qp_solvers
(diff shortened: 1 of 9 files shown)


Transurgeon (approved):
wow nice bug fix!
I am wondering why we missed this? Definitely a mishap on our part, but I presume its because we didn't test much the commercial solvers in the CI.
This only fails for macOS also right?

PTNobel:
1. It's also order dependent; so if you load knitro first it can work
2. Yes only MacOS since it doesn't have an OMP impl by default

Issue #3310, opened 2026-05-01
Multiple attributes for a variable not handled correctly


EducatedGuest:
**Describe the bug**
Multiple attributes for a variable are not handled correctly. The attributes are `complex=True` and `PSD=True`.

**To Reproduce**
```
import cvxpy as cp

X = cp.Variable((2, 2), complex=True, PSD=True)
obj = cp.Minimize(cp.real(cp.trace(X)))

p1 = cp.Problem(obj)
print("No PSD:", p1.solve(), p1.status)

p2 = cp.Problem(obj, [X >> 0])
print("With PSD:", p2.solve(), p2.status)
```

**Expected behavior**
The problems should be identical and lead to the same optimal value, that is, obj.value = 0.

**Output**

> No PSD: -inf unbounded
> With PSD: 1.4030706107720334e-12 optimal

**Version**
 - CVXPY Version: 1.8.2, 1.9.0.dev0+232.4cdf938, 1.9.0.dev0+248.100fda0

**Additional context**
On the master branch (1.9.0.dev0+248.100fda0):

> Solving p1 leads to -inf
> Solving p2 results in an error, KeyError: 148

On version 1.9.0.dev0+232.4cdf938:

> No PSD: -inf unbounded
> With PSD: 1.4030706107720334e-12 optimal

This is a minimal example extracted from a higher-dimensional problem where the same issue occurs.

curious7-web:
I've started tracing the canonicalization pipeline for this. It looks like the PSD=True attribute is getting misrouted or dropped during compilation when complex=True is also present, which is triggering the KeyError: 148 on the latest master. Digging into the reductions module and aiming to have a PR up for this.

PTNobel:
Hmm... The bug is probably in complex2real when we construct a new variable. We need to add a new constraint. I suspect sparsity and maybe some other attributes would also be broken... Thanks for catching this! @curious7-web any ETA on the PR? I'm hoping to release 1.9.1 next week

PTNobel:
It's this file: https://github.com/cvxpy/cvxpy/blob/master/cvxpy/reductions/complex2real/canonicalizers/variable_canon.py

 we need cases for every attribute I think...

curious7-web:
[...]

PR #3342, opened 2026-05-23
Pass attributes to real variables in Complex2Real canonicalizer

curious7-web:
## Description
This draft PR explores patching `variable_canon.py` directly to fix the issue where attributes (like `PSD=True`) get dropped when complex variables are split into real and imaginary parts.

**Context:** As discussed in the issue #3310, promoting `CvxAttr2Constr` higher up in the solving chain catches the `PSD` attribute but triggers 126 downstream regressions (mostly in DGP and perspective functions). 

**Current State:** This targeted patch copies the attributes directly. It catches `PSD`, solves the test case perfectly, and brings regressions down to just 1 failing test (`test_qp_solvers.py`). 

**The Blocker:** Blindly passing the `attributes` dictionary forces the `imag` variable to inherit properties like `PSD` or `hermitian`. This makes the imaginary part symmetric, when mathematically it must be skew-symmetric, which over-constrains it to zero and breaks that QP test.

Opening this as a draft so we can figure out the cleanest way to map specific attributes (like `symmetric` vs `skew_symmetric`) to the real/imaginary components without making the canonicalizer a mess.

Issue link: [#3310 ]

## Type of change
- [x] Bug fix

## Contribution checklist
- [ ] Write unittests.
- [x] Run the unittests and check that they’re passing. *(Note: 1 known failure documented above).*


SteveDiamond:
I like the more targeted bug fix approach here. It's also good that the Complex2Real reduction can stand on its own and not require assumptions about what attributes variables have. It may be the case that determining the attributes of the real and imaginary parts requires a lot of case specific logic that analyzes the attributes of the original complex variable. That's ok as long as the logic is clear and separated out well into it's own helper function.

SteveDiamond:
Can you add a bunch of test problems that break with the non-patched complex2real reduction?

curious7-web:
Updated the PR (Ready for review whenever you have time. cc @SteveDiamond @PTNobel ✌️)

The main change is that `complex=True, PSD=True` and `complex=True, NSD=True` are now handled explicitly during Complex2Real canonicalization. The variable attributes are split more carefully in `variable_canon.py`, and Complex2Real now adds the corresponding real block PSD representation for complex PSD/NSD variables:

```text
[ Re(X)   Im(X)  ]
[-Im(X)   Re(X)  ] >> 0
```

for PSD, and the negated version for NSD.

I also made sure this block expansion is only applied to complex `Variable` leaves, not parameters/constants or purely real leaves. The added regression tests check the user-facing behavior directly by comparing the attribute form against the equivalent explicit cone-constraint form, e.g. `complex=True, PSD=True` versus `[X >> 0]`, on bounded problems.

PTNobel:
Can we add errors for invalid attributes like nonneg+complex?

PTNobel:
Can we pull the transform for complex+PSD into a helper? There should be another implementation of it to handle explicit X>>0 constraints

curious7-web:
Thanks @PTNobel, updated.

I moved the complex PSD block construction into `psd_canon.py` and reused it from the attribute path, so the `complex=True, PSD=True` case now goes through the same block-matrix transform as explicit `X >> 0`.

I also added validation tests for invalid complex or imaginary sign attributes like `nonneg`, `nonpos`, `pos`, and `neg`.

cvxpy/reductions/complex2real/canonicalizers/psd_canon.py
```diff
@@ -19,14 +19,18 @@
 from cvxpy.atoms import bmat
 
 
+def complex_psd_matrix(real_part, imag_part):
+    """Return the real block representation of a complex PSD expression."""
+    if imag_part is None:
+        return real_part
+    if real_part is None:
+        real_part = np.zeros(imag_part.shape)
```
PTNobel: Could we use a sparse arrays here? I guess maybe it doesn't matter, but feels wasteful to have a huge dense array and I think our bmat supports sparse arrays

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
```diff
@@ -14,32 +14,71 @@
...
     elif expr.is_imag():
-        # Purely imaginary
+        # Purely imaginary.
         imag = Variable(expr.shape, var_id=real2imag[expr.id])
         return None, imag
+
     elif expr.is_complex() and expr.is_hermitian():
         n = expr.shape[0]
-        real = Variable((n, n), var_id=expr.id, symmetric=True)
+        real_attr, imag_attr = _split_complex_attributes(expr.attributes)
```
PTNobel: Where is imag_attr used? I feel like it isn't threaded through

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
```diff
@@ -14,32 +14,71 @@
...
+    real_attr = attributes.copy()
+    imag_attr = attributes.copy()
+
+    for attr in ['complex', 'hermitian']:
+        real_attr.pop(attr, None)
+        imag_attr.pop(attr, None)
+
+    if attributes.get('hermitian') or attributes.get('PSD') or attributes.get('NSD'):
+        imag_attr.pop('symmetric', None)
+        imag_attr['skew_symmetric'] = True
```
PTNobel: Is this a new attribute

curious7-web:
Thanks, updated. You were right about `skew_symmetric`; it should not be treated like a Variable attribute, so I removed that path and kept the Hermitian imaginary part built through `skew_symmetric_wrap(...)`.

I also cleaned up the unused `imag_attr` bit and switched the zero block in the PSD transform to sparse.

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
PTNobel: Dropping it explictly isn't correct, right? For example sparsity should be setting some entries to 0, but we're now allowing the imaginery part to be arbitrary.

PTNobel:
> I also cleaned up the unused imag_attr bit 

I do not think you did this correctly; and I am concerned we don't have enough tests.

Are we correctly handling sparsity + complex and sparsity + imaginary etc? Do we have tests for these?

curious7-web:
Good point, I missed that case.

I checked the allowed combinations more directly. `complex=True` with `sparsity` and `imag=True` with `sparsity` are valid, while `hermitian=True` with `sparsity` and `complex=True, PSD=True` with `sparsity` already raise from the leaf attribute checks.

I added tests for the valid cases: complex+sparsity, imag+sparsity, and imag+diag. I also updated the pure imaginary branch to preserve valid split attributes instead of dropping them.

cvxpy/tests/test_complex.py
```diff
@@ -171,6 +171,55 @@ def test_imag_sign_attributes_raise(self) -> None:
...
+        real, imag = variable_canon(X, None, None, real2imag)
+
+        self.assertEqual(real.attributes.get("sparsity"), sparsity)
+        self.assertEqual(imag.attributes.get("sparsity"), sparsity)
+        self.assertFalse(real.attributes.get("complex"))
+        self.assertFalse(imag.attributes.get("complex"))
+
+    def test_imag_sparsity_attribute_preservation(self) -> None:
+        """Test sparsity attribute is preserved when splitting an imaginary variable."""
+        import cvxpy.lin_ops.lin_utils as lu
```
PTNobel: Imports belong at the top.

curious7-web: Thanks, fixed. I moved the new test imports to module scope.

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
```diff
@@ -14,32 +14,66 @@
...
-        imag = Variable(expr.shape, var_id=real2imag[expr.id])
+        # Purely imaginary.
+        _, imag_attr = _split_complex_attributes(expr.attributes)
+        imag = Variable(expr.shape, var_id=real2imag[expr.id], **imag_attr)
         return None, imag
+
     elif expr.is_complex() and expr.is_hermitian():
         n = expr.shape[0]
-        real = Variable((n, n), var_id=expr.id, symmetric=True)
+        real_attr, _ = _split_complex_attributes(expr.attributes)
```
PTNobel, line 69: Why is throwing out the imag_attrs valid here?

PTNobel: At the very least the answer is non-obvious and we should comment it. I'm worried that new attributes will trigger the same bugs.

PTNobel (commented):
I still don't understand how this PR works. I'm not sure what next steps are, but I flagged my biggest point of confusion.

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
curious7-web: Thanks for pointing out the confusing part. The PR definitely got harder to follow after a few iterations.

The current split is:

* General complex / purely imaginary variables split into full-size real and imaginary `Variable`s, so attrs like `diag` and `sparsity` can be preserved there.
* Hermitian variables are different: the imaginary part is represented by a compact strict-upper-triangular vector and expanded with `skew_symmetric_wrap(...)`, so `imag_attr` is not forwarded in that branch.
* Complex PSD/NSD variables use the same block-real PSD transform as explicit complex PSD constraints.

I added a docstring and inline comment for the Hermitian branch to make that part clearer.

curious7-web:
One extra thing I checked locally: the structural Hermitian combinations that would make `imag_attr` ambiguous are already rejected before Complex2Real runs.

For example, `hermitian=True` with `diag`, `sparsity`, `PSD`, `NSD`, or `symmetric` raises during leaf attribute validation. For the plain Hermitian case, `variable_canon` gives the real part a full matrix `Variable` with `symmetric=True`, while the imaginary part is a `skew_symmetric_wrap` expression.

So I think the Hermitian branch is safe to avoid forwarding `imag_attr`, but I added the comment because that reasoning was not obvious from the code.

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
PTNobel: Maybe I just need to look at the list of attributes and understand if any of them should carry over.

cvxpy/expressions/leaf.py
```diff
@@ -159,6 +159,15 @@ def __init__(
                 "A CVXPY Variable cannot have more than one of the following attributes: "
                 f"{dim_reducing_attr}"
             )
+        if self.attributes['complex'] or self.attributes['imag']:
+            sign_attrs = [
+                k for k in ['nonneg', 'nonpos', 'pos', 'neg']
```
PTNobel: What about bounds?

curious7-web: Good point. I checked this locally.

For `hermitian=True, bounds=[0, 1]`, the bounds are forwarded to the real split variable. The imaginary part is still represented as a `skew_symmetric_wrap` expression rather than a full `Variable`, so there is no `imag_attr`/bounds path to forward there.

For `complex=True` or `imag=True` with bounds, I agree that seems closer to the sign-attribute validation question. Should I add bounds to the invalid complex/imag attribute validation in this PR, or keep this PR focused on the PSD/attribute-routing fix and handle bounds separately?

PTNobel: I think bounds are like nonneg and make no sense for complex/imag variables. I'm quite worried this PR has not covered all attributes and isn't building infrastructure to keep the bug from happening again. 

I have no idea how we'd do that tbh , but I think we need to think about it

PTNobel: Maybe a well placed comment where attributes are defined that points to the attribute handling in the complex codepath or maybe an allow list for attributes that can be shared with complex/imag? I'm open to ideas

curious7-web: Updated this to make the attribute routing opt-in instead of copy-then-remove.

`_split_complex_attributes` now uses an allowlist (`diag`, `sparsity`) so new attributes have to be added deliberately after checking their Complex2Real semantics. PSD/NSD are handled explicitly on the real side only. I also added `bounds` to the invalid attribute validation for `complex` and `imag` variables, same as the sign attrs.

The focused complex/attribute tests pass locally.

cvxpy/expressions/leaf.py
```diff
@@ -162,6 +162,19 @@ def __init__(
                 "A CVXPY Variable cannot have more than one of the following attributes: "
                 f"{dim_reducing_attr}"
             )
+        if self.attributes['complex'] or self.attributes['imag']:
+            sign_attrs = [
```
PTNobel: Can we use set intersection? I think it's more readable

PTNobel (commented):
30s glance, I'm going to need to spend a bit more time with this PR before I'm ready to approve. I think it is way closer tho and really appreciate the iterations.

cvxpy/expressions/leaf.py
curious7-web: Thanks, updated this to use set intersection.

cvxpy/reductions/complex2real/complex2real.py
```diff
@@ -275,6 +279,30 @@ def apply(self, problem):
...
+            real_part, imag_part = canon_expr
+
+            if real_part is None or imag_part is None:
+                continue
+
+            block_mat = complex_psd_matrix(real_part, imag_part)
+
+            if attrs.get("PSD"):
+                constrs.append(PSD(block_mat))
+            else:
```
PTNobel: Can we make this an explicit if NSD

cvxpy/reductions/complex2real/canonicalizers/variable_canon.py
PTNobel: Aren't both sparsity and diagonal legal to put on a hermetian matrix?

PTNobel (commented):
I think you're mishandling the combo of hermitian (but not definite) and sparsity or diagonal.

At the very least, I don't see any tests for that case.

PTNobel:
Sorry for how long it took me to get back to this!

curious7-web:
Verified locally. Both raise at construction:

hermitian+sparsity → ValueError: A CVXPY Variable cannot have more than one of the following attributes: ['diag', 'symmetric', 'PSD', 'NSD', 'hermitian', 'sparsity']
hermitian+diag → same error

They're mutually exclusive via the dim_reducing_attr check in leaf.py, so that combination can't reach the Hermitian branch in variable_canon.py. I'll add explicit regression tests for both to make this documented instead of implicit.

curious7-web:
Thanks, this should address both remaining points: I added explicit regression tests for the `hermitian=True` + `sparsity` and `hermitian=True` + `diag=True` cases, both of which are rejected during leaf attribute validation, so they never reach the Hermitian Complex2Real branch; I also updated the PSD/NSD handling to use an explicit `elif attrs.get("NSD")` instead of falling through `else`. Ready for another look whenever you have time.

PTNobel (approved):
Great work! Thanks for putting up with all my repeated questions :)

PR #3346, opened 2026-05-26
Apply MOSEK/KNITRO fixes to test_qp_solvers, rewrite as pytest

PTNobel:
## Description

`test_qp_solvers.py` had three issues that were already fixed in `test_conic_solvers.py` (#3332, #3341) but never propagated. This PR mirrors those fixes and rewrites the file as plain pytest while it's open.

### Bugfixes

1. **`is_mosek_available()`** used `env.getlicense()`. In MOSEK 11 that method returns a license token, not a rescode, so the comparison spuriously failed and MOSEK QP tests were silently skipped. Now uses a trivial solve, matching `is_mosek_available()` in `test_conic_solvers.py`. To avoid duplication, `is_mosek_available` and `is_knitro_available` are imported from `test_conic_solvers`.

2. **`is_knitro_available()`** eagerly imported `knitro` and called `KN_new`. On macOS that loads the native KNITRO runtime and its bundled LLVM `libomp.dylib` into the test process at collection time — the exact problem #3341 fixed for the conic suite. Detection now reads `ARTELYS_LICENSE` / `ARTELYS_LICENSE_NETWORK_ADDR`.

3. **KNITRO QP tests** were not marked `@pytest.mark.knitro` and not run in a separate pytest invocation. Now `@pytest.mark.knitro` lives on every KNITRO-bearing `pytest.param`, and `test_optional_solvers.yml` runs `test_qp_solvers.py` in two passes (`-m "not knitro"` then `-m knitro`), matching the existing split for `test_conic_solvers.py`.

### Rewrite

The file drops `unittest` entirely. Old structure: `BaseTest` subclasses (`QPTestBase` / `TestQp` / `TestConicQuadObj`) with ~20 helper methods called from two giant `test_all_solvers` loops. New structure:

- Each QP correctness check is its own pytest test (e.g. `test_quad_form`), parametrized over a single `QP_PARAMS` list of `(solver, use_quad_obj)` pairs.
- `QP_PARAMS` covers every native QP solver with `use_quad_obj=False` plus every conic solver that supports quadratic objectives with `use_quad_obj=True`. KNITRO is excluded from the conic-side because its conic + `use_quad_obj=True` path is unstable in CI.
- License-driven skips (MOSEK, XPRESS) and the `knitro` mark are attached per-`pytest.param`. No more class-level decorators or runtime `filter_licensed_solvers` calls.
- Five identical warm-start tests collapse to one parametrized `test_warm_start`. The four infeasibility tests collapse to two. (XPRESS and Gurobi warm-start tests stay separate — they exercise genuinely different code paths.)
- Custom `assertItemsAlmostEqual` / `assertAlmostEqual` helpers are gone; tests use `np.testing.assert_allclose` directly.

A `test_quad_form` failure for CLARABEL in the conic + `use_quad_obj` path now shows up as `test_quad_form[CLARABEL-quadobj]` instead of being buried inside a `test_all_solvers` loop.

Issue link (if applicable): N/A — follow-up cleanup noted on the 1.9.1 backport before tagging.

## Type of change

- [x] Bug fix

- [ ] Write unittests.

Locally (no MOSEK/KNITRO licenses available):

```
$ uv run pytest -rs -m "not knitro" cvxpy/tests/test_qp_solvers.py
======================== 94 passed, 15 skipped in 1.59s ========================
```

### DIFF
--- a/.github/workflows/test_optional_solvers.yml
+++ b/.github/workflows/test_optional_solvers.yml
@@ -64,7 +64,10 @@ jobs:
         run: uv run pytest -rs -m knitro cvxpy/tests/test_conic_solvers.py
       - name: Run test_qp_solvers
         if: always()
-        run: uv run pytest -rs cvxpy/tests/test_qp_solvers.py
+        run: uv run pytest -rs -m "not knitro" cvxpy/tests/test_qp_solvers.py
+      - name: Run test_qp_solvers (KNITRO)
+        if: always()
+        run: uv run pytest -rs -m knitro cvxpy/tests/test_qp_solvers.py
 
     env:
       RUNNER_OS: ${{ matrix.os }}
--- a/cvxpy/tests/test_qp_solvers.py
+++ b/cvxpy/tests/test_qp_solvers.py
@@ -14,9 +14,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import os
 import unittest
 
 import numpy as np
+import pytest
 import scipy.sparse as sp
 from scipy.linalg import lstsq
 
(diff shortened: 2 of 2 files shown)


PTNobel:
I ended up spending an hour today chasing a big in OSQP that was actually in HiGHZ because of the single monolithic test QP architecture that we have right now and I had missed that the failure was after the point I thought it was.

This PR will be a huge improvement next time we have a QP solver bug

.github/workflows/test_optional_solvers.yml
```diff
@@ -61,7 +61,10 @@ jobs:
         run: uv run pytest -rs -m knitro cvxpy/tests/test_conic_solvers.py
       - name: Run test_qp_solvers
         if: always()
-        run: uv run pytest -rs cvxpy/tests/test_qp_solvers.py
+        run: uv run pytest -rs -m "not knitro" cvxpy/tests/test_qp_solvers.py
+      - name: Run test_qp_solvers (KNITRO)
```
Transurgeon, line 68: why are we separating and special casing knitro? I didn't quite catch this.
Otherwise the PR looks good.

PTNobel: Knitro ships its own copy of OpenMP on MacOS so it cannot be loaded in the same process as any other solver that uses OMP. This change brings QPs in line with NLP and conic tests, tho it's not strictly necessary since I don't think we have a second QP solver using OMP right now.

Transurgeon: oh I see, its the same issue you mentioned earlier but now its for QP solvers. Looks good then.

Transurgeon (approved):

Issue #2205, opened 2023-08-07
Slow compiling time for an unconstrained quadratic problem 


MichaRandomName:
Dear CVXPY Team,

I have posted this question already on [Stackoverflow](https://stackoverflow.com/questions/76789288/cvxpy-shows-warning-objective-contains-too-many-subexpressions-and-takes-a-lon), but it seems that nobody could help me, so I'll give it a shot here in a more concise and altered version.

I have tried to solve a simple unconstrained quadratic problem with CVXPY, which solves the problem as expected, as long as the matrix size is small enough (adjustable by the parameters `N_r, N_t` and `N_s`). The following code shows a simplified version of my optimization problem, which takes about 12 seconds to compile and solve.  

```python
import numpy as np
import cvxpy as cp
from scipy.linalg import dft
import time as time

start_time = time.time()

N_r = 8
N_t = 1
N_s = 10

x = np.random.randint(2, size = N_s*N_r*N_t)

H = dft(N_s*N_r*N_t)*1j
H_H = H.conj().T

err = np.random.random(N_r)
Err = np.kron(np.diag(np.ones(N_s*N_t)), np.diag(err))

y = H_H@Err@H@x

var = cp.Variable(shape = (N_r), complex = True)
Err_est = cp.kron(np.diag(np.ones(N_s*N_t)), cp.diag(var))

res = cp.sum_squares(H_H@Err_est@H@x - y)

prob = cp.Problem(cp.Minimize(res))
prob.solve(solver = cp.CVXOPT, verbose = True)

print(f'Result: {np.abs(err-var.value)}')
print(f'\nProcessing duration: {time.time()-start_time:.2f} seconds')
``` 
Changing the size of matrix `Err` and `Err_est` by setting `N_t = 2`, increases the compiling and consequently the execution time to 200 seconds, where the solver time in both cases remains at 30 ms. This code also invokes the warnings: 

1. UserWarning: Objective contains too many subexpressions. Consider vectorizing your CVXPY code to speed up compilation. warnings.warn("Objective contains too many subexpressions.
2. UserWarning: Constraint #0 contains too many subexpressions. Consider vectorizing your CVXPY code to speed up compilation. warnings.warn(f"Constraint #{i} contains too many subexpressions.

[...]

Issue #1132, opened 2020-09-28
Getting a segfault when trying to solve an SDP


iamsalil:
**Issue description**
I am trying to solve an SDP. There is only one variable, G, in the optimization problem. When the size of G is small (50x50), the problem is solved. When the size of G becomes large (300x300 for example), the problem segfaults.

**Code Example**
```python
import cvxpy as cp
import numpy as np

def impute_distmat(D, W, alpha=1, verbose=False):
    # Define V and e
    n = D.shape[0]
    x = -1/(n + np.sqrt(n))
    y = -1/np.sqrt(n)
    V = np.ones((n, n-1))
    V[0, :] *= y
    V[1:, :] *= x
    V[1:, :] += np.eye(n-1)
    e = np.ones((n, 1))

    # Solve optimization problem
    G = cp.Variable((n-1, n-1), PSD=True)
    objective = cp.Maximize(cp.trace(G) - alpha*cp.norm(cp.multiply(W, cp.kron(e, cp.reshape(cp.diag(V@G@V.T), (1, n))) + cp.kron(e.T, cp.reshape(cp.diag(V@G@V.T), (n, 1))) - 2*V@G@V.T - D), p='fro'))
    prob = cp.Problem(objective, [])
    prob.solve(verbose=verbose)

    if G.value is None:
        D_rc = None
    else:
        B = V@G.value@V.T
        D_rc = np.kron(e, np.reshape(np.diag(B), (1, n))) + np.kron(e.T, np.reshape(np.diag(B), (n, 1))) - 2*B

    return prob.status, prob.value, D_rc

#############################################
# Construct D
n = 300
np.random.seed(0)
points = np.random.rand(5, n)
xtx = points.T@points
xtxd = np.diag(xtx)
e = np.ones((n, ))
D = np.outer(e, xtxd) - 2*xtx + np.outer(xtxd, e)
# Construct W
W = np.ones((n, n))
# Optimization Problem
opt_status, opt_value, D_rc = impute_distmat(D, W, verbose=True)
```

**Issue**
The segfault occurs at the line `prob.solve(verbose=verbose)`

**Output**
The output is:
Segmentation fault (core dumped)

**Version**
 - OS: Issue occurring on WSL and native Linux
- Python Version: 3.5.6
- CVXPY Version: 1.0.8
- NumPy Version: 1.15.2

==========
[...]

PR #3348, opened 2026-05-27
DRAFT: diffengine backend for ignore_dpp

Transurgeon:
## Description

This PR adds a new backend for the ignore_dpp compilation path.
The new backend is based on SparseDiffengine, a package for structured and sparse differentiation (it is also used in computing general derivatives for DNLP).

We treat the problem of matrix stuffing as the Jacobian of affine constraints A and the Hessian of a quadratic objective P. For parameters, we treat them as they are, symbolic constants. So their values are updated on each solve. Since the sparsity pattern of A and P remain fixed across the family of parameters, we can cache expensive computations on the first iteration, and walk the same expression trees to only retrieve the values on subsequent iterations. Empirically, this leads to similar performance improvements from first solve to subsequent ones than the DPP path.

The main design change is to modify the behaviour of the ignore_dpp path and the EvalParams reduction.
Instead of replacing all Parameter nodes with constants, we now only do that for purely parametric expressions.
One motivating example is ``log_det(P)`` would previously get passed through the canonicalisers and create one SDP cone and some exponential cones, even though the argument is constant (for a single instance).

## Results
This table compares the first compilation time between all backends.

| benchmark | n_vars | CPP | SCIPY | COO | DIFFENGINE | d/CPP |
|---|---:|---:|---:|---:|---:|---:|
| finance.CVaRBenchmark | 769 | 38.07 | 26.71 | 25.92 | 26.14 | 0.69x |
| finance.FactorCovarianceModel | 25,250 | 5.12 | 4.39 | 4.66 | 1.24 | 0.24x |
| gini_portfolio.Cajas | 2,450 | 19.66 | 2.82 | 2.69 | 1.05 | 0.05x |
| gini_portfolio.Murray | 245,400 | 5.93 | 3.77 | 3.75 | 3.52 | 0.59x |
| gini_portfolio.Yitzhaki | 44,900 | 3.32 | 3.01 | 3.25 | 0.51 | 0.15x |
| high_dim_convex_plasticity.ConvexPlasticity | 15,000 | 2.29 | 2.26 | 2.31 | 2.25 | 0.98x |
| huber_regression.HuberRegression | 3,000 | 11.41 | 10.39 | 11.20 | 1.77 | 0.16x |
| matrix_stuffing.ConeMatrixStuffingBench | 5,000 | 11.28 | 9.67 | 16.93 | 3.70 | 0.33x |
| matrix_stuffing.ParamConeMatrixStuffing | 200 | 3.32 | 7.80 | 0.30 | 0.17 | 0.05x |
| matrix_stuffing.ParamSmallMatrixStuffing | 300 | 8.12 | 28.32 | 0.45 | 0.26 | 0.03x |
| matrix_stuffing.SmallMatrixStuffing | 4,000 | 6.88 | 7.34 | 10.67 | 2.92 | 0.42x |
| optimal_advertising.OptimalAdvertising | 250,000 | 6.34 | 2.29 | 2.05 | 5.43 | 0.86x † |
| quantum_hilbert_matrix.QuantumHilbertMatrix | 4,096 | 4.01 | 4.06 | 4.64 | 1.69 | 0.42x |
| sdp_segfault_1132 (issue #1132) | 9,801 | 63.20 | 10.97 | 40.20 | 13.75 | 0.22x |
| semidefinite_programming | 40,000 | 5.70 | 2.36 | 3.68 | 1.98 | 0.35x |
| simple_LP.SimpleLPBenchmark | 10,000,000 | 40.05 | 35.30 | 37.43 | 6.38 | 0.16x |
| simple_LP.SimpleScalarParametrizedLP | 2,000,000 | 7.41 | 6.60 | 7.04 | 1.53 | 0.21x |
| simple_QP.LeastSquares | 2,000 | 14.39 | 15.09 | 16.99 | 2.34 | 0.16x |
| simple_QP.SimpleQPBenchmark | 1,600 | 18.01 | 15.98 | 19.83 | 3.21 | 0.18x |
| simple_QP.UnconstrainedQP (issue #2205) | 18 | 20.33 | 16.22 | 29.45 | 0.20 | 0.01x |
| slow_pruning (issue #1668) | 2,000 | 17.62 | 11.79 | 12.49 | 5.16 | 0.29x |
| svm_l1_regularization | 501 | 24.06 | 14.94 | 14.12 | 5.10 | 0.21x |
| tv_inpainting | 786,432 | 5.85 | 4.94 | 4.66 | 1.64 | 0.28x |

† OptimalAdvertising is the noisiest row across runs (an earlier suite run under machine
contention read 1.58x). Re-measured isolated, 3 fresh cold compiles per backend
(`results_probe_optadv.txt`): CPP median 5.92s (5.81–5.96), DIFFENGINE median 5.22s
(5.10–5.70) → **0.88x**, consistent with the suite row. The other rows are single cold
compiles; treat small ratios near 1x as within noise.

Issue link (if applicable): https://github.com/cvxpy/cvxpy/discussions/3221

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -196,12 +196,17 @@ def _build_solving_chain(
...
     is_dpp = problem.is_dpp(dpp_context, quad_form_dpp=quad_form_dpp)
+
+    use_diffengine = False
     if ignore_dpp or not is_dpp:
         if not ignore_dpp and enforce_dpp:
             raise DPPError(DPP_ERROR_MSG)
         if not ignore_dpp:
             warn(DPP_ERROR_MSG)
-        reductions = [EvalParams()] + reductions
+        if ignore_dpp and ConeMatrixStuffing(use_diffengine=True).accepts(problem):
```
SteveDiamond: The check ``ConeMatrixStuffing(use_diffengine=True).accepts(problem)`` does not seem right here. It doesn't take reformulations into account.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -235,7 +239,8 @@ def _build_solving_chain(
         reductions.append(SOCDim3())
 
     reductions += [
-        ConeMatrixStuffing(quad_obj=quad_obj, canon_backend=canon_backend),
+        ConeMatrixStuffing(quad_obj=quad_obj, canon_backend=canon_backend,
+                           use_diffengine=use_diffengine),
```
SteveDiamond: Why is use_diffengine different than canon_backend? Isn't it a specific canonicalization backend, so another canon_backend?

SteveDiamond:
This looks promising, but could you add benchmark timings? There were some benchmarks used for the COO backend PRs, or you could make more.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -55,6 +55,50 @@
 from cvxpy.utilities.coeff_extractor import CoeffExtractor
 
 
+def lower_and_order_constraints(constraints):
+    """Lower Equality/Inequality to Zero/NonNeg, normalize axis-1 cones, reorder.
+
+    Returns the constraints reordered as Zero, NonNeg, SOC, PSD, SvecPSD, ExpCone,
+    PowCone3D, PowConeND, along with a (new_id -> old_id) map. Because all lowering
+    preserves the original `constr_id`, the map is an identity.
```
PTNobel: I thought we were moved away from reused ids between objects

cvxpy/reductions/solvers/solving_chain.py
Transurgeon: I made the change now. I thought it would be different since it doesn't do the same thing as current canon backends.. but I think its nicer if it can just be a drop-in flag.

cvxpy/reductions/solvers/solving_chain.py
Transurgeon: yes agreed, I had missed that and made it more inline with existing code.
Actually I think the new backend should only be added once it has full feature parity with cvxcore.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
Transurgeon: done, I am now using the code from master and decided to not refactor this part of the class.

Transurgeon:
These are the benchmark results for the parametric resolve times.

| benchmark | n_vars | dpp | eval_params | diffengine (rebuilt) | **de_cached** |
|---|---:|---:|---:|---:|---:|
| finance.FactorCovarianceModel | 25,250 | 0.144 | 0.870 | 0.879 | 0.282 |
| high_dim_convex_plasticity | 15,000 | 2.145 | 2.074 | 2.090 | 2.133 |
| matrix_stuffing.ParamConeMatrixStuffing | 200 | 0.0031 | 0.130 | 0.128 | **0.0010** |
| matrix_stuffing.ParamSmallMatrixStuffing | 300 | 0.0048 | 0.191 | 0.186 | **0.0015** |
| simple_LP.SimpleScalarParametrizedLP | 2,000,000 | 0.382 | 1.113 | 1.099 | **0.246** |
| svm_l1_regularization | 501 | 0.296 | 2.396 | 2.288 | 0.542 |
| SimpleFullyParametrizedLP (1M params) | 1,000,000 | 0.255 | 0.607 | 0.606 | **0.161** |
| ParametrizedQP (14.4M params) | 2,400 | 1.429 | 2.869 | 2.870 | **1.011** |

The diff engine currently gets rebuilt because we must "fold" trees that are purely parametric (I.e. ``log_det(P)``) into constants. Therefore we cannot cache the C_problem that is built for the diff engine and reuse it properly. This is a bit of an issue, so I am thinking about how to get around it now.

PTNobel:
Full disclosure, I'm getting worried about eventually having to get two reviews for this over 2000 line PR. I know this is a prototype and so I don't think you need to make it reviewable, but be aware that I'll probably ask you to break it into smaller units with clear boundaries when it comes time to merge.

Transurgeon:
> Full disclosure, I'm getting worried about eventually having to get two reviews for this over 2000 line PR. I know this is a prototype and so I don't think you need to make it reviewable, but be aware that I'll probably ask you to break it into smaller units with clear boundaries when it comes time to merge.

yeah of course, I think I will break it up into three PRs.
Thanks for the heads up.
Also the line changes are mainly defining the new DiffEngineConeProgram class and lots of new tests.

PR #3353, opened 2026-05-27
Dedup identical subtrees in Dcp2Cone canonicalization

PTNobel:
## Description
Currently, `cp.Problem(cp.Minimize(cp.norm1(x)), [cp.norm1(x) <= 1])` canonicalizes the two separately-constructed `cp.norm1(x)` expressions into independent epigraph variables and constraint pairs. This PR adds a per-`apply()` common-subexpression cache to `Dcp2Cone` so structurally identical Expression subtrees share one canonicalized expression and one set of auxiliary constraints within a single reduction pass.

For the example above, the canonicalized data the solver sees shrinks from 3 scalar variables / 5 constraint rows down to 2 scalar variables / 3 constraint rows.

### How it works
- Cache lives on the `Dcp2Cone` instance and is reset at the top of every `apply()` call.
- Keyed on a structural digest of each Expression subtree: atom type, shape, hashable `get_data()`, child keys, leaf cvxpy ids for `Variable`/`Parameter`, and shape+dtype+value bytes for `Constant`.
- `affine_above` is folded into the key only when a quad-canon-eligible descendant could make canonicalization depend on it; otherwise identical subtrees deduplicate across objective and constraint contexts (which start with different `affine_above` values).
- Only Expression subtrees are cached — objectives, user constraints, and `partial_problem` subtrees are excluded so their IDs flow into `inverse_data` unchanged.
- If a structural key cannot safely be built for an unusual `get_data()` entry, that subtree is silently skipped rather than risking incorrect reuse.

## Type of change
- [x] Bug fix

- [x] Write unittests.

## Test plan
- [x] New `cvxpy/tests/test_dcp2cone_cse.py` covers scalar dedup, vector dedup, distinct-subtree non-merge (`norm1(x)` vs `norm1(-x)`), end-to-end solve equivalence, and dedup across a shared `Parameter` subtree.
- [x] `pytest cvxpy/tests/test_problem.py cvxpy/tests/test_qp_solvers.py cvxpy/tests/test_atoms.py cvxpy/tests/test_dgp2dcp.py cvxpy/tests/test_dqcp.py` all green.

### DIFF
--- a/cvxpy/reductions/dcp2cone/dcp2cone.py
+++ b/cvxpy/reductions/dcp2cone/dcp2cone.py
@@ -75,11 +75,6 @@ class Dcp2Cone(Canonicalization):
         inverse_data = InverseData(problem)
 
         self._cse_cache = {}
-        # Counter incremented whenever canonicalize_expr takes the quad branch
-        # (producing a SymbolicQuadForm). Subtrees that touch the quad branch
-        # are not cached, because downstream code identifies SymbolicQuadForm
-        # instances by Python id and assumes each occurrence is distinct.
-        self._quad_canon_count = 0
 
         canon_objective, canon_constraints = self.canonicalize_tree(
             problem.objective, True)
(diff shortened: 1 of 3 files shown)


PTNobel:
Combined into #3355, which now targets master directly.

PR #3355, opened 2026-05-27
Dedup identical subtrees in Dcp2Cone canonicalization

PTNobel:
## Description

When the same Expression subtree appears in two places in a DCP problem, for example `cp.norm1(x)` in both the objective and a constraint, `Dcp2Cone` previously emitted a fresh set of auxiliary variables and epigraph constraints per occurrence. This PR adds a per-`apply()` structural-key cache so `Dcp2Cone` canonicalizes each structurally identical subtree once and reuses the canonical expression on later occurrences.

Worked example for the reported case:
```python
cp.Problem(cp.Minimize(cp.norm1(x)), [cp.norm1(x) <= 1])
```
The canonicalized problem now has one epigraph variable `t` shared across the objective and the constraint, with one pair of `t >= x`, `t >= -x` inequalities.

The DNLP changes from earlier revisions were removed from this PR. The original DNLP work is preserved on branch `ptn/dnlp2smooth-cse-dnlp-followup` for follow-up issue discussion.

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.


SteveDiamond:
This generally looks really good. @Transurgeon can you review the changes to DNLP?

SteveDiamond (commented):
Reviewed with differential testing of shared-vs-unshared canonicalization across many atoms/solvers — the CSE correctness looks solid (the DAG-mutation risk is correctly defended by the fresh-placeholder-id fix, and the `_affine_above_relevant` key logic mirrors `canonicalize_tree`). A few inline notes below; the main one is the O(N²) canonicalization regression on deep expression trees.

cvxpy/reductions/subexpr_cache.py
```diff
@@ -0,0 +1,137 @@
...
+
+class UncacheableError(Exception):
+    """Raised when a structural cache key cannot be safely built.
+
+    Callers should treat this as "skip the cache for this subtree" rather than
+    risking incorrect reuse.
+    """
+
+
+def expr_key(expr):
```
SteveDiamond: **Performance: O(N²) canonicalization regression on deep trees.**

`expr_key` recursively walks the *entire* subtree to build a key, and `canonicalize_tree` calls it at *every* node with no memoization → O(N²) where canonicalization was O(N). The key tuples are themselves O(subtree)-sized, so per-node hashing/storage is also O(subtree) → O(N²) memory.

Measured on a depth-N `cp.abs(-(-(…-x)))` chain (toggling only the cache-key construction):

| depth | master | this PR | slowdown |
|---|---|---|---|
| 100 | 0.72 ms | 3.79 ms | 5× |
| 400 | 3.11 ms | 68.7 ms | 22× |
| 800 | 7.36 ms | 311 ms | 42× |

Master scales ×2 per doubling (linear); this PR scales ×4 (quadratic), and the factor widens with N. Flat n-ary sums are unaffected (CVXPY flattens associative ops) — it bites genuinely deep trees: nested `abs`/`reshape`/neg/index chains.

**Suggested fix:** thread a per-apply `{id(expr): key}` memo through `expr_key` so each node's key is built once and child keys are reused (one bottom-up O(N) pass). Same change applies to the Dnlp2Smooth call site.

cvxpy/reductions/dcp2cone/dcp2cone.py
```diff
@@ -145,3 +179,43 @@ def canonicalize_expr(self, expr, args, affine_above: bool) -> tuple[Expression,
...
+        """
+        try:
+            structural = expr_key(expr)
+        except UncacheableError:
+            return None
+        if self._affine_above_relevant(expr):
+            return (structural, bool(affine_above))
+        return (structural, None)
+
+    def _affine_above_relevant(self, expr) -> bool:
```
SteveDiamond: This is a *second*, independent O(N²) whole-subtree walk: `_make_cache_key` calls `_affine_above_relevant` at every node (when `quad_obj=True`), and it recurses over the full subtree each time. Instrumented call count tracks N²/2 exactly (5,155 calls at depth 100; 80,605 at depth 400).

It can fold into the same bottom-up pass suggested for `expr_key`: a node is relevant iff it is quad-eligible or any child is relevant — O(1) amortized per node.

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
```diff
@@ -76,6 +93,18 @@ def canonicalize_tree(self, expr, affine_above: bool) -> tuple[Expression, list]
         -------
         A tuple of the canonicalized expression and generated constraints.
         """
+        # Only Expression subtrees are eligible for the CSE cache. Objectives
+        # and user constraints are intentionally excluded so their IDs flow
+        # through to inverse_data unchanged.
+        cache_key = None
+        if isinstance(expr, Expression):
+            try:
+                cache_key = expr_key(expr)
```
SteveDiamond: Same O(N²) pattern as Dcp2Cone: `expr_key(expr)` is recomputed from scratch at every node with no reuse of child keys already built. The memoization fix (per-apply `{id(expr): key}` memo) should be applied here too.

cvxpy/reductions/subexpr_cache.py
```diff
@@ -0,0 +1,137 @@
...
+    3. Fallback: id of the Constant wrapper. The common large-data pattern of
+       binding ``c = cp.Constant(arr)`` and reusing ``c`` deduplicates here.
+    """
+    value = expr.value
+    try:
+        arr = np.asarray(value)
+    except (TypeError, ValueError):
+        return ("const", id(expr))
+    if arr.size <= _CONSTANT_VALUE_HASH_MAX_SIZE:
+        return ("const-val", arr.shape, str(arr.dtype), arr.tobytes())
```
SteveDiamond, line 169: **Sparse Constants are silently mis-keyed here.** CVXPY Constants frequently wrap SciPy sparse matrices, and `np.asarray(sparse)` returns a 0-dim `object` array of size 1 — which passes the `arr.size <= 64` guard. `arr.tobytes()` then returns the 8 bytes of `id(value)` (verified), *not* the contents, under the `'const-val'` label, with the real shape recorded as `()`.

This is not a correctness bug — within one `apply()` all Constants are live, so ids are unique and no false merge can occur — but it (a) defeats CSE entirely for every sparse-constant subtree and (b) is a fragile footgun: a "by-value" path that actually encodes a transient pointer and discards shape.

Suggest guarding `arr.dtype != object` before the `tobytes()` value branch, and handling sparse explicitly (e.g. key on `(data.tobytes(), indices.tobytes(), indptr.tobytes(), shape)`), falling back to `id(expr)` (the wrapper, kept alive by the problem tree) rather than `id(value)`.

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
```diff
@@ -76,6 +93,18 @@ def canonicalize_tree(self, expr, affine_above: bool) -> tuple[Expression, list]
         -------
         A tuple of the canonicalized expression and generated constraints.
         """
+        # Only Expression subtrees are eligible for the CSE cache. Objectives
+        # and user constraints are intentionally excluded so their IDs flow
+        # through to inverse_data unchanged.
+        cache_key = None
+        if isinstance(expr, Expression):
```
SteveDiamond: Minor / latent: Dcp2Cone explicitly excludes `partial_problem` from the cache (so embedded constraint ids flow through unchanged), but here the eligibility check is a bare `isinstance(expr, Expression)` — and `PartialProblem` *is* an `Expression`. It's currently latent (the NLP path doesn't support `partial_optimize` and would fail earlier), but it's an inconsistency that becomes a real id-collapse trap if NLP `partial_optimize` support is ever added. Worth mirroring the Dcp2Cone guard now.

cvxpy/utilities/replace_quad_forms.py
```diff
@@ -29,10 +33,17 @@ def replace_quad_forms(expr, quad_forms):
...
+    appearing at multiple positions (e.g. a shared SymbolicQuadForm reused
+    across an objective expression) gets a distinct row in the coefficient
+    matrix downstream. Previously the placeholder inherited
+    ``quad_form.id``, which caused two occurrences to collapse onto a single
+    row and silently halve the quadratic coefficient.
+    """
     quad_form = expr.args[idx]
-    placeholder = Variable(quad_form.shape,
-                           var_id=quad_form.id)
+    placeholder = Variable(quad_form.shape)
```
SteveDiamond, line 45: The fresh-id change is correct. One trivial follow-up: the comment at `coeff_extractor.py:131` ("`var_id` is the placeholder's ID (= the SymbolicQuadForm's ID)") is now stale — the placeholder no longer shares the quad form's id. The logic is unaffected (`orig_id` is recomputed from `quad_forms[var_id][2].args[0].id`), but updating that comment in this PR would prevent someone from re-introducing the `var_id=quad_form.id` assumption.

cvxpy/reductions/subexpr_cache.py
PTNobel: D'oh. Thanks for the catch...

cvxpy/reductions/subexpr_cache.py
PTNobel: > CVXPY Constants frequently wrap SciPy sparse matrices, and np.asarray(sparse) returns a 0-dim object array of size 1

This seems like a separate bug?

cvxpy/utilities/replace_quad_forms.py
PTNobel: Fixed.

cvxpy/reductions/dnlp2smooth/dnlp2smooth.py
PTNobel: Good idea; fixed.

cvxpy/reductions/subexpr_cache.py
PTNobel: Fixed regardless.

PTNobel:
All my experiments show linear memory etc now!

Transurgeon:
nice progress on this PR and really nice job getting it to work!
I am a bit busy with other things now but this is definitely on my mind to review soon :).

Transurgeon (dismissed):
@PTNobel I got a chance to look at the PR now.
I think we should remove the changes for DNLP since there is not much benefit if we can't reuse the derivative computation. I will open an issue for that.

Also one clarifying question:
- this PR allows us to not create new auxiliary variables, but we still can't reuse affine subexpressions right?

As an example, I was thinking of this [control of car](https://github.com/cvxpy/examples/blob/main/nlp_examples/control_of_car.py#L41-L46) example. Where we could reuse ``x_curr`` across all constraints, instead of having to reform the indexing expression every time.
Im not saying we should do this (if its not done), but I am just trying to understand the scope of the PR.
After you answer this question and make the DNLP change, I think this is looking good to merge!

PTNobel:
> this PR allows us to not create new auxiliary variables, but we still can't reuse affine subexpressions right?

I am not sure that I know what it would mean to reuse affine subexpressions.

This ensures that identical (in formation) affine subexpressions refer to the *same object* when passed to the next step, which means that canonical_form returns the same LinOp object. 

At this point, we do **not** cache canonicalizing LinOp objects. Therefore, this doesn't do anything. However, this PR makes it way easier to make that change (it becomes local to the backend rather than needing to be done across the codebase).

PTNobel:
#3423 goes ahead and adds the LinOp cache to SCIPY and COO.

rileyjmurray (approved):

PR #3357, opened 2026-05-29
Fix XPRESS QP test gating to run under the Community license

PTNobel:
## Description
`is_xpress_available()` in `test_qp_solvers.py` gated every license-sensitive XPRESS QP test, but it checked the license via `xpress.env().getlicense()`. `xpress.env` was **removed in modern xpress (9.x)**, so the call raised `AttributeError`, which the bare `except Exception: return False` swallowed. The helper therefore *always* returned False, and pytest silently skipped every XPRESS-parametrized QP correctness test.

The practical consequence: the inequality-row and constraint-dual code paths in `xpress_qpif.py` were never exercised (the only XPRESS QP test that ran was the unconstrained-MIP warm-start test), leaving the interface at 74.5% coverage despite XPRESS being installed and licensed in CI.

This PR detects availability by creating a `problem()` object — that is where XPRESS acquires its license — which works under the Community license bundled with the `xpress` package, mirroring how the MOSEK/KNITRO helpers construct an env/context.

The Community license caps problems at 200 rows+columns. Five standard QP correctness tests exceed that limit, so `_solve()` now catches XPRESS's specific size-limit `SolverError` and skips those tests with a clear reason — a full XPRESS license still runs all of them.

Verified locally with `xpress==9.9.0` (Community license):
- XPRESS QP tests: 18 passed, 5 skipped (oversized), 0 failed
- Full `test_qp_solvers.py`: 244 passed, 20 skipped, 0 failed (no change for other solvers)
- `xpress_qpif.py` coverage: 74.5% → 87%, with the inequality-row and constraint-dual paths now covered

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/tests/test_qp_solvers.py
+++ b/cvxpy/tests/test_qp_solvers.py
@@ -50,14 +50,21 @@ from cvxpy.tests.test_conic_solvers import is_knitro_available, is_mosek_availab
 # canonical, KNITRO-OpenMP-safe implementations live in one place.
 
 def is_xpress_available():
-    """Check if XPRESS is installed and a license is available."""
+    """Check if XPRESS is installed and a usable license can be acquired.
+
+    The Community license bundled with the ``xpress`` package is sufficient for
+    the small problems in this suite, so we only confirm that a problem object
+    can be created -- that is where XPRESS acquires its license. The previous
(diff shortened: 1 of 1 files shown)


Transurgeon (approved):
nice!

PR #3358, opened 2026-05-29
Return COPT dual Farkas infeasibility certificates (conic interface)

PTNobel:
## Description
COPT silently discarded its dual Farkas ray on infeasible problems, so COPT returned `dual_value = None` on infeasible problems while OSQP/HIGHS/CLARABEL return an infeasibility certificate. Neither COPT interface requested the ray (`ReqFarkasRay`) nor read it (`Info.DualFarkas`); `invert()` called `failure_solution(status, attr)` with no duals.

This PR fixes both COPT interfaces to request the ray, read it on infeasibility, negate it for CVXPY's sign convention, split it `[eq; ineq]` like the optimal duals, and pass it through `failure_solution` — mirroring OSQP. Each interface now shares a `_dual_vars()` helper between the optimal-dual and Farkas paths. The dualized PSD path and MIPs are excluded (no dual ray).

### Conic interface (`copt_conif.py`) — verified ✅
Handles infeasible **LPs**, which COPT detects correctly. COPT now passes the full `StandardTestInfeasibleProblems` LP tests (status, value, *and* a valid Farkas certificate `y >= 0, A^T y = 0, b^T y < 0`). Added `test_copt_infeasible_lp_{ineq,eq}`. No regression — 28 COPT conic tests pass.

### QP interface (`copt_qpif.py`) — latent, pending COPT ⚠️
The same fix is applied, but it can't be exercised yet: for infeasible **QPs**, COPT (`loadMatrix` + `loadQ` build path) fails to detect infeasibility and returns a constraint-violating `optimal` instead of `infeasible`, so the new branch is never reached. Reproduced in raw `coptpy` (no CVXPY): with `loadMatrix` then `loadQ`, an infeasible QP solves to a spurious `optimal`; loading `Q` first detects infeasibility but `loadMatrix` then resets the quadratic term — so it's not reorderable on our side. Being reported to Cardinal. The extraction code is in place and activates once COPT detects QP infeasibility. Feasible COPT QP solves are unaffected (23 COPT QP tests pass).

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
@@ -334,16 +334,17 @@ class COPT(ConicSolver):
                 model.loadExpCone(nexpcone, None,
                                   range(A.shape[1] - nexpconedim, A.shape[1]))
 
+        # Request a dual Farkas ray so an infeasibility certificate is available
+        # for infeasible problems (not the dualized PSD path, and not MIPs). Set
+        # it before the user-settings loop so an explicit value in solver_opts wins.
+        if not dims[s.PSD_DIM] and not (data[s.BOOL_IDX] or data[s.INT_IDX]):
+            model.setParam(copt.COPT.Param.ReqFarkasRay, 1)
+
         # Set parameters
         for key, value in solver_opts.items():
             # Ignore arguments unique to the CVXPY interface.
             if key not in self.INTERFACE_ARGS:
                 model.setParam(key, value)
-
-        # Request a dual Farkas ray so an infeasibility certificate is available
-        # for infeasible problems (not the dualized PSD path, and not MIPs).
-        if not dims[s.PSD_DIM] and not (data[s.BOOL_IDX] or data[s.INT_IDX]):
-            model.setParam(copt.COPT.Param.ReqFarkasRay, 1)
 
         if 'save_file' in solver_opts:
             model.write(solver_opts['save_file'])
--- a/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/copt_conif.py
@@ -138,6 +138,29 @@ class COPT(ConicSolver):
 
         return data, inv_data
 
+    def _dual_vars(self, solution, inverse_data):
+        """Map the stacked ``[eq; ineq]`` dual vector to a CVXPY dual dict.
+
+        Shared by the optimal duals and the infeasibility (Farkas) certificate,
+        which use the same row layout.
+        """
+        eq_dual = utilities.get_dual_values(
+            solution[s.EQ_DUAL],
+            utilities.extract_dual_value,
+            inverse_data[COPT.EQ_CONSTR])
+        leq_dual = utilities.get_dual_values(
+            solution[s.INEQ_DUAL],
+            utilities.extract_dual_value,
+            inverse_data[COPT.NEQ_CONSTR])
+        for con in inverse_data[self.NEQ_CONSTR]:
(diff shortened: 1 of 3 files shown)


cvxpy/reductions/solvers/qp_solvers/copt_qpif.py
```diff
@@ -222,6 +236,13 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
         if solution[s.STATUS] == s.USER_LIMIT and not model.hasmipsol:
             solution[s.STATUS] = s.INFEASIBLE_INACCURATE
 
+        # On infeasibility, return the dual Farkas ray as the certificate,
+        # negated to match CVXPY's sign convention (as done for optimal duals).
+        if (solution[s.STATUS] in (s.INFEASIBLE, s.INFEASIBLE_INACCURATE)
+                and vtype is None and model.hasdualfarkas):
+            solution['y'] = -np.array(
+                model.getInfo(copt.COPT.Info.DualFarkas, model.getConstrs()))
```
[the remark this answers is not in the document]
PTNobel: No? This would be bad? We should error if hasdualfarkas is True and we can't load the Farkas lemma

PTNobel:
@terryweng03 can you review this PR? I think you know the COPT interface better than me.

cvxpy/reductions/solvers/qp_solvers/copt_qpif.py
PTNobel, line 214: Shouldn't we reaise a SolverError?

cvxpy/reductions/solvers/conic_solvers/copt_conif.py
```diff
@@ -330,6 +340,11 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
             if key not in self.INTERFACE_ARGS:
                 model.setParam(key, value)
 
+        # Request a dual Farkas ray so an infeasibility certificate is available
+        # for infeasible problems (not the dualized PSD path, and not MIPs).
+        if not dims[s.PSD_DIM] and not (data[s.BOOL_IDX] or data[s.INT_IDX]):
```
PTNobel: I can't figure out how to get unboundedness rays for PSD problems

Transurgeon:
pinging also @wujianjack who originally developed the COPT interface.
This will allow the user to (eventually) debug infeasibility better.
We are (passively) working on an API which will allow users to do that.

terryweng03:
> @terryweng03 can you review this PR? I think you know the COPT interface better than me.

Dear Parth, thanks very much for developing this. Wujian and I will review it later. 
Also, we reproduced the bug you reported, and we will investigate how to fix it.

terryweng03:
@PTNobel Hi Parth! The bug has been fixed, and we released COPT 8.0.5 last Saturday. Thanks very much for sharing this.

The code looks good to me! I have one question: Dual Farkas in COPT can only be computed for LP. As far as I know, if a problem is LP, cvxpy will only use ConicSolver to solve it. Should we add this part in QPSolver as well?

PTNobel:
No need for the QP interface to have it then. Do y'all have any IIS mechanisms for non-LPs?

Transurgeon (approved):
Looks good per approval from @terryweng03 !

PTNobel:
I'll drop the QP changes before I merge

terryweng03:
> No need for the QP interface to have it then. Do y'all have any IIS mechanisms for non-LPs?

Our IIS mechanisms also support MILP (you could refer https://guide.coap.online/copt/en-doc/infeasible.html)

PR #3359, opened 2026-05-29
Fix #3333: Support parsing SCS solver options

### LINKED ISSUE #3333: Support parsing direct SCS solvers options and parameters
Right now, only a number of solvers such as MOSEK (through `mosek_params`) or GLOP (`parameters_proto`) is allowed to pass solver options directly into cvxpy solvers. I was wondering if this feature can be extended to cvxpy.SCS? It's a fairly popular solver that still concurrently maintained. Thank you!

curious7-web:
## Description

This updates `SCS.parse_solver_options` to pull `scs_params` out of `solver_opts` with `.pop()` and merge those options into the arguments passed to SCS. This makes `scs_params` behave like other solver-specific option dictionaries such as `mosek_params`.

I also added `test_scs_custom_params` to check that a custom SCS option, such as `max_iters`, is passed through and affects the solve.

Issue link: #3333

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/scs_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/scs_conif.py
@@ -193,6 +193,8 @@ class SCS(ConicSolver):
 
     @staticmethod
     def parse_solver_options(solver_opts):
+        scs_params = solver_opts.pop("scs_params", {})
+        solver_opts.update(scs_params)
         import scs
         if Version(scs.__version__) < Version('3.0.0'):
             if "eps_abs" in solver_opts or "eps_rel" in solver_opts:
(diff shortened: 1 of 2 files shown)


SteveDiamond:
A little bit of context around this topic of solver args. The initial design for problem.solve(solver=XXX, *kwargs) was that kwargs would be passed to the solver. This is how the current SCS interface works, so I don't believe adding scs_params increases functionality. In the long run it may be worth revisiting how solver arguments work and moving more generally to a ``solver_args`` argument rather than the generic kwargs. A couple solver interfaces like Mosek have a solver_params argument already because the kwarg interface wasn't a good fit.

curious7-web:
@SteveDiamond
Got it, thanks for the context. I see the distinction now: SCS already accepts direct solver kwargs, so this is more of an API consistency question than missing functionality.

Happy to close this if you’d prefer to keep the SCS path as-is and revisit solver option dictionaries in a broader solver args cleanup later.

SteveDiamond:
Yeah let's close this for the moment.

PR #3362, opened 2026-06-02
Draft: Post-solver feasibility design note

curious7-web:
## Description

This is a draft design note for the post-solver feasibility check direction listed on the CVXPY GSoC project ideas page.

The full design note is in `scratch/post_solver_feasibility_design.md`.

It focuses on:

- possible API surfaces;
- lazy vs eager feasibility checking;
- how `constraint.violation()` values should be aggregated;
- a small local benchmark comparing one vectorized constraint against many scalar constraints.

This is not intended as a ready-to-merge implementation yet. I’m opening it as a draft to collect design feedback before turning the spike into a cleaner implementation PR.

## Type of change

- [x] Other (design discussion / spike)

AI assistance was used for brainstorming the design-note structure and reviewing wording.

### DIFF
--- a/scratch/post_solver_feasibility_design.md
+++ b/scratch/post_solver_feasibility_design.md
@@ -175,7 +175,7 @@ The absolute times are small in this toy benchmark, but the scaling difference s
 
 A first PR could stay small:
 
-* Internal helper to normalize each constraint violation to a scalar and compute the max original-constraint violation.
+* Internal helper to normalize each primal constraint violation to a scalar and compute the max original-constraint violation.
 * Tests for a solved feasible problem.
 * Tests for manually corrupted variable values.
 * Tests covering scalar normalization behavior across core constraint classes.
@@ -192,3 +192,15 @@ Possible follow-up work could add warning behavior through a solve flag:
 ```python
 prob.solve(check_feasibility=True, feasibility_tol=...)
 ```
+
+## Future scope
+
+The first version of this design focuses on primal constraint violation only.
+
+However, the API should be documented carefully because the meaning of `prob.violation()` may expand in the future if related checks can be defined consistently. Possible follow-up directions include:
+
+* **Dual feasibility:** checking whether unpacked constraint dual values satisfy the expected dual-domain conditions.
+* **Complementarity:** checking primal-dual complementarity where it can be evaluated cleanly for a constraint type.
+* **Duality gap / stationarity:** investigating whether these can be computed reliably in the original problem space.
+
+These are left out of the first version because they require more care than a primal-side violation check, especially for reductions, functional convex problems, subgradients, and solver-specific dual information.
--- a/scratch/post_solver_feasibility_design.md
+++ b/scratch/post_solver_feasibility_design.md
@@ -2,7 +2,7 @@
 
 ## Goal
 
-Explore a small post-solver feasibility-check layer for CVXPY. The first API direction considered here is a raw scalar primal-side constraint violation value, which could later support tolerance-based warnings or feasibility reporting.
+Explore a small post-solver validation layer for CVXPY. The first API direction considered here is a raw scalar violation value computed from checks on the original problem after solution values are unpacked. The first pass should cover primal constraint violation, and audit whether dual feasibility and complementarity can be included cleanly.
 
 ## Main design question
 
@@ -82,18 +82,18 @@ Reason:
 * Lowest risk.
 * No default overhead.
 * No new default warnings.
-* Avoids mixing raw violation reporting with tolerance-based feasibility decisions.
+* Separates raw post-solver validation values from tolerance-based feasibility decisions.
 * Easier to extend later into warning behavior or a solve flag.
 
 ## Minimal first API
 
-For the first public API, prefer exposing only a scalar violation value rather than a full feasibility report:
+For the first public API, prefer exposing a scalar violation value rather than a full feasibility report:
 
 ```python
 violation = prob.violation()
 ```
 
-This should be interpreted as a CVXPY-side primal constraint violation check on the original constraints after solution values are unpacked. It should not be framed as an infeasibility certificate.
+This should be interpreted as a CVXPY-side post-solver validation value computed from the original problem after solution values are unpacked. The initial implementation should define exactly which checks are included. It should not be framed as an infeasibility certificate.
 
 Tolerance-based behavior, warnings, or an `is_feasible` boolean can be considered later once the raw violation definition is clear.
 
@@ -109,7 +109,7 @@ all original problem constraints after solution values are unpacked.
 
 Initial audit: `Constraint.violation()` is documented as the distance from the current expression value to the constraint domain. Before building a problem-level API, the main open question is whether each individual constraint violation should first be normalized to a scalar with a clearly documented norm/distance convention. Once each constraint reports a scalar consistently, `prob.violation()` can aggregate those scalar values directly.
 
-This API should only report primal-side violation of returned values against the original constraints. It should not be presented as an infeasibility certificate or as a substitute for solver certificates such as a Farkas ray.
+This API should report post-solver validation against the original problem data. It should not be presented as an infeasibility certificate or as a substitute for solver certificates such as a Farkas ray.
 
 Constraint types to check:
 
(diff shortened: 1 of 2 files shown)


scratch/post_solver_feasibility_design.md
```diff
@@ -0,0 +1,191 @@
...
+## Definition of violation
+
+Initial definition:
+
+```text
+max_violation is the maximum value returned by constraint.violation()
+across all original problem constraints after solution values are unpacked.
+```
+
+Initial audit: `Constraint.violation()` is documented as the distance from the current expression value to the constraint domain. Existing tests cover violation behavior for equality, PSD, power cone, boolean/inequality-style, and linear cone cases. One open question is aggregation: some constraints return vector residuals, while others return scalar norms, so the feasibility report needs a clear rule for reducing all constraint violations to one `max_violation` value.
```
PTNobel: I feel like we should normalize all the violations to return the same type; which is probably a scalar rather than a vector?

Similarly it isn't obvious to me what "distance" means; it could be $\ell_1$, $\ell_2$ or $\ell_\infty$. When we have a consistent definition there, I think it will naturally follow out how to aggregate them.

scratch/post_solver_feasibility_design.md
```diff
@@ -0,0 +1,191 @@
...
+
+Before proposing a public API, the next step is to check the existing coverage report and tests for `constraint.violation()` across the main constraint classes, especially equality, inequality, PSD, SOC, exponential, and power constraints.
+
+Constraint types to check:
```
PTNobel, line 114: Approx cones are maybe most important to check.

scratch/post_solver_feasibility_design.md
```diff
@@ -0,0 +1,191 @@
...
+- If lazy, it needs cache/staleness rules.
+- Could be confused with `solver_stats`, even though this is CVXPY-side validation rather than solver-returned metadata.
+
+## Current preference
+
+Start with a lazy explicit method:
+
+```python
+prob.solve()
+report = prob.feasibility_report(tol=1e-6)
```
PTNobel: What does the tolerance do?

I feel like a static prob.violation should be fine if we're aggregating the violations consistently, rather than having the user specify a tolerance.

PTNobel:
Is there a relationship between violation and infeasibility that we should bridge with this API? Is there a connection between how large the constraints are violated and the direction of the Farkas ray in measuring infeasibility?

curious7-web:
Thanks, this is helpful.

I think the cleaner split is that `prob.violation()` should probably be a raw primal-side constraint violation number, not a feasibility/infeasibility certificate and not tolerance-based by default.

One thing I want to make sure I’m understanding correctly: would you prefer each individual `constraint.violation()` to return a scalar first, with a clearly documented norm/distance convention for what “distance” means, and then have `prob.violation()` aggregate those scalar values? That seems cleaner than having `prob.violation()` handle a mix of scalar norms and vector residuals.

I’ll also update the audit section to focus more on approximate cones like exp/power, since that seems like where the ambiguity matters most.

PTNobel:
> would you prefer each individual constraint.violation() to return a scalar first, with a clearly documented norm/distance convention for what “distance” means, and then have prob.violation() aggregate those scalar values?

This makes sense to me. 

> I’ll also update the audit section to focus more on approximate cones like exp/power, since that seems like where the ambiguity matters most.

Exp is not an approx cone. Power can be, but we're trying to move away from that. The notable approx cones are all quantum ones where we use quadrature approximation.

curious7-web:
Thanks for the correction;  I misunderstood what counted as an approximate cone here.

I’ll update the note to separate exp/power from the actual approximate cone cases, especially the quantum cones using quadrature approximation.

curious7-web:
Updated the note to make the first direction a raw scalar `prob.violation()` API, separated from tolerance/warning behavior. I also revised the violation section around scalar normalization and documented norm/distance conventions, and separated exp/power from the actual approximate cone cases like quantum cones with quadrature approximation.

scratch/post_solver_feasibility_design.md
```diff
@@ -0,0 +1,194 @@
...
+
+```text
+Assuming each constraint violation is first normalized to a scalar,
+prob.violation() would return the maximum scalar violation across
+all original problem constraints after solution values are unpacked.
+```
+
+Initial audit: `Constraint.violation()` is documented as the distance from the current expression value to the constraint domain. Before building a problem-level API, the main open question is whether each individual constraint violation should first be normalized to a scalar with a clearly documented norm/distance convention. Once each constraint reports a scalar consistently, `prob.violation()` can aggregate those scalar values directly.
+
+This API should only report primal-side violation of returned values against the original constraints. It should not be presented as an infeasibility certificate or as a substitute for solver certificates such as a Farkas ray.
```
PTNobel: Shouldn't we also check dual feasiblity, at least at the problem level? Should we also check duality gap? Is that feasible to do in the original problem space?

curious7-web: That makes sense as a broader validation question.

My instinct is to keep the first API narrowly scoped to primal-side violation of the original constraints, since that can be evaluated directly from the returned primal variable values.

Dual feasibility and duality gap seem like a separate design problem. For some problem classes they may be computable from the original objective and constraint dual values, but making that consistent across CVXPY reductions, cone canonicalization, and solvers seems like a larger design question than this primal violation check.

Would it make sense to treat `prob.violation()` as the small primal-side API first, document that scope explicitly, and leave dual feasibility / duality gap as a separate follow-up design question?

PTNobel: Dual feasiblity should be trivial, we already assign the dual variables to the .dual_variable. value and I don't see why we shouldn't be able to quickly and easily evaluate that.

I think duality gap might be hard, but maybe complementarity could be easy...

curious7-web: Ah I was overthinking dual feasibility because of the original-space/canonicalized-space distinction, but if the relevant duals are already unpacked onto `constraint.dual_value`, then it should be reasonable to include in the audit.

I’ll update the note to separate the checks more clearly: primal constraint violation, dual feasibility from the unpacked dual values, complementarity as a likely feasible next check, and duality gap as the part that may need more design.

PTNobel: I mean complementarity and duality gap measure the same thing iirc

PTNobel: Wait I checked BV and they do not; they're equivalent for conic programs, not functional convex programs. For those we also need the gradient of the Lagrangian. Which is probably easier to compute than the dual objective function, but definitely out of scope here.

PTNobel: Maybe we should document that Problem.violation may change in the future to include those other conditions?

PTNobel: Of course stationarity is presumably finding the minimum norm vector in a set because of subgradients. Not sure how to approach that.

curious7-web: Good point about BV and subgradients. That complexity around functional convex programs and stationarity is exactly why I was hesitant to fold the dual side into the first pass.

I updated the note with a Future scope section to make the split clearer: first version focuses on primal constraint violation, while leaving room for dual feasibility, complementarity, duality gap, and stationarity later if those can be defined consistently.

If this direction looks reasonable, my next step will be to turn the primal-side part into a small implementation PR and keep the dual/optimality checks as follow-up design work.

PTNobel: I see no reason to delay complimentarity and dual feasiblity. They're both super obvious how they work

PTNobel: Duality gap and stationarity it makes sense to delay

curious7-web: Got it, I think we’re on the same page now.

I’ll update the note so dual feasibility and complementarity are part of the first-pass scope along with primal violation. I’ll only leave duality gap and stationarity in the deferred section.

PTNobel:
This looks great.

curious7-web:
Sounds good, thanks for working through the scope with me. I’ll mark this as ready for review so you can decide whether the design note and scratch benchmark are worth keeping in-tree, or whether this should just be closed as a design discussion.

For the implementation, I’ll open a separate PR once I have the first-pass helper/tests in a clean state.

PTNobel:
We don't keep design docs or scratch code in tree, so I'll close this.

WilliamZhang20:
@curious7-web would you be open to me helping out?

curious7-web:
Sure, happy to coordinate. I’m planning to open the implementation PR separately — feel free to comment here or on the new PR so we don’t duplicate work.

PTNobel:
Maybe we could open an issue or discussion with checkboxes of the work to do?

WilliamZhang20:
That would be great!

curious7-web:
Great idea, @PTNobel – I opened a tracking issue here: #3418.

I’m planning to start with the violation/residual audit and the primal-side helper/API. @WilliamZhang20, feel free to claim one of the remaining audit/check items there so we can coordinate cleanly.

PR #3363, opened 2026-06-02
Deduplicate Xpress column names to allow duplicate Variable.name()

### LINKED ISSUE #3361: duplicate colnames with xpress_qpif
**Describe the bug**
The change to generate column names based on Variable.name() attributes that was added to xpress_qpif in https://github.com/cvxpy/cvxpy/pull/2957 created an implicit requirement that Variable.name() attributes need to be unique across a model. I am not aware of any cvxpy functionality with that requirement prior to this change (correct me if I am wrong).

Although this seems like a straightforward assumption, there are actually a number of model generation approaches that produce Variables with duplicate names. One example would be a scenario-based formulation where scenarios are generated from a common function and then connected via additional constraints.

Of course this can be avoided if users update how they build their models before upgrading cvxpy, but this functionality in the xpress interface doesn't seem to warrant breaking existing code.  

**To Reproduce**
Attempt to solve any cvxpy model with at least 2 variables that have the same name through xpress_qpif.py:
```
xpress.SolverError: ?1030 Error: Duplicate column names are not allowed - column 0 would have same name as column 1: my_variable_name_x_000000000
```

**Expected behavior**
Hoping for at least one of these outcomes:
 1. use of column names based on Variable.name() is controllable via some keyword that can be passed into Problem.solve(), with the behavior disabled by default
 2. column name generation is updated to deduplicate common column names with an integer counter 

**Version**
 - CVXPY Version: latest

ghackebeil:
## Description
Resolves https://github.com/cvxpy/cvxpy/issues/3361 where the column name generation added to the xpress interfaces was resulting in an xpress error for models with duplicate variables names. Adds a function to deduplicate names with a `__dup<n>` suffix.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/xpress_conif.py
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from collections.abc import Iterable
+
 import numpy as np
 
 import cvxpy.settings as s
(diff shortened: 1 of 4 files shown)


ghackebeil:
Unclear if the benchmark regressions are related to this change. It's possible, if they are using the xpress interface.

This is partly why I view the symbolic name generation as something that should be off by default. It's a debugging mechanism.

SteveDiamond (approved):
LGTM

PR #3365, opened 2026-06-04
Add dense and parametric quad_form bindings to the NLP diffengine

Transurgeon:
convert_quad_form now uses the format-string make_quad_form binding with three paths: sparse constant P (CSR), dense constant P (permuted_dense), and parametric P (fed to the dense path as a matrix-valued child that is re-evaluated each solve). Previously only constant sparse P was supported and a parametric P raised NotImplementedError.

Adds TestQuadFormDiffEngine (DerivativeChecker-based, no NLP solver required) covering dense/sparse/parametric P over a leaf variable and dense/sparse P over a composition (x[:k]); removes the obsolete test_parameterized_quad_form_errors_clearly.

Bumps the sparsediffpy pin to >=0.4.0,<0.5.0, which provides the format-string make_quad_form binding the new paths require.

## Description
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


dance858:
Context for non-sparsediff-people: we want to distinguish between sparse and dense matrices to be able to treat dense blocks as dense in the underlying diff engine. This PR adds support for quad_form(x, P) when P is a dense numpy matrix.

Questions for William:

1. Do we have a test for a nonlinear composition? Like quad_form(sin(cp.multiply(x, x)), P) where P is a dense matrix? 
2. Do we have a test for quad_form with a parameter matrix P that updates the parameter? Maybe we put that in the parameter tests file?
3. Do we have a test such as quad_form(x, cp.diag(p)) where p is a vector parameter?

I would probably also remove any comment that mentions permuted_dense, since that's an internal detail in sparsediffengine and not relevant for CVXPY developers.

SteveDiamond:
Can you update this MR to fix the CI?

Transurgeon:
> Can you update this MR to fix the CI?

the CI is failing because we need to make a new release of SparseDiffPy.
Once we merged the PR for dense quadform support: https://github.com/SparseDifferentiation/SparseDiffEngine/pull/96. The CI should be good to go again.

Transurgeon:
@dance858 we didn't have any of the tests you mentioned, I've added them now. They all pass locally, but let's make a new release of SparseDiffPy first before merging this.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -105,13 +106,18 @@ def convert_rel_entr(expr, children):
 
 
 def convert_quad_form(expr, children):
-    """Convert scalar quadratic form x.T @ P @ x."""
+    """Convert scalar quadratic form x.T @ P @ x.
+
+    Constant P uses the native sparse (CSR) or dense binding. A parametric P (affine
+    in the parameters, independent of x, so the Hessian is still 2P) is fed to the
```
PTNobel: Even if P has known sparsity?

cvxpy/settings.py
```diff
@@ -202,6 +202,11 @@
 # When problem is DPP and total parameter size >= this threshold, use COO backend
 DPP_PARAM_THRESHOLD = 1000
 
+# Nonzero fraction at/below which a constant dense array is treated as sparse: the diff
+# engine routes such a constant operand (matmul A or quad_form P) to its sparse binding
+# rather than building a dense Jacobian/Hessian.
+SPARSE_DENSITY_THRESHOLD = 0.05
```
PTNobel, line 220: @rileyjmurray any opinions on what this number should be

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: what do you mean here, are you talking about the sparsity attribute?

PTNobel: Or it is a block matrix; yeah.

Transurgeon: I think this comment might be suggesting the wrong thing, I'll remove it.
I think it's referring to a purely local Hessian (which would then be 2P in this case).

Transurgeon:
@PTNobel I updated the sparsediffpy version and now the CI is passing.
I think the 5% sparsity threshold is fine, but open to using another magic number instead.
After that we can merge this :).

PTNobel (approved):

PTNobel:
I will run some benchmarks locally to see if there's any obvious framework for this value...

PTNobel:
Quick IPOPT benchmark to calibrate `SPARSE_DENSITY_THRESHOLD`.

sweep

Forced dense vs sparse dispatch on random least-squares, median of 3 solves, n=200–1200 × density 0.005–0.30. Break-even density is ~0.13 across all sizes, so the current 0.05 leaves the 0.05–0.12 band (where sparse is still ~1.1–1.4× faster) on the dense path. Suggests bumping to ~0.10.

PTNobel:
Wait I think I tested the wrong thing

PTNobel:
I benchmarked the other use of the threshold for left matmul

PTNobel:
Corrected for the `quad_form` P dispatch — the one above benchmarked `sum_squares`, i.e. the left-matmul path.

quad_form P dispatch sweep

PTNobel:
Seems like we should have different thresholds tbh

PTNobel:
Ran more tests: constraint count doesn't affect the threshold at all

PTNobel:
Swept problem size, #constraints, and solver (IPOPT/UNO). Only solver moved the crossover, and only for the left-matmul path.

- **quad_form**: sparse-P wins to ~0.85 density; unchanged by size, #constraints, or solver.
- **left-matmul**: crossover ~0.38 (IPOPT) vs ~0.77/parity (UNO); unchanged by size or #constraints.

left-matmul IPOPT vs UNO
quad_form IPOPT vs UNO
left-matmul vs #rows
quad_form vs #constraints

Transurgeon:
@PTNobel thanks for the analysis!
I propose that we go for 10% since it covers more cases where sparse would be faster like your analysis suggests.

Overall, we want to help the user who might have mis-used numpy matrices (or maybe aren't even aware of scipy?)
for very sparse and structured matrices. This can make a big difference (between almost instant derivatives and unusable solver interface) and we hope to help the user in this case.

However, I feel like in the general case, when this difference is not as noticeable, it is probably fine to just pass to the diffengine what the user wrote (even if it might be slightly faster to use sparse, for example).

@dance858 what do you think about this? We should aim for a number that's decently broad and keep things as simple as possible. I also don't think this is a path that will get hit often, atleast I hope it won't :).

PTNobel:
I think we should always sparsify the quad_form looking at this, and use 0.10 for the other purposes tbh.

PTNobel:
One of my core design principles is that we don't require users to understand what is and isn't performance critical when we can avoid it. I'm worried that this is leaving a ton of performance on the table, especially with uno

Transurgeon:
> I think we should always sparsify the quad_form looking at this, and use 0.10 for the other purposes tbh.

can you explain the experiments you did again? Are you using the new sparsediffpy for it?
I think its a bit surprising to me that we should always sparsify, that doesn't seem correct.
Also what problem are you running the benchmark on.. would it be possible to share a script of it?

PTNobel:
Yes I can share. I'm away from my laptop right now. 

I'm not surprised that we should always sparsify when the block goes in the Hessian exactly. The solver doesn't consider dense blocks in the Hessian and instead always treats it as sparse

PTNobel:
```
Name: sparsediffpy
Version: 0.5.0
Location: /home/ptnobel/Research/cvxpy/.worktrees/pr3365-sparse-threshold-perf/.venv/lib/python3.13/site-packages
Requires: numpy
Required-by: cvxpy
```

PTNobel:
https://github.com/cvxpy/cvxpy/blob/ptn/pr3365-sparse-threshold-perf/tools/benchmark_sparse_density_threshold.py

Here's the code; I think this has all the experiments, but it is possible it doesn't and I have to dig up the others.

dance858:
That's very interesting, thank you Parth!

If the objective is just cp.quad_form(x, P) where x is a variable, the code for the hessian (which is often the most expensive part in the diff backend) is identical for the sparse and dense case (just a memcpy of Ps-values, times a scalar dual variable Ipopt specifies). So I am a bit surprised to see that one (the sparse dispatch in this case) is significantly faster than another. I'm traveling so I don't have time to look into exactly what is going on right now, but it is super interesting. I don't know if you want to look into it further @Transurgeon? At least we should open an "issue" in the sparsediffengine repository so we don't forget about it. Maybe we do something stupid on our end. 

In my experience, taking dense blocks into account really helps when there are nonlinear compositions. If we do \phi(Ax) where \phi is an elementwise atom, the first term of the Hessian is A_global^T D A_global, where D is the Hessian of \phi and A_global is the Jacobian (which is equal to A if x is the only variable in the problem). When A is a numpy dense matrix, A_global will be a matrix consisting of just one dense block. Then we internally compute A_global^T D A_global by using dense linear algebra on the dense block instead of treating A_global as a generic sparse matrix. On the trimmed logistic regression example (section 9.3 of https://arxiv.org/abs/2606.02896), for example, this gave a speedup of a factor 10 in total solve time, because computing the derivatives was the bottleneck on that problem. 

But since the Hessian of just cp.quad_form(x, P) is so simple, I would expect the sparse dispatch and the dense dispatch to be as fast. So these results are very interesting.

Note that in the end, we always give the solver a sparse matrix for the Jacobian and Hessian. So the solver sees the exact same input in the sparse and dense branches.

PTNobel:
Wouldn't you have explicit zeros in the dense case? 

I'll try and spend some time debugging this.

dance858:
> Wouldn't you have explicit zeros in the dense case?

Oh yes that's right!  If you have a sparse matrix with zeros and you convert it to dense before the diff backend, you'll have explicit zeros. But if you have a dense matrix and treat it as sparse, you'll have the same number of nonzeros in the two branches.

PTNobel:
Yes, my experiments were adjusting the density of the matrix and then comparing the two codepaths.

Transurgeon:
@PTNobel and @dance858 I opened an issue on the diff engine. Let's come back to the bindings if we realise that the sparse path is truly better than dense in all these cases (but we will first try to speedup the dense path when it makes sense). But like I said before I think this code is just to catch some edge cases of sparse data being passed as a dense matrix, so this analysis seems fitting for a follow-up task.

I'll merge this for now so I can keep iterating on the ignore_dpp backend.

PR #3366, opened 2026-06-04
Route super-sparse constant left-matmul to the sparse diffengine binding

Transurgeon:
A constant dense matmul operand stored as a numpy array but mostly zeros (e.g. the murray benchmark, ~0.2% nonzero for a difference matrix, [link](https://github.com/cvxpy/benchmarks/blob/main/benchmark/gini_portfolio.py#L50-L91)) was always sent down the dense permuted_dense path, which materializes a fully dense Jacobian/Hessian and is orders of magnitude slower and heavier than the sparse CSR binding.

convert_matmul now checks the nonzero density of a constant dense left operand and, when it falls below SPARSE_MATMUL_DENSITY_THRESHOLD (5%), converts it to CSR and dispatches to the existing sparse left-matmul path. The check is gated on pure constants (param_node is None): sparsifying a parametric matrix would freeze its sparsity pattern to the current value.

The density count runs once per matmul node at C-graph construction time, not per solver iteration.

## Description
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.


PTNobel:
> The check is gated on pure constants (param_node is None): sparsifying a parametric matrix would freeze its sparsity pattern to the current value.

Should we add a warning or something telling the user if they're using a very sparse matrix that things can be faster if they tell us the sparsity pattern? Maybe we could hash the sparsity pattern and if it isn't changed between two runs we warn?

PTNobel:
Could the same optimization be applied to the other paths?

Edit: I guess not really... it would just be a question of if we should convert to sparse earlier in the pipeline.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
```diff
@@ -29,6 +30,11 @@
 )
 from cvxpy.reductions.solvers.nlp_solvers.diff_engine.registry import ATOM_CONVERTERS
 
+# A constant dense matmul operand whose nonzero fraction falls below this threshold is
+# routed to the sparse (CSR) binding instead of the dense permuted_dense path, which would
+# otherwise build a dense Jacobian/Hessian and be orders of magnitude slower.
+SPARSE_MATMUL_DENSITY_THRESHOLD = 0.05
```
PTNobel: Shouldn't this go in cvxpy.settings or something? I don't think we do magic constants floating around much.

Also I swear we have a closely related constant already...

cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_matmul_sparse.py
```diff
@@ -127,3 +127,87 @@ def test_sparse_and_dense_matmul2(self):
...
+    @staticmethod
+    def _hessian_lower_nnz(prob, check=False):
+        checker = DerivativeChecker(prob)
+        if check:
+            assert all(checker.run().values())
+        else:
+            checker._init_coo()
+        return len(checker.hess_rows)
+
+    def test_super_sparse_dense_routes_to_sparse(self):
```
PTNobel: None of your tests refer to the constant? How are we testing it?

PTNobel (commented):
Overall, looks great. Nice perf opt.

cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_matmul_sparse.py
Transurgeon: changed the tests to refer to the constant.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
Transurgeon: I took another look and we don't. I moved the constant to cvxpy.settings.

Transurgeon:
> Could the same optimization be applied to the other paths?

Do you mean other backends? I took a look at that and they perform well because they treat the problem data as sparse from the beginning.

> Edit: I guess not really... it would just be a question of if we should convert to sparse earlier in the pipeline.

Also to clarify, this is to catch "extreme" cases where the user defined a dense matrix (like the Murray benchmark.. where it is a super large difference matrix in numpy) which is actually sparse. By catching this upstream before converting it to C, it allows us to do sparse matmul.

Transurgeon:
@PTNobel @dance858 I cleaned up this PR, should be ready to merge now.
I left the parameter warning for now because using ``sparsity=...`` makes CVXPY do the reformulation to ``sparse_coeff @ dense_param_vec``.
But maybe it could still be worth raising a warning so that the user is aware of this API?
It just isn't something that we can easily "control" (i.e. convert ourselves to a sparse matmul instead of dense).
I hope that makes sense?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
```diff
@@ -44,6 +46,13 @@ def convert_matmul(expr, children, var_dict, n_vars, param_dict):
         param_node = children[0] if left_arg.parameters() else None
         if sparse.issparse(A):
             return make_sparse_left_matmul(param_node, children[1], A)
+        # A constant dense matrix that is mostly zeros: route it to the sparse CSR binding
+        # to avoid building a dense Jacobian/Hessian. Restricted to constants
+        # (param_node is None) because sparsifying a parametric matrix would freeze its
+        # sparsity pattern to the current value.
+        density = np.count_nonzero(A) / A.size if A.size else 1.0
+        if param_node is None and density < s.SPARSE_MATMUL_DENSITY_THRESHOLD:
+            return make_sparse_left_matmul(None, children[1], sparse.csr_matrix(A))
```
PTNobel: We are trying to migrate to the sparray over spmatrix. What's going on here?

PTNobel (commented):
https://github.com/cvxpy/cvxpy/blob/master/cvxpy%2Fexpressions%2Fleaf.py#L151

Ah ha. We did have this. Should this use the same setting? They're measuring the same thing I think

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
Transurgeon: that's a good point, I will change it. But also I want to say that its probably fine because the ``make_sparse_left_matmul`` function is only used to pass the indices to C. There are no more downstream usages of the scipy matrix after this.

Transurgeon:
> https://github.com/cvxpy/cvxpy/blob/master/cvxpy%2Fexpressions%2Fleaf.py#L151
> 
> Ah ha. We did have this. Should this use the same setting? They're measuring the same thing I think

yeah nice catch to remember this usage! I can change the code here to use the new constant in the settings.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
PTNobel: Yeah, it's mostly a style thing here

Transurgeon:
@PTNobel could you have another look please? I think we can merge this and edit the constant value (if needed) in the quadform bindings PR since it also uses it.

PTNobel (approved):

PR #3375, opened 2026-06-10
Pass lower Hessian triangle to HiGHS QP interface

SteveDiamond:
## Description
The HiGHS QP interface passes `sp.triu(P)` with `HessianFormat.kTriangular`, but HiGHS' triangular Hessian format requires the **lower** triangle. highspy < 1.14.0 silently drops all upper-triangle off-diagonal entries and reports `kOptimal` for the wrong (diagonal-Hessian) problem, so any QP with cross terms in the quadratic returns a silently wrong answer:

```python
x = cp.Variable(2)
P = np.array([[2., 1.], [1., 2.]])
prob = cp.Problem(cp.Minimize(cp.quad_form(x, P) - 2*cp.sum(x)))
prob.solve(solver=cp.HIGHS)
print(prob.value)   # -0.5 before this fix; true optimum -2/3 (CLARABEL/OSQP/SCS agree)
```

highspy >= 1.14.0 accepts either triangle, so passing the lower triangle is correct on all versions. With highspy 1.12 installed, 6 existing tests in `test_qp_solvers.py` fail before this change and pass after. Introduced in #3302.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
+++ b/cvxpy/reductions/solvers/qp_solvers/highs_qpif.py
@@ -197,11 +197,14 @@ class HIGHS(QpSolver):
             # entries whose upper and lower triangles may differ by floating-
             # point epsilon after canonicalization, which HiGHS >= 1.14.0
             # rejects as asymmetric.  See https://github.com/cvxpy/cvxpy/issues/3301
-            P_upper = sp.triu(P, format="csc")
+            # HiGHS' triangular Hessian format requires the LOWER triangle:
+            # highspy < 1.14.0 silently drops upper-triangle off-diagonal
+            # entries, while highspy >= 1.14.0 accepts either triangle.
+            P_lower = sp.tril(P, format="csc")
             hessian.format_ = hp.HessianFormat.kTriangular
-            hessian.start_ = P_upper.indptr
-            hessian.index_ = P_upper.indices
-            hessian.value_ = P_upper.data
+            hessian.start_ = P_lower.indptr
+            hessian.index_ = P_lower.indices
+            hessian.value_ = P_lower.data
 
         solver = hp.Highs()
 
--- a/cvxpy/tests/test_qp_solvers.py
+++ b/cvxpy/tests/test_qp_solvers.py
@@ -611,6 +611,22 @@ def test_highs_cvar():
     assert problem.status == cp.OPTIMAL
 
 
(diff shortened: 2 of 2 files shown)


PTNobel:
We bumped the HiGHS minimum version because of this bug. I don't see any reason to support old HiGHS versions.

SteveDiamond:
**Verification that the test fails without the fix** (per review practice of test-first evidence — the branch history bundles test+fix in one commit, so here is the split run):

Applying only this PR's test file onto current master (`6a27fc27f`) and running the new test:
```
$ git checkout master && git checkout fix-highs-qp-hessian-triangle -- cvxpy/tests/test_qp_solvers.py
$ pytest cvxpy/tests/test_qp_solvers.py::test_highs_hessian_cross_terms
FAILED - AssertionError: HIGHS returns -0.5 for the cross-term QP; analytic optimum is -2/3
1 failed in 0.39s
```
With the fix (this branch): `1 passed`. Additionally, 6 pre-existing tests in `test_qp_solvers.py` fail on master with highspy 1.12 installed and pass with this change.

SteveDiamond:

SteveDiamond:
Note on the test-first run for this PR: the test-only head came back **all green**, as expected — CI installs the dev extra, which pins `highspy >= 1.14`, and HiGHS 1.14+ accepts either Hessian triangle, so the bug cannot reproduce on CI. The bug only affects the unpinned `cvxpy[HIGHS]` extra / older highspy installs (e.g. highspy 1.12, where this PR's new test fails and 6 pre-existing tests in `test_qp_solvers.py` also fail; all pass with the fix):

```
$ pip show highspy | grep Version   # 1.12.0
$ pytest cvxpy/tests/test_qp_solvers.py::test_highs_hessian_cross_terms   # on master + test only
FAILED - AssertionError: HIGHS returns -0.5; analytic optimum is -2/3 (CLARABEL/OSQP/SCS agree)
```

Passing `sp.tril` is correct on every highspy version, so this fix closes the gap for old installs without affecting new ones. A runtime version floor for highspy (or a pin in the `HIGHS` extra) would be a reasonable follow-up if preferred.

PTNobel:
My preference is to just close this PR

Transurgeon:
yeah I agree with Parth, we shouldn't update the interface to support older versions.

PR #3376, opened 2026-06-10
Fix lambda_sum_largest k-doubling for real arguments in Complex2Real

SteveDiamond:
## Description
`lambda_sum_largest_canon` in the Complex2Real reduction doubles `k` unconditionally, but the matching `real /= 2` correction (and the 2n×2n Hermitian dilation that makes the doubling correct) only applies when the argument has an imaginary part. Since Complex2Real runs over the whole problem, applying `lambda_sum_largest` to a real symmetric matrix in any problem containing an unrelated complex expression rebuilt the atom with `2k` on the original n×n matrix — silently wrong results, or spurious `unbounded` when `2k > n`:

```python
X = cp.Variable((3,3), symmetric=True)
z = cp.Variable(complex=True)                       # unrelated complex leaf
p = cp.Problem(cp.Maximize(cp.trace(X)),
               [cp.lambda_sum_largest(X, 2) <= 5, X >> 0, X << 10*np.eye(3),
                cp.abs(z) <= 1])
p.solve(solver=cp.CLARABEL)
# before: 30.0 with lambda_sum_largest(X, 2) == 20 (constraint violated)
# after:  7.5, matching the same problem without z
```

The fix moves `real.k *= 2` inside the `imag_args[0] is not None` branch.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
+++ b/cvxpy/reductions/complex2real/canonicalizers/matrix_canon.py
@@ -117,10 +117,12 @@ def lambda_sum_largest_canon(expr: lambda_sum_largest,
                              imag_args: list[Expression | None], real2imag):
     """Canonicalize sum of k largest eigenvalues with Hermitian matrix input.
     """
-    # Divide by two because each eigenvalue is repeated twice.
     real, imag = hermitian_canon(expr, real_args, imag_args, real2imag)
-    real.k *= 2
     if imag_args[0] is not None:
+        # The Hermitian dilation repeats each eigenvalue twice, so sum the
+        # 2k largest eigenvalues and divide by two. A real argument is left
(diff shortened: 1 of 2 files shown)


PTNobel (approved):

SteveDiamond:
**Verification that the test fails without the fix** (per review practice of test-first evidence — the branch history bundles test+fix in one commit, so here is the split run):

Applying only this PR's test file onto current master (`6a27fc27f`) and running the new test:
```
$ git checkout master && git checkout fix-lambda-sum-largest-complex -- cvxpy/tests/test_complex.py
$ pytest cvxpy/tests/test_complex.py::TestComplex::test_lambda_sum_largest_real_arg
E   AssertionError: 30.00000005323264 != 7.499999998339203 within 4 places
1 failed in 0.39s
```
With the fix (this branch): `1 passed`.

PR #3378, opened 2026-06-10
Apply ATOM_EVAL_TOL near-integer rounding in floor.numeric

SteveDiamond:
## Description
`ceil.numeric` rounds near-integer inputs within `ATOM_EVAL_TOL` before applying `np.ceil` (added in #1300), but `floor.numeric` never received the same treatment:

```python
cp.ceil(cp.Constant(3 + 1e-10)).value    # 3.0 (tolerant)
cp.floor(cp.Constant(3 - 1e-10)).value   # 2.0 before this fix
```

Since solvers return solutions with ~1e-9 noise, DQCP objective values containing `floor` were off by one whenever the optimum sat at an integer. This applies the identical rounding used by `ceil`.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/ceil.py
+++ b/cvxpy/atoms/elementwise/ceil.py
@@ -108,7 +108,8 @@ class floor(Elementwise):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        return np.floor(values[0])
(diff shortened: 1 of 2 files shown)


SteveDiamond:

PTNobel:
Duplicate of #3397

PR #3379, opened 2026-06-10
Add missing numpy_numeric decorator to entr.numeric

SteveDiamond:
## Description
`entr` is the only elementwise atom whose `numeric` is not decorated with `@Elementwise.numpy_numeric`, which densifies sparse inputs. Sparse constants therefore reached `scipy.special.xlogy` undensified and crashed:

```python
c = cp.Constant(sp.csc_array(np.array([[0.5, 0.], [0., 2.0]])))
cp.entr(c).value
# TypeError: ufunc 'xlogy' not supported for the input types...
# (cp.exp(c).value works fine on the same input)
```

## Type of change
- [x] Bug fix

- [x] Write unittests.


SteveDiamond:

PTNobel (approved):

PR #3380, opened 2026-06-10
Fix IndexError in Elementwise.is_symmetric for scalar and 1-D atoms

SteveDiamond:
## Description
`Elementwise.is_symmetric` indexes `self.shape[0] == self.shape[1]` without an ndim check, so it crashes for every scalar or 1-D elementwise atom:

```python
cp.abs(cp.Variable()).is_symmetric()   # IndexError: tuple index out of range
```

Now square 2-D atoms keep the argument-based check, and everything else falls back to the `Expression` default (`True` only for scalars, matching `cp.Variable().is_symmetric()`).

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/elementwise.py
+++ b/cvxpy/atoms/elementwise/elementwise.py
@@ -42,8 +42,10 @@ class Elementwise(Atom):
     def is_symmetric(self) -> bool:
         """Is the expression symmetric?
         """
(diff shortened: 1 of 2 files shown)


SteveDiamond:

PTNobel (approved):

PR #3381, opened 2026-06-10
Fix sign of variables with a partial boolean index list

SteveDiamond:
## Description
`Leaf.is_nonneg` returns `... or self.attributes['boolean']`, and the boolean attribute may be a list of indices constraining only some entries. The truthy list made the whole variable NONNEGATIVE (and `is_nonneg()` returned the list itself instead of a bool), corrupting DCP analysis:

```python
x = cp.Variable(2, boolean=[(0,)])      # only x[0] boolean; x[1] unconstrained
x.is_nonneg(), x.sign                   # [(0,)], NONNEGATIVE — wrong

f = cp.abs(cp.maximum(x[1], -1.0))      # nonconvex, accepted as DCP via the bogus sign
prob = cp.Problem(cp.Minimize(f), [x[1] <= -3, x[0] == 0])
prob.solve(solver=cp.HIGHS)             # 'infeasible' — truly feasible (x=[0,-3], f=1)
```

A partially-boolean variable now has unknown sign; fully boolean variables (`boolean=True`) remain nonnegative. The `is True` check matches the existing precedent in the variable-bounds path.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -285,8 +285,10 @@ class Leaf(expression.Expression):
 
     def is_nonneg(self) -> bool:
         """Is the expression nonnegative?"""
-        return (self.attributes['nonneg'] or self.attributes['pos'] or
-                self.attributes['boolean'])
+        # The boolean attribute may be an index list constraining only some
+        # entries, in which case the leaf's sign is unknown.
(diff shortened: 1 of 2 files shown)


SteveDiamond:

PTNobel (approved):

PR #3383, opened 2026-06-10
Register DGP quad_form canonicalizer under the QuadForm class

SteveDiamond:
## Description
The DGP `CANON_METHODS` table keyed `quad_form_canon` on the `quad_form` factory **function**, but `Dgp2Dcp.canonicalize_expr` dispatches on `type(expr)`, which is the `QuadForm` **class** — the only non-class key in the table. So the canonicalizer was unreachable: `Dgp2Dcp` accepted the atom (`QuadForm.is_atom_log_log_convex` is True), then fell through to `expr.copy(args)` with log-space arguments, producing `QuadForm(u, log(P))`:

```python
P = np.array([[np.exp(4), 1.0], [1.0, np.exp(1)]])   # log(P) is PSD
x = cp.Variable(2, pos=True)
prob = cp.Problem(cp.Minimize(cp.quad_form(x, P)), [x[0]*x[1] >= 1.0])
prob.solve(gp=True, solver=cp.CLARABEL)
# before: 59.32 (silently wrong) — true GP optimum 26.365
```

When elementwise `log(P)` is not PSD, it instead crashed with `Cannot reduce problem to cone program` despite `is_dgp()` being True.

`quad_form_canon` itself also imported the `hstack` *module* instead of the atom, so it would have crashed on first use — confirming the path was never executed. Both fixed, with the explicit log-space formulation as the test oracle.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/dgp2dcp/canonicalizers/__init__.py
+++ b/cvxpy/reductions/dgp2dcp/canonicalizers/__init__.py
@@ -23,7 +23,7 @@ from cvxpy.atoms.one_minus_pos import one_minus_pos
 from cvxpy.atoms.pf_eigenvalue import pf_eigenvalue
 from cvxpy.atoms.pnorm import Pnorm, PnormApprox
 from cvxpy.atoms.prod import Prod
-from cvxpy.atoms.quad_form import quad_form
+from cvxpy.atoms.quad_form import QuadForm
 from cvxpy.atoms.quad_over_lin import quad_over_lin
 from cvxpy.constraints.finite_set import FiniteSet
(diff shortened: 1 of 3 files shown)


PTNobel (commented):
Can we add a test that enumerates all canonicalizer maps and asserts they are subclasses of Atom? I've seen this bug a few times in other areas too

SteveDiamond:

PTNobel:
I think we should close this in favor of #3406

PTNobel:
Duplicate of #3406

PR #3386, opened 2026-06-10
Fix np.matmul semantics for 1-D operands in batched matmul

PTNobel:
## Description
Bug: in N-D (batched) matmul, a 1-D operand violated np.matmul
semantics. w (3,) @ A (2,3,4) silently returned shape (2,2,4) with
wrong values, and A (2,3,4) @ v (4,) raised ValueError; numpy gives
(2,4) and (2,3) respectively.

Root cause: MulExpression._broadcast_batch_dims treated a 1-D operand
as a matrix with empty batch dims and broadcast it along the batch
dimensions, instead of applying numpy's documented promotion rule
(prepend/append a 1, matmul, then drop the inserted axis).

Fix: _broadcast_batch_dims now promotes a 1-D operand to 2-D via
reshape ((k,) -> (1,k) for the lhs, (k,) -> (k,1) for the rhs) before
batch broadcasting, and the matmul() entry point (now used by
__matmul__/__rmatmul__) removes the inserted axis from the result.
Shapes, values, and the canonicalization/solve path all match numpy.

Tests: added TestNDRmulEdgeCases::test_1d_operand_batched covering
constant values for (3,)@(2,3,4), (2,3,4)@(4,), and 4-D batches, plus
solve-path checks with 1-D variables on either side, for the SCIPY and
COO backends.

## Type of change
- [x] Bug fix

- [x] Write unittests.


cvxpy/atoms/affine/binary_operators.py
```diff
@@ -119,8 +119,20 @@ def bounds_from_args(self) -> tuple[np.ndarray, np.ndarray]:
         return bounds_utils.mul_bounds(lb1, ub1, lb2, ub2)
 
 
-def matmul(lh_exp, rh_exp) -> "MulExpression":
+def matmul(lh_exp, rh_exp) -> Expression:
     """Matrix multiplication."""
+    lh_exp = Expression.cast_to_const(lh_exp)
```
PTNobel: Should be Expression.const

rileyjmurray: Expression.cast?

PTNobel: Yes that's what I meant I promise

PTNobel:
```suggestion
    lh_exp = Expression.cast(lh_exp)
```

cvxpy/atoms/affine/binary_operators.py
```diff
@@ -119,8 +119,20 @@ def bounds_from_args(self) -> tuple[np.ndarray, np.ndarray]:
         return bounds_utils.mul_bounds(lb1, ub1, lb2, ub2)
 
 
-def matmul(lh_exp, rh_exp) -> "MulExpression":
+def matmul(lh_exp, rh_exp) -> Expression:
     """Matrix multiplication."""
+    lh_exp = Expression.cast(lh_exp)
+    rh_exp = Expression.cast_to_const(rh_exp)
```
PTNobel:
```suggestion
    rh_exp = Expression.cast(rh_exp)
```

Transurgeon:
I had Claude review this PR and these are the comments:

**1. `cvxpy/expressions/expression.py:841` — the deprecated `*`-as-matmul path bypasses the new squeeze (should fix)**

The star-matmul branch in `__mul__` still calls `cvxtypes.matmul_expr()` directly, so `*` and `@` now disagree for 1-D operands in batched matmul: with `X = cp.Variable((2,3,4))` and `v = np.arange(4.0)`, `X * v` yields shape `(2,3,1)` while `X @ v` yields `(2,3)`. Since the deprecation warning tells users the two spellings are equivalent, migrating `*` → `@` silently changes result shapes. One-line fix:

```python
return cvxtypes.matmul()(self, other)
```

**2. `MulExpression` docstring vs. direct construction**

With the squeeze living only in the `matmul()` wrapper, directly constructing `MulExpression` (which is publicly exported) keeps the promoted singleton axis: `MulExpression(w, A3)` gives `(2,1,4)` where `np.matmul` and `@` give `(2,4)`. That makes the docstring's "exactly as those of NumPy's matmul" claim false for ND inputs — and it can bite silently, e.g. `y == MulExpression(w, A3)` with `y` of shape `(2,4)` broadcasts into a `(2,2,4)` constraint with no error. Could we either note the promoted-axis behavior in the docstring and point callers at `matmul()`, or treat direct construction as internal-only? (All in-tree callers of `matmul_expr()` are ≤2-D, so this is about the public/third-party surface.)

**3. `matmul()` wrapper — consider reusing `u.shape.mul_shapes`**

The axis-dropping arithmetic re-derives what `u.shape.mul_shapes` (`cvxpy/utilities/shape.py`) already encodes — it's the single source for np.matmul shape semantics (promotion, batch broadcast, removal of the inserted axes) and is what `shape_from_args` uses. Computing the target shape once from the original operand shapes keeps the rule in one place and drops the `max(ndim) > 2` special-casing (in the ≤2-D case the reshape is a no-op):

```python
def matmul(lh_exp, rh_exp) -> Expression:
    """Matrix multiplication."""
    lh_exp = Expression.cast(lh_exp)
    rh_exp = Expression.cast(rh_exp)
    prod = MulExpression(lh_exp, rh_exp)
    target_shape = u.shape.mul_shapes(lh_exp.shape, rh_exp.shape)
    if prod.shape != target_shape:
        # A 1-D operand was promoted per np.matmul semantics; drop the inserted axis.
        prod = reshape(prod, target_shape, order='F')
    return prod
```

Otherwise the "which axis gets dropped" logic is duplicated between here and `mul_shapes` and the two copies can drift.

**4. Constant 1-D operands and the CPP backend**

For a *constant* 1-D operand, `_broadcast_batch_dims` wraps the `Constant` in `reshape` and (usually) `broadcast_to` expression atoms. Since `broadcast_to._supports_cpp()` is `False`, a plain `w @ A3` with numpy inputs is forced off the default CPP backend, and every non-cached canonicalization carries the extra affine atoms. When the operand `is_constant()`, building a new `Constant` from `np.reshape`/`np.broadcast_to` (O(1) views) would leave a plain leaf in the tree and keep CPP eligibility. Not blocking, but it affects the most common data-on-one-side case.

**5. Error messages show internal shapes (minor)**

Because promotion and batch broadcasting happen before shape validation, a mismatch like `np.ones(5) @ A` with `A` of shape `(2,3,4)` now raises "Incompatible dimensions (2, 1, 5) (2, 3, 4)" — neither shape appears in the user's code (they wrote `(5,)`). Validating the inner-dimension match against the original shapes before promoting would keep the diagnostic actionable.

**6. Test nit**

In `test_1d_operand_batched`, the two solve-path blocks are mirror copies with renamed variables; the constant checks above already use a compact `for lh, rh in [...]` loop and the solve blocks could fold into the same pattern. Also worth adding one case with a 1-D `Parameter` operand — it works today (I checked SCIPY and COO), but nothing in the suite pins that down, and it exercises the parametric reshape path in the backends.

Transurgeon (approved):

PR #3389, opened 2026-06-10
Only merge positive-ratio rows in CVXOPT inequality presolve

PTNobel:
## Description
compress_matrix, used by the CVXOPT interface to drop redundant
inequality rows when the equality matrix is rank deficient, treated a
row as redundant whenever it was any scalar multiple of a kept row with
a matching right-hand side ratio. For inequalities Gx <= h a negative
multiple describes the opposite half-space (e.g. z >= 1 stored as
-z <= -1 versus z <= 1), so dropping it enlarges the feasible set:
min z s.t. z <= 1, z >= 1 plus duplicated equalities was reported
unbounded instead of optimal at 1.

Require a positive ratio before merging rows. compress_matrix's only
caller compresses the LEQ block of G in cvxopt_conif.py, so no equality
usage relies on merging negative multiples.

Tests: unit test that compress_matrix keeps negative-multiple rows, and
a CVXOPT regression test for the formerly-unbounded problem.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/compr_matrix.py
+++ b/cvxpy/reductions/solvers/compr_matrix.py
@@ -29,7 +29,8 @@ def get_row_nnz(mat, row):
 def compress_matrix(A, b, equil_eps: float = 1e-10):
     """Compresses A and b by eliminating redundant rows.
 
-    Identifies rows that are multiples of another row.
+    Identifies rows that are positive multiples of another row,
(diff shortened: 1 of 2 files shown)


PTNobel:
Folded into #3390, which now carries both commits (this fix plus the dependent dual-recovery fix) against master.

PR #3391, opened 2026-06-10
Fix COPT cone ordering and constraint-dual alignment for PSD problems

PTNobel:
## Description
Two COPT-interface dual/cone-bookkeeping bugs, one commit each.

**Commit 1 — Fix COPT cone-order mismatch for problems with PSD and exp cones**
CVXPY stacks constraint rows in SCS order (zero, nonneg, SOC, PSD,
exp), but COPT's loadConeMatrix expects (zero, nonneg, SOC, exp, PSD),
as the _add_psd_bound_rows docstring already noted. The PSD path passed
A and b straight through, so exponential cone rows were interpreted as
PSD entries and problems mixing PSD and exp cones returned garbage
(e.g. "unbounded" for a bounded problem).

Permute the stacked rows so the exp block precedes the PSD block before
calling loadConeMatrix, and apply the inverse permutation to the
recovered dual vector (the dualized model's variables follow COPT's row
order). The rowmap-based primal recovery indexes columns of A and is
unaffected by the row permutation.

Add regression tests for PSD+exp (values and duals, verified against
CLARABEL) and PSD+SOC+exp; existing PSD-only and exp-only COPT tests
cover non-regression of those paths.

**Commit 2 — Fix COPT constraint-dual misalignment for PSD problems with bounds**
Bug: for PSD problems with finite variable bounds, COPT returned wrong
dual_values for every cone constraint after the NonNeg block (SOC, PSD,
exp). Primal values were unaffected.

Root cause: the PSD path adds variable bounds as explicit inequality
rows appended to the NonNeg block (loadConeMatrix has no bounds).
The dual recovery returned those bound-row duals inside INEQ_DUAL, so
the contiguous walk in utilities.get_dual_values read every subsequent
constraint's dual at an offset shifted by the number of bound rows.

Fix: _add_psd_bound_rows now reports how many rows it inserted, and the
dual recovery drops those entries from the dual vector before splitting
into EQ_DUAL/INEQ_DUAL.

Tests: two regressions in TestCOPT (PSD+bounds and SOC+PSD+bounds)
checking dual_values against analytic KKT solutions; both fail without
the fix. Verified duals also match CLARABEL on the same problems.

Found by an automated bug-hunting audit of master; each finding was reproduced against an independent oracle before fixing (commit 2's tests fail with all-zero / shifted PSD duals without the fix).

## Type of change
- [x] Bug fix

- [x] Write unittests.


PTNobel:
I think we should expand our tests for everything based on this bug not just test COPT more.

PR #3392, opened 2026-06-10
Fix sparse symmetry checks to compare entry positions

PTNobel:
## Description
is_sparse_symmetric and is_sparse_skew_symmetric in
cvxpy/interface/matrix_utilities.py sorted the strictly-lower and
strictly-upper triangle values independently and compared only the
value arrays, never the positions. Any asymmetric sparse matrix whose
triangle value multisets matched was misclassified as symmetric,
Hermitian, or skew-symmetric, e.g. A with A[1,0]=5, A[0,2]=5 passed
is_sparse_symmetric and quad_form accepted it while the dense path
correctly rejected it.

After the existing lexicographic sort, also require the lower-triangle
coordinates to equal the transposed upper-triangle coordinates before
comparing values. Sum duplicate COO entries and eliminate explicit
zeros first so structurally redundant but genuinely (skew-)symmetric
matrices still classify correctly.

Adds a regression test covering the asymmetric symmetric/Hermitian/
skew repros, duplicate and explicit-zero inputs, and a randomized
sparse-vs-dense agreement battery.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/interface/matrix_utilities.py
+++ b/cvxpy/interface/matrix_utilities.py
@@ -332,8 +332,9 @@ def is_sparse_symmetric(m, complex: bool = False) -> bool:
     if m.shape[0] != m.shape[1]:
         raise ValueError('m must be a square matrix')
 
-    if not isinstance(m, sp.coo_array):
-        m = sp.coo_array(m)
+    m = sp.coo_array(m)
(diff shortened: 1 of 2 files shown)


Transurgeon (approved):
nice!

PR #3394, opened 2026-06-10
Fix cp.moveaxis to handle negative and scalar axes like numpy

PTNobel:
## Description
cp.moveaxis produced wrong shapes/values for negative axes (e.g.
source=[0], destination=[-1] gave (3,2,4) instead of (3,4,2) for a
(2,3,4) input) and raised an opaque ValueError for negative sources.

Root cause: source/destination were never normalized to non-negative
axes, so `n not in source` failed to exclude negative sources and
`order.insert(dest, src)` with a negative dest inserted at the wrong
position. Scalar axes were also rejected, unlike np.moveaxis.

Fix: normalize both arguments with numpy's public
np.lib.array_utils.normalize_axis_tuple (numpy >= 2.0 is required),
mirroring np.moveaxis exactly, including AxisError for out-of-range
axes, ValueError for duplicates, and acceptance of scalar/tuple axes.

Tests: parametrized regression tests in TestND_Expressions compare
shape and value against np.moveaxis for negative/scalar/identity moves
on 3-D and 4-D constants, assert error messages match numpy for
out-of-range, duplicate, and mismatched-length axes, and check a
CLARABEL solve through a moveaxis constraint.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/transpose.py
+++ b/cvxpy/atoms/affine/transpose.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import numpy as np
+from numpy.lib.array_utils import normalize_axis_tuple
 
 import cvxpy.lin_ops.lin_op as lo
 import cvxpy.lin_ops.lin_utils as lu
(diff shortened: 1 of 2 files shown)


Transurgeon (approved):
nice!

PR #3397, opened 2026-06-10
Apply near-integer rounding tolerance in floor.numeric

PTNobel:
## Description
ceil.numeric rounds its input to ATOM_EVAL_TOL decimals before
ceiling (added in #1300 to absorb DQCP solver noise), but
floor.numeric in the same module was never given the same treatment.
As a result cp.floor(x).value at x = 2.9999999999 returned 2.0 while
ceil returned 3.0, so DQCP objective values containing floor were off
by one under ordinary solver noise.

Apply the identical np.around(..., decimals) step in floor.numeric.
ceil and floor are the only numeric() implementations in the
ceil/floor family; no other atom in the module has the omission.

Add a test next to the existing ceil/floor DQCP tests checking that
values within 1e-10 of an integer floor/ceil to that integer while
exact integers and clearly-non-integer values are unchanged.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/elementwise/ceil.py
+++ b/cvxpy/atoms/elementwise/ceil.py
@@ -108,7 +108,8 @@ class floor(Elementwise):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
-        return np.floor(values[0])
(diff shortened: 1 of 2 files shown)


Transurgeon (approved):
nice.

PR #3398, opened 2026-06-10
Fix DPP AssertionError for cumsum of parameter-only expressions

PTNobel:
## Description
Bug: a DPP problem containing cumsum(p) @ x, with p a Parameter,
passed is_dpp() but crashed during canonicalization with a bare
AssertionError (coeff_extractor.py: assert affine_expr.is_dpp()).
The same crash occurred in constraint position.

Root cause: cumsum_canon unconditionally replaced cumsum(X) with a
fresh auxiliary Variable defined via difference constraints, even when
X contained no variables. That turned the parameter-affine factor
cumsum(p) into a Variable, so cumsum(p) @ x became Variable @ Variable,
which is not DPP-representable.

Fix: when the cumsum argument contains no variables (constant or
parameter-affine), keep the expression parameter-affine by applying an
explicit lower-triangular ones operator along the cumsum axis
(I_post kron L kron I_pre on the F-order vectorization, which handles
1-D, 2-D, and N-D arguments and all axis values including None). The
existing variable path is unchanged, so variable arguments still avoid
dense triangular matrices; parameter-only arguments are typically
small, so the O(n^2) operator is an acceptable trade-off.

Audited the other eliminate_pwl canonicalizers that introduce
auxiliary variables (cummax etc.); they are nonlinear atoms whose
epigraph variables remain valid for parameter-only arguments, and no
concrete failure could be reproduced, so they are left unchanged.

Tests: new regression tests in cvxpy/tests/test_dpp.py cover the
objective and constraint positions, parameter re-solve (DPP cache),
and 2-D parameter cumsum along axis 0, 1, and None.

## Follow-up fix (second commit)

The new canon path exposed a pre-existing master bug in the SCIPY and
COO canonicalization backends (the test_backends CI jobs failed with a
silently wrong objective, -0.25 instead of -11.5): when reshaping a
parametric constant from column to matrix format for matmul stuffing,
both backends deduplicated entries by param_idx alone, keeping a
single entry per parameter slice. That is only correct for broadcast
copies. A parametric expression such as A @ p has several genuine
entries per parameter slice, all of which must be kept — e.g. on
master, ((A @ p) @ x) is stuffed with wrong coefficients on both
Python backends, independent of this PR.

The second commit dedups only when a broadcast factor was divided out,
keyed on the (param_idx, position) pair, and adds backend regression
tests for (A @ p) @ x and broadcast_to(p, (2, 2)) @ x.

## Type of change
- [x] Bug fix

- [x] Write unittests.


Transurgeon:
I had Claude review this PR and these are the comments:

**1. `cumsum_canon.py` — dense O(n²) operator is a measured perf/memory regression, and the sparse constructor already exists (must-fix)**

The no-variables path builds a dense `n×n` `np.tril(np.ones(...))` before sparsifying, replacing the old O(n) Variable+constraints canonicalization with an O(n²) one. Measured on this branch: canonicalizing `x >= cumsum(p)` takes ~4.2s at n=3000 and ~104s at n=10000 (vs ~0.01s before), and around n≥30000 the dense intermediate (~7 GB) OOMs. There's no size guard behind the "typically small" assumption, and parameter-free constants take this path too.

The good news: the sparse operator already exists in `cvxpy/atoms/affine/cumsum.py` — `_sparse_triu_ones` (written explicitly "to avoid allocating a dense dim x dim matrix") and `_grad`'s `sp.kron(sp.kron(eye(post), triu), eye(pre))`; the operator built here is exactly that gradient's transpose. Sharing one `_cumsum_axis_op` helper between the atom and this canon fixes the regression and keeps the two constructions from drifting in axis/F-order conventions.

**2. `coo_backend.py` ~1518 / `scipy_backend.py` ~531 — keep-first dedup semantics**

The dedup keeps the *first* entry per `(param_idx, pos)` group. That's correct when coincident entries are broadcast copies (identical data), but if entries at the same key ever carried different data — e.g. COO duplicates built additively — the extras get silently dropped rather than summed. As far as I can tell no production path reaches this branch with differing data today, so this is latent, but an `assert` that dropped copies carry identical data (or summing per group) would make the invariant explicit rather than assumed.

**3. Duplicated dedup block across backends**

The 5-line pair-key dedup (combined key, `np.unique(..., return_index=True)`, triple fancy-index) plus its comment is duplicated verbatim in `scipy_backend.py:530-534` and `coo_backend.py:1517-1521`. Since per-file copies drifting apart is essentially how the original param_idx-only bug arose, could this live once in the shared backend base?

**4. `coo_backend.py` ~1444-1476 — docstring still documents the removed behavior**

The "Concrete Example" and "Position Calculation" sections of `reshape_parametric_constant`'s docstring still describe the old semantics this PR removes: dedup to one entry per `param_idx`, "param_idx directly encodes the position… new_row = param_idx % new_m", and a strided duplicate layout that contradicts the consecutive-copies assumption `pos // broadcast_factor` relies on. Worth rewriting alongside the code — a maintainer following the documented rule would reintroduce exactly the coefficient-dropping bug this PR fixes.

**5. `scipy_backend.py` ~516 (coo twin ~1498) — broadcast branch justification doesn't match reachability**

The docstring justifies the `broadcast_factor > 1` branch with the ND matmul case `P(m,k) @ X(B,k,n)` → `broadcast_to(P, (B,m,k))`, but that case is intercepted earlier by the `is_batch_varying` routing and never reaches this function. The branch appears to be defensive-only today — stating that (and what *does* reach it, e.g. direct `get_constant_data` calls) would prevent someone from extending ND-matmul handling here, or removing the interception believing this covers it.

**6. `test_python_backends.py` ~3280 — stale test docstring**

`test_coo_reshape_vs_reshape_parametric_constant`'s docstring still says the function "deduplicates based on param_idx" while the rewritten assertions directly below verify the opposite (`nnz == 4`, all entries kept). Easy to update while here; as written, it invites a future "re-fix" back to the buggy behavior.

**7. `test_python_backends.py` ~123 — RUST backend coverage**

The new `(A @ p) @ x` regression test parametrizes SCIPY and COO. CPP checks out pre- and post-PR, but the experimental RUST backend is never exercised for the parametric-lhs shape the new cumsum canonicalization emits — if it mishandles it the way pre-PR SCIPY/COO did, `canon_backend='RUST'` would return silently wrong coefficients with nothing to catch it. Worth adding to the parametrize list (with the usual skip if the extension isn't installed).

**8. `cumsum_canon.py` ~59-61 — nit: format round-trips and the 1-D common case**

`sp.kron` returns BSR, so the operator is materialized twice in BSR and then re-wrapped via `sp.csc_array` — passing `format='csc'` to the krons and dropping the extra wrap avoids the copies. In the dominant 1-D case `pre == post == 1`, both krons and both F-order reshapes are identity node churn; a short-circuit (`op = tril`, no reshapes) removes it. (Both fold away naturally if this moves to a shared sparse helper per comment 1.)

Transurgeon (approved):

PR #3400, opened 2026-06-10
Include canon_backend in the compilation cache key

PTNobel:
## Description
Re-solving a Problem with a different canon_backend silently reused the
previously compiled canonicalization: Cache.make_key only keyed on
(solver, gp, ignore_dpp, use_quad_obj), so get_problem_data treated the
second solve as a cache hit and never constructed a solving chain with
the requested backend.

Fix: add canon_backend to Cache.make_key and its single call site in
get_problem_data. The value is keyed as passed (None for the default),
so the common None -> None re-solve still hits the cache, while None and
an explicit backend are kept distinct (None enables auto-COO selection
for large DPP problems, so they are not equivalent).

Audited the other arguments construct_solving_chain consumes: solver,
gp, ignore_dpp, and solver_opts['use_quad_obj'] were already in the key;
verbose only affects logging; remaining solver_opts are forwarded to the
solver at solve time and do not change canonicalization. enforce_dpp
only gates raising DPPError versus warning for non-DPP problems and
never changes the compiled chain, so it is left out of the key.

Tests: TestCanonBackendCacheKey in test_backend_selection.py asserts a
backend switch re-canonicalizes (via the cached chain's
ConeMatrixStuffing.canon_backend) and that same-backend and default
re-solves still reuse the cached chain object.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -115,8 +115,8 @@ class Cache:
         self.param_prog = None
         self.inverse_data = None
 
-    def make_key(self, solver, gp, ignore_dpp, use_quad_obj):
-        return (solver, gp, ignore_dpp, use_quad_obj)
+    def make_key(self, solver, gp, ignore_dpp, use_quad_obj, canon_backend):
+        return (solver, gp, ignore_dpp, use_quad_obj, canon_backend)
 
     def gp(self):
         return self.key is not None and self.key[1]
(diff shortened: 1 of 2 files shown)


PTNobel:
`canon_backend` shouldn't change the compiled form right? I think this is a hallucination.

PTNobel:
This was a fable hallucination

PR #3402, opened 2026-06-10
Project leaf values with multiple entrywise attributes

PTNobel:
## Description
Leaf.project short-circuited with the raw value whenever a leaf had
more than one attribute, which made the nonneg+nonpos -> 0 branch dead
code and value validation vacuous: contradictory values such as
Variable(2, nonneg=True, nonpos=True).value = [1, -1] or
Variable(2, integer=True, nonneg=True).value = [0.5, -3] were accepted
despite the documented ValueError.

Restructure project so entrywise attributes (boolean, integer, nonneg,
nonpos, pos, neg, bounds) compose: discrete entries are rounded first,
then sign and bound clipping follow, with discrete entries clipped to
the integral endpoints inside the bounds so the composition is an exact
projection even for fractional bounds. Combinations involving a
structural attribute (symmetric, PSD, NSD, diag, hermitian, imag,
complex, sparsity) plus other attributes keep the previous pass-through
behavior, now documented in a comment; single structural attributes
move unchanged into _project_structural.

Adds test_project_multiple_attributes in test_expressions.py covering
the repros, fractional bounds exactness, and the structural
pass-through.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -447,34 +447,64 @@ class Leaf(expression.Expression):
         """
         if not self.is_complex():
             val = np.real(val)
-        # Skip the projection operation for more than one attribute
-        if self.num_attributes > 1:
-            return val
+        structural = ('imag', 'complex', 'diag', 'hermitian', 'symmetric',
+                      'PSD', 'NSD', 'sparsity')
+        if any(self.attributes[key] for key in structural):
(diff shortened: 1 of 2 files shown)


PTNobel:
Folded into #3403, which now carries both commits (this fix plus the dependent coordinate-tuple fix) against master.

PR #3406, opened 2026-06-10
Fix unreachable DGP canonicalizers; add registry key sanity test

PTNobel:
## Description
DGP problems using cp.norm1 (or cp.norm(x, 1)) raised DGPError even
though a norm1 canonicalizer is registered in dgp2dcp: the norm1 atom
never declared is_atom_log_log_convex, so the DGP curvature check
rejected it before canonicalization. For positive arguments, norm1 is
a sum of the entries (a posynomial), hence log-log convex; monotonicity
is already correct via is_incr. Declare is_atom_log_log_convex = True
and is_atom_log_log_concave = False, matching Pnorm and norm_inf.

Auditing the rest of the dgp2dcp registry uncovered a second latent
bug: the quad_form entry was keyed on the quad_form *function* instead
of the QuadForm class, so lookup by type(expr) always missed and the
fallback expr.copy(args) silently built QuadForm(log_x, log_P), giving
a wrong optimal value (e.g. 27.717 instead of 27.596). Registering
QuadForm exposed a third bug in the never-executed quad_form_canon:
it imported the cvxpy.atoms.affine.hstack module instead of the hstack
function. Both are fixed; DGP quad_form now matches the analytic
optimum.

All other registered atoms (norm_inf, Pnorm, sum, max/min, maximum/
minimum, exp, log, xexp, cumprod, geo_mean, prod, trace, power, div,
mul, one_minus_pos, quad_over_lin) were smoke-tested in small DGP
problems and solve correctly.

Adds regression tests for norm1 and quad_form under gp=True with
analytic optima. Test suites test_dgp.py, test_dgp2dcp.py, and
test_curvature.py pass.

Also add a registry-wide sanity test (test_canon_methods.py) asserting every
canonicalizer registry key is an Expression/Constraint class, so function-keyed
entries like this one can never silently reappear. The test immediately caught
one more instance: complex2real's CANON_METHODS keyed the bmat *function*; that
entry was unreachable dead code (bmat() builds Vstack/Hstack expressions, which
have their own entries) and is removed here.

Notes:
- The new `test_canon_methods.py` fails on master on two counts: the dgp2dcp `quad_form` function key fixed here, and the dead complex2real `bmat` function key removed here.
- Import blocks in the two touched `__init__.py` files were re-sorted by the repo's ruff-isort hook (mechanical). #3385 touches the same complex2real import block; the changed lines are disjoint, so the merges should be clean.

Found by an automated bug-hunting audit of master; each finding was reproduced against an independent oracle before fixing, and the new tests fail on master.

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/norm1.py
+++ b/cvxpy/atoms/norm1.py
@@ -55,6 +55,16 @@ class norm1(AxisAtom):
         """
         return False
 
+    def is_atom_log_log_convex(self) -> bool:
+        """Is the atom log-log convex?
+        """
+        return True
+
+    def is_atom_log_log_concave(self) -> bool:
+        """Is the atom log-log concave?
+        """
+        return False
+
     def is_incr(self, idx) -> bool:
         """Is the composition non-decreasing in argument idx?
         """
--- a/cvxpy/reductions/complex2real/canonicalizers/__init__.py
+++ b/cvxpy/reductions/complex2real/canonicalizers/__init__.py
@@ -16,7 +16,7 @@ limitations under the License.
 
 import numpy as np
 
-from cvxpy.atoms import (MatrixFrac, Pnorm, PnormApprox, QuadForm, abs, bmat, conj, conv,
+from cvxpy.atoms import (MatrixFrac, Pnorm, PnormApprox, QuadForm, abs, conj, conv,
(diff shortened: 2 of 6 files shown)


Transurgeon (approved):
nice catch!

Issue #706, opened 2019-04-18
Stuff constraints in parallel for large problems


akshayka:
Matrix stuffing is currently serial, but it's actually embarrassingly parallel: the for loop over constraints in  `MatrixStuffing.apply` can be parallelized. Specifically, we can get significant speed-ups by making many parallel calls to cvxcore, instead of batching all the expressions together and making a single call. Alternatively, we could get a similar speed-up by instead parallelizing cvxcore itself, which might be the preferred method.

PR #3423, opened 2026-06-18
Memoize shared LinOps in Python canonicalization backends

PTNobel:
## Description

Adds a per-`build_matrix` memo to the Python canonicalization backend path (`SCIPY` and `COO`) keyed by Python LinOp object identity. When the same LinOp object appears multiple times, the backend now reuses the already-lowered TensorView instead of recursively lowering that LinOp tree again.

Cached TensorViews are cloned when stored and on cache hits, so mutating backend operations such as negation, row selection, and accumulation cannot alter the cached snapshot or another occurrence's view.

In order to control cache size, we first walk the full tree counting duplicates, and then only store objects that are needed later.

Issue link (if applicable):

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/lin_ops/backends/base.py
+++ b/cvxpy/lin_ops/backends/base.py
@@ -56,6 +56,7 @@ This module contains the abstract base classes used by all backends:
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections import Counter
 from collections.abc import Callable
 from dataclasses import dataclass
 from enum import Enum
@@ -501,6 +502,7 @@ class PythonCanonBackend(CanonBackend):
     ) -> sp.csc_array | sp.csr_array:
         self.id_to_col[-1] = self.var_length
 
+        self._lin_op_counts = self._count_reusable_lin_ops(lin_ops)
         self._lin_op_tensor_cache: dict[int, TensorView] = {}
         try:
             constraint_res = []
@@ -515,9 +517,42 @@ class PythonCanonBackend(CanonBackend):
                 row_offset += lin_op_rows
             tensor_res = self.concatenate_tensors(constraint_res)
         finally:
+            del self._lin_op_counts
             del self._lin_op_tensor_cache
             self.id_to_col.pop(-1)
         return tensor_res.flatten_tensor(self.param_size_plus_one, order=order)
+
+    def _count_reusable_lin_ops(self, lin_ops: list[LinOp]) -> Counter[int]:
+        raw_counts: Counter[int] = Counter()
+        for lin_op in lin_ops:
+            self._count_lin_op_tree(lin_op, raw_counts)
+
+        process_counts: Counter[int] = Counter()
+        seen_cacheable: set[int] = set()
+        for lin_op in lin_ops:
+            self._count_processed_lin_ops(lin_op, raw_counts, process_counts, seen_cacheable)
+        return process_counts
+
+    def _count_lin_op_tree(self, lin_op: LinOp, counts: Counter[int]) -> None:
+        counts[id(lin_op)] += 1
+        for arg in lin_op.args:
+            self._count_lin_op_tree(arg, counts)
+
+    def _count_processed_lin_ops(
+        self,
+        lin_op: LinOp,
+        raw_counts: Counter[int],
+        process_counts: Counter[int],
+        seen_cacheable: set[int],
+    ) -> None:
+        lin_op_id = id(lin_op)
+        process_counts[lin_op_id] += 1
+        if raw_counts[lin_op_id] > 1:
+            if lin_op_id in seen_cacheable:
+                return
+            seen_cacheable.add(lin_op_id)
+        for arg in lin_op.args:
+            self._count_processed_lin_ops(arg, raw_counts, process_counts, seen_cacheable)
 
     def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorView:
         """
@@ -533,9 +568,11 @@ class PythonCanonBackend(CanonBackend):
         The processed node as a TensorView.
         """
 
+        counts = getattr(self, "_lin_op_counts", None)
+        should_cache = counts is not None and counts[id(lin_op)] > 1
         cache = getattr(self, "_lin_op_tensor_cache", None)
         cache_key = id(lin_op)
-        if cache is not None and cache_key in cache:
+        if should_cache and cache is not None and cache_key in cache:
             return self._copy_tensor_view(cache[cache_key], empty_view)
 
         # Leaf nodes
@@ -571,7 +608,7 @@ class PythonCanonBackend(CanonBackend):
                 assert res is not None
                 result = res
 
-        if cache is not None:
+        if should_cache and cache is not None:
             cache[cache_key] = self._copy_tensor_view(result, empty_view)
         return result
 
--- a/cvxpy/lin_ops/backends/base.py
+++ b/cvxpy/lin_ops/backends/base.py
@@ -501,18 +501,22 @@ class PythonCanonBackend(CanonBackend):
     ) -> sp.csc_array | sp.csr_array:
         self.id_to_col[-1] = self.var_length
 
-        constraint_res = []
-        total_rows = sum(np.prod(lin_op.shape) for lin_op in lin_ops)
-        row_offset = 0
-        for lin_op in lin_ops:
-            lin_op_rows = np.prod(lin_op.shape)
-            empty_view = self.get_empty_view()
-            lin_op_tensor = self.process_constraint(lin_op, empty_view)
-            constraint_res.append(lin_op_tensor.get_tensor_representation(row_offset, total_rows))
-            row_offset += lin_op_rows
-        tensor_res = self.concatenate_tensors(constraint_res)
-
-        self.id_to_col.pop(-1)
+        self._lin_op_tensor_cache: dict[int, TensorView] = {}
+        try:
+            constraint_res = []
+            total_rows = sum(np.prod(lin_op.shape) for lin_op in lin_ops)
+            row_offset = 0
+            for lin_op in lin_ops:
+                lin_op_rows = np.prod(lin_op.shape)
+                empty_view = self.get_empty_view()
+                lin_op_tensor = self.process_constraint(lin_op, empty_view)
+                constraint_res.append(lin_op_tensor.get_tensor_representation(row_offset,
+                                                                               total_rows))
+                row_offset += lin_op_rows
+            tensor_res = self.concatenate_tensors(constraint_res)
+        finally:
+            del self._lin_op_tensor_cache
+            self.id_to_col.pop(-1)
         return tensor_res.flatten_tensor(self.param_size_plus_one, order=order)
 
     def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorView:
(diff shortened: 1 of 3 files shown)


cvxpy/lin_ops/backends/base.py
```diff
@@ -529,38 +568,76 @@ def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorVie
...
+                data=data.data.copy(),
+                row=data.row.copy(),
+                col=data.col.copy(),
+                param_idx=data.param_idx.copy(),
+                m=data.m,
+                n=data.n,
+                param_size=data.param_size,
+            )
+        if hasattr(data, "copy"):
+            return data.copy()
```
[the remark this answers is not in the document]
PTNobel: Fixed.

PTNobel:
@claude merge in master

cvxpy/lin_ops/backends/base.py
```diff
@@ -501,20 +502,58 @@ def build_matrix(
...
+                lin_op_tensor = self.process_constraint(lin_op, empty_view)
+                constraint_res.append(lin_op_tensor.get_tensor_representation(row_offset,
+                                                                               total_rows))
+                row_offset += lin_op_rows
+            tensor_res = self.concatenate_tensors(constraint_res)
+        finally:
+            del self._lin_op_counts
+            del self._lin_op_tensor_cache
+            self.id_to_col.pop(-1)
         return tensor_res.flatten_tensor(self.param_size_plus_one, order=order)
```
Transurgeon, line 524: I am not a big fan of try catches generally.. and also don't really understand why we need one here.

cvxpy/lin_ops/backends/base.py
```diff
@@ -501,20 +502,58 @@ def build_matrix(
...
+    def _count_reusable_lin_ops(self, lin_ops: list[LinOp]) -> Counter[int]:
+        raw_counts: Counter[int] = Counter()
+        for lin_op in lin_ops:
+            self._count_lin_op_tree(lin_op, raw_counts)
+
+        process_counts: Counter[int] = Counter()
+        seen_cacheable: set[int] = set()
+        for lin_op in lin_ops:
+            self._count_processed_lin_ops(lin_op, raw_counts, process_counts, seen_cacheable)
+        return process_counts
```
Transurgeon, line 534: why do we need a count of these lin_ops? Can't we use a cache directly somehow?

cvxpy/lin_ops/backends/base.py
```diff
@@ -529,38 +568,66 @@ def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorVie
...
+        return empty_view.create_new_tensor_view(
+            variable_ids,
+            self._copy_tensor_data(view.tensor),
+            view.is_parameter_free,
+        )
+
+    def _copy_tensor_data(self, data: Any) -> Any:
+        if isinstance(data, dict):
+            return {key: self._copy_tensor_data(value) for key, value in data.items()}
+        if data is None:
```
Transurgeon, line 633: do we need to copy the tensor data? Isn't it the whole point to try to avoid copying data and reusing duplicate lin_op trees?

cvxpy/tests/test_python_backend_linop_memo.py
```diff
@@ -0,0 +1,115 @@
...
+    x = lu.create_var((3,), var_id=1)
+    calls = 0
+    original = backend_cls.get_variable_tensor
+
+    def spy_get_variable_tensor(self, shape, variable_id):
+        nonlocal calls
+        calls += 1
+        return original(self, shape, variable_id)
+
+    monkeypatch.setattr(backend_cls, "get_variable_tensor", spy_get_variable_tensor)
```
Transurgeon, line 80: this test is really weirdly written, I don't quite like both the style, the naming conventions and its purpose.

Transurgeon (changes requested):

cvxpy/lin_ops/backends/base.py
PTNobel: This isn't a try catch, it's a try finally. This makes sure that if we error we still drop the caches which aren't needed anymore before propagating the error.

cvxpy/lin_ops/backends/base.py
PTNobel: We only cache linops that appear multiple times to control memory growth.

PTNobel: Also because we mutate, we need to copy the linop into the cache. That's expensive if we're never going to access it

cvxpy/lin_ops/backends/base.py
PTNobel: We mutate linops, we have to copy them or rearchitect the backends. The goal is to avoid recomputing linops, plus with the counts we never have to copy them more than the minimum times necessary.

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: It counts the number of times each linop is canonicalized via monkeypatch and then asserts they are all 1.

PTNobel:
@Transurgeon apologies, but what changes are requested? I saw design questions in your comments; if you have concerns about the design lmk and I'll revise.

Transurgeon:
> @Transurgeon apologies, but what changes are requested? I saw design questions in your comments; if you have concerns about the design lmk and I'll revise.

yes you are correct, they were more questions. I should have left them under a comment, apologies as well.
I'll take another look this weekend and let you know if I have more specific design changes.. thanks for your fast answers, they make things a bit more clearer now.

cvxpy/tests/test_python_backend_linop_memo.py
```diff
@@ -0,0 +1,115 @@
...
+
+    matrix = _problem_matrix([x, x], backend)
+
+    assert calls == 1
+    assert matrix.shape == (2 * 3 * (3 + 1), 1)
+    assert matrix.nnz == 2 * 3
+
```
Transurgeon, line 87: I looked at the test more carefully and want to ask shouldn't leaf nodes be deduplicated already?

cvxpy/tests/test_python_backend_linop_memo.py
```diff
@@ -0,0 +1,115 @@
...
+        calls += 1
+        return original(self, view, empty_view)
+
+    monkeypatch.setattr(backend_cls, "_copy_tensor_view", spy_copy_tensor_view)
+
+    matrix = _problem_matrix([x], backend)
+
+    assert calls == 0
+    assert matrix.shape == (3 * (3 + 1), 1)
+    assert matrix.nnz == 3
```
Transurgeon, line 136: I don't think this test is really useful.. but happy to hear your thoughts on why we should keep it.

cvxpy/tests/test_python_backend_linop_memo.py
```diff
@@ -0,0 +1,115 @@
...
+        return original(self, view, empty_view)
+
+    monkeypatch.setattr(backend_cls, "_copy_tensor_view", spy_copy_tensor_view)
+
+    matrix = _problem_matrix([neg_x, neg_x], backend)
+
+    assert calls == 2
+    assert matrix.shape == (2 * 3 * (3 + 1), 1)
+    assert matrix.nnz == 2 * 3
+
```
Transurgeon, line 158: it seems like this is also testing caching leaf nodes. since neg mutates the tensor object.
I think generally, we should have greater test coverage with more types of expressions (maybe some indexing as well, and nested expressions).
We should also add more tests that use parameters and ensure that they work as well.

cvxpy/tests/test_python_backend_linop_memo.py
Transurgeon, line 1: I think in general these tests start making a bit more sense to me.
But I'd like to also have some where we test correctness rather than the internal calls. For example, we could derive some expected A matrices by hand and ensure that we get the correct result compared to building it with memoization.

cvxpy/lin_ops/backends/base.py
```diff
@@ -529,38 +568,66 @@ def process_constraint(self, lin_op: LinOp, empty_view: TensorView) -> TensorVie
         The processed node as a TensorView.
         """
 
+        counts = getattr(self, "_lin_op_counts", None)
+        should_cache = counts is not None and counts[id(lin_op)] > 1
+        cache = getattr(self, "_lin_op_tensor_cache", None)
```
Transurgeon, line 580: we define these as attributes of the Backend Class on line 504-505 already so I don't know if its necessary to be defensive here (probably not).

cvxpy/lin_ops/backends/base.py
Transurgeon: that's a great point, sorry for my misunderstanding. But now Claude is suggesting we could initialise these two new attributes in a new ``_init_`` function for PythonCanonBackend, since that will only call ``build_matrix`` once. Then we don't really need to del these attributes anymore and maybe can also avoid the try/finally.

Transurgeon (commented):
I used Claude to do a more thorough review now and found a few things that could be improved.
- firstly, we should have more and different tests which cover parameters and other expressions
- secondly, there are some design improvements and suggestions. If possible, it would actually be nice to avoid caching leaf nodes since those should be de-duplicated already right?
- thirdly, we should add some simple benchmarks results to verify that this memoization does the bring lots of performance benefits (at the very least locally, for simple but potentially large problems)

Honestly, this is going to be an awesome improvement to the backends, thanks for getting it started @PTNobel ! But I think there is still quite a bit of work left before I feel this is ready to merge.

cvxpy/lin_ops/backends/base.py
```diff
@@ -501,20 +502,58 @@ def build_matrix(
...
+                empty_view = self.get_empty_view()
+                lin_op_tensor = self.process_constraint(lin_op, empty_view)
+                constraint_res.append(lin_op_tensor.get_tensor_representation(row_offset,
+                                                                               total_rows))
+                row_offset += lin_op_rows
+            tensor_res = self.concatenate_tensors(constraint_res)
+        finally:
+            del self._lin_op_counts
+            del self._lin_op_tensor_cache
+            self.id_to_col.pop(-1)
```
Transurgeon, line 522: this is also not related to the PR, we should remove it. (the ``self.id_to_col.pop(-1)`` line)

cvxpy/lin_ops/backends/base.py
```diff
@@ -501,20 +502,58 @@ def build_matrix(
...
+        seen_cacheable: set[int],
+    ) -> None:
+        lin_op_id = id(lin_op)
+        process_counts[lin_op_id] += 1
+        if raw_counts[lin_op_id] > 1:
+            if lin_op_id in seen_cacheable:
+                return
+            seen_cacheable.add(lin_op_id)
+        for arg in lin_op.args:
+            self._count_processed_lin_ops(arg, raw_counts, process_counts, seen_cacheable)
```
Transurgeon: here we only loop through the args, but constants and parameters appear in ``.data``, for lots of Lin_ops this makes it so we don't cache the shared parametrised sub-trees. We should count both for the args and for the data maybe?

Transurgeon:
also I am worried about the ability to parallelise the backend with these new changes (I guess it is a trade-off to consider). For example cvxcore probably has some form of parallelisation (see https://github.com/cvxpy/cvxpy/issues/706).

cvxpy/lin_ops/backends/base.py
PTNobel: That line was in the original code, I'm just indenting it. Why should it be deleted? I think it's needed for correctness. The new indent is part of the adding the try/finally.

cvxpy/lin_ops/backends/base.py
PTNobel: Hmm do we really guarantee that build_matrix will only be called once? If that's not documented I'm not comfortable just depending on existing behavior.

Also, I guess I preferred this approach so we can always drop the cache ASAP if Python wants the memory; doesn't the backend object stick around after build_matrix is called?

PTNobel: One of my design goals is minimizing how much this caching changes our peak memory use hence the caution and eviction rules.

cvxpy/lin_ops/backends/base.py
PTNobel: This makes sense to me

cvxpy/lin_ops/backends/base.py
PTNobel: Since we del them there, I thought it did since I wasn't too confident this function is never called, but I'm happy to just make them always present

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: I don't think they were before my last PR since I thought we created a new LinOp object for each time it appears. I'm pretty sure LinOps were trees not DAGs

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: I would prefer to keep tests that validate that we aren't caching unnecessarily. I'm open to other approaches, but I thought this worked since we copy info the cache

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: I don't think that's true? The purpose of this test is to verify the leaf node wasn't cached

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: Now I'm worried I'm misunderstanding your question, what do you mean by deduplicated?

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: I'm certainly open to more tests, but I don't understand what design you are pursuing with these suggestions.

Why indexing for example? Architecturally it's the same as neg a mutating LinOp iirc

PTNobel: Agreed on parameters

cvxpy/tests/test_python_backend_linop_memo.py
PTNobel: I guess, don't we already have a ton of tests for canonicalization correctness? I don't see why we would add more tests. If memorization is breaking the output of the canon process it'll break the rest of our test suite.

It felt duplicative to repeat those in this file

PTNobel:
> the ability to parallelise the backend with these new changes

I mean this only affects the pure python backends. We can't get any performance benefit from parallelization here on GIL'd python. I guess I'm not scared of sacrificing that.

(Also design that enables parallelization: on the tree walk to count, we can topologically sort the linops that need to be in the cache, populate the cache in parallel and then parallelize the evaluation of the rest of the linops with a read only cache)

PR #3428, opened 2026-06-19
Install from source using `pip install -e .` fails install numpy on windows (fix provided)

Whning0513:
Add Windows build tools note to install docs explaining DISTUTILS_USE_SDK=1 and vcvarsall.bat workaround for building numpy dependency from source.

**Files changed:**
- `doc/source/install/index.rst`

### DIFF
--- a/doc/source/install/index.rst
+++ b/doc/source/install/index.rst
@@ -72,6 +72,22 @@ or a conda environment.
 
         1. Clone the official `CVXPY git repository`_, or a newly minted fork of the CVXPY repository.
         2. Navigate to the top-level of the cloned directory.
+
+        .. note::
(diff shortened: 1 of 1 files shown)


PTNobel:
I'm not sure we support installing cvxpy without numpy installed?

Whning0513:
@CLAassistant please recheck

Whning0513:
It's not about removing the numpy dependency -- it's about documenting the Windows build tools needed to compile numpy from source when doing `pip install -e .`. On Windows, users often hit obscure build failures because they don't have the right Visual Studio build tools or SDK environment variables set. The PR just adds those steps to the install docs.

Whning0513:
@PTNobel Thanks for the reply. You're right — this is about `pip install -e .` failing when numpy isn't already installed. I'll close this PR.

PR #3431, opened 2026-06-19
BUG: fix None.value crashes in power, finite_set, and inverse modules

Whning0513:
Fix 6 potential crashes from .value access on Parameter objects without set values: PowCone3D.__init__, PowConeND.__init__, Power.sign_from_args, Power.numeric, PowerApprox.__init__ (approx_error), FiniteSet.residual, and dqcp2dcp inverse.py power_inv.

**Files changed:**
- `cvxpy/atoms/elementwise/power.py`
- `cvxpy/constraints/finite_set.py`
- `cvxpy/constraints/power.py`
- `cvxpy/reductions/dqcp2dcp/inverse.py`

### DIFF
--- a/cvxpy/atoms/elementwise/power.py
+++ b/cvxpy/atoms/elementwise/power.py
@@ -190,12 +190,17 @@ class Power(Elementwise):
 
     @Elementwise.numpy_numeric
     def numeric(self, values):
+        if self.p.value is None:
+            raise ValueError(
+                "Power exponent must have a value set. "
+                "Received a Parameter exponent without a value."
+            )
         return np.power(values[0], float(self.p.value))
 
     def sign_from_args(self) -> tuple[bool, bool]:
         """Returns sign (is positive, is negative) of the expression.
         """
-        if self.p.value == 1:
+        if self.p.value is not None and self.p.value == 1:
             # Same as input.
             return (self.args[0].is_nonneg(), self.args[0].is_nonpos())
         else:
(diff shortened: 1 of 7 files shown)


cvxpy/atoms/elementwise/power.py
PTNobel, line 1: If `p.value` is not set, that's a bug in construction of the class. I don't see why we need to be defensive here.

PTNobel (commented):
Are any of these possible to hit in correct user code? I don't think so.

Whning0513:
@CLAassistant please recheck

Transurgeon:
@Whning0513 could you add some tests that fail without the changes in this PR?
Like Parth, I am not sure that these new checks you are adding will ever get hit.

Whning0513:
Added a follow-up commit on `fix-math-17` with regression tests that exercise the previously questioned paths. The new coverage includes unset exponent/alpha parameters in the power atom, `PowCone3D`, `PowConeND`, `FiniteSet.residual`, and the DQCP inverse path, so these cases now fail without the fix and pass with it.

Transurgeon:
@Whning0513 could you install the pre-commit and run ruff? It seems like your PR diff has many more line changes than I would expect. Thank you!

Transurgeon:
hello @Whning0513 I will actually close this PR because on second thought I agree with @PTNobel that if these values are not set they should already error in the construction of the atom. Feel free to open a new PR if this isn't done already (although I would be very doubtful if we hadn't added those checks). Thanks for your PR nonetheless!

Issue #3418, opened 2026-06-16
Post-solver validation: first-pass tracking issue


curious7-web:
This tracks follow-up implementation work from the design discussion in #3362.

The first-pass scope from that thread is:
* primal constraint violation
* dual feasibility
* complementarity

Duality gap and stationarity are deferred for now, since those appear to need more careful design.

### Tasks
- [x] Audit current `Constraint.violation()` / `residual` behavior across constraint classes
- [x] Decide/document a consistent scalar convention for constraint violation
- [ ] Add problem-level primal violation helper/API
- [x] Audit `constraint.dual_value` behavior across constraint classes
- [x] Add dual feasibility checks where well-defined
- [ ] Add complementarity checks where well-defined
- [x] Add tests for representative constraint types
- [x] Defer duality gap and stationarity to follow-up design work

### Notes
One open design point from #3362 is that some constraints expose vector-valued residuals while others already reduce violations to scalar norms. The first implementation step should clarify that behavior before adding problem-level aggregation.

I’m planning to start with the violation/residual audit and the primal-side helper/API.

WilliamZhang20:
Great stuff! I'll look at auditing `constraint.dual_value` behavior and adding dual feasibility checks.

PR #3434, opened 2026-06-19
Make constraint residuals and violations consistent

curious7-web:
## Description

This PR follows up on #3418 and adds the first internal helper layer for primal-side post-solver validation.

`Constraint.violation()` can currently return scalar, vector, or matrix-like values depending on the constraint type. Before adding a public problem-level API, this PR normalizes individual constraint violations to a scalar using the Euclidean norm, then adds a helper for the maximum scalar violation across a collection of original constraints.

This intentionally does not expose `prob.violation()` yet. The goal is to make the scalar convention explicit and test-covered before making it public.

Issue link: #3418

## Type of change

- [ ] Bug fix

- [x] Write unit tests.
- [x] Run the unit tests and check that they’re passing.


curious7-web:
Small scope note: this PR only normalizes the existing `Constraint.violation()` outputs to a scalar and adds aggregation over original constraints. It does not add or change violation implementations for additional constraint classes.

I kept this internal for now so we can agree on the scalar convention before exposing a public problem-level API. Follow-up PRs from #3418 can extend coverage and API surface once that direction is settled.

PTNobel:
My preference would be to change `.violation` on the constraint classes to be consistent.

Could you make a table of existing constraints and what their violations currently are? Scalars/vectors/etc? And then for the ones that are scalars how they're reduced?

I think the only logical reductions are:

* $\|v\|_1 / n$
* $\|v\|_2 / \sqrt{n}$
* $\|v\|_\inf$

My gut feeling is $\ell_\inf$ is the best option, but I would want the other maintainers to weigh in.

curious7-web:
Thanks, that makes sense. I audited the current `.violation()` behavior across the main constraint classes.

The main finding is that base `Constraint.violation()` currently returns `self.residual` directly, so the return shape depends on how each constraint class defines `residual`.

| Constraint class | Current `.violation()` behavior | Shape / type | Current reduction |
|---|---|---|---|
| `Zero` / `Equality` | `abs(expr.value)` | array-like | none |
| `Inequality` (`<=` / `>=`) | elementwise positive violation | array-like | none |
| explicit `NonPos` | `np.linalg.norm(residual, ord=2)` | scalar | ℓ2 |
| explicit `NonNeg` | `np.linalg.norm(residual, ord=2)` | scalar | ℓ2 |
| `PSD` | negative minimum eigenvalue residual | scalar in tested case | scalar residual already |
| `SOC` | distance to SOC projection | scalar in tested case | scalar residual already |
| `ExpCone` | projection problem objective value | scalar | scalar residual already |
| `PowCone3D` | projection problem objective value | scalar | scalar residual already |

So the current behavior is mixed: some constraints return elementwise arrays, some explicitly reduce using ℓ2, and several cone constraints already expose scalar projection distances through `residual`.

Given that, I agree the cleaner next step is probably to make `.violation()` itself consistent across constraint classes rather than adding a separate helper layer. I’ll wait for maintainer input on whether the convention should be ℓ1/n, ℓ2/n, or ℓ∞ before changing behavior.

<details>
<summary>Runtime audit script used</summary>

```python
import numpy as np
import cvxpy as cp


def summarize(name, c):
    print(f"\n=== {name} ===")
    print("class:", type(c).__name__)

    try:
        r = c.residual
        print("residual:", r)
        print("residual type:", type(r).__name__)
        if r is not None:
            print("residual shape:", np.asarray(r).shape)
    except Exception as e:
        print("residual error:", type(e).__name__, e)

    try:
        v = c.violation()
        print("violation:", v)
        print("violation type:", type(v).__name__)
        if v is not None:
            print("violation shape:", np.asarray(v).shape)
    except Exception as e:
        print("violation error:", type(e).__name__, e)


# Equality / Zero
x = cp.Variable(2)
c = x == np.array([0.0, 0.0])
x.value = np.array([3.0, -4.0])
summarize("Equality / Zero: x == 0", c)

# Inequality <=
x = cp.Variable(2)
c = x <= np.array([0.0, 0.0])
x.value = np.array([3.0, -4.0])
summarize("Inequality <=: x <= 0", c)

# Inequality >=
x = cp.Variable(2)
c = x >= np.array([0.0, 0.0])
x.value = np.array([3.0, -4.0])
summarize("Inequality >=: x >= 0", c)

# PSD
X = cp.Variable((2, 2), symmetric=True)
c = X >> 0
X.value = np.array([[-1.0, 0.0], [0.0, 2.0]])
summarize("PSD: X >> 0", c)

# SOC
t = cp.Variable()
y = cp.Variable(2)
c = cp.SOC(t, y)
t.value = 1.0
y.value = np.array([2.0, 0.0])
summarize("SOC: ||y||_2 <= t", c)

# ExpCone
a = cp.Variable()
b = cp.Variable()
z = cp.Variable()
c = cp.ExpCone(a, b, z)
a.value = 1.0
b.value = 1.0
z.value = 1.0
summarize("ExpCone", c)

# PowCone3D
u = cp.Variable()
v = cp.Variable()
w = cp.Variable()
c = cp.PowCone3D(u, v, w, alpha=0.5)
u.value = 1.0
v.value = 1.0
w.value = 2.0
summarize("PowCone3D", c)

**Comment by odow** [comment]
You might be interested in in how JuMP defined the violation distances

https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/distance_to_set.jl

**Comment by curious7-web** [comment]
Thanks for the pointer @odow, this is helpful.

Looking through the MOI implementation, the useful reference point seems to be that violation is treated as a scalar distance-to-set quantity rather than an elementwise residual. In many cases that is the minimal L2/Euclidean distance, while some cone methods are documented as projection upper bounds when exact projection is expensive.

That makes me think the CVXPY issue is partly that `.violation()` currently mixes two ideas:

| Case | Current CVXPY behavior |
|---|---|
| `Zero` / `Equality` | returns elementwise residuals |
| `Inequality` (`<=` / `>=`) | returns elementwise residuals |
| explicit `NonPos` / `NonNeg` | reduces with L2 norm |
| `PSD`, `SOC`, `ExpCone`, `PowCone3D` | already return scalar residuals in the cases I tested |

So I agree the cleaner direction is probably to make `.violation()` itself consistently return a scalar distance-style value, rather than adding a separate helper layer first.

The remaining design question is the scalar convention. MOI gives a strong argument for an L2/distance-to-set interpretation, while `ℓ∞` may be attractive for solver-tolerance-style “worst entry” reporting. I’ll wait for maintainer input before changing the PR direction.

**Comment by PTNobel** [comment]
Alright here's what we settled on:

1. residual+value must always be in the constraint (so inside a cone, =0 for equality, etc) 
2. Violation is `$||residual||_∞$` for nonspectral constraints, and it's the operator norm for spectral constraints

**Comment by curious7-web** [comment]
Thanks for the clarification.

Would you prefer that I redirect this draft PR toward that convention, or close this one and open a cleaner PR focused specifically on making `residual` / `violation()` consistent across constraint classes?

**Comment by PTNobel** [comment]
Feel free to redirect this one.

**Comment by curious7-web** [comment]
Updated this draft PR to remove the earlier helper-layer approach.

Since `Constraint.violation()` now returns the scalar violation directly under the settled convention, the separate validation helpers were redundant and have been removed.

The PR is now focused on:

- making nonspectral `Constraint.violation()` return a scalar infinity norm of the residual,
- making `Constraint.value()` check that scalar violation,
- changing equality/zero and inequality residuals to signed feasibility corrections,
- removing the old `NonPos` / `NonNeg` L2 violation overrides,
- updating tests for scalar violations and correction residuals.

**Comment by PTNobel** [comment]
You appear to have committed a bunch of underrated formatting changes. Please revert those.

**Comment by curious7-web** [comment]
Sorry about the formatting noise... reverted those. The diff should now only contain the logical changes to violation(), residual, and the related tests.

File: cvxpy/constraints/constraint.py [hunk 1]
```diff
@@ -233,9 +225,12 @@ def violation(self):
         """
         residual = self.residual
         if residual is None:
-            raise ValueError("Cannot compute the violation of a constraint "
+            raise ValueError("Cannot compute the violation of an constraint "
```
PTNobel: I don't think this change is grammatically correct

cvxpy/constraints/constraint.py
```diff
@@ -207,23 +207,15 @@ def residual(self):
...
-        expression's value and its projection onto the domain of the
-        constraint:
-
-        .. math::
-
-            ||\\Pi(v) - v||_2
-
-        where :math:`v` is the value of the constrained expression and
-        :math:`\\Pi` is the projection operator onto the constraint's domain .
+        For nonspectral constraints, the violation is the infinity norm of
```
PTNobel, line 212: Should we include the other case too?

PTNobel (commented):
Overall seems fine for the changes included; what's the status on the spectral changes?

curious7-web:
Thanks, fixed the grammar and added wording for the spectral case.

For status: this PR keeps the existing spectral residual behavior unchanged and focuses on making the nonspectral / linear constraint convention consistent. My understanding is that spectral constraints already use scalar operator-norm-style residuals, so the new scalar reduction leaves that case unchanged. Happy to handle any spectral-specific cleanup here too if you’d prefer.

cvxpy/constraints/constraint.py
```diff
@@ -207,23 +207,17 @@ def residual(self):
...
+        the constraint residual. Spectral constraints may return scalar
```
PTNobel:
```suggestion
        the constraint residual. Spectral constraints return
```

cvxpy/constraints/constraint.py
```diff
@@ -207,23 +207,17 @@ def residual(self):
...
+        operator-norm residuals directly, which are unchanged by this scalar
```
PTNobel:
```suggestion
        the operator-norm of the residual.
```

cvxpy/constraints/nonpos.py
```diff
@@ -263,4 +244,4 @@ def residual(self):
         """
         if self.expr.value is None:
             return None
-        return np.maximum(self.expr.value, 0)
+        return -np.maximum(self.expr.value, 0)
```
PTNobel, line 247: What's the sign convention here? Do we add the residual to the left or right side

PTNobel (approved):

cvxpy/constraints/nonpos.py
curious7-web: The convention I’m using is to add the residual to the constrained expression value.

For NonPos, the constrained expression is expr <= 0, so the residual satisfies:

`expr.value + residual <= 0`.

That is why `NonPos.residual` is `-np.maximum(expr.value, 0)`: if expr.value = 3, the residual is -3, and 3 + (-3) = 0 is feasible.

PTNobel: This is the Inequality class, not NonPos.

curious7-web: Ah, you’re right. I mixed up the class name there.

The same sign convention applies to `Inequality`: since it represents expr <= 0, the residual is added to `expr.value` to project it back to feasibility. So if expr.value = 3, the residual is -3, and 3 + (-3) = 0 is feasible.

PTNobel:
FYI, this is a breaking API change so it needs a second review, feel free to mark it ready and hopefully someone gets to it soon. (I'm also waiting on reviews right now, we're as always short staffed)

Transurgeon (approved):
this looks great and very clean! approving and merging.

Issue #3442, opened 2026-07-06
`cp.nlp.cos/cp.nlp.sin` with numpy array-indexed variable causes segfault in DNLP Hessian initialisation


bmeyers:
## Summary

When building a DNLP problem where `cp.nlp.cos` or `cp.nlp.sin` receives an argument formed by **numpy array indexing** of a CVXPY variable (e.g. `theta[rows, 0]` where `rows` is a `np.ndarray`), the process segfaults inside `init_hessian_coo_lower_tri` during `prob.solve(..., nlp=True)`. No Python exception is raised — the process hard-crashes with no diagnostic output.

The identical problem structure built using **scalar integer indexing** in a Python loop works correctly.

---

## Environment

```
cvxpy:   1.9.1  (latest in pip)
Python:  3.13.2
OS:      macOS (Apple Silicon)
solver:  IPOPT via cyipopt
```

---

### Crashes (numpy array indexing)

```python
import numpy as np
import cvxpy as cp

nb   = 9
nnz  = 27
rows = np.array([0, 0, 0, 1, 1, 1, 2, 2, 2,   # example sparsity pattern
                 3, 3, 3, 4, 4, 4, 5, 5, 5,
                 6, 6, 6, 7, 7, 7, 8, 8, 8])
cols = np.array([0, 1, 2, 0, 1, 3, 0, 2, 4,
                 1, 3, 5, 2, 4, 6, 3, 5, 7,
                 4, 6, 7, 5, 6, 8, 6, 7, 8])

theta = cp.Variable((nb, 1), name='theta')
v     = cp.Variable((nb, 1), name='v')
x     = cp.Variable(nnz,     name='x')

# Crashes: numpy array indexing into a 2-D CVXPY variable
C  = cp.nlp.cos(theta[rows, 0] - theta[cols, 0])
vv = cp.multiply(v[rows, 0], v[cols, 0])

prob = cp.Problem(cp.Minimize(cp.sum(x)), [x == cp.multiply(vv, C)])
prob.solve(solver=cp.IPOPT, nlp=True)   # segfault in init_hessian_coo_lower_tri
print(prob.status)                       # never reached
```

**Result:** process crashes, no output after `prob.solve(...)`.

Crash location from C traceback:
```
File ".../cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py", line 78
    in init_hessian_coo_lower_tri
[...]

PR #3446, opened 2026-07-07
Add regression tests for duplicate-index gather multiply (#3442)

Transurgeon:
## Description
Regression tests for #3442: a DNLP problem containing `cp.multiply(a[idx], b[idx])` where `idx` has **duplicate** entries (the issue's 9-bus power flow indexes bus variables by branch, `rows = [0, 0, 0, 1, ...]`) corrupted the heap during derivative-structure initialization and segfaulted, while the scalar-indexed loop formulation worked.

The root cause was in the diff engine, not cvxpy: `sparse_index_alloc` sized the gather Jacobian by the *source's* nnz, an undercount when duplicated indices select the same row more than once (`x[[0,0,1,2,2,3]]` on a size-4 variable gathers 6 entries from a 4-nnz source). Fixed in SparseDifferentiation/SparseDiffEngine#105, shipping in sparsediffpy 0.6.0.

This PR adds the cvxpy-side coverage:

- `TestDuplicateIndexGathers::test_multiply_duplicate_index_gathers_derivatives` — solver-free `DerivativeChecker` run over the issue's shape (`multiply(v[rows], v[cols])` and `nlp.cos(theta[rows] - theta[cols])`), cross-checking values, gradient, Jacobian, and Hessian against finite differences.
- `TestAffineDiffEngine::test_multiply_duplicate_index_gathers_solve` — IPOPT solve of the array-indexed formulation, asserted to agree with the scalar-indexed loop formulation (the issue's workaround).

Both are gated on `sparsediffpy >= 0.6.0`: on the currently pinned 0.5.x the bug corrupts the process heap, so the tests skip loudly (naming the engine fix) rather than crash CI. They activate automatically when the sparsediffpy pin is bumped.

Issue link: #3442

## Type of change
- [x] Bug fix

- [x] Write unittests.

(Unittests run against a dev build of the fixed engine: the derivative-checker test passes locally. The IPOPT solve test has NOT been run locally — IPOPT is not installed here — it skips locally and will first be exercised by the NLP CI workflow once the sparsediffpy pin is bumped. No benchmark run — test-only change, no library code touched.)

### DIFF
--- a/cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_affine_vector_atoms.py
+++ b/cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_affine_vector_atoms.py
@@ -22,6 +22,37 @@ from cvxpy.reductions.solvers.defines import INSTALLED_SOLVERS
 from cvxpy.tests.nlp_tests.derivative_checker import DerivativeChecker
 
 
+def _duplicate_gather_problem():
+    """gh-3442 shape: multiply of gathers whose index arrays contain
+    DUPLICATES, so each gather Jacobian has more nonzeros than its source
+    variable. Corrupted the heap in derivative-structure initialization."""
+    nb = 4
+    rows = np.array([0, 0, 1, 2, 2, 3])
+    cols = np.array([1, 2, 2, 0, 3, 1])
+    theta = cp.Variable(nb, name='theta', bounds=[-0.5, 0.5])
+    v = cp.Variable(nb, name='v', bounds=[0.9, 1.1])
+    x = cp.Variable(rows.size, name='x')
+    C = cp.nlp.cos(theta[rows] - theta[cols])
+    vv = cp.multiply(v[rows], v[cols])
+    prob = cp.Problem(cp.Minimize(cp.sum(x)), [x == cp.multiply(vv, C)])
+    return prob, theta, v, x, rows, cols
+
+
+class TestDuplicateIndexGathers:
+    # Regression tests for gh-3442 (solver-free: the crash was in the diff
+    # engine's jacobian/hessian structure init, before any solver ran).
+    # The fix (gather-Jacobian allocation for duplicated indices,
+    # SparseDiffEngine#105) ships in sparsediffpy 0.6.0, the pinned floor.
+    def test_multiply_duplicate_index_gathers_derivatives(self):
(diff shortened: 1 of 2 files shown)


cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_affine_vector_atoms.py
```diff
@@ -21,6 +21,51 @@
...
+    from importlib.metadata import version as _pkg_version
+    _ENGINE_FIXED = tuple(
+        int(p) for p in _pkg_version("sparsediffpy").split(".")[:2]) >= (0, 6)
+except Exception:  # pragma: no cover - not installed / unparsable version
+    _ENGINE_FIXED = False
+
+_ENGINE_SKIP_REASON = (
+    'requires sparsediffpy >= 0.6.0 (duplicate-index gather fix, '
+    'SparseDiffEngine#105).'
+)
```
Transurgeon: I'll clean this PR up once we make a new release of the diffengine.

PTNobel (approved):
My preference is to merge this PR after the SparseDiff release with the simplifications.

Transurgeon:
> My preference is to merge this PR after the SparseDiff release with the simplifications.

yes of course, that's what I had in mind too. We should ensure the CI passes as well.

cvxpy/tests/nlp_tests/stress_tests_diff_engine/test_affine_vector_atoms.py
Transurgeon: Done: sparsediffpy 0.6.0 (with the SparseDiffEngine#105 gather fix) is on PyPI, so the `_ENGINE_FIXED` version sniffing and both skip markers are gone — the pin floor is now `>= 0.6.0, < 0.7.0` and the tests run unconditionally.

Transurgeon:
sparsediffpy **0.6.0 is released on PyPI**, so this is cleaned up as promised: the version-sniffing gate is removed, the requirement is bumped to `sparsediffpy >= 0.6.0, < 0.7.0` (same one-liner as #3447; whichever merges second sees a no-op), and CI now actually exercises the regression tests. The solver-free DerivativeChecker test passes locally against the released wheel; the IPOPT solve test runs in the `run_nlp_tests` job here.

PR #3447, opened 2026-07-08
Kron diffengine converter

Transurgeon:
## Description
Adds native Kronecker-product support to the DNLP diff engine. `cp.kron(A, B)` with one variable-free operand (which `kron`'s DCP rules already require; it may be a `Parameter`) converts to the engine's `make_left_kron` / `make_right_kron` bindings introduced in sparsediffpy 0.6.0.

The converter passes the variable-free operand's structurally nonzero entries as **active blocks** (column-major flat indices), so the engine materializes output rows only for nonzero blocks — `O(nnz(result))` instead of dense coefficient matrices. A parametric operand gets all blocks, since its values change between solves.

Also bumps the sparsediffpy requirement to `>= 0.6.0, < 0.7.0` (the release with the kron bindings, now on PyPI).

Issue link (if applicable): n/a

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
+++ b/cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
@@ -54,6 +54,8 @@ def convert_kron(expr, children):
     variable-free operand's structurally nonzero entries (column-major flat
     indices). A parametric operand gets all blocks since its values change."""
     a, b = expr.args
+    if not (a.is_constant() or b.is_constant()):
+        raise ValueError("kron requires at least one variable-free operand.")
     const_is_left = a.is_constant()
     const_expr = a if const_is_left else b
     const_node = children[0] if const_is_left else children[1]
@@ -72,6 +74,8 @@ def convert_kron(expr, children):
             active = np.unique(coo.row[mask] + coo.col[mask] * n_rows)
             active = active.astype(np.int32)
         else:
+            # No format conversion: read the nonzero pattern straight off the
+            # dense values, same result as the sparse branch above.
             flat = np.asarray(val, dtype=np.float64).flatten(order="F")
             active = np.flatnonzero(flat).astype(np.int32)
 
--- a/cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
+++ b/cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
@@ -47,18 +47,37 @@ def convert_conv(expr, children):
 
 
 def convert_kron(expr, children):
-    """Convert cp.kron(A, B). One operand is variable-free (kron requires it);
-    it is the "parameter" side that scales the variable-carrying operand. The
-    native node re-evaluates that operand each solve, so a parametric A or B is
-    supported (not just a constant). ``const_is_left`` tells the engine which
-    operand is the parameter; (p, q) and (r, s) are A's and B's dims."""
+    """Convert cp.kron(A, B); the variable-free operand (kron requires one) is
+    re-evaluated each solve, so it may be parametric.
+
+    The engine materializes output rows only for the ``active`` blocks: the
+    variable-free operand's structurally nonzero entries (column-major flat
+    indices). A parametric operand gets all blocks since its values change."""
     a, b = expr.args
     const_is_left = a.is_constant()
-    param_node = children[0] if const_is_left else children[1]
+    const_expr = a if const_is_left else b
+    const_node = children[0] if const_is_left else children[1]
     var_node = children[1] if const_is_left else children[0]
-    p, q = a.shape
-    r, s = b.shape
-    return _diffengine.make_kron(param_node, var_node, int(const_is_left), p, q, r, s)
+    p, q = normalize_shape(a.shape)
+    r, s = normalize_shape(b.shape)
+    n_rows = p if const_is_left else r
+
+    if const_expr.parameters():
+        active = np.arange(const_expr.size, dtype=np.int32)
+    else:
+        val = const_expr.value
+        if sparse.issparse(val):
+            coo = val.tocoo()
+            mask = coo.data != 0  # drop stored-but-zero entries
+            active = np.unique(coo.row[mask] + coo.col[mask] * n_rows)
+            active = active.astype(np.int32)
+        else:
+            flat = np.asarray(val, dtype=np.float64).flatten(order="F")
+            active = np.flatnonzero(flat).astype(np.int32)
+
+    if const_is_left:
+        return _diffengine.make_left_kron(const_node, var_node, p, q, r, s, active)
+    return _diffengine.make_right_kron(const_node, var_node, p, q, r, s, active)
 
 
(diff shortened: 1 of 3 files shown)


Transurgeon:
@dance858 here are the python tests for kron.

dance858:
Awesome! Let's maybe add one test when the left argument is a parameter and also a test when the right argument is a parameter?

Transurgeon:
@PTNobel could we merge this? I think it is ready.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -46,6 +46,40 @@ def convert_conv(expr, children):
...
 
+def convert_kron(expr, children):
+    """Convert cp.kron(A, B); the variable-free operand (kron requires one) is
+    re-evaluated each solve, so it may be parametric.
+
+    The engine materializes output rows only for the ``active`` blocks: the
+    variable-free operand's structurally nonzero entries (column-major flat
+    indices). A parametric operand gets all blocks since its values change."""
+    a, b = expr.args
+    const_is_left = a.is_constant()
```
PTNobel, line 59: Shouldn't you check that one of them is const as validation

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -46,6 +46,40 @@ def convert_conv(expr, children):
...
+    var_node = children[1] if const_is_left else children[0]
+    p, q = normalize_shape(a.shape)
+    r, s = normalize_shape(b.shape)
+    n_rows = p if const_is_left else r
+
+    if const_expr.parameters():
+        active = np.arange(const_expr.size, dtype=np.int32)
+    else:
+        val = const_expr.value
+        if sparse.issparse(val):
```
PTNobel, line 71: Should we use the sparse threshold to coerce mostly 0 matrices to sparse

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -46,6 +46,40 @@ def convert_conv(expr, children):
...
+            coo = val.tocoo()
+            mask = coo.data != 0  # drop stored-but-zero entries
+            active = np.unique(coo.row[mask] + coo.col[mask] * n_rows)
+            active = active.astype(np.int32)
```
PTNobel: We've had overflow issues with int32 before, why the small type?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -46,6 +46,40 @@ def convert_conv(expr, children):
...
+        else:
+            flat = np.asarray(val, dtype=np.float64).flatten(order="F")
+            active = np.flatnonzero(flat).astype(np.int32)
```
PTNobel: Is this a dense -> sparse coersion?

Ditto on 32 bits

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: done.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: remove the int32 cast.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: I am not sure on the coercion. what do you mean exactly?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: that's a great suggestion! but this converter for kron already selects the non-zero indices by using ``np.flatnonzero``. Also another answer to your question would be that we don't have "sparse" kron and "dense" kron operators in the diff engine so there's no point to apply any sort of conversion to sparse here.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
PTNobel: It is still here?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
PTNobel: Okay.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
PTNobel: Re coersion, you answered above. This is converting a dense array to sparse one.

Please address the 32 bit issue.

PTNobel (dismissed):

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon: @PTNobel sorry it took so long to come back to this. I found out today that this type conversion actually already happens (see the registry file, for example ``extract_flat_indices_from_index``). I think the engine only supports int32 for the number of nonzeros in the matrices. 
Do you think that's an issue? if so we might need to open an issue in the engine and make a bigger change over there. Otherwise, I think this would be fine to merge now.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
```diff
@@ -46,6 +46,45 @@ def convert_conv(expr, children):
...
+            coo = val.tocoo()
+            mask = coo.data != 0  # drop stored-but-zero entries
+            active = np.unique(coo.row[mask] + coo.col[mask] * n_rows)
+        else:
+            # No format conversion: read the nonzero pattern straight off the
+            # dense values, same result as the sparse branch above.
+            flat = np.asarray(val, dtype=np.float64).flatten(order="F")
+            active = np.flatnonzero(flat)
+
+    active = active.astype(np.int32)
```
PTNobel, line 81: Flagging this as an open q tho the original line has been changed

Transurgeon: @PTNobel I created an issue about the int32 conversion here: https://github.com/SparseDifferentiation/SparseDiffPy/issues/25. Could you have another quick look at this PR and merge it please?

PTNobel: I removed my Request Changes so someone else can approve. I'm out of town so I don't have the time to do a review (and I have no memory of this PR so I don't think my review will be faster or anything)

SteveDiamond (approved):

Issue #1043, opened 2020-05-31
Segmentation fault (core dumped) while using CVXPY library


upendra05:
**Describe the bug**
I am trying to solve an optimization problem using CVXPY library. My code runs properly on a dummy dataset but crashes when I key in the actual dataset. The error I got is

> `Segmentation fault      (core dumped)`

**To Reproduce**
```
import cvxpy as cp
import numpy as np
from numpy import linalg as LA   

X_source = np.load("seen_class_vgg16_features.npy")
Y_source = np.load("seen_class_labels.npy")

#Shuffling the two feature matrices along with their corresponding labels

from sklearn.utils import shuffle 
Xs, ys = shuffle(X_source, Y_source, random_state=0)

#Now loading all classes attributes

all_classes_attributes = np.load("all_classes_attributes.npy")

ys =  [int(i) for i in ys]
ys = np.asarray(ys)

Ys = []

for i in ys:
    Ys.append(all_classes_attributes[i-1])

Ys = np.asarray(Ys)

# Define and solve the CVXPY problem.
Ds = cp.Variable(shape=(312,4096))

def objective_fn(Xs, Ys, Ds, lamda1, lamda2):
    return (cp.square(cp.norm(Xs-Ys*Ds,'fro')) + lamda1*cp.square(cp.norm(Ds,'fro')) + lamda2*(cp.square(cp.norm(Ds,'fro')) - 1))

lamda1 = cp.Parameter(nonneg=True)
lamda2 = cp.Parameter(nonneg=True)

Ds_values = []

lamda1_values = []
lamda2_values = []

problem = cp.Problem(cp.Minimize(objective_fn(Xs,Ys,Ds,lamda1,lamda2)))

lambd1_values = np.logspace(-2, 3, 2)
lambd2_values = np.logspace(-2, 3, 2)

for lam1 in lambd1_values:
    for lam2 in lambd2_values:
        lamda1.value = lam1
        lamda2.value = lam2
        problem.solve(qcp=True, verbose=True)
        print("Optimal value of Ds = ",Ds.value)
        lamda1_values.append(lam1)
        lamda2_values.append(lam2)
        Ds_values.append(Ds.value)
```

**Expected behavior**
[...]

PR #3448, opened 2026-07-08
Add an opt-in DIFFENGINE canon backend for parameter-free problems

Transurgeon:
## Description

Adds an opt-in `DIFFENGINE` canonicalization backend that builds the stuffed cone matrices by evaluating the canonicalized expression trees with the C diff engine (`sparsediffpy`), instead of constructing parameter tensors. One extraction at `x = 0` yields `(q, d, A, b, P)`, which are encoded as the single-column constant-slice tensors of a stock parameter-free `ParamConeProg` — so formatting, solver interfaces, and inversion all run the unmodified tensor-path code. No new solver-facing API, no change to any default behavior.

**Scope is deliberately narrow:**
- Selected only via an explicit `canon_backend="DIFFENGINE"` argument.
- Parameter-free compilation only: a problem whose stuffing still carries Parameters raises a `ValueError` pointing at `ignore_dpp=True` (which evaluates parameters before canonicalization). Parametric variable bounds — which `EvalParams` does not bake — are rejected loudly on both the bounds-attribute and lowered-constraint routes. N-D problems are rejected at chain construction, mirroring the explicit-CPP treatment.
- DPP-path parametric support (symbolic re-extraction on re-solves, caching) is deferred to follow-up PRs.

**Why:** for problems whose expression trees make the tensor pipeline blow up, evaluating the tree directly is far cheaper. Cold `get_problem_data` on the cvxpy benchmarks repo (this branch, CLARABEL, `ignore_dpp=True`, BLAS pinned to 1 thread, one isolated subprocess per measurement):

| benchmark | CPP | DIFFENGINE |
|---|---|---|
| gini_portfolio Cajas | 21.45 s | **1.42 s** |
| SDPSegfault1132 (issue #1132) | 60.24 s | **23.05 s** |
| QuantumHilbertMatrix | 4.47 s | **1.85 s** |
| ConvexPlasticity | **0.38 s** | 0.67 s |
| CVaRBenchmark | **36.96 s** | 64.40 s |

It is not uniformly faster — hence opt-in, not a default. The CVaR row is dominated by the single-column tensor encode/decode on a 25.7M-nnz constraint matrix (several O(nnz) sparse passes); that cost is specific to this PR's stock-`ParamConeProg` encoding, has optimization headroom (direct flat-column construction), and disappears for re-solves once the parametric follow-up lands. `test_cvxcore_issues.py` replays scaled-down reproductions of the issue-tracker cases (#1132, #2205, #1043) on the new backend.

**Correctness:** `test_diffengine_backend.py` verifies stuffed-data equality against the CPP backend for QP/LP/SOC fixtures, solution+dual parity on SOC/PSD problems, the rejection paths, and MIP attribute plumbing. The full stuffed-data equivalence was additionally checked externally on the cvxpy benchmarks (FactorCovarianceModel, ConvexPlasticity, HuberRegression: `P/c/A/b` identical to the CPP path, matching nnz).

Three commits: converter groundwork (shared with the existing NLP path) + `sparsediffpy >= 0.6.1, < 0.7.0` pin, the backend itself, and `canon_backend` added to the solve cache key (an explicit backend switch on a re-solve must not reuse the chain compiled for another backend).

The backend implementation lives entirely in the `diff_engine` package (`cone_stuffing.py`); `ConeMatrixStuffing.apply` gains only an 11-line dispatch, which is the stable seam future backend work extends without touching the stuffing reduction again.

Issue link (if applicable): #2205, #1132

## Type of change
- [ ] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -786,10 +786,12 @@ class Problem(u.Canonical):
             When True, DPP problems will be treated as non-DPP,
             which may speed up compilation. Defaults to False.
         canon_backend : str, optional
-            'CPP' (default) | 'SCIPY'
+            'CPP' (default) | 'SCIPY' | 'COO' | 'DIFFENGINE'
             Specifies which backend to use for canonicalization, which can affect
             compilation time. Defaults to None, i.e., selecting the default
-            backend.
+            backend. The DIFFENGINE backend keeps parameters symbolic and
+            re-evaluates them on each solve; it requires expressions of
+            dimension at most 2.
         verbose : bool, optional
             If True, print verbose output related to problem compilation.
         solver_opts : dict, optional
@@ -928,10 +930,12 @@ class Problem(u.Canonical):
             When True, DPP problems will be treated as non-DPP,
             which may speed up compilation. Defaults to False.
         canon_backend : str, optional
-            'CPP' (default) | 'SCIPY'
+            'CPP' (default) | 'SCIPY' | 'COO' | 'DIFFENGINE'
             Specifies which backend to use for canonicalization, which can affect
             compilation time. Defaults to None, i.e., selecting the default
-            backend.
+            backend. The DIFFENGINE backend keeps parameters symbolic and
+            re-evaluates them on each solve; it requires expressions of
+            dimension at most 2.
         solver_opts: dict, optional
             Additional arguments to pass to the solver.
 
@@ -998,10 +1002,12 @@ class Problem(u.Canonical):
             When True, DPP problems will be treated as non-DPP,
             which may speed up compilation. Defaults to False.
         canon_backend : str, optional
-            'CPP' (default) | 'SCIPY'
+            'CPP' (default) | 'SCIPY' | 'COO' | 'DIFFENGINE'
             Specifies which backend to use for canonicalization, which can affect
             compilation time. Defaults to None, i.e., selecting the default
-            backend.
+            backend. The DIFFENGINE backend keeps parameters symbolic and
+            re-evaluates them on each solve; it requires expressions of
+            dimension at most 2.
         kwargs : dict, optional
             A dict of options that will be passed to the specific solver.
             In general, these options will override any default settings
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -116,8 +116,8 @@ class Cache:
         self.param_prog = None
         self.inverse_data = None
 
-    def make_key(self, solver, gp, ignore_dpp, use_quad_obj):
-        return (solver, gp, ignore_dpp, use_quad_obj)
+    def make_key(self, solver, gp, ignore_dpp, use_quad_obj, canon_backend):
+        return (solver, gp, ignore_dpp, use_quad_obj, canon_backend)
 
     def gp(self):
         return self.key is not None and self.key[1]
@@ -822,13 +822,13 @@ class Problem(u.Canonical):
         # Convert solver argument to upper case.
         if isinstance(solver, str):
             solver = solver.upper()
-        # Cache includes ignore_dpp and solver_opts['use_quad_obj']
-        # because they alter compilation.
+        # Cache includes ignore_dpp, solver_opts['use_quad_obj'], and
+        # canon_backend because they alter compilation.
         if solver_opts is None:
             use_quad_obj = None
         else:
             use_quad_obj = solver_opts.get('use_quad_obj', None)
-        key = self._cache.make_key(solver, gp, ignore_dpp, use_quad_obj)
+        key = self._cache.make_key(solver, gp, ignore_dpp, use_quad_obj, canon_backend)
         if key != self._cache.key:
             self._cache.invalidate()
             solving_chain = self._construct_chain(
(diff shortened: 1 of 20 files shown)


cvxpy/reductions/solvers/nlp_solvers/diff_engine/__init__.py
```diff
@@ -12,13 +12,12 @@
...
 
-"""
 CVXPY integration layer for the DNLP diff engine.
 
-This module converts CVXPY expressions to C expression trees
-for automatic differentiation.
+This module converts CVXPY expressions to C expression trees for automatic
+differentiation, shared between the NLP solver path (C_problem) and the DCP
+matrix-stuffing path (DiffengineConeProgram).
 """
```
Transurgeon, line 21: maybe we should call this structured differentiation?
After lots of reading, it seems like automatic kind of refers to something else?
Otherwise we can use algorithmic too.

cvxpy/problems/problem.py
Transurgeon, line 1: let's make sure these changes are needed, and if so, it would be good to understand why?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/c_problem.py
```diff
@@ -35,23 +35,46 @@ def __init__(self, cvxpy_problem: cp.Problem, verbose: bool = True):
...
         inverse_data = InverseData(cvxpy_problem)
+        self._build_capsule(
+            cvxpy_problem.objective.expr,
+            [c.expr for c in cvxpy_problem.constraints],
+            cvxpy_problem.parameters(),
+            inverse_data,
+            verbose,
+        )
+
+    @classmethod
```
Transurgeon: is the class method really helping that much? If there is a cleaner way to do this, I wouldn't mind. But if this adds simplicity, let's do it, why not.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/cone_stuffing.py
```diff
@@ -0,0 +1,118 @@
...
+            "does not support parametric variable bounds.")
+    if problem.parameters():
+        raise ValueError(
+            f"The {s.DIFFENGINE_CANON_BACKEND} canonicalization backend "
+            "does not yet support parametric problems. Solve with "
+            "ignore_dpp=True to evaluate parameters before "
+            "canonicalization. If you already passed ignore_dpp=True, the "
+            "remaining parameters most likely come from parametric "
+            "variable bounds, which the DIFFENGINE backend does not "
+            "support.")
```
Transurgeon, line 76: this is temporary and just a first step so that the PR can be more easily reviewed.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/cone_stuffing.py
```diff
@@ -0,0 +1,118 @@
...
+    n = inverse_data.x_length
+    # Encode as the single-column (constant-slice) tensors ParamConeProg
+    # decodes: [q; d] for the objective, [A | b] flattened column-major
+    # for the constraints, P flattened column-major for the quadratic
+    # term. Everything downstream then matches the tensor path exactly.
+    q_tensor = sp.csr_array(np.concatenate([q, [d]])[:, None])
+    A_tensor = sp.csc_array(
+        sp.hstack([A, sp.csc_array(b[:, None])]).reshape((-1, 1), order='F'))
+    P_tensor = (sp.csc_array(P.reshape((n * n, 1), order='F'))
+                if P is not None else None)
```
Transurgeon: do we need to do this? and also why are they called A_tensor and P_tensor, respectively?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
Transurgeon, line 1: are the changes here really necessary and that useful? If not, let's clean them up and maybe add them to the quadform PR. Also make sure you can reuse cvxpy API when possible, for example I am pretty there is such a thing as ``.is_constant()`` that could be used instead of is_plain_constant (maybe?)

cvxpy/reductions/solvers/nlp_solvers/diff_engine/helpers.py
Transurgeon, line 1: the changes here should maybe go in the quadform bindings PR. to keep this PR cleaner.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon, line 1: I think let's make these into their own functions and APIs in the diffengine, it will be much cleaner.

cvxpy/settings.py
```diff
@@ -206,6 +206,11 @@ def _env_var_to_bool(name: str, default: bool) -> bool:
 RUST_CANON_BACKEND = "RUST"
 CPP_CANON_BACKEND = "CPP"
 COO_CANON_BACKEND = "COO"  # 3D COO sparse tensor backend: O(nnz) operations for large parameters
+# Builds the problem matrices directly from the expression trees via the C
+# diff engine. Explicit opt-in via canon_backend="DIFFENGINE"; supports
+# parameter-free compilation only (pass ignore_dpp=True so parameters are
+# evaluated before canonicalization).
```
Transurgeon: we don't need that many comments, I think a one liner should suffice. that is straight to the point.

cvxpy/problems/problem.py
Transurgeon: so I think the reasoning makes sense. Before all backends produced the same thing, the tensor parameter maps. But now the diffengine does something different so we need to add that as something to detect to see if the cache can be reused.

SteveDiamond (approved):

PR #3449, opened 2026-07-08
Default ignore_dpp/non-DPP solves to the DIFFENGINE canon backend

Transurgeon:
## Description

Supersedes and consolidates #3484 and #3450 (both can be closed in favour of this).

Staging this as three PRs did not survive contact with the code. #3449 defaulted to DIFFENGINE on the `ignore_dpp` / non-DPP branch *because* `EvalParams` substitutes the parameter values, leaving stuffing parameter-free — and #3484 then removed `EvalParams` from exactly that path, demoting the case to a fallback and inverting #3449's own assertions (`test_complex_dpp.py` went from `assert EvalParams in reduction_types` to `assert EvalParams not in`). #3484 in turn delivered nothing measurable on its own, since it left the program uncached. Measured mean compile time per re-solve on a non-DPP parametric problem (`(p*p)*sum(x)` plus a 400×250 least-squares term):

| stage | first compile | re-solve compile | cached |
|---|---|---|---|
| master | 19.1 ms | 12.4 ms | no |
| #3449 alone | 18.5 ms | 10.4 ms | no |
| #3484 alone | 24.2 ms | 9.4 ms | no |
| this PR | 21.3 ms | **3.2 ms** | yes |

On this route parameters now stay symbolic through canonicalization, and stuffing returns a `DiffengineParamConeProg` that re-extracts `(q, d, A, b, P)` at the current values — so a re-solve costs one engine extraction instead of a full chain rebuild.

**Routing** (unchanged elsewhere):

| condition | backend | parameters |
|---|---|---|
| explicit `canon_backend` | as requested | symbolic if DIFFENGINE |
| `ignore_dpp` / non-DPP, parametric, eligible | DIFFENGINE | symbolic, program cached |
| `ignore_dpp` / non-DPP, parameter-free or ineligible | `EvalParams` + DIFFENGINE / tensor | n/a |
| parametric + DPP | CPP, or COO above `DPP_PARAM_THRESHOLD` | tensor |

### Two regressions fixed

Both were present in the staged PRs and are fixed here, each pinned by a test that fails if the guard is removed.

**Unsupported atoms hard-failed.** Auto-selected DIFFENGINE raised on atoms with no converter, so a non-DPP problem using `cp.concatenate` or `cp.diag(x, k != 0)`
failed where master solves — including when the user passed no flags at all. A `_supports_diffengine()` capability check, mirroring the existing
`_supports_cpp()` pattern, keeps those on the tensor backends; an eFFENGINE"` still fails loudly.

| | before | now |
|---|---|---|
| `concatenate`, non-DPP, no flags | `NotImplementedError` | 24.000
| `diag(x, 1)`, non-DPP, no flags | `NotImplementedError` | 15.0000 (= master) |

**Cached programs served stale numbers.** `safe_to_cache` stopped asking whether `EvalParams` was in the chain and started reading a flag derived from
`problem.parameters()`. But `perspective` keeps its function in `exgs`, so its parameters are invisible there, while `perspective_canon`
freezes their current values into numeric cone data. A cached re-solve then returned a wrong answer with no error: `2.499988` where the truth is `1.211103`.
Caching is now disabled by inspecting the atom directly, which alsoecond, visible parameter puts the problem on the symbolic route
(where there is no `EvalParams` to detect).

### Cleanups

- `ConicSolver.format_constraints`' block-diagonal restructuring matrix is now a module-level `build_restruct_blocks()`, with `build_restruct_mat_sparse()`
materializing the sign-only `IdentityOperator` / `NegativeIdentityOneConeFormat` calls it directly instead of recovering R by pushing a
stub `[I_m | 0]` program through `format_constraints`, which also stops routing an `m(m+1)`-row probe tensor through the reshape path. No solver interface is
affected — nothing overrides `format_constraints`, and its `cls` was unused.
- The five getter/setter pairs deferring `encode_cone_tensors` collapse into one `_lazy_tensor` descriptor. The deferral stays: eager re-encoding measures at ~2×
the cost of the re-extraction it accompanies (n=250: 2.38 ms vs 1.2.7 ms), which would erase most of what caching saves.

Issue link (if applicable): #2205, #1132

## Type of change
- [ ] Bug fix

- [x] Write unittests.
- [ ] Run the benchmarks to make sure your change doesn’t introduce

### DIFF
--- a/cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
+++ b/cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
@@ -37,12 +37,8 @@ def perspective_canon(expr, args, solver_context: SolverInfo | None = None):
     chain = aux_prob._construct_chain(solver=solver, solver_opts=solver_opts, ignore_dpp=True)
     chain.reductions = chain.reductions[:-1]  # skip solver reduction
     prob_canon = chain.apply(aux_prob)[0]  # grab problem instance
-    # get cone representation of c, A, and b for some problem.
-
-    q = prob_canon.q.toarray().flatten()[:-1]
-    d = prob_canon.q.toarray().flatten()[-1]
-    Ab = prob_canon.A.toarray().reshape((-1, len(q) + 1), order="F")
-    A, b = Ab[:, :-1], Ab[:, -1]
+    # get cone representation of q, A, and b for the parameter-free aux problem.
+    q, d, A, b = prob_canon.apply_parameters()
 
     # given f in epigraph form, aka epi f = \{(x,t) | f(x) \leq t\}
     # = \{(x,t) | Fx +tg + e \in K} for K a cone, the epigraph of the
--- a/cvxpy/reductions/dcp2cone/dcp2cone.py
+++ b/cvxpy/reductions/dcp2cone/dcp2cone.py
@@ -219,9 +219,12 @@ class Dcp2Cone(Canonicalization):
         A tuple of the canonicalized expression and generated constraints.
         """
         # Constant trees are collapsed, but parameter trees are preserved.
-        if isinstance(expr, Expression) and (
-                expr.is_constant() and not expr.parameters()):
-            return expr, []
+        # Canonicalizing parametric constants is sound here: chains where it
+        # would not be fold them to CallbackParam leaves first (see
+        # CallbackParamFold).
+        if isinstance(expr, Expression) and expr.is_constant():
+            if not expr.parameters():
+                return expr, []
 
         if self.quad_obj and affine_above and type(expr) in self.quad_canon_methods:
             # Special case for power.
--- a/cvxpy/reductions/fold_callback_params.py
+++ b/cvxpy/reductions/fold_callback_params.py
@@ -1,0 +1,81 @@
+"""
+Copyright, the CVXPY authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from cvxpy import problems
+from cvxpy.expressions.constants.callback_param import CallbackParam
+from cvxpy.reductions.reduction import Reduction
+from cvxpy.utilities import scopes
+
+
+def _fold_expr(expr):
+    """Replace maximal non-affine variable-free parametric subtrees.
+
+    Such a subtree (e.g. ``power(t, 2)``, ``floor(t)``, ``log_det(P)``)
+    becomes a ``CallbackParam`` evaluating it on each access. Parameter-affine
+    subtrees (a bare ``Parameter``, ``2 * p + A``) stay symbolic.
+    """
+    if not expr.parameters():
+        return expr
+    if not expr.variables():
+        with scopes.dpp_scope():
+            if expr.is_affine():
+                return expr
+        return CallbackParam(
+            callback=lambda e=expr: e.value,
+            shape=expr.shape,
+            nonneg=expr.is_nonneg(),
+            nonpos=expr.is_nonpos(),
+        )
+    new_args = [_fold_expr(arg) for arg in expr.args]
+    if all(id(new) == id(old) for new, old in zip(new_args, expr.args)):
+        return expr
+    return expr.copy(new_args)
+
+
+class CallbackParamFold(Reduction):
+    """Fold non-affine parametric-constant subtrees into CallbackParams.
+
+    Runs before canonicalization on chains where parameters stay symbolic
+    across solves: canonicalizing such subtrees would be unsound there
+    (``x <= power(t, 2)`` would epigraph-relax vacuously), and unlike
+    ``EvalParams`` the folded values refresh per solve, so compiled programs
+    stay cacheable.
+    """
+
+    def accepts(self, problem) -> bool:
+        return True
+
+    def apply(self, problem):
+        if len(problem.objective.parameters()) > 0:
+            obj_expr = _fold_expr(problem.objective.expr)
+            objective = type(problem.objective)(obj_expr)
+        else:
+            objective = problem.objective
+
+        constraints = []
+        for c in problem.constraints:
+            args = [_fold_expr(arg) for arg in c.args]
+            if all(id(new) == id(old) for new, old in zip(args, c.args)):
+                constraints.append(c)
+            else:
+                data = c.get_data()
+                if data is not None:
+                    constraints.append(type(c)(*(args + data)))
+                else:
+                    constraints.append(type(c)(*args))
+        return problems.problem.Problem(objective, constraints), []
+
+    def invert(self, solution, inverse_data):
+        return solution
--- a/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
+++ b/cvxpy/reductions/solvers/conic_solvers/conic_solver.py
@@ -19,8 +19,8 @@ import scipy.sparse as sp
 
 import cvxpy.settings as s
 from cvxpy.constraints import PSD, SOC, ExpCone, NonNeg, PowCone3D, PowConeND, SvecPSD, Zero
+from cvxpy.problems.param_prob import ParamProb
 from cvxpy.reductions.cvx_attr2constr import convex_attributes
-from cvxpy.reductions.dcp2cone.cone_matrix_stuffing import ParamConeProg
 from cvxpy.reductions.solution import Solution, failure_solution
 from cvxpy.reductions.solvers import utilities
 from cvxpy.reductions.solvers.solver import Solver
@@ -117,7 +117,7 @@ class ConicSolver(Solver):
     EXP_CONE_ORDER = None
 
     def accepts(self, problem):
-        return (isinstance(problem, ParamConeProg)
+        return (isinstance(problem, ParamProb)
                 and (self.MIP_CAPABLE or not problem.is_mixed_integer())
                 and not convex_attributes([problem.x])
                 and (len(problem.constraints) > 0 or not self.REQUIRES_CONSTR)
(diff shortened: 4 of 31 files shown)


cvxpy/reductions/dcp2cone/dcp2cone.py
```diff
@@ -58,6 +59,10 @@ def __init__(self, problem=None, quad_obj: bool = False, solver_context=None) ->
         # solver_context : The solver context: supported constraints and bounds.
         self.solver_context = solver_context
 
+        # Set during apply() when the cone quad_form canon factorizes a
+        # parametric P.value (see safe_to_cache in problem.py).
+        self._param_quad_form_factorized = False
+
```
Transurgeon: I am wondering if we can't just use the quad_form_dpp scope value somehow?

cvxpy/reductions/dcp2cone/canonicalizers/perspective_canon.py
```diff
@@ -30,11 +32,18 @@ def perspective_canon(expr, args, solver_context: SolverInfo | None = None):
...
+        aux_prob, _ = EvalParams().apply(aux_prob)
     # Does numerical solution value of epigraph t coincide with expr.f numerical
     # value at opt?
     solver_opts = {"use_quad_obj": False}
     solver = solver_context.solver_name if solver_context is not None else None
-    chain = aux_prob._construct_chain(solver=solver, solver_opts=solver_opts, ignore_dpp=True)
+    # Pin to CPP: this aux chain unpacks the raw .q/.A tensors below, so it
+    # must not ride backend-default changes.
+    chain = aux_prob._construct_chain(solver=solver, solver_opts=solver_opts, ignore_dpp=True,
+                                      canon_backend=CPP_CANON_BACKEND)
```
Transurgeon: we could cleanup the comments. also it would be important to understand what's really happening here, since diffengine should work (in theory).

cvxpy/reductions/solvers/conic_solvers/conic_solver.py
Transurgeon, line 1: I like these changes/cleanups but maybe they are not really in scope for this PR?
could we potentially avoid anything that changes too much the approach of applying R, the restructuring matrix? I think our goal will be to do a clean refactor of it as a new reduction, either later or in parallel.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/cone_stuffing.py
```diff
@@ -47,33 +47,41 @@
...
-    with the C diff engine instead of building parameter tensors.
+    """Stuff a problem by evaluating the expression trees with the C diff
+    engine instead of building parameter tensors.
 
-    Produces the same artifact as the tensor path: a ParamConeProg whose
+    A parameter-free problem produces a stock ParamConeProg whose
     single-column tensors hold the constant slice, so every downstream
     consumer (formatting, solvers, inversion) runs the stock code path.
+    A parametric problem produces a DiffengineParamConeProg, which keeps the
+    engine program alive and re-extracts on every apply_parameters().
```
Transurgeon: I don't like the word stock here. Also I don't know if we should have this distinction between non param and parametrized problems? Maybe it would be quite nice as a code structure for us to have a consistent story and use the same code path for both?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
```diff
@@ -129,13 +131,17 @@ def convert_symbolic_quad_form(expr, var_dict, n_vars, param_dict):
...
         x_c = convert_expr(x, var_dict, n_vars, param_dict)
+        n = x.size
+        if P.parameters():
+            # P is affine in the parameters and independent of x (Hessian still 2P):
+            # feed it as a matrix-valued child evaluated each solve. Peel value-identity
+            # Wrap atoms (e.g. psd_wrap) the converter can't build directly.
+            P_inner = P
+            while isinstance(P_inner, Wrap):
+                P_inner = P_inner.args[0]
+            P_c = convert_expr(P_inner, var_dict, n_vars, param_dict)
```
Transurgeon: I don't like having a while here. let's clean it up. Also please remind me why the changes to this quadform converter are necessary.

cvxpy/reductions/solvers/nlp_solvers/diff_engine/parametric_program.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class _Tensors(NamedTuple):
+    """ParamConeProg's coefficient tensors for one set of matrices."""
+
+    q: sp.csr_array
+    A: sp.csc_array
+    P: sp.csc_array | None
+    reduced_A: ReducedMat
+    reduced_P: ReducedMat
```
Transurgeon: I really don't like that these two are separate. Can't we just combine them into one? Why do we need the distinction? Can you go over the full pipeline of what's happening after this? So that we can check and ensure that there won't be any issues?
For example the matrix tuple just adds the constant parts of the problem data (i.e. d and b). These could easily be extended from the tensors no?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/parametric_program.py
```diff
@@ -0,0 +1,204 @@
...
+        super().__init__(None, x, None, variables, var_id_to_col, constraints,
+                         parameters, param_id_to_col, P=None,
+                         formatted=formatted,
+                         lower_bounds=lower_bounds, upper_bounds=upper_bounds)
+        # The parameter vector the stored matrices were extracted at.
+        # Kept per instance, NOT on the shared extractor: a restructured copy
+        # and its raw sibling can hold matrices from different parameter
+        # vectors. Extraction-once: construction itself extracted at the
+        # current values, so the first apply_parameters() short-circuits.
+        self._extracted_param_vec = self._param_vec()
```
Transurgeon: let's cleanup this comment, and also explain the distinction to me again? Why is this needed? to have an extracted and an original param_vec?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/parametric_program.py
```diff
@@ -0,0 +1,204 @@
...
+    def P(self):
+        return self._encode_tensors().P
+
+    @property
+    def reduced_A(self):
+        return self._encode_tensors().reduced_A
+
+    @property
+    def reduced_P(self):
+        return self._encode_tensors().reduced_P
```
Transurgeon: are the properties needed? can't we set these as attributes when we initialize the class?

cvxpy/reductions/solvers/nlp_solvers/diff_engine/registry.py
Transurgeon, line 1: let's revert the changes here if they are not necessary

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -107,6 +111,53 @@ def _fallback_solver(problem_form: ProblemForm) -> Solver:
...
+    if isinstance(expr, list):
+        # PartialProblem carries its constraints as a list argument.
+        return any(_freezes_parameters_at_canon(e) for e in expr)
+    if isinstance(expr, perspective):
+        for hidden in (expr.f, expr.f_recession):
+            if hidden is not None and (hidden.parameters()
+                                       or _freezes_parameters_at_canon(hidden)):
+                return True
+    return any(_freezes_parameters_at_canon(arg)
+               for arg in getattr(expr, 'args', ()))
```
Transurgeon: I don't like this naming and whatever. Can you explain if its actually necessary or only for perspective? I think we should be careful adding code that runs all the time for special cases.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -212,12 +263,48 @@ def _build_solving_chain(
             f"The {DIFFENGINE_CANON_BACKEND} backend cannot be used with "
             "problems that have expressions of dimension greater than 2.")
 
+    # Canonicalizers that compile a sub-problem substitute the parameter
+    # values in force at canon time, so such programs cannot be cached.
+    uncached_param_prog = any(
+        _freezes_parameters_at_canon(expr)
+        for expr in problem.constraints + [problem.objective.expr])
```
Transurgeon: why do we have this flag? what does it mean to have an uncached param prog? Isn't that only triggered when we change the canon_backend key?

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -254,8 +341,17 @@ def _build_solving_chain(
...
+        # The symbolic program must reach the solver already formatted, or
+        # ConicSolver.apply would rebuild it as a stock ParamConeProg and lose
+        # re-extraction. Only added on this route: appending it for every conic
+        # solver would change what _construct_chain(...)[:-1] yields, which
+        # callers rely on to obtain the unformatted program.
+        reductions.append(ConeFormat(solver_instance))
     reductions.append(solver_instance)
-    return SolvingChain(reductions=reductions, solver_context=solver_context)
+    return SolvingChain(reductions=reductions, solver_context=solver_context,
+                        uncached_param_prog=uncached_param_prog)
```
Transurgeon: I don't like the changes to the Solving Chain, its a very common class and we shouldn't have to make changes to it ideally.

cvxpy/reductions/fold_callback_params.py
Transurgeon, line 1: this is really nice, but potentially we could consider putting it in a separate PR? What do you think about that?

cvxpy/tests/test_cone_restructuring.py
Transurgeon, line 1: I don't understand the tests here. we could consider putting it in the separate cone format PR?

cvxpy/tests/test_diffengine_backend.py
```diff
@@ -182,19 +191,21 @@ def test_psd_solve_matches_default(self) -> None:
...
+            self.assertAlmostEqual(x.value, val, places=5)
 
     def test_ignore_dpp_parametric_solves_and_tracks_values(self) -> None:
-        """With ignore_dpp=True, EvalParams bakes the parameters and the
-        backend compiles the resulting parameter-free problem each solve."""
+        """With ignore_dpp=True the values are substituted, so the backend
+        compiles a parameter-free problem on each solve."""
         p = cp.Parameter()
         x = cp.Variable()
         prob = cp.Problem(cp.Minimize(cp.square(x - p)))
```
Transurgeon: here shouldn't the problem be parametrized? the comment says it is parameter-free?

cvxpy/tests/test_dpp.py
```diff
@@ -142,7 +143,7 @@ def test_solve_dpp_problem(self) -> None:
         x.value = 3
         self.assertAlmostEqual(problem.solve(cp.SCS), 6)
 
-    def test_chain_data_for_non_dpp_problem_evals_params(self) -> None:
+    def test_chain_data_for_non_dpp_problem_keeps_params_symbolic(self) -> None:
```
Transurgeon: this name is way too long no?

cvxpy/tests/test_ignore_dpp.py
```diff
@@ -0,0 +1,775 @@
...
+
+def _stuffing_backend(chain) -> str | None:
+    stuffing = [r for r in chain.reductions
+                if isinstance(r, ConeMatrixStuffing)][0]
+    return stuffing.canon_backend
+
+
+class TestIgnoreDppSelection(BaseTest):
+    """The ignore_dpp / non-DPP branch defaults to the DIFFENGINE backend when
+    the stuffed problem is parameter-free, and falls back silently otherwise."""
```
Transurgeon: we should update this comment? I think diffengine should support parametrized problems currently.

Transurgeon (commented):
Left a first round of review for myself.

PR #3452, opened 2026-07-11
Add DCP violation explanations for non-DCP problems

WilliamZhang20:
When DCP checks fail, identify the offending subexpressions and explain the composition, objective-sense, or constraint-type rule that was violated via richer DCPError messages and cp.explain_dcp.

In a future PR, one could add suggestions for rewriting problems/constraints to follow DCP, or even auto-rewrite to adhere to DCP or to use higher-level atoms instead (e.g. CVaR, norm2, etc.).

## Description

## Type of change
- [ ] Bug fix

- [x] Write unittests.


cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,18 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
+    def explain_dcp(self) -> str:
+        """Explain why this expression fails DCP, if it does.
+
+        Returns
+        -------
+        str
+            A human-readable explanation of DCP violations, or a short
+            success message if the expression follows DCP.
+        """
+        from cvxpy.utilities.debug_tools import explain_dcp
```
PTNobel: We prohibit inline imports unless needed to avoid a circular import. I don't think that's the case here?

cvxpy/problems/problem.py
```diff
@@ -291,6 +291,18 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
+    def explain_dcp(self) -> str:
+        """Explain why this problem fails DCP, if it does.
+
+        Returns
+        -------
+        str
+            A human-readable explanation of DCP violations, or a short
+            success message if the problem follows DCP.
+        """
+        from cvxpy.utilities.debug_tools import explain_dcp
```
PTNobel: Ditto

doc/source/faq/index.rst
```diff
@@ -30,13 +30,38 @@ The problems you solve in CVXPY must follow the rules of disciplined convex prog
...
+
 How do I find DCP errors?
 -------------------------
 You can test whether a problem, objective, constraint, or expression satisfies the DCP
 rules by calling ``object.is_dcp()``.
 If the function returns ``False``,
 there is a DCP error in that object.
 
+To see *why* an object fails DCP without calling ``solve()``, use
+``cp.explain_dcp(object)`` (also available as ``object.explain_dcp()`` on
```
PTNobel: I don't like having two identical ways to do the same thing. Is there a reason you did it this way?

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+    if decr:
+        return "nonincreasing"
+    return "nonmonotonic"
+
+
+def _atom_display_name(expr) -> str:
+    """Human-readable atom name for diagnostics."""
+    name = type(expr).__name__
+    # Power / PowerApprox are the implementation of power, sqrt, and square.
+    if name in ("Power", "PowerApprox") and hasattr(expr, "p"):
```
PTNobel: Please use `isinstance`. Should we even be doing this tho? Maybe this should be figured out by the power class.

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+    return name
+
+
+def format_expr_for_diagnostics(expr) -> str:
+    """Pretty-print an expression for DCP diagnostic messages.
+
+    Uses user labels when set, and renders ``power(., 0.5)`` / ``power(., 2)``
+    as ``sqrt`` / ``square`` so messages match how users typically write code.
+    """
+    if getattr(expr, "_label", None) is not None:
```
PTNobel: Do we need to access internal structures like this? Also, can an expr ever not have a _label?

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+        return expr._label
+
+    cls_name = type(expr).__name__
+    args = getattr(expr, "args", [])
```
PTNobel: I really don't like getattr use. Please explain why it is necessary here.

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+
+    if cls_name in ("Power", "PowerApprox"):
```
PTNobel: use isinstance.

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+        inner = format_expr_for_diagnostics(args[0])
+        try:
+            p = float(expr.p.value)
+        except (TypeError, ValueError, AttributeError):
+            return f"power({inner}, {expr.p.value})"
+        if p == 0.5:
```
PTNobel: Isn't this logic completely duplicate with a function above

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+            return f"sqrt({inner})"
+        if p == 2.0:
+            return f"square({inner})"
+        return f"power({inner}, {expr.p.value})"
+
+    if not args:
+        return expr.name() if hasattr(expr, "name") else str(expr)
```
PTNobel: Again why this defensiveness

PTNobel: name is an @abstractmethod on Expression, it is not possible for an expr to not have one. What types are you worried about here? Please, add type annotations to the function.

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+        return f"{left} {op_name} {right}"
+
+    data = []
+    if hasattr(expr, "get_data") and expr.get_data() is not None:
+        data = [str(elem) for elem in expr.get_data()]
+    parts = formatted_args + data
+    return f"{cls_name}({', '.join(parts)})"
+
+
+def format_constraint_for_diagnostics(constraint) -> str:
```
PTNobel: This cannot be the only place we implement this. Why is this here and not elsewhere?

cvxpy/utilities/debug_tools.py
```diff
@@ -24,6 +26,423 @@ def node_count(expr) -> int:
...
+    expr : Expression
+        A node that fails ``is_dcp()`` while every child passes (or a leaf).
+
+    Returns
+    -------
+    str or None
+        A reason string, or None if no specific explanation is available.
+    """
+    # Constraints are handled separately.
+    from cvxpy.constraints.constraint import Constraint
```
PTNobel: Inline imports are prohibited.

PTNobel (changes requested):
I started a review; I did not finish. Please ensure that PRs in the future follow our coding conventions. I flagged the major violations in my PR.

I am also extremely worried that much of this code is duplicate elsewhere, please confirm that there is no similar/identical logic elsewhere and let me know.

I look forward to fully reviewing after your edits!

cvxpy/constraints/nonpos.py
```diff
@@ -224,6 +244,17 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
         return self.expr.is_convex()
 
+    def dcp_failure_reason(self) -> str | None:
+        expr = self.expr
+        if expr.is_convex():
+            return None
+        lhs = self.args[0].format_labeled()
+        rhs = self.args[1].format_labeled()
+        return (
+            f"The inequality {lhs} <= {rhs} requires (lhs - rhs) to be convex, "
```
PTNobel: We normally state this as lhs convex, rhs concave. Also I don't think that we should refer to lhs and rhs but instead use the label.

PTNobel: It would also be nice to case 1. lhs not convex, 2. rhs not concave, 3. Both, separately.

cvxpy/constraints/zero.py
```diff
@@ -58,6 +58,16 @@ def is_dcp(self, dpp: bool = False) -> bool:
                 return self.args[0].is_affine()
         return self.args[0].is_affine()
 
+    def dcp_failure_reason(self) -> str | None:
+        expr = self.args[0]
+        if expr.is_affine():
+            return None
+        pretty = expr.format_labeled()
+        return (
+            f"Equality constraints require an affine expression, "
```
PTNobel: Shouldn't this say Zero?

cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,56 @@ def is_dcp(self, dpp: bool = False) -> bool:
                 return self.is_convex() or self.is_concave()
         return self.is_convex() or self.is_concave()
 
+    def atom_name(self) -> str:
+        """Short display name for this expression's atom (defaults to the class name)."""
+        return type(self).__name__
```
PTNobel: Hmm, I'm a little hesitant to add this as a public method. At least, it should return None if the expression is not an Atom. Maybe, we could have a return None here and in Atom this impl?

cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,56 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
+            return None
+
+        try:
+            atom_convex = self.is_atom_convex()
+            atom_concave = self.is_atom_concave()
+        except (AttributeError, NotImplementedError):
+            return None
+
+        atom_name = self.atom_name()
+        if not atom_convex and not atom_concave:
```
PTNobel: Can you factor the not out with DeMorgan? I think it'll be more readable

cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,56 @@ def is_dcp(self, dpp: bool = False) -> bool:
                 return self.is_convex() or self.is_concave()
         return self.is_convex() or self.is_concave()
 
+    def atom_name(self) -> str:
+        """Short display name for this expression's atom (defaults to the class name)."""
+        return type(self).__name__
+
+    def dcp_failure_reason(self) -> str | None:
```
PTNobel: Doesn't this impl belong in the Atom class? It immediately falls over if it's called on not an Atom

cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,56 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
+            atom_concave = self.is_atom_concave()
+        except (AttributeError, NotImplementedError):
+            return None
+
+        atom_name = self.atom_name()
+        if not atom_convex and not atom_concave:
+            pretty = self.format_labeled()
+            return (
+                f"{atom_name} is neither a convex nor a concave atom, "
+                f"so DCP cannot verify the curvature of {pretty}."
```
PTNobel: Should we list its `curvatures` maybe? idk or suggest NLP solving?

cvxpy/expressions/expression.py
```diff
@@ -406,6 +406,56 @@ def is_dcp(self, dpp: bool = False) -> bool:
...
+            if reason is not None:
+                reasons.append(reason)
+        if atom_concave:
+            reason = debug_tools.explain_composition_failure(self, "concave")
+            if reason is not None:
+                reasons.append(reason)
+
+        if not reasons:
+            return None
+        if len(reasons) == 1:
```
PTNobel: You can just fall through to the join. No need to special case

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+    if incr and decr:
+        return "constant"
+    if incr:
+        return "nondecreasing"
+    if decr:
+        return "nonincreasing"
+    return "nonmonotonic"
+
+
+def explain_composition_failure(expr: Expression, atom_curvature: str) -> str | None:
```
PTNobel: type for the second arg could be Literal["convex", "concave"]

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+        msg += f"\n{indent}{str_expr}"
+        if reason is not None:
+            msg += f"\n{indent}    Reason: {reason}"
+    return msg
+
+
+def _passes_discipline(expr, prop_name: str) -> bool:
+    if prop_name == "is_dcp":
+        return expr.is_dcp()
+    return expr.is_dgp()
```
PTNobel: uhh what is this function for? It's missing dqcp, DNLP, etc

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+) -> list[tuple[str, str | None]]:
+    """Collect minimal nodes that fail ``prop_name`` while all children pass."""
+    if res is None:
+        res = []
+    if len(expr.args) == 0 and _passes_discipline(expr, prop_name):
+        return res
+    if (not _passes_discipline(expr, prop_name) and
+            all(_passes_discipline(child, prop_name) for child in expr.args)):
+        if discipline_type == DCP:
+            str_expr = expr.format_labeled()
```
PTNobel: Maybe expr_label would be a better name?

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+            reason = expr.dcp_failure_reason()
+        else:
```
PTNobel: Are you assuming that if not DCP, dgp? That is definitely not okay

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+            str_expr = str(expr)
+            reason = None
+            if isinstance(expr, cvxtypes.variable()):
+                str_expr += " <-- needs to be declared positive"
```
PTNobel: Are you checking that it isn't? and maybe we should say `pos=True` somewhere in this description?

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+    prop_name: str,
+    discipline_type: str,
+    res: list[tuple[str, str | None]] | None = None,
+) -> list[tuple[str, str | None]]:
+    """Collect minimal nodes that fail ``prop_name`` while all children pass."""
+    if res is None:
+        res = []
+    if len(expr.args) == 0 and _passes_discipline(expr, prop_name):
+        return res
+    if (not _passes_discipline(expr, prop_name) and
```
PTNobel: I cannot figure out what this if is for

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+
+def _passes_discipline(expr, prop_name: str) -> bool:
+    if prop_name == "is_dcp":
+        return expr.is_dcp()
+    return expr.is_dgp()
+
+
+def _find_non_prop_leaves(
+    expr,
+    prop_name: str,
```
PTNobel: Maybe should this be a Literal[...]?

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+    discipline_type: str,
```
PTNobel: Also possibly a literal type would be wise?

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+    return res
+
+
+def explain_expression_dcp(expr: Expression) -> str:
+    """Explain why an expression fails DCP."""
+    if expr.is_dcp():
+        return "Expression follows DCP rules."
+    violations = _find_non_prop_leaves(expr, "is_dcp", DCP)
+    if not violations:
+        return "Expression does not follow DCP rules."
```
PTNobel, line 182: Wait why would this be if there are no violations

cvxpy/utilities/debug_tools.py
```diff
@@ -20,60 +30,197 @@
...
+    for child in expr.args:
+        res = _find_non_prop_leaves(child, prop_name, discipline_type, res)
+    return res
+
+
+def explain_expression_dcp(expr: Expression) -> str:
+    """Explain why an expression fails DCP."""
+    if expr.is_dcp():
+        return "Expression follows DCP rules."
+    violations = _find_non_prop_leaves(expr, "is_dcp", DCP)
```
PTNobel: Is it really leaves you are finding? And not nodes?

PTNobel (commented):
Huge improvement. I will review tests and docs when the impl is done

cvxpy/expressions/expression.py
WilliamZhang20: Fixed — `debug_tools` is imported at module top in `expression.py` (and elsewhere). No inline import remains.

cvxpy/constraints/nonpos.py
WilliamZhang20: Done — messages use labelled sides, state convex LHS / concave RHS, and case LHS-only / RHS-only / both.

cvxpy/utilities/debug_tools.py
WilliamZhang20: It’s the minimal-violator check: node fails discipline, but every child passes — those are the only nodes we report. Extracted as _is_minimal_violator in 5f882fc to make that clearer.

cvxpy/utilities/debug_tools.py
WilliamZhang20: Fixed — explicit DCP / DGP / else: raise.

cvxpy/utilities/debug_tools.py
WilliamZhang20: It’s only a dispatcher for the existing DCP/DGP error builder. Typed as Literal["is_dcp","is_dgp"] and raises otherwise. DQCP/DNLP out of scope here.

cvxpy/expressions/expression.py
PTNobel: When are these functions called? I feel like we could delete them and just have the atom class?

PR #3453, opened 2026-07-13
composed parametric coefficients must refresh through cached DIFFENGINE re-solves

Transurgeon:
## Description
Regression tests pinning the *composed parametric coefficient* contract on the DIFFENGINE cached re-solve paths, split out of #3450 for separate review.

A parameter expression that scales a constant matrix — `(p * A) @ x`, `quad_form(x, g * Sigma)`, or the `I/p` matrix the quadratic-objective canonicalization builds for `quad_over_lin` — reaches the C engine as a *composite side subtree*. On a cached program, those subtrees must re-evaluate when parameter values change. This PR adds canaries for exactly that, on both cached paths (`ignore_dpp` and the explicit-DPP backend), plus a scaled-scalar control that is correct by construction.

Test-design note: the formulations deliberately make the **optimal point** depend on the parameter. CVXPY recomputes the reported objective from the live expression at unpack time, so a parameter-invariant argmin would mask stale problem data.

**These tests fail against released `sparsediffpy 0.6.0`** (what CI installs today) — that is the point of this PR: it demonstrates the stale-value bug and documents what the engine release must guarantee. Against sparsediffpy 0.6.0 + cvxpy at this branch:

```
E  AssertionError: np.float64(0.4999...) != 0.005 within 4 places        # (p*A)@x stuck at p=1 optimum
E  AssertionError: np.float64(39.99...) != -0.01666... within 4 places   # quad_form(x, g*Sig) stale Q
E  AssertionError: array(-0.5) != -500.0 within 3 places                 # quad_over_lin symbolic I/p
3 failed, 1 passed  (the scaled-scalar control passes, as designed)
```

All four pass with the engine fix ([SparseDiffEngine#107](https://github.com/SparseDifferentiation/SparseDiffEngine/pull/107), which also carries C-level tests for the same shapes). Marked **draft** until an engine release containing that fix exists and the `sparsediffpy` pin is bumped accordingly.

Stacked on #3450 (pr-c) → #3449 (pr-b) → #3448 (pr-a); only the last commit (`ca08a26d1`) is new relative to #3450.

Issue link (if applicable): SparseDifferentiation/SparseDiffEngine#107

## Type of change
- [ ] Bug fix

- [x] Write unittests.

(The two unchecked boxes are deliberate: the new unittests *fail by design* against the released engine — see above — and pass locally against the fixed engine build; no benchmark-relevant code is touched.)

### DIFF
--- a/cvxpy/cvxcore/python/canonInterface.py
+++ b/cvxpy/cvxcore/python/canonInterface.py
@@ -290,6 +290,11 @@ def get_problem_matrix(linOps,
 
     # Allow to switch default backends through an environment variable for CI
     default_canon_backend = get_default_canon_backend()
+    # DIFFENGINE is not a tensor backend: it is resolved in
+    # construct_solving_chain and never reaches this helper. Other consumers
+    # (affine _grad, cone2cone transforms) fall back to the standard default.
+    if default_canon_backend == s.DIFFENGINE_CANON_BACKEND:
+        default_canon_backend = s.DEFAULT_CANON_BACKEND
     canon_backend = default_canon_backend if not canon_backend else canon_backend
 
     if canon_backend == s.CPP_CANON_BACKEND:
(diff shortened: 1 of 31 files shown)


Transurgeon:
Superseded by https://github.com/cvxpy/cvxpy/pull/3454 — the same contract as backend-agnostic tests based directly on master (no stack dependency). The DIFFENGINE-specific canaries (cache identity, explicit backend) remain part of #3450.

PR #3454, opened 2026-07-13
Add regression tests: composed parametric coefficients refresh across re-solves

Transurgeon:
## Description
Regression tests for a correctness contract that no current test pins: when a **composed parameter expression** scales a coefficient, reassigning the parameter must be reflected in the next solution, whatever the canonicalization backend caches across solves. Three shapes:

- `(p * A) @ x` — a scaled constant matrix as a linear coefficient;
- `quad_form(x, g * Sigma)` — a scaled constant matrix as the quadratic form;
- `quad_over_lin(x, p)` — a quadratic matrix (`I/p`) *derived* from the parameter by canonicalization.

Each is exercised on the default path and with `ignore_dpp=True`, with formulations whose **optimal point depends on the parameter** — the reported objective is recomputed from the live expression at unpack time, so a parameter-invariant argmin would silently mask stale problem data (we learned this the hard way while writing them).

All tests pass on master (58 insertions in `cvxpy/tests/test_dpp.py`, no source changes). Motivation: on the DIFFENGINE backend stack (#3448/#3449/#3450), these exact shapes served **stale values** across cached re-solves with released `sparsediffpy 0.6.0` (e.g. optimum stuck at `0.5` instead of `0.005` after `p: 1 -> 100`); fixed at the engine level in [SparseDiffEngine#107](https://github.com/SparseDifferentiation/SparseDiffEngine/pull/107). These backend-agnostic tests guard the contract for every backend, current and future.

Issue link (if applicable): #3455

## Type of change
- [ ] Bug fix

- [x] Write unittests.

(Tests only; no benchmark-relevant code touched.)

---

**Update — NLP path**: the same contract is *violated on master today* via the DNLP path (`solve(nlp=True)`), which caches the compiled diff-engine problem across solves: with the pinned `sparsediffpy 0.6.0`, the repro in #3455 returns the p=1 optimum after `p -> 10`. This PR therefore also adds NLP-path tests (`tests/nlp_tests/test_nlp_parameters.py`, IPOPT-gated) marked `xfail` with a pointer to [SparseDiffEngine#107](https://github.com/SparseDifferentiation/SparseDiffEngine/pull/107); they xpass on an engine build containing the fix, and the marker should be dropped when the pin moves to a release that includes it. Fixes are engine-side only — no cvxpy source change needed.

### DIFF
--- a/cvxpy/tests/nlp_tests/test_nlp_parameters.py
+++ b/cvxpy/tests/nlp_tests/test_nlp_parameters.py
@@ -460,3 +460,54 @@ class TestNlpParameters:
 
         assert np.allclose(param_sol1, hardcoded_sol1)
         assert np.allclose(param_sol2, hardcoded_sol2)
+
+
+@pytest.mark.skipif('IPOPT' not in INSTALLED_SOLVERS, reason='IPOPT is not installed.')
+class TestComposedParametricRefreshNlp:
+    """Cached NLP oracles must serve fresh values when a composed parameter
+    expression scales a coefficient (see TestComposedParametricRefresh in
+    tests/test_dpp.py for the conic-path counterparts).
+
+    The NLP path caches the compiled diff-engine problem across solves and
+    pushes new parameter values into it, so these tests exercise the engine's
+    composite param_source refresh. They fail against sparsediffpy releases
+    that predate the refresh fix (SparseDiffEngine#107): the re-solve then
+    optimizes the problem frozen at the first parameter value.
+    """
+
+    XFAIL_REASON = (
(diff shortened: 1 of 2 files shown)


Transurgeon:
@dance858 here are the python side tests, two are for DNLP and the others for CVXPY's conic DPP canonicalisation (they should still pass once we add the diffengine backend). Let me know if there are more that you would like to see!

dance858:
Awesome! Very subtle bug indeed. I feel like most bug fixes we do are related to parameters hehe.... It's great that you use SparseDiff as the backend because that certainly expands our test suite

SteveDiamond (approved):

PR #3458, opened 2026-07-15
Add problem-level constraint violation helper

curious7-web:
Stacked on top of #3434. This is a draft follow-up and should be reviewed only after #3434 is finalized, since it depends on the scalar constraint-level `Constraint.violation()` convention.

This PR adds `Problem.violation()` as a thin problem-level wrapper over the scalar constraint violation convention from #3434.

It returns the maximum scalar violation across the problem’s constraints, or `0.0` for unconstrained problems.

Tests cover:
- unconstrained problems returning `0.0`
- returning the maximum violation across multiple constraints
- raising when constraint expressions are None-valued


PTNobel:
FYI @rileyjmurray is not supportive of adding this API till we've figured out how to measure stationarity/duality gap.

curious7-web:
Thanks for the heads up @PTNobel.

I don’t want to move the public `Problem.violation()` API ahead of the broader stationarity / duality-gap design. I’ll leave this as a parked draft follow-up for now and focus on getting #3434 finalized first.

PR #3460, opened 2026-07-16
Replace Sum shape inference with exact tuple formula

### LINKED ISSUE #3459: Replace Sum shape inference with an exact formula
### Problem

`Sum.shape_from_args` currently infers its output shape by constructing a dummy NumPy array and calling `np.sum` on it. PR #3238 fixes the immediate signaling-NaN / floating-point exception issue from #3235, but the broader implementation still relies on allocation and a dummy reduction just to compute a shape tuple.

That is unnecessary work during compilation and keeps shape inference tied to a runtime array operation instead of explicit shape logic.

### Proposed solution

Replace the dummy `np.sum`-based shape inference with a direct tuple-based formula that computes the output shape from:

- the input expression shape
- `axis`
- tuple axes
- negative axes
- `axis=None`
- `keepdims=True` / `False`
- scalar-shape cases

The implementation should match NumPy’s `sum` shape behavior without allocating a dummy array or performing a reduction.

### Context

Follow-up from #3238 and #3235.

curious7-web:
## Description
This PR refactors Sum.shape_from_args by replacing the dummy np.sum array allocation with an explicit tuple-based formula. 

Following the maintainer suggestion, this implementation utilizes NumPy's `numpy.lib.array_utils.normalize_axis_tuple` to natively handle edge cases like negative indices, shorthands, and out-of-bounds validation without runtime overhead.

Issue link (if applicable): Closes #3459

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/atoms/affine/sum.py
+++ b/cvxpy/atoms/affine/sum.py
@@ -20,6 +20,7 @@ from types import GeneratorType
 
 import numpy as np
 from numpy.exceptions import AxisError
+from numpy.lib.array_utils import normalize_axis_tuple
 
 import cvxpy.interface as intf
 import cvxpy.lin_ops.lin_op as lo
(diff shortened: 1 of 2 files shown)


Transurgeon (approved):
really nice fix @curious7-web ! thanks a lot for your work :).

PR #3462, opened 2026-07-19
Fix matrix_frac crash on a matrix argument with an ndarray P

chuenchen309:
## Description

`cp.matrix_frac(X, P)` computes `tr(Xᵀ P⁻¹ X)`, and a matrix `X` (shape `(n, k)`, `k > 1`) is a supported, tested case. But when `P` is a numpy `ndarray` — the most natural way to pass a constant matrix, e.g. `cp.matrix_frac(X, np.eye(n))` — it raises `ValueError: Invalid dimensions for arguments to quad_form`, and a `Problem` using it cannot even be constructed.

The `matrix_frac` wrapper takes a `QuadForm` shortcut whenever `P` is an `ndarray`, but `QuadForm` only accepts a vector `x`. The same `X, P` succeed through every other path — `cp.Constant(P)`, a `Variable` `P`, and `P.tolist()` — all returning the correct `tr(Xᵀ P⁻¹ X)`. So the failure is triggered purely by the *type* of `P`, not the math.

The fix restricts the `QuadForm` shortcut to a vector `x`; a matrix `X` falls through to `MatrixFrac`, which already handles an `ndarray` `P` by casting it to a `Constant`. A vector `x` still uses the shortcut unchanged, and mismatched dimensions still raise the proper error.

Verified: `cvxpy/tests/test_atoms.py` passes (122), the value matches `np.trace(X.T @ inv(P) @ X)` and the `Constant(P)` path, and a full `Problem(Minimize(cp.matrix_frac(Variable((n, k)), P)))` now constructs and solves to `optimal`. Added a regression test.

Issue link (if applicable): none — found during atom-behavior verification.

## Type of change
- [x] Bug fix

- [ ] Add our license to new files. *(N/A — no new files; only `cvxpy/atoms/matrix_frac.py` and `cvxpy/tests/test_atoms.py` are edited.)*
- [x] Write unittests.
- [x] Run the unittests and check that they’re passing. *(`test_atoms.py`: 122 passed.)*
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression. *(Not run; the change only narrows a dispatch condition and adds no work on the existing vector path.)*

---

*Disclosure: this change was found, fixed, tested and described by an AI coding agent (Claude Code) running on this account. The account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.*

### DIFF
--- a/cvxpy/atoms/matrix_frac.py
+++ b/cvxpy/atoms/matrix_frac.py
@@ -146,7 +146,9 @@ class MatrixFrac(Atom):
 
 @wraps(MatrixFrac)
 def matrix_frac(X, P) -> Expression:
-    if isinstance(P, np.ndarray):
+    # The QuadForm shortcut only accepts a vector x, so restrict it to that case;
+    # a matrix X must go through MatrixFrac (which also handles an ndarray P).
+    if isinstance(P, np.ndarray) and Expression.cast(X).shape in [(P.shape[0], 1), (P.shape[0],)]:
         invP = LA.inv(P)
         return QuadForm(X, (invP + np.conj(invP).T) / 2.0)
     else:
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -356,6 +356,17 @@ class TestAtoms(BaseTest):
         atom = cp.matrix_frac(self.x, self.A)
         self.assertEqual(atom.shape, tuple())
         self.assertEqual(atom.curvature, s.CONVEX)
+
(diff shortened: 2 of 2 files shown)


PTNobel:
Hello! Thank you for your contribution! We cannot review it until you sign the CLA.

Additionally, we require use of our PR template, even for agent drafted PRs.

PTNobel:
Hello, please sign the CLA so we can review your PR!

PR #3466, opened 2026-07-21
Make bmat promote scalars and 1-D blocks like numpy.block

### LINKED ISSUE #2328: Unable to concatenate a cp.Variable vector
**Describe the bug**
I'm trying to make an LMI constraint like:

[[scalar, variable2D'],
[variable2D, identity2D]] >>0

but is impossible due to the fact that the solver does not concatenate the elements. The entire matrix at the end has a shape = (2,2) and not (3,3) as expected. 

**To Reproduce**
```
import cvxpy as cp
import numpy as np

# scalar
u = 3
# variable
x = cp.Variable(2)
c_vert = cp.vstack([u, x])
```

**Expected behavior**
`c_vert` must be a vector with 3 elements. First one as known scalar and second ones from `x` variable vector.

**Output**
```
  File "/home/*/.local/lib/python3.8/site-packages/cvxpy/atoms/affine/vstack.py", line 29, in vstack
    return Vstack(*arg_list)
  File "/home/*/.local/lib/python3.8/site-packages/cvxpy/atoms/atom.py", line 50, in __init__
    self.validate_arguments()
  File "/home/*/.local/lib/python3.8/site-packages/cvxpy/atoms/affine/vstack.py", line 74, in validate_arguments
    raise ValueError(("All the input dimensions except"
ValueError: All the input dimensions except for axis 0 must match exactly.
```

**Version**
 - OS: Linux Ubuntu 20.04
 - CVXPY Version: 1.4.1

enzo-boulin:
## Description
Fixes #2328

### Problem
`cp.bmat` raised `ValueError: all the input arrays must have same number of dimensions` when a block row mixed a scalar or vector with a matrix — e.g. the LMI / Schur-complement construction from the issue:

```python
U = cp.Variable((2, 1))
cp.bmat([[4, U.T], [U, np.identity(2)]])   # used to raise
```

`numpy.block` handles this by promoting every block to a common dimensionality; `bmat` stacked them as-is.

### Fix
Promote scalars to `(1, 1)` and 1-D blocks to `(1, n)` before stacking, matching `numpy.block` semantics. Existing all-matrix usage is unchanged, so the change is backwards compatible.

### Verification
- The example above now returns shape `(3, 3)` and is numerically identical to `numpy.block` on constant inputs.
- Works inside an actual SDP (`M >> 0`) and stays affine with `Variable` inputs.
- Added regression test `test_bmat_promotes_scalars`.
- Full `test_atoms.py` passes; `ruff` clean.

Issue link (if applicable): [2328](https://github.com/cvxpy/cvxpy/issues/2328)

## Type of change
- [ ] Bug fix

## Contribution checklist
- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -1707,6 +1707,29 @@ class TestAtoms(BaseTest):
                                       np.array([[1, 2]]).T])])
         self.assertItemsAlmostEqual(expr, const)
 
+    def test_bmat_promotes_scalars(self) -> None:
+        """bmat should promote scalars and 1-D blocks like numpy.block.
+
+        Regression test for https://github.com/cvxpy/cvxpy/issues/2328.
+        """
+        # Mixing a scalar, a vector, and a matrix (e.g. an LMI) used to raise.
+        U = np.array([[10.0], [20.0]])
+        expr = cp.bmat([[4, U.T], [U, np.identity(2)]])
+        ref = np.block([[4, U.T], [U, np.identity(2)]])
+        self.assertEqual(expr.shape, (3, 3))
+        self.assertItemsAlmostEqual(expr.value, ref)
+
+        # A scalar and a 1-D vector on the same row.
+        expr = cp.bmat([[1, np.array([5.0, 6.0])]])
+        self.assertEqual(expr.shape, (1, 3))
+        self.assertItemsAlmostEqual(expr.value, np.array([[1.0, 5.0, 6.0]]))
+
+        # Works with Variables and stays affine.
+        x = cp.Variable((2, 1))
+        expr = cp.bmat([[4, x.T], [x, np.identity(2)]])
+        self.assertEqual(expr.shape, (3, 3))
+        self.assertTrue(expr.is_affine())
+
     def test_conv(self) -> None:
         """Test the conv atom.
         """
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -1730,6 +1730,16 @@ class TestAtoms(BaseTest):
         self.assertEqual(expr.shape, (3, 3))
         self.assertTrue(expr.is_affine())
 
+        # A scalar Variable is promoted just like a scalar constant.
+        s = cp.Variable()
+        expr = cp.bmat([[s, x.T], [x, np.identity(2)]])
+        self.assertEqual(expr.shape, (3, 3))
+        self.assertTrue(expr.is_affine())
+        s.value = 7.0
+        x.value = np.array([[10.0], [20.0]])
+        ref = np.block([[7.0, x.value.T], [x.value, np.identity(2)]])
+        self.assertItemsAlmostEqual(expr.value, ref)
+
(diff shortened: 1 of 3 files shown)


cvxpy/tests/test_atoms.py
```diff
@@ -1707,6 +1707,29 @@ def test_bmat(self) -> None:
...
+        self.assertEqual(expr.shape, (3, 3))
+        self.assertItemsAlmostEqual(expr.value, ref)
+
+        # A scalar and a 1-D vector on the same row.
+        expr = cp.bmat([[1, np.array([5.0, 6.0])]])
+        self.assertEqual(expr.shape, (1, 3))
+        self.assertItemsAlmostEqual(expr.value, np.array([[1.0, 5.0, 6.0]]))
+
+        # Works with Variables and stays affine.
+        x = cp.Variable((2, 1))
```
PTNobel, line 1728: Should we test promotion of scalar variables?

enzo-boulin: yes of course. I added a case covering a scalar Variable being promoted (0-d). I checked and it stays affine and solves inside an SDP

PTNobel:
Does this make our API the same as np.block now? Or does it support more features we haven't covered

enzo-boulin:
> Does this make our API the same as np.block now? Or does it support more features we haven't covered

It matches for the common (2-D) case but isn't a full drop-in.

This PR fixes the dimension-promotion rule inside a 2-D block matrix: scalars → (1,1) and 1-D blocks → (1, n) before stacking, which is what np.block does for a list-of-lists. So the LMI/Schur-complement pattern from #2328 now agrees with np.block numerically.

The remaining gaps :

- Requires a list of lists. np.block also accepts a bare array (np.block(x)), a flat single-level list (np.block([a, b]) -> concatenate along the last axis), and arbitrary nesting depth. bmat expects exactly two levels and raises otherwise.
- Fixed concatenation axes. bmat is built on hstack/vstack, which concatenate along axis 1 and axis 0. That equals np.block for 2-D, but np.block concatenates the innermost list along the last axis and the next along the second-to-last, so for ≥3-D blocks the results differ (e.g. np.block([a3, a3]) → (2, 2, 4) vs bmat → (2, 4, 2)) and nesting depth isn't used to infer output dims.

So no new features beyond promotion, this just closes the promotion gap within bmat's existing 2-D list-of-lists contract. I can add a note to the docstring spelling out these limitations

PTNobel:
Ah of course. No need for docs, but if you could file a "implement cp.block" issue that includes the text of that comment and want to add a line to your pr that adds a link to that issue to the contributing wishlist, it would be nice to make progress on a long term solution!

enzo-boulin:
> Ah of course. No need for docs, but if you could file a "implement cp.block" issue that includes the text of that comment and want to add a line to your pr that adds a link to that issue to the contributing wishlist, it would be nice to make progress on a long term solution!

Sure ! It is my first time doing this so I hope I did what you meant.

On a side note I accidentally opened the first issue with my work account, I closed it and reopened the same issue with my personal account, sorry for that.

PTNobel (approved):

PR #3467, opened 2026-07-26
Fix SCIP interface misclassifying single-SOC problems as continuous LPs

HugoDelatte:
## Description

The SCIP interface applies special settings to continuous LPs so that dual values can be recovered:

```python
# scip_conif.py, _set_params
is_mip = data[s.BOOL_IDX] or data[s.INT_IDX]
has_soc_constr = len(dims[s.SOC_DIM]) > 1
if not (is_mip or has_soc_constr):
    model.setPresolve(SCIP_PARAMSETTING.OFF)
    model.setHeuristics(SCIP_PARAMSETTING.OFF)
    model.disablePropagation()
```

`dims[s.SOC_DIM]` is a list of cone sizes, so a problem with exactly **one** SOC constraint has `len(...) == 1` and the `> 1` check misclassifies it as a continuous LP. The same check is repeated in `_solve` to gate dual recovery via `getDualsolLinear`.

This contradicts the intended policy from scipopt/scip#22, where it was agreed these settings should apply only to continuous LPs, not SOCPs. Problems with two or more cones already skip this path but single-cone SOCPs (e.g. the classic portfolio problem `Minimize(norm(A @ w, 2))`) don't.

Consequences of taking the LP path on a single-cone SOCP:

- With PySCIPOpt >= 6.0.0 (SCIP 10), `model.optimize()` crashes with a **native integer-divide-by-zero inside libscip**. Reproduced on Windows (process exit code `0xC0000094`). Because it is a native fault, it kills the Python process and cannot be caught with try/except.
- With PySCIPOpt 5.x (SCIP 9), the same path happened to return a correct primal, which is why this went unnoticed.
- Mixed-integer problems (including MISOCPs) are unaffected: `is_mip` already routes them away from this path, and they work fine on PySCIPOpt 6.

Minimal reproduction:

```python
import cvxpy as cp
import numpy as np

rng = np.random.default_rng(0)
A = rng.normal(size=(30, 10))
w = cp.Variable(10)

problem = cp.Problem(
    cp.Minimize(cp.norm(A @ w, 2)),  # exactly one SOC cone
    [cp.sum(w) == 1, w >= 0],
)
problem.solve(solver=cp.SCIP)
print(problem.status, problem.value)
```

- PySCIPOpt 6.2.1 (SCIP 10) on Windows / Python 3.13 / cvxpy 1.9.2: hard crash during `optimize()`.
- PySCIPOpt 5.7.1 (SCIP 9), same environment: solves to `optimal`.
- The same problem with binary variables added (MISOCP) solves fine on PySCIPOpt 6.

Fix:

Treat any SOC constraint as non-LP, and apply this consistently to both the solver settings and the dual-recovery gate (computed once so the two sites cannot drift apart again):

```python
has_soc_constr = len(dims[s.SOC_DIM]) > 0
```

Resulting behavior:

| Problem | LP dual settings | Dual recovery |
|---|---|---|
| Continuous LP | yes | yes |
| Continuous SOCP (any number of cones) | no | no |
| MIP / MISOCP | no (unchanged) | no (unchanged) |

Behavior change:

Continuous problems with a single SOC previously returned dual values for their linear constraints. After this change they return no duals, consistent with problems that have two or more cones. The interface documents that duals are supported for LPs only, the code comment in `_solve` notes these values are not always correct, and every SCIP SOCP test in the suite already runs with `duals=False`.

Tests:

- Added a regression test in `cvxpy/tests/test_conic_solvers.py::TestSCIP` with a continuous SOCP containing exactly one cone (the existing SOCP tests did not cover this case, which is how the bug survived).
- SCIP LP dual tests still pass. SCIP SOCP/MI tests pass with `duals=False` where applicable (`test_scip_socp_0` aligned with the other SOCP tests).

## Type of change
- [x] Bug fix

- [x] Write unittests.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2523,7 +2523,7 @@ class TestSCIP(unittest.TestCase):
         StandardTestLPs.test_lp_4(solver="SCIP")
 
     def test_scip_socp_0(self) -> None:
-        StandardTestSOCPs.test_socp_0(solver="SCIP")
+        StandardTestSOCPs.test_socp_0(solver="SCIP", duals=False)
 
     def test_scip_socp_1(self) -> None:
         StandardTestSOCPs.test_socp_1(solver="SCIP", places=2, duals=False)
@@ -2536,6 +2536,18 @@ class TestSCIP(unittest.TestCase):
         StandardTestSOCPs.test_socp_3ax0(solver="SCIP", duals=False)
         # axis 1
         StandardTestSOCPs.test_socp_3ax1(solver="SCIP", duals=False)
+
+    def test_scip_socp_single_cone(self) -> None:
+        # SOCPs with a single cone must not use the continuous-LP dual path.
+        np.random.seed(0)
+        A = np.random.randn(30, 10)
+        w = cp.Variable(10)
+        problem = cp.Problem(
+            cp.Minimize(cp.norm(A @ w, 2)),
+            [cp.sum(w) == 1, w >= 0],
+        )
+        problem.solve(solver="SCIP")
+        self.assertEqual(problem.status, cp.OPTIMAL)
 
     def test_scip_mi_lp_0(self) -> None:
         StandardTestLPs.test_mi_lp_0(solver="SCIP")
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2526,7 +2526,7 @@ class TestSCIP(unittest.TestCase):
         StandardTestSOCPs.test_socp_0(solver="SCIP", duals=False)
 
     def test_scip_socp_1(self) -> None:
-        StandardTestSOCPs.test_socp_1(solver="SCIP", places=2, duals=False)
+        StandardTestSOCPs.test_socp_1(solver="SCIP", places=3, duals=False)
 
     def test_scip_socp_2(self) -> None:
         StandardTestSOCPs.test_socp_2(solver="SCIP", places=2, duals=False)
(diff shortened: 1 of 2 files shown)


PTNobel:
Please sign the CLA and use our PR template so we can review your PR!

HugoDelatte:
@PTNobel CLA singed and PR updated to follow the template.
Thank you

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2537,6 +2537,18 @@ def test_scip_socp_3(self) -> None:
         # axis 1
         StandardTestSOCPs.test_socp_3ax1(solver="SCIP", duals=False)
 
+    def test_scip_socp_single_cone(self) -> None:
```
PTNobel: Does it make sense to add this test problem to all the solver interfaces?

PTNobel (commented):
Change looks great, one q about the tests

cvxpy/tests/test_conic_solvers.py
HugoDelatte: Good question. I looked into it and SCIP is the only interface that changes solver settings based on problem structure to recover duals (setPresolve/setHeuristics/disablePropagation when it classifies the problem as a continuous LP). So it doesn't have an analogue in the other interfaces.

Also worth correcting my original framing: socp_0, socp_1 and socp_2 all have exactly one cone ([3], [4], [2]), so they were already on the buggy path but they don't fail because they're small enough that SCIP still solves them with presolve, heuristics and propagation disabled and the new problem is large enough to catch it. So what's actually distinctive about the new problem is its size so I've renamed it accordingly.

Also the fix does improve accuracy: test_scip_socp_1 goes from places=2 to places=3, so I've tightened it too in this PR.

PTNobel: I still think we should just make it a generic test problem we hold all solvers too

Transurgeon (approved):
changes look good and minimal, thanks!

cvxpy/tests/test_conic_solvers.py
HugoDelatte: Agreed, it would be the first continuous SOCP test combining equality and inequality constraints with a dense, moderately sized problem. Do you want me to apply this test to all solvers in this PR?

Transurgeon: could you open a new PR if you don't mind? I think it would be more suited as a separate one, since this is a bug fix.

HugoDelatte: Yes I'll do that

PTNobel: Thank you Hugo!

PR #3470, opened 2026-07-27
Stop internal parameter checks warning about sparse .value reads

### LINKED ISSUE #3367: Spurious sparse-`.value` RuntimeWarning emitted by Problem.solve() internal Parameter check
## Summary

`Problem.solve()` reads `parameter.value` on **sparse** Parameters during its own pre-solve validation loop, which emits a spurious

```
RuntimeWarning: Reading from a sparse CVXPY expression via `.value` is discouraged. Use `.value_sparse` instead
```

The warning is triggered by CVXPY's *internal* code, not by user code, so a user who only ever uses `.value_sparse` (as intended) still gets warned on every solve. Downstream packages that legitimately use sparse-pattern Parameters (e.g. [DCCP](https://github.com/cvxgrp/dccp), which stores sparse gradients via `.value_sparse`) currently have to suppress this warning by hand.

## Reproduction (CVXPY only)

```python
import scipy.sparse as sp
import cvxpy as cp

x = cp.Variable(2)
P = cp.Parameter((2, 2), sparsity=([0, 1], [0, 1]))
P.value_sparse = sp.eye(2).tocoo()          # set the value the recommended way

prob = cp.Problem(cp.Minimize(cp.sum_squares(P @ x - 1)))
prob.solve()                                 # <-- emits the RuntimeWarning
```

Tested on **cvxpy 1.9.1**.

## Root cause

`Problem._solve()` validates that every Parameter has a value:

```python
# cvxpy/problems/problem.py:1019-1020
for parameter in self.parameters():
    if parameter.value is None:
        raise error.ParameterError(...)
```

`Leaf.value` warns whenever the leaf is sparse:

```python
# cvxpy/expressions/leaf.py:527-530
if self.sparse_idx is None:
    return self._value
else:
    warn('Reading from a sparse CVXPY expression via `.value` is discouraged.'
         ' Use `.value_sparse` instead', RuntimeWarning)
    ...
```

So the internal "is the value set?" check on a sparse Parameter trips the very warning that tells users not to read `.value` on sparse leaves.

## Suggested fix

Have CVXPY's internal value-presence checks avoid the warning path for sparse leaves, e.g. test `parameter._value is None` directly, or use `parameter.value_sparse` / a non-warning internal accessor when `sparse_idx is not None`. A quick grep for internal `.value is None` / `.value` reads on parameters would likely surface a few more such sites (the verbose pre-solve loop above is the one that reproduces here).

## Impact

- Spurious warning on every `solve()` for any problem containing a sparse Parameter, even when the user only ever assigns via `.value_sparse`.
- Forces downstream libraries to wrap solves in `warnings.catch_warnings()` to keep their users' output clean.

AbdullahRasheed45:
## Description
Reading `.value` on a leaf that has a sparsity pattern warns and points the user at `.value_sparse`. Several of CVXPY's own checks read parameter values through that same property, so a user who only ever touches `.value_sparse` — as the warning asks — still gets warned on every `solve()`, from CVXPY's internals rather than from their own code. Downstream packages that legitimately use sparse-pattern parameters (DCCP is the example in the issue) currently suppress it by hand.

Reproducing on `main`, this emits the warning four times:

```python
x = cp.Variable(2)
P = cp.Parameter((2, 2), sparsity=([0, 1], [0, 1]))
P.value_sparse = sp.eye(2).tocoo()
cp.Problem(cp.Minimize(cp.sum_squares(P @ x - 1))).solve()
```

Tracing every warning back to its caller turned up five internal sites, one more than the issue identifies:

- `Problem._solve` — `if parameter.value is None`
- `Atom.value` — `if any([p.value is None for p in self.parameters()])`
- `CvxAttr2Constr.apply` — `if param.value is not None`
- `CvxAttr2Constr.update_parameters` — `if param.value is not None`
- `Expression._value_impl` — `return self.value`, reached for a leaf

The fifth (`update_parameters`) only fires on the **DPP re-solve** path, so it survives a fix that only covers the first solve. It is already visible as a stray warning in the existing `test_sparse_parameter` test output on `main`.

Four of the five only need to know *whether* a value was assigned, not what it is.

## Change

- Split the densification in `Leaf.value` out into `_value_dense()`, which does the same work without warning, so the warning now sits in `value` alone and the logic is not duplicated.
- Add a `Leaf._has_value` predicate for the assigned-or-not checks.
- Point the four checks at `_has_value` and override `Leaf._value_impl` to use `_value_dense()`.

Reading or writing `.value` directly on a sparse leaf still warns — that is the case the warning exists for, and it is covered by the existing `test_sparsity_read_value`.

`_has_value` is equivalent to the old check in both cases: for a dense leaf `value` returns `_value` directly, and for a sparse leaf `value` returns `None` exactly when `_value is None`.

Issue link: closes #3367

## Type of change
- [x] Bug fix

- [x] Write unittests.

Notes on the two unchecked boxes, rather than leaving them silent:

- **License headers** — no new files were added, so there was nothing to add a header to.
- **Benchmarks** — I did not run them. The change removes work rather than adding it (`_has_value` is an attribute test where the old code densified a sparse leaf into a full `np.zeros` array just to compare against `None`), so I would expect it to be neutral-to-faster on the sparse-parameter path and unchanged elsewhere. Happy to run them if you would like that confirmed before merge.

## Tests

Added to `cvxpy/tests/test_attributes.py`:

- `test_solve_does_not_warn_on_sparse_parameter` — asserts no sparse-`.value` warning escapes `solve()`, and separately re-solves after updating `value_sparse` to cover the DPP path that the fifth call site sits on. Fails on `main`.
- `test_sparse_parameter_without_value_still_raises` — the checks being changed are the ones that reject unset Parameters, so this pins that `ParameterError` is still raised.

`pytest cvxpy/tests/test_attributes.py cvxpy/tests/test_problem.py cvxpy/tests/test_expressions.py cvxpy/tests/test_param_cone_prog.py cvxpy/tests/test_param_quad_prog.py cvxpy/tests/test_parametric_bounds.py` → **331 passed, 4 skipped**. `ruff` clean on all five files.

### DIFF
--- a/cvxpy/atoms/atom.py
+++ b/cvxpy/atoms/atom.py
@@ -450,7 +450,7 @@ class Atom(Expression):
 
     @property
     def value(self) -> ExpressionValue | None:
-        if any([p.value is None for p in self.parameters()]):
+        if any([not p._has_value for p in self.parameters()]):
             return None
         return self._value_impl()
 
--- a/cvxpy/expressions/constants/callback_param.py
+++ b/cvxpy/expressions/constants/callback_param.py
@@ -53,3 +53,8 @@ class CallbackParam(Parameter):
     @value.setter
     def value(self, _val):
         raise NotImplementedError("Cannot set the value of a CallbackParam.")
+
+    @property
+    def _has_value(self) -> bool:
+        """A CallbackParam's value comes from its callback, not from _value."""
+        return self.value is not None
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -602,16 +602,32 @@ class Leaf(expression.Expression):
     @property
     def value(self) -> expression.ExpressionValue | None:
         """The numeric value of the expression."""
-        if self.sparse_idx is None:
-            return self._value
-        else:
+        if self.sparse_idx is not None:
             warn('Reading from a sparse CVXPY expression via `.value` is discouraged.'
                   ' Use `.value_sparse` instead', RuntimeWarning)
-            if self._value is None:
-                return None
-            val = np.zeros(self.shape, dtype=self._value.dtype)
-            val[self.sparse_idx] = self._value
-            return val
+        return self._value_dense()
+
+    def _value_dense(self) -> expression.ExpressionValue | None:
+        """The value of the leaf, densified when sparse, without warning.
+
+        `value` steers users towards `value_sparse` for sparse leaves. CVXPY's own
+        internals need the same data without emitting that warning at the user.
+        """
+        if self.sparse_idx is None:
+            return self._value
+        if self._value is None:
+            return None
+        val = np.zeros(self.shape, dtype=self._value.dtype)
+        val[self.sparse_idx] = self._value
+        return val
+
+    def _value_impl(self) -> expression.ExpressionValue | None:
+        return self._value_dense()
+
+    @property
+    def _has_value(self) -> bool:
+        """Whether a value has been assigned, without reading it through `value`."""
+        return self._value is not None
 
     @value.setter
     def value(self, val) -> None:
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -1017,7 +1017,7 @@ class Problem(u.Canonical):
         if verbose:
             print(_HEADER)
         for parameter in self.parameters():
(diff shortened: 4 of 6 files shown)


PTNobel:
Running CI... Will hopefully get to reviewing before the weekend

PR #3475, opened 2026-08-07
Make support function lifting solver-independent and defer it to canonicalization time

nimish:
## Description

This follows up on #3268 by removing the SCS-specific assumption from support-function canonicalization. In doing so, it removes some back and forth transformations through full matrix and triangular forms.

This PR:

- Makes the conic lift lazy and not SCS-specific
- Uses SvecPSD where possible
- Handles per-solver scaling requirements
- Keeps the old scs specific helpers and implementations that were public for backwards compat

## Type of change

- [ ] Bug fix

## Contribution checklist

- [x] Add our license to new files. (No new files were added.)
- [x] Write unittests.
- [x] Run the benchmarks to make sure your change doesn’t introduce a regression -- created a separate microbenchmark for this (https://github.com/cvxpy/benchmarks/pull/35) that shows an improvement from ~170ms to 10ms. 

## Verification

- Full test suite: 2547 passed, 852 skipped.
- `ruff check cvxpy`: passed.
- Pyright on the changed Python files: passed.


cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
```diff
@@ -1,18 +1,43 @@
...
+    scaling, the dual cone needs a factor of 1/2 on off-diagonal entries.
+    """
+    if solver_context.psd_sqrt2_scaling:
+        return arg
+
+    if solver_context.psd_triangle_kind == TriangleKind.LOWER:
+        rows, cols = np.tril_indices(n)
+    else:
+        rows, cols = np.triu_indices(n)
+    order = np.argsort(np.ravel_multi_index((rows, cols), (n, n), order='F'))
```
PTNobel: What is the argsort for

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
```diff
@@ -1,18 +1,43 @@
...
+    weights = np.where((rows == cols)[order], 1.0, 0.5)
+    weights = np.tile(weights, arg.size // weights.size)
```
PTNobel: Is this for batch handling? I don't get the second argument

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
```diff
@@ -42,18 +67,25 @@ def suppfunc_canon(expr, args, solver_context: SolverInfo | None = None):
...
-        local_cons.append(curmat >> 0)
+    for psdsel, source_con in psdsels:
+        eta_block = eta[psdsel]
+        if isinstance(source_con, SvecPSD):
+            n = source_con.cone_sizes()[0]
+            dual_arg = _svec_psd_dual_arg(eta_block, n, solver_context)
+            local_cons.append(SvecPSD(dual_arg, n=n))
+        else:
+            eta_mat = reshape(eta_block, source_con.shape, order='F')
+            local_cons.append(
```
PTNobel: This is a really expensive addition to a problem; shouldn't symmetry be guaranteed in some cases?

cvxpy/tests/test_suppfunc.py
PTNobel, line 1: The complexity of this test seems way too high. Can you explain what's going on? What was missing in the old tests you're covering now?

cvxpy/transforms/suppfunc.py
```diff
@@ -47,6 +58,96 @@ def scs_coniclift(x, constraints):
...
+    x: cvxpy.Variable
+    constraints: list of cvxpy.constraints.constraint.Constraint
+        Each Constraint object must be DCP-compatible.
+
+    Notes
+    -----
+    This function DOES NOT work when ``x`` has attributes, like ``PSD=True``,
+    ``diag=True``, ``symmetric=True``, etc...
+    """
+    from cvxpy.atoms.affine.sum import sum
```
PTNobel: Inline imports are strictly prohibited in the project without a justification for their necessity

cvxpy/transforms/suppfunc.py
```diff
@@ -1,9 +1,19 @@
...
+from cvxpy.constraints.second_order import SOC
 from cvxpy.expressions.variable import Variable
 from cvxpy.reductions.cvx_attr2constr import CONVEX_ATTRIBUTES
+from cvxpy.utilities.solver_context import SolverInfo
+
+if TYPE_CHECKING:
+    from cvxpy.reductions.dcp2cone.cone_matrix_stuffing import ConeDims
 
 
 def scs_coniclift(x, constraints):
```
PTNobel, line 25: I thought we were trying to remove this function?

cvxpy/transforms/suppfunc.py
```diff
@@ -47,6 +58,96 @@ def scs_coniclift(x, constraints):
...
+    # ^ The objective value is only used to make sure that "x"
+    # participates in the problem. So, if constraints is an
+    # empty list, then the support function is the standard
+    # support function for R^n.
+    problem_form = ProblemForm(prob)
+    cones = problem_form.cones(quad_obj=False).copy()
+    _, exact_targets, approx_targets = expand_cones(
+        cones, solver_context.solver_supported_constraints)
+
+    # This deliberately mirrors the parameter-free DCP portion of the normal
```
PTNobel: Is there no way to use the existing solving chain?

cvxpy/transforms/suppfunc.py
```diff
@@ -47,6 +58,96 @@ def scs_coniclift(x, constraints):
...
+    # solving chain, but stops before solver-specific cone formatting. Bounds
+    # must be explicit because they are part of the set being dualized.
+    reductions = []
+    if complex2real.accepts(prob):
+        reductions.append(complex2real.Complex2Real())
+    reductions.extend([
+        Dcp2Cone(quad_obj=False, solver_context=solver_context),
+        CvxAttr2Constr(reduce_bounds=True),
```
PTNobel: Wouldn't this make attributes work? I don't understand

cvxpy/transforms/suppfunc.py
```diff
@@ -47,6 +58,96 @@ def scs_coniclift(x, constraints):
...
+    x_indices = np.arange(x_offset, x_offset + x.size)
+    x_selector = np.zeros(shape=(A.shape[1],), dtype=bool)
+    x_selector[x_indices] = True
+    A_x = A[:, x_selector]
+    A_other = A[:, ~x_selector]
+    A = sparse.hstack([A_x, A_other])
+    return A, b, _cone_selectors(cone_prog.cone_dims, cone_prog.constraints)
+
+
 def scs_cone_selectors(K):
```
PTNobel, line 137: I thought we were trying to get rid of this

cvxpy/transforms/suppfunc.py
```diff
@@ -94,6 +195,67 @@ def scs_cone_selectors(K):
     return selectors
 
 
+def _cone_selectors(
+    K: "ConeDims", constraints: list[Constraint],
+) -> dict[str, np.ndarray | list]:
+    """
+    Parse a ConeDims object from an unformatted ParamConeProg.
```
PTNobel: What is an "unformatted" ParamConeProg? I'm unfamiliar with the term

cvxpy/transforms/suppfunc.py
```diff
@@ -94,6 +195,67 @@ def scs_cone_selectors(K):
...
+
+    Returns
+    -------
+    selectors : dict
+        Keyed by strings, which specify cone types. Values are numpy
+        arrays, or lists of numpy arrays. The numpy arrays give row indices
+        of the affine operator (A, b) stored by the ParamConeProg.
+    """
+    if K.p3d or K.pnd:
+        msg = "SuppFunc doesn't yet support feasible sets represented \n"
```
PTNobel, line 211: This is a much more problematic restriction now that cp.power uses power cones

PTNobel: At least we should file an issue and add it to the wishlist

cvxpy/transforms/suppfunc.py
```diff
@@ -182,17 +343,35 @@ def __call__(self, y) -> SuppFuncAtom:
         sigma_at_y = SuppFuncAtom(y, self)
         return sigma_at_y
 
-    def _compute_conic_repr_of_set(self) -> None:
+    def _conic_repr_of_set(
+        self, solver_context: SolverInfo,
+    ) -> tuple[
+        sparse.sparray | sparse.spmatrix,
```
PTNobel: spmatrix should be impossible. Please confirm

cvxpy/transforms/suppfunc.py
```diff
@@ -182,17 +343,35 @@ def __call__(self, y) -> SuppFuncAtom:
...
+        np.ndarray,
+        dict[str, np.ndarray | list],
+    ]:
+        # Proper cross-chain caching is deferred for now; this only caches
+        # canonicalizations that share the same SolverInfo instance.
```
PTNobel: That's correct

cvxpy/transforms/suppfunc.py
PTNobel, line 1: This is a lot of code duplication. Much more than I'm comfortable with. Any fixes possible?

PTNobel (commented):
Awesome first cut! Looking forward to the next round.

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
nimish: Yes it is, I'll make that more explicit (it can just be num_cones()).

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
nimish: Yep, let me check and avoid adding it if unnecessary.

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
nimish: The indices from tril/triu are row-major and not column major; let me see if I can simplify this logic and remove it since we just care about on/off-diagonal and they should just be the same.

cvxpy/transforms/suppfunc.py
nimish: Yep, fixing.

cvxpy/transforms/suppfunc.py
nimish: Thanks for the review, going through these. In general: I left in the scs specific methods since they are public and to remove them would be technically altering the public API. If we aren't concerned about keeping that around, then I'm happy to eliminate them.

If we're ok with refactoring a bit of the existing solving chain, that'd reduce the next set of duplication (and remove the need for inline imports to avoid circular imports). We would need to start with just solverinfo and stop after matrix stuffing, but that could be extracted as a shared helper.

I'll clean up the testsuite as well. The old tests didn’t exercise the new solver-dependent branches: full versus triangular PSD storage, scaled versus unscaled svec, vectorized cone row ordering, or moving the lift from construction to canonicalization. But I think it could be simplified a bit.

cvxpy/transforms/suppfunc.py
nimish: Clarified this, it just meant the representation before ConicSolver.format_constraints
rearranged it.

cvxpy/transforms/suppfunc.py
PTNobel: Ugh. I had missed that it was part of our public API. Let me look at it some more... I actually think we may have broken the API when we introduced SvecPSD, so maybe it's fine. 

I'm open to solver chain refactoring. I wrote the current one so I hopefully still remember it.

That said, feel free to post a plan here before you make edits so I can decide if I'm comfortable with the direction...

Thank you re tests!

rileyjmurray: I think it's fair to say that `SuppFunc` is the only public-API function form this file. I'd be low-key shocked if someone was using helper functions from this file in "production." I think we can mark this file's helper functions with a deprecation warning saying that they'll be removed in cvxpy 1.10.

nimish: The refactor I had in mind would be pretty small:

* Extract just the common DCP-to-cone part of the solving chain into a private helper that'd build the reductions through ConeMatrixStuffing. Things that differ between callers would need to remain as separate args (quad_obj, native bounds, the canonicalization backend, SOC-dim-3 handling).
* The normal solving chain would call that helper and then append the solver interface. SuppFunc would call it with quad_obj=False and bounds lowered to constraints, then use the resulting ParamConeProg directly.
* Once both paths use the helper, remove the duplicated reduction assembly and inline imports from suppfunc.py.


Pushed up a number of small fixes to the other comments, including deprecationwarnings to the scs functions.

cvxpy/transforms/suppfunc.py
nimish: Yeah it's not obvious. CvxAttr2Constr handles attributes on auxiliary variables, but `x` is a special case because we pull its columns back out after stuffing. If `x` has an attribute, that reduction can change its variable ID or store it using fewer coordinates, while the code below still assumes the original `x.id` and `x.size`. I tried removing the guard and those cases still fail. I’ll clarify the comment.

cvxpy/reductions/dcp2cone/canonicalizers/suppfunc_canon.py
nimish: Added a symmetry check

cvxpy/reductions/solvers/conic_solvers/scs_conif.py
```diff
@@ -62,6 +62,10 @@ def scs_psdvec_to_psdmat(vec: Expression, indices: np.ndarray) -> Expression:
     cannot be used, because this function builds a cvxpy Expression,
     rather than a numpy ndarray.
     """
+    warn(
+        "scs_psdvec_to_psdmat is deprecated and will be removed in CVXPY 1.10.",
```
PTNobel: Let's make this 1.11, I don't think we'll release this PR till 1.10.

PTNobel: Also I don't think we publicly expose this function at the root namespace? I think the only thing that needs deprecating is the public methods on supp func?

Fine to leave this tho... At the least it should be a comment here

cvxpy/tests/test_suppfunc.py
```diff
@@ -167,23 +169,37 @@ def test_psd_native_solver_context(self) -> None:
...
+            cp.Minimize(symmetric_epigraph), symmetric_constraints)
+        symmetric_prob.solve(solver=cp.CLARABEL)
+        self.assertAlmostEqual(symmetric_prob.value, 1.0, places=6)
+
     def test_unscaled_svec_uses_dual_weights(self) -> None:
-        X = cp.Variable((3, 3))
-        sigma = cp.suppfunc(X, [X >> 0, cp.trace(X) <= 1])
-        expr = sigma(np.eye(3))
+        x = cp.Variable(12)
+        sigma = cp.suppfunc(x, [SvecPSD(x, n=3)])
```
PTNobel, line 190: Is svecpsd publicly exposed?

cvxpy/transforms/suppfunc.py
PTNobel: Could we just make a dummy solver for SupFunc and make that the canonicalization target? If you have to get matrix stuffing, why not?

Would that be cleaner?

cvxpy/transforms/suppfunc.py
PTNobel: Ah. Of course. Thank you.

cvxpy/transforms/suppfunc.py
PTNobel: No need to say anything about deferring.

PTNobel (commented):
Awesome progress

cvxpy/tests/test_suppfunc.py
nimish: Yes, through `cvxpy.constraints`, though this test was specifically to test the internal behavior.

cvxpy/transforms/suppfunc.py
nimish: So I tried both -- the shared helper is simpler. The dummy solver ends up needing fixup to handle derived aux bounds (since it claims BOUNDED_VARIABLES=False) which makes it more complicated in the end, IMO

nimish: @PTNobel ended up pushing up the shared helper. Also cleaned up the local imports -- most of were lifted to module level, but a few were left (like in scs_coniclift) that otherwise would cause a nasty circular import.

cvxpy/transforms/suppfunc.py
nimish: @PTNobel I ended up just implementing this in another commit (not pushed yet). It's not a huge change overall since most of the work is done by the Power Cone classes already. Should I just roll it into this PR?

cvxpy/reductions/solvers/conic_solvers/scs_conif.py
nimish: Done

cvxpy/tests/test_suppfunc.py
PTNobel: cvxpy.constraints.* are not part of the public API.

cvxpy/transforms/suppfunc.py
PTNobel: Thank you. Will look carefully.

PTNobel (commented):
Reviewed the changes and they all look great.

cvxpy/reductions/solvers/solving_chain.py
```diff
@@ -216,29 +217,47 @@ def _build_solving_chain(
...
+def _cone_matrix_stuffing_reductions(
+    problem_form: ProblemForm,
+    solver_context: SolverInfo,
+    *,
+    quad_obj: bool,
+    reduce_bounds: bool,
+    canon_backend: str | None = None,
+    soc_dim3_only: bool = False,
+) -> list[Reduction]:
+    """Build the solver-independent reductions through matrix stuffing."""
```
PTNobel: They're not solver-independent since all of solver_context, quad_obj, reduce_bounds, and soc_dim3_only are solver related.

cvxpy/tests/test_suppfunc.py
PTNobel: We can leave the test tho

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+            with self.assertRaisesRegex(SolverError, "mixed-integer set"):
+                prob.get_problem_data(solver=cp.CLARABEL)
+
+    def test_vectorized_soc_row_order(self) -> None:
+        x = cp.Variable(2)
+        X = cp.reshape(x, (1, 2), order='F')
+        sigma = cp.suppfunc(x, [cp.SOC(np.array([1.0, 2.0]), X)])
+        y = cp.Constant([2.0, 1.0])
+
+        epigraph, constraints = suppfunc_canon(
```
PTNobel: Why can't this be tested directly against sigma? Why are we pulling the canon out?

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+        prob = cp.Problem(cp.Minimize(epigraph), constraints)
+        prob.solve(solver=cp.CLARABEL)
+        self.assertAlmostEqual(prob.value, 4.0, places=6)
+
+    def test_vectorized_expcone_row_order(self) -> None:
+        x = cp.Variable(2)
+        sigma = cp.suppfunc(x, [cp.exp(x) <= np.exp([1.0, 2.0])])
+        y = cp.Constant([2.0, 1.0])
+
+        epigraph, constraints = suppfunc_canon(
```
PTNobel: ditto

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+            sigma(y), [y], _solver_context(CLARABEL))
+        prob = cp.Problem(cp.Minimize(epigraph), constraints)
+        prob.solve(solver=cp.CLARABEL)
+        self.assertAlmostEqual(prob.value, 4.0, places=6)
+
+    def test_construction_is_lazy(self) -> None:
+        x = cp.Variable(1)
+        with patch("cvxpy.transforms.suppfunc._coniclift") as lift:
```
PTNobel: We do not need a test that it is lazy. I'm not worried about us accidnetally merging a not lazy impl

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+            with patch("cvxpy.problems.problem.Problem.get_problem_data") as get_data:
+                cp.suppfunc(x, [x <= 1])
+        lift.assert_not_called()
+        get_data.assert_not_called()
+
+    def test_constraints_are_snapshotted(self) -> None:
```
PTNobel, line 251: ooh this is good. THanks

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+        constraints = [x <= 1]
+        sigma = cp.suppfunc(x, constraints)
+        constraints.append(x <= 0)
+        epigraph = cp.Variable()
+        prob = cp.Problem(cp.Minimize(epigraph), [sigma(1) <= epigraph])
+
+        prob.solve(solver=cp.CLARABEL)
+        self.assertAlmostEqual(epigraph.value, 1.0, places=6)
+
+    def test_svec_psd_canonicalization_is_cached(self) -> None:
```
PTNobel: Sorry what is this testing?

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+        X = cp.Variable((2, 2))
+        sigma = cp.suppfunc(X, [X >> 0, cp.trace(X) <= 1])
+        Y = cp.Variable((2, 2))
+        solver_context = _solver_context(SCS)
+
+        with patch("cvxpy.transforms.suppfunc._coniclift", wraps=_coniclift) as lift:
+            _, constraints = suppfunc_canon(sigma(Y), [Y], solver_context)
+            suppfunc_canon(sigma(Y), [Y], solver_context)
```
PTNobel: Why not just construct a problem and call get_problem_data?

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+        np.testing.assert_allclose(old_A.toarray(), A.toarray())
+        np.testing.assert_allclose(old_b, b)
+        self.assertEqual(old_selectors["nonneg"].size, 1)
+
+        vec = cp.Variable(3)
+        with pytest.warns(CvxpyDeprecationWarning, match="scs_psdvec_to_psdmat"):
+            psdmat = scs_psdvec_to_psdmat(vec, np.arange(3))
+        self.assertEqual(psdmat.shape, (2, 2))
+
+    def test_power_cone_remains_unsupported(self) -> None:
```
PTNobel: We don't need this.

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +148,177 @@ def test_psd_dualcone(self) -> None:
...
+        solver_context = _solver_context(SCS)
+
+        with patch("cvxpy.transforms.suppfunc._coniclift", wraps=_coniclift) as lift:
+            _, constraints = suppfunc_canon(sigma(Y), [Y], solver_context)
+            suppfunc_canon(sigma(Y), [Y], solver_context)
+
+        self.assertEqual(lift.call_count, 1)
+        self.assertEqual(sum(isinstance(con, SvecPSD) for con in constraints), 1)
+
+    def test_scs_compatibility_api(self) -> None:
```
PTNobel, line 275: I am fine with just a test that it warns or leaving it untested.

PTNobel (commented):
A few comments on the test file after a more thorough review of it. Once this is fixed should be good to merge!

Thanks for the great work Nimish.

cvxpy/reductions/solvers/solving_chain.py
nimish: Updated comment

cvxpy/tests/test_suppfunc.py
nimish: We can, changed it to just testing against sigma vs the specific function

cvxpy/tests/test_suppfunc.py
nimish: Removed

cvxpy/tests/test_suppfunc.py
nimish: Renamed, it just tests that we reuse coniclift whenever possible

cvxpy/tests/test_suppfunc.py
nimish: Done.

cvxpy/tests/test_suppfunc.py
nimish: Removed

cvxpy/tests/test_suppfunc.py
```diff
@@ -115,6 +146,150 @@ def test_psd_dualcone(self) -> None:
...
+
+    def test_auxiliary_variable_bound_attribute(self) -> None:
+        x = cp.Variable(1)
+        aux = cp.Variable(1, nonneg=True)
+        sigma = cp.suppfunc(x, [x == aux, aux <= 2])
+        solver_context = _solver_context(COPT)
+
+        for direction, expected in [(1.0, 2.0), (-1.0, 0.0)]:
+            y = cp.Constant([direction])
+            epigraph, constraints = suppfunc_canon(
```
PTNobel, line 210: Why are we hand canonicalizing here?

PTNobel (approved):
I missed one thing on my last review... I'm away from my laptop or I'd fix it myself.

cvxpy/tests/test_suppfunc.py
nimish: I wanted to test the solvers that have native variable bounds, so this was a bit of a hack to canonicalize with COPT and solve with Clarabel.

PTNobel: can't you just use an affine set and HiGHS?

nimish: Yep, made that change.

PTNobel:
Sorry I meant to merge this earlier, but I was waiting for CI to run. Please don't do any more merge commits unless there's a blocker, and just ping me instead

nimish:
@PTNobel checking in on this: anything further to do here? It's getting increasingly stale w.r.t. main branch

PTNobel:
Can you do another merge commit and ping me? I'll run CI once pinged and then merge tonight

nimish:
@PTNobel done

Issue #1712, opened 2022-03-17
DOC: user guide should include "migrating from Matlab and CVX to CVXPY"


rileyjmurray:
Migrating from Matlab to the scientific computing ecosystem in Python can be a painful process. CVXPY is a common Python entry-point for scientists and engineers who know Matlab. We should have the user guide provide pointers on the differences between CVX and CVXPY. We should also emphasize the crucial differences between numpy ndarrays, SciPy (sparse) matrices, cvxpy Expressions, and Matlab arrays. 

I provided some pointers along these lines here: https://github.com/cvxpy/cvxpy/discussions/1711#discussioncomment-2377676.

prasu-baran:
Hello @rileyjmurray !,

I am intrested in working in this updation and I’d like clarification on:
- Should this be documentation-only, or are minor UX improvements (e.g., clearer bmat error messages) acceptable?
- Is the 2D-only expression model and restricted broadcasting considered a fixed design principle?
- Where would you prefer this guidance to live in the docs?
Could you assist further ?
Thanks!

Issue #2618, opened 2024-11-09
Port examples to Marimo and add new section for n-dimensional expressions


Transurgeon:
Recently, in #2606 , @aryamanjeendgar has introduced a marimo notebook for the new quantum information atoms. 
Using WebAssembly it is possible to run marimo notebooks entirely on the web, see [here](https://docs.marimo.io/guides/wasm.html) for more details. In addition, @phschiele has also added cvxpy to [pyodide](https://pyodide.org/en/stable/usage/packages-in-pyodide.html) which makes it possible to have cvxpy examples in marimo notebooks entirely on the web!

It would be awesome to have an interactive page for the examples where the user would be able to run things directly instead of being redirected to a google colab link.

In addition, in the upcoming release we are starting to support n-dimensional expressions and it would be great to have concrete examples for its usage (ideas include power systems, networking, etc.).

langestefan:
I think this can be closed? 

https://www.cvxpy.org/examples/index.html#basic-examples

Quite a few marimo examples here, which run directly in the browser using WASM.

PR #3476, opened 2026-08-09
DOC: add user guide page for MATLAB and CVX users

nicoloangileri:
## Description

Adds a "CVXPY for MATLAB and CVX users" page to the user guide, as requested in #1712.

The page covers the differences that most often cause silent errors when migrating: the four array types in play (MATLAB matrix, NumPy ndarray, SciPy sparse, CVXPY `Expression`) and which to use for data vs unknowns; shape and broadcasting semantics; `@` vs `*` vs `cp.multiply`; the Fortran-order default of `reshape`/`vec`; 0-based indexing; constructs CVX accepts and CVXPY rejects; and how to read results back out, mapped from `cvx_optval`/`cvx_status`.

Two notes on accuracy:

- The linked comment in #1711 states that CVXPY expressions are limited to two dimensions. That is no longer the case (`cp.Variable((2, 3, 4))` works on 1.9.2), so the page says so explicitly rather than repeating it.
- `reshape`/`vec` currently default to Fortran order, matching MATLAB but not NumPy, and the default is scheduled to change to `'C'`. The page flags this in a warning admonition, since MATLAB users' intuition is correct today and will silently stop being correct.

Every code snippet was executed against CVXPY 1.9.2 and the outputs in the page are real. `sphinx -b html` builds the page with no new warnings.

Issue link: #1712

## Type of change
- [ ] Bug fix

- [x] Add our license to new files. — N/A, no new `.py` files
- [x] Write unittests. — N/A, documentation only
- [x] Run the unittests and check that they're passing. — N/A, documentation only
- [x] Run the benchmarks to make sure your change doesn't introduce a regression. — N/A, no code changes

### DIFF
--- a/doc/source/tutorial/matlab/index.rst
+++ b/doc/source/tutorial/matlab/index.rst
@@ -1,0 +1,239 @@
+.. _matlab:
+
+CVXPY for MATLAB and CVX users
+==============================
+
+Many CVXPY users arrive from MATLAB, and most of them arrive from `CVX
+<https://cvxr.com/cvx/>`_. The modeling ideas carry over almost unchanged: you
+declare variables, build an objective and a list of constraints, and hand the
+problem to a solver. What does not carry over is the array semantics
+underneath. In MATLAB every array is a matrix, and a plain list of numbers is a
+row or column vector. In Python it is not, and CVXPY inherits NumPy's rules.
+
+This page collects the differences that cause the most trouble. It assumes you
+have read :ref:`the introductory tutorial <tutorial>`.
+
+A problem side by side
+----------------------
+
+A bounded least-squares problem in CVX:
+
+.. code:: matlab
+
+    cvx_begin
+        variable x(n)
+        minimize( sum_square(A*x - b) )
+        subject to
+            0 <= x <= 1
+    cvx_end
+
+and in CVXPY:
+
+.. code:: python
+
+    import cvxpy as cp
+
+    x = cp.Variable(n)
+    objective = cp.Minimize(cp.sum_squares(A @ x - b))
+    constraints = [0 <= x, x <= 1]
+    prob = cp.Problem(objective, constraints)
+    prob.solve()
+
+Four differences are already visible:
+
+* CVXPY has no ``cvx_begin``/``cvx_end`` block. Expressions are ordinary Python
+  objects that you can build, store, and pass around.
+* Constraints are a Python list, not statements in a block.
+* ``0 <= x <= 1`` is **not** valid in CVXPY. Python evaluates chained
+  comparisons in a way CVXPY cannot intercept, so you must write the two
+  constraints separately. CVXPY raises an exception if you try.
+* Matrix multiplication is ``@``, not ``*``. See
+  :ref:`multiplication <matlab-multiplication>` below.
+
+Four kinds of array
+-------------------
+
+Code that mixes MATLAB habits with CVXPY tends to confuse four distinct types:
+
+``numpy.ndarray``
+    The default numeric array in Python. Unlike a MATLAB matrix it can have any
+    number of axes, including one. A NumPy array of shape ``(n,)`` is
+    one-dimensional: it is neither a row nor a column vector.
+
+SciPy sparse arrays
+    Usable anywhere a dense constant is, including as the left operand of ``@``.
+    CVXPY keeps them sparse internally, so prefer them for large structured data.
+
+CVXPY ``Expression``
+    A symbolic object: a :py:class:`~cvxpy.expressions.variable.Variable`, a
+    :py:class:`~cvxpy.expressions.constants.parameter.Parameter`, a
+    :py:class:`~cvxpy.expressions.constants.constant.Constant`, or anything built
+    from them. Expressions have a ``.shape``, but no numeric value until the
+    problem is solved.
+
+MATLAB matrix
+    Always two-dimensional, always column-major. Neither property holds in
+    Python.
+
+The rule of thumb: use NumPy arrays and SciPy sparse arrays for **data**, and
+CVXPY expressions for **unknowns**. Mixing them in one expression is fine and
+expected — ``A @ x + b`` with ``A`` sparse, ``b`` dense and ``x`` a variable
+works.
+
+.. _matlab-shapes:
+
+Shapes and broadcasting
+-----------------------
+
+This is the single most common source of silent errors for MATLAB users.
+
+In MATLAB, ``zeros(4,1)`` and a "vector of length 4" are the same thing. In
+NumPy — and therefore in CVXPY — shapes ``(4,)``, ``(4, 1)`` and ``(1, 4)`` are
+three different objects, and arithmetic between mismatched shapes does not fail:
+it *broadcasts*.
+
+.. code:: python
+
+    import cvxpy as cp
+    import numpy as np
+
+    A = np.random.randn(4, 5)
+    x = cp.Variable(5)
+
+    (A @ x - np.zeros(4)).shape       # (4,)     — as intended
+    (A @ x - np.zeros((4, 1))).shape  # (4, 4)   — a matrix, silently
+
+The second line produces no error and no warning. The residual has become a
+4-by-4 matrix, and any objective built on it optimizes something you did not
+write:
+
+.. code:: python
+
+    b = np.arange(4.)
+
+    correct = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b)))
+    wrong   = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b.reshape(-1, 1))))
+
+    correct.solve()   # 0.0
+    wrong.solve()     # 20.0
+
+Both problems are scalar-valued, both solve to optimality, and only one of them
+is the least-squares problem. The only difference is the shape of ``b``.
+
+The habit that prevents this: **do not create explicit row or column vectors in
+Python unless something specifically requires one.** Write ``cp.Variable(n)``
+rather than ``cp.Variable((n, 1))``, and keep data one-dimensional with
+``b.ravel()`` when it arrives from a source that produced a column.
+
+When you do need a block structure, build it explicitly with
+:py:func:`~cvxpy.bmat` rather than relying on broadcasting to assemble it.
+
+.. _matlab-multiplication:
+
+Multiplication
+--------------
+
+MATLAB uses ``*`` for matrix multiplication and ``.*`` for elementwise. Python
+splits this three ways:
+
+============================  ==========================================
+Operation                     CVXPY
+============================  ==========================================
+matrix-matrix, matrix-vector  ``A @ x``
+scalar times anything         ``2 * x``
+elementwise                   ``cp.multiply(a, x)``
+============================  ==========================================
+
+Using ``*`` for matrix multiplication has been deprecated since CVXPY 1.1 and
+emits a warning. It is worth fixing rather than silencing: on operands that
+happen to have equal shapes, the intent is genuinely ambiguous.
+
+Reshaping: CVXPY follows MATLAB, NumPy does not
+-----------------------------------------------
+
+MATLAB is column-major: ``reshape`` fills the first column, then the second.
+NumPy is row-major by default. :py:func:`~cvxpy.reshape` and
+:py:func:`~cvxpy.vec` currently default to Fortran (column-major) order, which
+matches MATLAB and *differs* from NumPy on the same data:
+
+.. code:: python
+
+    M = np.arange(6).reshape(2, 3)
+
+    M.reshape(3, 2)                             # [[0, 1], [2, 3], [4, 5]]
+    cp.reshape(cp.Constant(M), (3, 2), order='C').value  # [[0, 1], [2, 3], [4, 5]]
+    cp.reshape(cp.Constant(M), (3, 2), order='F').value  # [[0, 4], [3, 2], [1, 5]]
+
+.. warning::
+
+    CVXPY's default order will change from ``'F'`` to ``'C'`` in a future
+    version, to match NumPy. Omitting ``order`` currently raises a
+    ``FutureWarning``. **Always pass** ``order`` **explicitly** — code that
+    relies on the current default will change meaning silently when it flips.
+
+Indexing
+--------
+
+Python indices start at ``0``, and slices exclude their endpoint, so MATLAB's
+``x(1:3)`` is ``x[0:3]`` — same three entries, different notation. Transposition
+is ``A.T``, not ``A'``.
+
+.. code:: python
+
+    M = cp.Variable((3, 4))
+
+    M[0, 0].shape    # ()      scalar
+    M[0].shape       # (4,)    first row, as a 1-D expression
+    M[:, 0].shape    # (3,)    first column, also 1-D
+    M[0:2].shape     # (2, 4)  first two rows
+
+Note that indexing a single row or column drops the axis, exactly as in NumPy
+and unlike MATLAB, where the result stays two-dimensional.
+
+More than two dimensions
+------------------------
+
+Older material states that CVXPY expressions are limited to two dimensions.
+This is no longer true: ``cp.Variable((2, 3, 4))`` is valid and N-dimensional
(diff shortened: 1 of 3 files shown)


doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,239 @@
...
+* CVXPY has no ``cvx_begin``/``cvx_end`` block. Expressions are ordinary Python
+  objects that you can build, store, and pass around.
+* Constraints are a Python list, not statements in a block.
+* ``0 <= x <= 1`` is **not** valid in CVXPY. Python evaluates chained
+  comparisons in a way CVXPY cannot intercept, so you must write the two
+  constraints separately. CVXPY raises an exception if you try.
+* Matrix multiplication is ``@``, not ``*``. See
+  :ref:`multiplication <matlab-multiplication>` below.
+
+Four kinds of array
```
rileyjmurray: `arrays`, plural

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,239 @@
...
+    M.reshape(3, 2)                             # [[0, 1], [2, 3], [4, 5]]
+    cp.reshape(cp.Constant(M), (3, 2), order='C').value  # [[0, 1], [2, 3], [4, 5]]
+    cp.reshape(cp.Constant(M), (3, 2), order='F').value  # [[0, 4], [3, 2], [1, 5]]
+
+.. warning::
+
+    CVXPY's default order will change from ``'F'`` to ``'C'`` in a future
+    version, to match NumPy. Omitting ``order`` currently raises a
+    ``FutureWarning``. **Always pass** ``order`` **explicitly** — code that
+    relies on the current default will change meaning silently when it flips.
```
rileyjmurray: I suggest we make this fact more prominent. The fact that our behavior is about to change is more important than the fact that we currently agree with matlab's convention.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,239 @@
...
+Note that indexing a single row or column drops the axis, exactly as in NumPy
+and unlike MATLAB, where the result stays two-dimensional.
+
+More than two dimensions
+------------------------
+
+Older material states that CVXPY expressions are limited to two dimensions.
+This is no longer true: ``cp.Variable((2, 3, 4))`` is valid and N-dimensional
+expressions are supported. Support is newer than the rest of the library, so
+not every atom accepts N-D input; check the documentation of the specific atom.
```
rileyjmurray: Be more specific. Say when we added support for these variables. Ideally, provide a reference to a tracking issue that says what cvxpy functions are(n't) confirmed to work with N-D input.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Things CVXPY rejects that CVX accepts
+-------------------------------------
+
+* **Chained constraints.** ``0 <= x <= 1`` and ``x == y == 2`` raise an
+  exception. Write them as separate entries in the constraint list.
+* **Strict inequalities.** ``<`` and ``>`` are not constraints in CVXPY, since
+  they have no meaning for a numerical solver.
```
rileyjmurray: Wait, does CVX accept strict inequalities?

Separately, I know that our docs have said "strict inequalities have no meaning for a numerical solver" for many years, but that's not actually true. If we define optimization problems in finite-precision (IEEE 754) arithmetic then strict inequalities are perfectly well-defined. The reason why we reject strict inequalities is that it's ill-posed as a mathematical problem. Then there's a secondary issue that because numerical solvers don't generate strictly feasible solutions there's no operational difference between strict and non-strict inequalities, and raising an error is more likely to alert a user to that fact.

rileyjmurray (commented):
Thanks so much for this PR! I opened the issue several ago. It's great to see it resolved.

nicoloangileri:
Thanks for the fast review, and for opening the issue in the first place.

On strict inequalities: yes, CVX accepts them. Its docs say "Strict inequalities < and > are accepted as well, but they are interpreted identically to their nonstrict counterparts", adding that their use is strongly discouraged and may be removed. So the difference is real, and arguably sharper than I first wrote it — CVX silently reinterprets the model, CVXPY errors. I've cited that.

You're right about the rationale, and I've replaced it: ill-posedness (an open feasible set need not attain its optimum) as the primary reason, with the point about solvers not returning strictly feasible points as secondary.

On N-D: dated to 1.6 and linked to the existing n-dimensional section rather than restating it. I couldn't find a tracking issue listing per-atom N-D compatibility — the closest open one is #2618, which is about something else. Rather than invent a reference, the page now says plainly that no such list exists and points at the existing experimental warning. Happy to open a tracking issue for it as a follow-up if you'd like one.

The reshape section now leads with the change, using the 1.7/1.8 timeline from the release notes. Plural fixed.

PTNobel:
>  I couldn't find a tracking issue listing per-atom N-D compatibility 

I thought we had finished around N-d support to atoms? With the exception of like bmat or something 

I would definitely support an issue with a list of remaining incompatibilities.

nicoloangileri:
You're right, and my wording was too strong — I've softened it.

I went through the atoms with a 3-D variable on 1.9.2. Coverage is broad: elementwise atoms, axis reductions, indexing, `reshape` / `vec` / `permute_dims` / `swapaxes` / `moveaxis` / `broadcast_to` / `squeeze`, `sum_largest` / `sum_smallest`, `power`, `quad_over_lin`, `dotsort`, `einsum` and batched `@` ((2,3,4) @ (2,4,3) -> (2,3,3)) all accept N-D input. `bmat` does too, incidentally — `bmat([[X, X]])` on a (2,3,4) variable gives (2,6,4) — so it doesn't look like the exception.

The ones I could still make fail, where N-D seems conceptually meaningful:

- `tv` — "tv cannot have input arrays with more than 2 dimensions" (explicit guard; 3-D TV is a real use case for volumetric/video denoising)
- `cvar` — "cvar input must be a 1d array"
- `geo_mean` — "x must be a row or column vector"
- `kron` — "kron requires both arguments to be 2-d"

I excluded things that are 2-D by definition (`trace`, `diag`, `upper_tri`, `partial_trace`) and 1-D by definition (`conv`, `outer`).

Caveat: this is 1.9.2 from PyPI, not master, so you may already have moved past some of these. The page now names those four rather than claiming no list exists. Happy to open the tracking issue with this as the starting list if it's useful just say the word and I'll file it.

nicoloangileri:
Gentle ping on this one it's been a couple of weeks since the last exchange.

No changes needed as far as I know: the four points from @rileyjmurray's review are addressed, and the N-D claim was corrected after @PTNobel pushed back on it. Since then #3493 landed and made `geo_mean` reduce along any axis, so the sentence in the page about atoms that still refuse N-D input is now down to `tv` and `cvar`. Happy to update that line before this merges, or leave it and follow up once #3494 settles whichever you'd rather.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+
+SciPy sparse arrays
+    Usable anywhere a dense constant is, including as the left operand of ``@``.
+    CVXPY keeps them sparse internally, so prefer them for large structured data.
+
+CVXPY ``Expression``
+    A symbolic object: a :py:class:`~cvxpy.expressions.variable.Variable`, a
+    :py:class:`~cvxpy.expressions.constants.parameter.Parameter`, a
+    :py:class:`~cvxpy.expressions.constants.constant.Constant`, or anything built
+    from them. Expressions have a ``.shape``, but no numeric value until the
```
PTNobel: Constants and Parameters both do or can have numeric values before solve. Also if you're reusing variable names between problems, you can keep the solution. We should be more careful here with wording

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+
+    A = np.random.randn(4, 5)
+    x = cp.Variable(5)
+
+    (A @ x - np.zeros(4)).shape       # (4,)     — as intended
+    (A @ x - np.zeros((4, 1))).shape  # (4, 4)   — a matrix, silently
+
+The second line produces no error and no warning. The residual has become a
+4-by-4 matrix, and any objective built on it optimizes something you did not
+write:
```
PTNobel:
```suggestion
intend to write:
```

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+    b = np.arange(4.)
+
+    correct = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b)))
+    wrong   = cp.Problem(cp.Minimize(cp.sum_squares(A @ x - b.reshape(-1, 1))))
+
+    correct.solve()   # 0.0
+    wrong.solve()     # 20.0
+
+Both problems are scalar-valued, both solve to optimality, and only one of them
+is the least-squares problem. The only difference is the shape of ``b``.
```
PTNobel:
```suggestion
is the intended least-squares problem. The only difference is the shape of ``b``.
```

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+
+The habit that prevents this: **do not create explicit row or column vectors in
+Python unless something specifically requires one.** Write ``cp.Variable(n)``
+rather than ``cp.Variable((n, 1))``, and keep data one-dimensional with
+``b.ravel()`` when it arrives from a source that produced a column.
+
+When you do need a block structure, build it explicitly with
+:py:func:`~cvxpy.bmat` rather than relying on broadcasting to assemble it.
```
PTNobel: I think we merged cp.block that's more powerful/general recently?

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+happen to have equal shapes, the intent is genuinely ambiguous.
+
+Reshaping: always pass ``order`` explicitly
+-------------------------------------------
+
+.. warning::
+
+    The default order of :py:func:`~cvxpy.reshape`, :py:func:`~cvxpy.vec` and
+    :py:func:`~cvxpy.flatten` is changing. It is Fortran (``'F'``) today, and
+    omitting ``order`` raises a ``FutureWarning``. CVXPY 1.7 plans to raise an
```
PTNobel: This plan was changed. We'll continue raising a warning for a while.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+    error when ``order`` is not specified, and 1.8 will switch the default to
+    ``'C'`` to match NumPy. **Pass** ``order`` **explicitly**: code that relies
+    on the current default will change meaning when the switch happens.
+
+The reason this matters more to MATLAB users than to anyone else is that the
```
PTNobel: Can we drop this paragraph? It's not much of a value add tbh

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+current default happens to agree with MATLAB and disagree with NumPy — so your
+existing intuition is right today and will stop being right.
+
+MATLAB is column-major: ``reshape`` fills the first column, then the second.
+NumPy is row-major by default. The three libraries therefore disagree on the
```
PTNobel: Is this true? CVXPY agrees with one of them, whichever you specify.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+More than two dimensions
+------------------------
+
+Older material states that CVXPY expressions are limited to two dimensions.
+That has not been true since CVXPY 1.6, which introduced
+:ref:`N-dimensional expressions <n-dimensional>`, so ``cp.Variable((2, 3, 4))``
+is valid today.
+
+Coverage is broad: elementwise atoms, axis reductions, indexing, reshaping and
+batched ``@`` all accept N-D input. Support is still marked experimental and a
```
PTNobel: I do not think we should continue calling it experimental. If we still have that written somewhere other than release notes, we should amend it.

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+Things CVXPY rejects that CVX accepts
+-------------------------------------
+
+* **Chained constraints.** ``0 <= x <= 1`` and ``x == y == 2`` raise an
+  exception. Write them as separate entries in the constraint list.
+* **Strict inequalities.** CVX accepts ``<`` and ``>``, but `interprets them
+  identically to their non-strict counterparts
+  <https://cvxr.com/cvx/doc/basics.html>`_. CVXPY raises an exception instead.
+  The reason is not that strict inequalities are meaningless in finite
+  precision — in IEEE 754 arithmetic they are perfectly well defined — but that
```
PTNobel: Can we drop at least this parenthetical and maybe the whole opening to this sentence? Seems overly cute

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+
+CVX writes results into ``cvx_optval``, ``cvx_status`` and the variables
+themselves. CVXPY attaches them to the objects you created:
+
+=====================  ==================================
+CVX                    CVXPY
+=====================  ==================================
+``cvx_optval``         ``prob.value``, also returned by ``prob.solve()``
+``cvx_status``         ``prob.status``
+value of ``x``         ``x.value`` — a NumPy ndarray
```
PTNobel, line 235: Maybe we should mention value_sparse? Idk maybe that's too advanced for this

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+dual variable          ``constraint.dual_value``
+=====================  ==================================
+
+``x.value`` is ``None`` until the problem has been solved, and stays ``None`` if
+the problem is infeasible or unbounded. Its shape matches the variable's shape,
```
PTNobel, line 240: Or there is no optimal feasible point I think. Min exp x also stays none right?

doc/source/tutorial/matlab/index.rst
```diff
@@ -0,0 +1,257 @@
...
+so a ``cp.Variable(5)`` yields an array of shape ``(5,)`` — one-dimensional
+again, not a column.
+
+Where to go next
+----------------
+
+* :ref:`functions` — the atom library, with the CVX-equivalent names.
+* :ref:`dcp` — the ruleset for what counts as a valid convex expression. It is
+  the same discipline CVX enforces, and the error messages are more informative.
+* :ref:`advanced` — solver selection, warm starts and problem transformations.
```
PTNobel: Let's not talk about warm starts, their documentation isn't the most accurate

nicoloangileri:
All applied, except one where the code disagreed with both of us.

**On `Minimize(exp(x))`.** It doesn't leave `x.value` as `None` — it reports `optimal` with a finite point: -15.6 on SCS, -26.3 on CLARABEL. The infimum isn't attained, but the solver stops at its tolerance rather than chasing it, so from the user's side it looks like an ordinary solve. Only genuinely infeasible or unbounded problems leave the value unset, so that sentence is unchanged. Happy to be shown a case where it does behave the way you expected.

**On the reshape roadmap** — thank you, I'd have shipped something stale. The box now says the switch to `'C'` is planned for 2.0 and the warning stays until then.

**On "experimental"**: dropped here, and since you mentioned amending it wherever else it appears, `tutorial/advanced/index.rst` carried the same wording and now says that a few atoms don't accept N-D input yet. That's in this PR — say the word if you'd rather it were separate.

**On Constants and Parameters** — you're right, the paragraph was wrong as written. It now says a `Constant` always carries a value, a `Parameter` carries one once assigned, and a `Variable` gets one from a solve and keeps it.

`cp.block` replaces the `bmat` reference. Both wording suggestions applied. Dropped the MATLAB-intuition paragraph, the parenthetical in the strict-inequality item, and the warm-start reference.

I left `value_sparse` out — you weren't sure either, and for a page aimed at someone arriving from MATLAB it felt like one concept too many. Easy to add if you'd rather.

Docs build clean, no new warnings.

PR #3481, opened 2026-08-11
Modify type annotation for constraints argument of Problem constructor.

### LINKED ISSUE #1785: Prefer Sequence typing over List
For example the Problem class is typed:

```
class Problem(u.Canonical):
    def __init__(
        self,
        objective: Union[Minimize, Maximize],
        constraints: Optional[List[Constraint]] = None) -> None:
```

Would this be better typed like so?

```
class Problem(u.Canonical):
    def __init__(
        self,
        objective: Union[Minimize, Maximize],
        constraints: Optional[Sequence[Constraint]] = None) -> None:
```

This way passing a tuple constraint will not result in type errors.

carlosgmartin:
## Description

Fixes https://github.com/cvxpy/cvxpy/issues/1785.

Modifies the type annotation for the [`constraints` argument of the `Problem` constructor](https://www.cvxpy.org/api_reference/cvxpy.problems.html#cvxpy.Problem.__init__.constraints). The current type, `list`, causes problems for type checkers.

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/problems/problem.py
+++ b/cvxpy/problems/problem.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import time
 from collections import namedtuple
+from collections.abc import Iterable
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


Transurgeon (approved):
nice, thanks!

Issue #849, opened 2019-10-26
Use values of variables as initial guess (in mixed integer proplems)


samufi:
An important feature when solving mixed integer problems is to be able to suggest a good feasible solution. Some solvers support this feature, e.g. MOSEK (see [https://docs.mosek.com/9.0/pythonapi/tutorial-mio-shared.html#specifying-an-initial-solution](url)).

I suggest using the `value` property of variables for that purpose. That is, if a `value` exists, it should be used as initial guess provided the `warm_start` argument is set to `True`. This would require a minor adjustment when the stacked problem is created: the original variable values must be handed over to the newly created problem. Handing over these values to the respective solvers must be done according to the respective solver interfaces.

According to my knownledge this feature does not exist in CVXPY, and the only alternative would be to use a different software instead. This would be unfortunate, as the required change is small.

I have implemented my suggested solution for MOSEK and tested it on a problem of interest to me. Affected files are `matrix_stuffing.py` (I have added a helper function to stack the values), `cone_matrix_stuffing.py` (the stacked value is handed over to the new variable), and `mosek_conif.py` (the value is saved in the `data` dictionary and then handed over to the solver). 

The files are attached (as `.txt`, since upload of `.py` files was not supported). Changes are highlighted with a `CHANGE` key word.

I would be happy to create a pull request, but I am not sure that my solution is in line with the envisioned development path of CVXPY. The implemented solution affects MOSEK only, all other solvers will behave as usual. If this inconsistency is unwanted, I would like to encourage this feature to be implemented on a wider scale. Nonetheless, some people may be helped by my solution already. 

Please let me know if I can be of any further help.

[cone_matrix_stuffing.txt](https://github.com/cvxgrp/cvxpy/files/3775304/cone_matrix_stuffing.txt)
[matrix_stuffing.txt](https://github.com/cvxgrp/cvxpy/files/3775305/matrix_stuffing.txt)
[mosek_conif.txt](https://github.com/cvxgrp/cvxpy/files/3775306/mosek_conif.txt)

SteveDiamond:
Thanks for the implementation! The proposed change is in line with future development plans for cvxpy. It's implemented in some of the cvxpy extensions for nonconvex problems, such as dccp.

samufi:
I am aware that this has been implemented in extensions (though I am not sure whether also for integer programs - I should have looked into that). Maybe it would make then more sense to copy their implementation rather than reimplementing the feature from scratch. 

[...]

PR #3483, opened 2026-08-14
Add MIP start support to the SCIP interface

nicoloangileri:
## Description

`SCIP.solve_via_data` accepts a `warm_start` flag and ignores it, so there is currently no way to hand SCIP an incumbent from Python. GUROBI already supports this through `data['init_value']`; this brings SCIP in line.

`apply()` collects `Variable.value` through `utilities.stack_vals` — the same helper GUROBI uses — and `solve_via_data` registers the values before the constraints are added.

Two things differ from the GUROBI path, both forced by SCIP's API:

- The start is registered with `createPartialSol()` rather than `createSol()`. Users typically set values on a subset of their variables, and SCIP's `completesol` heuristic is built to fill in the rest. GUROBI uses `GRB.UNDEFINED` as its "not provided" sentinel; SCIP has no equivalent, so NaN is what lets the interface tell an unset entry apart from a genuine zero.
- `addSol()` rather than `trySol()`. At the point where CVXPY builds the model SCIP is still in the problem creation stage, and `trySol` raises `method cannot be called at this time in solution process` there.

**Verification that the values reach the solver**, rather than being silently discarded — from SCIP's own statistics output:

    without a start:  completesol : 0  0  0
    with a start:     completesol : 1  1  1
                      Primal Bound: ... found by <completesol>

The heuristic is never called without a start, is called once with one, and the resulting incumbent is attributed to it.

**Caveat, stated plainly:** I could not measure a speedup on the instances I tried — SCIP's own heuristics found better incumbents than the greedy ones I supplied. This exposes a capability that is currently not expressible at all; it is not a performance improvement by default.

**Compatibility:** behaviour changes only when `warm_start=True` and variable values are set, which previously did nothing for SCIP. The optimum is unaffected (covered by tests); the search path may differ.

Scope note: this covers the `Variable.value` path from #849. Re-using a previous solve via `solver_cache`, which GUROBI also implements, is not addressed here — the SCIP interface does not populate `solver_cache` at all.

Partial fix for #849

## Type of change
- [ ] Bug fix

- [x] Write unittests. — four, in `TestSCIP`: `init_value` contents, that a start does not change the optimum, `warm_start` with no values set, and an unusable start
- [x] Run the unittests and check that they're passing. — 45/45 SCIP, 142/142 in `test_conic_solvers.py`, no regressions
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally: the benchmark suite exercises the default solvers and does not touch the SCIP interface. The CI benchmark job will cover it.

### DIFF
--- a/cvxpy/tests/test_conic_solvers.py
+++ b/cvxpy/tests/test_conic_solvers.py
@@ -2606,6 +2606,43 @@ class TestSCIP(unittest.TestCase):
         assert np.allclose(data['init_value'][:2], [2.0, 3.0])
         assert np.isnan(data['init_value'][2])
 
+    @staticmethod
+    def completesol_calls(prob) -> int:
+        """How many times SCIP ran the heuristic that consumes a partial start.
+
+        Read from SCIP's own statistics: the ``completesol`` row of the primal
+        heuristics table, whose third numeric column is the call count.
+        """
+        model = prob.solver_stats.extra_stats["model"]
+        path = tempfile.mktemp(suffix=".txt")
+        try:
+            model.writeStatistics(path)
+            with open(path) as stats:
+                for line in stats:
+                    if line.strip().startswith("completesol"):
+                        return int(line.split(":", 1)[1].split()[2])
+        finally:
+            if os.path.exists(path):
+                os.unlink(path)
+        return 0
+
+    def test_scip_mip_start_reaches_solver(self) -> None:
+        """The start is handed to SCIP, not silently dropped.
+
+        Both solves pass ``warm_start=True``; the only difference is whether
+        the user set any variable values. SCIP's ``completesol`` heuristic
+        exists to complete a partial start, so it runs in the second case
+        and not in the first.
+        """
+        prob, x, y = self.get_mip_problem()
+        prob.solve(solver=cp.SCIP, warm_start=True)
+        assert self.completesol_calls(prob) == 0
+
+        prob, x, y = self.get_mip_problem()
+        x.value = np.array([2.0, 3.0])
+        prob.solve(solver=cp.SCIP, warm_start=True)
+        assert self.completesol_calls(prob) > 0
+
     def test_scip_mip_start(self) -> None:
         """A MIP start does not change the optimum."""
         prob, x, y = self.get_mip_problem()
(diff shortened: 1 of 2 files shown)


cvxpy/tests/test_conic_solvers.py
```diff
@@ -2584,6 +2584,49 @@ def get_simple_problem(self):
...
+        assert np.allclose(data['init_value'][:2], [2.0, 3.0])
+        assert np.isnan(data['init_value'][2])
+
+    def test_scip_mip_start(self) -> None:
+        """A MIP start does not change the optimum."""
+        prob, x, y = self.get_mip_problem()
+        expected = prob.solve(solver=cp.SCIP, warm_start=False)
+
+        x.value = np.array([2.0, 3.0])
+        assert prob.solve(solver=cp.SCIP, warm_start=True) == expected
```
Transurgeon, line 2652: maybe you could test that the number of iterations decreases? I am not sure if there is an easy way to get the number of iterations for SCIP though.. it differs from solver to solver.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2584,6 +2584,49 @@ def get_simple_problem(self):
...
+        prob, x, y = self.get_mip_problem()
+        x.value = None
+        y.value = None
+        assert prob.solve(solver=cp.SCIP, warm_start=True) == 9.0
+
+    def test_scip_mip_start_infeasible_values(self) -> None:
+        """An unusable start is discarded by SCIP, not propagated."""
+        prob, x, y = self.get_mip_problem()
+        x.value = np.array([99.0, 99.0])
+        assert prob.solve(solver=cp.SCIP, warm_start=True) == 9.0
```
Transurgeon, line 2666: this test is not really showing that the warm-start is being used or disregarded. Could we maybe try to show a log output? I guess this is not trivial, but could be worth giving it a try?

nicoloangileri:
I tried this and I don't think it can carry a test. `getNLPIterations` is available — CVXPY already stores it as `NUM_ITERS` — but on a problem small enough for the suite the count is identical with and without a start, and on larger instances it moves in both directions: on a 250-variable multi-knapsack I measured 288 nodes without a start vs 271 with one under a 1s limit, and 1126 vs 1171 under 3s. SCIP's own heuristics usually find a better incumbent than the one supplied, so any threshold would be flaky.

I went with the `completesol` call count instead — see the other thread.

nicoloangileri:
Agreed, and it turned out to be doable without parsing logs. The SCIP model is reachable from Python via `prob.solver_stats.extra_stats["model"]`, so its statistics can be read back with `writeStatistics`. The `completesol` heuristic exists specifically to complete a partial start, so its call count separates the two cases:

    without values set:  completesol : 0  0  0
    with values set:     completesol : 1  1  1

`test_scip_mip_start_reaches_solver` asserts exactly that. Both solves pass `warm_start=True` and differ only in whether any variable values were set, so the counter isolates the start itself rather than anything else in the pipeline.

I added the same assertion to the two robustness tests, which also pins down what happens to an unusable start: SCIP is offered it, tries to complete it, and the result is unaffected.

Transurgeon (approved):
LGTM!

PR #3488, opened 2026-08-18
Vectorize Gurobi conic interface model building, drop pre-9.5 code paths

SteveDiamond:
## Description

The Gurobi conic interface built its model one API call at a time. This vectorizes construction using Gurobi's matrix API, which `gurobi_qpif.py` has used since #2541.

- Variable types are assigned by index instead of testing `i in data[s.BOOL_IDX]` per column, which was quadratic in the number of integer variables.
- All problem variables are created in one `addMVar`.
- All SOC cones are built together in a new `add_model_soc_constrs`: one `addMVar` for the auxiliary variables, one `addMConstr` for the coupling rows (`[A | I] @ [x; aux] == b`), and one `addQConstr` per cone. This closes the standing `TODO: add all SOC constrs at once!`. Gurobi has no bulk API for quadratic constraints, so the cones still need one `addQConstr` each.
- Warm starts and primal extraction read `MVar` attributes in bulk.

The model handed to Gurobi is unchanged — same auxiliary variables, same bounds, same names, same constraint ordering. Only the number of Python-level API calls differs.

### Dropping the pre-9.5 code paths

The last commit removes the three-way `addMConstr` / `addMConstrs` / scalar-loop fallback, finishing the cleanup #2541 did for the QP interface. **This drops Gurobi < 9.5 for LP, MILP, SOCP and MISOCP**, which work today through that fallback; QPs have already required 9.5+ since #2541. Gurobi 9.5 shipped in November 2021 and the fallbacks have never had CI coverage, since CI installs `gurobipy` from pip.

### Results

Model construction time, best of 3, Gurobi 13.0.2, with `optimize` patched to a no-op so this measures only the building:

| problem | before | after | speedup |
|---|---|---|---|
| SOCP: 5000 cones × dim 10 | 244.0 ms | 86.2 ms | 2.8× |
| SOCP: 1 cone × dim 50000 | 154.7 ms | 59.0 ms | 2.6× |
| MISOCP: 1000 booleans | 18.3 ms | 4.8 ms | 3.8× |
| MISOCP: 6000 booleans | 498.0 ms | 28.5 ms | 17.5× |
| LP: 50000 rows | 18.9 ms | 19.1 ms | 1.0× |

MISOCP construction is now linear rather than quadratic in the number of integer variables, so the gap widens with size. Pure LPs are unchanged — that path already used `addMConstr`. Solve time is unaffected.

(Measured under the pip `gurobipy` restricted license, which caps solving but not model building. These are build-only numbers, not end-to-end solve times.)

## Type of change
- [ ] Bug fix
- [x] Other (performance; the last commit also drops Gurobi pre-9.5 support — see above)

- [ ] Add our license to new files. *(no new files)*
- [x] Write unittests.


PTNobel:
Can you clean up the PR description CC seems appears to have generated the most verbose nonsense I've seen in a while lol

cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
```diff
@@ -183,80 +183,55 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
-                    obj=c[i],
-                    name="x_%d" % i,
-                    vtype=vtype,
-                    lb=lb[i],
-                    ub=ub[i])
-            )
+        bool_idx = data[s.BOOL_IDX]
+        int_idx = data[s.INT_IDX]
+        is_mip = bool(bool_idx or int_idx)
+        # Assign variable types by index. Testing ``i in bool_idx`` for every
```
PTNobel: I don't like discussing deleted implementations in comments

cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
```diff
@@ -278,13 +253,15 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
                 model.setParam("DualReductions", 0)
                 model.optimize()
             solution["value"] = model.ObjVal
-            solution["primal"] = np.array([v.X for v in variables])
+            # Only the n problem variables are inverted; the auxiliary SOC
```
PTNobel: "inverted"?

cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
```diff
@@ -308,102 +285,72 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
...
+
+        aux_lb = np.full(num_rows, -gp.GRB.INFINITY)
+        aux_lb[heads] = 0.0
+        aux_names = ["soc_x_%d" % i for i in range(soc_start, soc_end)]
+        for head in heads:
+            aux_names[head] = "soc_t_%d" % (soc_start + head)
+        aux_vars = model.addMVar(num_rows, obj=0.0, lb=aux_lb,
+                                 ub=gp.GRB.INFINITY, name=aux_names)
+
+        # aux == vec - mat @ x, written as [mat | I] @ [x; aux] == vec so that
```
PTNobel, line 327: Should we be dualizing

cvxpy/tests/test_conic_solvers.py
```diff
@@ -2196,6 +2197,97 @@ def test_gurobi_mi_socp_1(self) -> None:
     def test_gurobi_mi_socp_2(self) -> None:
         StandardTestSOCPs.test_mi_socp_2(solver='GUROBI')
 
+    def test_gurobi_multiple_socp_cones(self) -> None:
+        """Several SOC cones of differing size, solved together.
```
PTNobel: Should this just be a new standard SOCP test

cvxpy/reductions/solvers/conic_solvers/gurobi_conif.py
SteveDiamond: The formulation in this MR is the same as the preexisting formulation. Dualizing is out of scope.

cvxpy/tests/test_conic_solvers.py
SteveDiamond: sure

SteveDiamond: I made it a standard test but only applied to gurobi in this PR to limit scope.

PTNobel (approved):

PTNobel (approved):

PR #3490, opened 2026-08-19
Add MIP start support to the HiGHS interface

nicoloangileri:
## Description

HiGHS could already be warm started from a previous solve held in `solver_cache`, but there was no way to hand it an incumbent of one's own: `solve_via_data` ignored `Variable.value` entirely. GUROBI and, since #3483, SCIP both support this, so this brings HiGHS in line and follows the same shape.

`apply()` collects the values through `utilities.stack_vals` with NaN marking entries the user did not set, and `solve_via_data` passes the known ones through the sparse overload of `setSolution`, which takes indices rather than a full solution vector. The existing `solver_cache` path keeps priority, mirroring GUROBI.

**Verification that the values reach the solver.** HiGHS says so itself, and only when a start is supplied:

    Attempting to find feasible solution by solving LP for user-supplied values of discrete variables

`test_highs_mip_start_reaches_solver` asserts on that line, in the same style as the existing `test_highs_warm_start`. Both solves in it pass `warm_start=True` and differ only in whether any variable values were set.

**On the rejected-guess branch.** HiGHS returns `kError` for a guess it cannot use. That is warned about rather than raised: the guess is a hint, not part of the model, and a stale value left on a variable should not stop an otherwise valid problem from being solved. Worth flagging that this branch is *not* reachable through a normal solve — CVXPY already validates `Variable.value` against bounds and integrality on assignment, so an out-of-range or fractional value is refused before it can reach the solver. Rather than leave it untested, a solver stub covers it, along with a second stub asserting that only the indices the user set are passed.

**Scope.** This covers the `Variable.value` path from #849 for HiGHS, as #3483 did for SCIP. Note that @PTNobel's #3123 approaches the same problem architecturally, with a `WARM_STARTABLE` flag and propagation to auxiliary variables; this is the incremental per-solver route and does not conflict with going that way later.

Partial fix for #849

## Type of change
- [ ] Bug fix

- [x] Write unittests. — five, in `TestHIGHS`: `init_value` contents, that a start does not change the optimum, that HiGHS reports using it, the rejected-guess warning, and that only known indices are passed
- [x] Run the unittests and check that they're passing. — 148/148 in `test_conic_solvers.py`, no regressions
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally: the benchmark suite exercises the default solvers and does not touch the HiGHS interface. The CI benchmark job will cover it.

### DIFF
--- a/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
@@ -223,13 +223,22 @@ class HIGHS(ConicSolver):
         Values are read from ``Variable.value`` and may be set on only some of
         the variables, so they are handed over through the sparse overload of
         ``setSolution``, which takes the indices that are known rather than a
-        full solution vector.
+        full solution vector. Indices are left as they come out of
+        ``flatnonzero``: HiGHS reports an out-of-range one as ``kError``,
+        whereas narrowing them here would wrap silently.
+
+        Only mixed-integer problems are given a start. On a pure LP the
+        simplex takes the same number of iterations with or without one.
 
         An unusable guess is reported by HiGHS as ``kError``. That is warned
         about rather than raised: the initial guess is a hint, not part of the
         model, and a stale value left on a variable should not stop an
         otherwise valid problem from being solved.
         """
+        if not (data[s.BOOL_IDX] or data[s.INT_IDX]):
+            # On a pure LP the simplex takes the same number of iterations with
+            # or without a primal point, so there is nothing to hand over.
+            return
         init_value = data.get("init_value")
         if init_value is None:
             return
@@ -237,9 +246,7 @@ class HIGHS(ConicSolver):
         known = np.flatnonzero(~np.isnan(init_value))
         if known.size == 0:
             return
-        status = solver.setSolution(
-            known.size, known.astype(np.int32), init_value[known]
-        )
+        status = solver.setSolution(known.size, known, init_value[known])
         if status == hp.HighsStatus.kError:
             warnings.warn(
                 "HiGHS rejected the initial guess taken from Variable.value; "
--- a/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
+++ b/cvxpy/reductions/solvers/conic_solvers/highs_conif.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import warnings
 from re import compile
 
 import numpy as np
@@ -157,6 +158,9 @@ class HIGHS(ConicSolver):
         data[s.INT_IDX] = [int(t[0]) for t in variables.integer_idx]
         inv_data["is_mip"] = data[s.BOOL_IDX] or data[s.INT_IDX]
 
+        # Initial guess, with NaN marking entries the user did not set.
+        data["init_value"] = utilities.stack_vals(problem.variables, np.nan)
+
         return data, inv_data
 
     def invert(self, results, inverse_data):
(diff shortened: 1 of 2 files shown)


nicoloangileri:
Note on CI before anyone spends time on it: `test_optional_solvers` is red here on ubuntu-22.04 and macos-14, but it is red on master too — the same two jobs fail on ae0fbff0, and they failed identically on #3483 before it was approved and merged.

The step-level pattern is the same as well: ubuntu fails both `Run test_conic_solvers` and `Run test_qp_solvers`, macos fails only `Run test_conic_solvers`. The QP suite goes through `highs_qpif.py`, which this PR does not touch, so at least that half cannot be caused by these changes. The other 50 checks pass.

I can't read the job logs without write access, so if part of the failure does turn out to be the HiGHS tests added here, point me at it and I'll fix it. The one place I'd look first is `test_highs_mip_start_reaches_solver`, which asserts on a HiGHS log line and would be sensitive to a different highspy version.

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
```diff
@@ -213,6 +217,36 @@ def invert(self, results, inverse_data):
...
+        """
+        init_value = data.get("init_value")
+        if init_value is None:
+            return
+        init_value = np.asarray(init_value, dtype=float)
+        known = np.flatnonzero(~np.isnan(init_value))
+        if known.size == 0:
+            return
+        status = solver.setSolution(
+            known.size, known.astype(np.int32), init_value[known]
```
PTNobel: Why 32 bit? We've had overflows with that before

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
```diff
@@ -213,6 +217,36 @@ def invert(self, results, inverse_data):
...
+        )
+        if status == hp.HighsStatus.kError:
+            warnings.warn(
+                "HiGHS rejected the initial guess taken from Variable.value; "
```
PTNobel, line 252: When does this happen?

PTNobel (commented):
I haven't looked at the tests yet. Overall, I'm a little hesitant; I encountered a situation recently where HiGHS did way worse with a warm start.

@jahall could you or someone from your team weigh in on whether we should encourage/expose warm starting from some but not all primal values? Should we do this only for MIPs?

nicoloangileri:
On the design question, I went and measured rather than guess, and it pointed at "MIPs only".

On a pure LP, `setSolution` with a primal point is a no-op for HiGHS. On a 900x1200 LP the simplex took **419 iterations in every case** — no start, all-zero, all-one, random, and partial — with an identical objective. Smaller instances behaved the same way. So gating on `is_mip` costs nothing, and I've pushed that: LPs are now left alone entirely, with a stub test pinning it down.

That also makes me doubt this is what you ran into. The `solver_cache` path predates this PR and passes a full `HighsSolution` from a previous solve, primal and dual together; this one only passes sparse primal values, only for MIPs, and only when the user has explicitly set `Variable.value` *and* passed `warm_start=True`. Nothing changes by default. If the case you hit was an LP, it can't have come from this path — and if it was the cache path, that's worth a separate look.

Happy to wait for @jahall on whether partial primal values are worth exposing at all. If the answer is that they aren't, I'd rather close this than have it merged on my say-so.

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
nicoloangileri: It didn't, before your other comment — which is a fair thing to have asked. CVXPY validates `Variable.value` against bounds and integrality on assignment (`leaf.py:763`), so an out-of-range or fractional value is refused long before it reaches HiGHS, and I kept the branch only as a guard, covered by a stub rather than left untested.

Removing the `int32` cast changes that: an out-of-range index now reaches HiGHS and comes back as `kError`, so the branch has a real trigger. It warns rather than raises because the guess is a hint, not part of the model — a stale value on a variable shouldn't stop an otherwise valid solve.

cvxpy/reductions/solvers/conic_solvers/highs_conif.py
nicoloangileri: Good catch, and the cast is gone. `HighsInt` is 32-bit in this build — `kHighsIInf == 2**31 - 1` — so the width was HiGHS's rather than a choice on my side, but narrowing it in Python was the unsafe half: numpy wraps silently, so index 2**31 + 5 became -2147483643. `flatnonzero` already returns `intp`, pybind11 hands that over, and HiGHS reports an index it can't use as `kError` instead of quietly corrupting it. Verified both behaviours before changing it.

nicoloangileri:
@PTNobel  it's been a few weeks without word from the HiGHS side, so I'd
rather ask how you'd like to handle it than leave it hanging.

The second half of your question is settled by the measurement above: on a
pure LP, setSolution with a primal point doesn't move the simplex (419
iterations with or without a start), so this only ever hands a start over
for MIPs.

That leaves the first half, whether to accept a start where only some
variables have values. If that's the part you're uneasy about, I can narrow
it to complete starts only: the start is passed when every variable has a
value, and skipped otherwise. That removes the partial case entirely, at the
cost of some flexibility.

So, whichever suits you: keep it as is, narrow it to complete starts, or
park it until someone from HiGHS weighs in.

PTNobel:
CVXPY development is quite slow at times, and I don't think we should rush a decision on this. I have a few questions for Julian, so I'll probably put an email to him together and see if he can weigh in.

PR #3491, opened 2026-08-20
Fixed PSD and NSD check to allow for complex coefficients

lfinocchiaro:
Minor change so that a complex valued PSD matrix does not fail the PSD validation check (which for now only works for real coefficients since only the transpose is done and not the complex conjugation)

### DIFF
--- a/cvxpy/expressions/leaf.py
+++ b/cvxpy/expressions/leaf.py
@@ -575,7 +575,7 @@ class Leaf(expression.Expression):
                   key in ['symmetric', 'PSD', 'NSD']]):
             if val.dtype.kind in 'ib':
                 val = val.astype(float)
-            val = val + np.swapaxes(val, -2, -1)
+            val = val + np.swapaxes(val, -2, -1).conj()
             val /= 2.
             if self.attributes['symmetric']:
                 return val
@@ -590,7 +590,7 @@ class Leaf(expression.Expression):
                 if not bad.any():
                     return val
                 w[bad] = 0
-            return (V * w[..., np.newaxis, :]) @ np.swapaxes(V, -2, -1)
+            return (V * w[..., np.newaxis, :]) @ np.swapaxes(V, -2, -1).conj()
         elif self.attributes['sparsity'] and not sparse_path:
             warn('Accessing a sparse CVXPY expression via a dense representation.'
                   ' Please report this as a bug to the CVXPY Discord or GitHub.',


Transurgeon (commented):
nice! could you add some regression tests please?

Transurgeon:
Output from Claude:

Thanks for the patch, and sorry for the slow turnaround here.

I went to add the regression test @Transurgeon asked for and couldn't write one, because this branch doesn't change any observable behaviour. `Leaf.project()` discards the imaginary part before `_project_structural` ever runs ([`leaf.py#L495-L496`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/expressions/leaf.py#L495-L496)):

```python
if not self.is_complex():
    val = np.real(val)
```

`is_complex()` is `attributes['complex'] or is_imag() or attributes['hermitian']`, and all three of those are *structural* attributes. So a leaf that is both complex and PSD/NSD/symmetric always has `num_attributes > 1`, and `project()` short-circuits at [`leaf.py#L500-L515`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/expressions/leaf.py#L500-L515) and returns the value unmodified before reaching the branch this PR edits. I enumerated every constructible attribute combination containing `symmetric`/`PSD`/`NSD` and none of them can deliver complex data to that code path.

For real input the two added `.conj()` calls are no-ops: `X + X.T.conj() == X + X.T`, and `eigh` of a real symmetric matrix returns a real `V`.

Running the case from the PR description against this branch gives the same result as master:

```python
import numpy as np, cvxpy as cp

H = np.array([[2.0, 1j], [-1j, 2.0]])          # Hermitian, eigenvalues 1 and 3
assert np.allclose(np.linalg.eigvalsh(H), [1., 3.])

p = cp.Parameter((2, 2), PSD=True)
print("project() ->\n", p.project(H, strict=False))
try:
    p.value = H
except ValueError as e:
    print("value assignment ->", e)
```

```
# identical on this branch and on master
project() ->
 [[2. 0.]
  [0. 2.]]
value assignment -> Parameter value must be positive semidefinite.
```

On the underlying need: complex Hermitian PSD matrices are already supported, through the `hermitian` attribute rather than `PSD`. `cp.Variable((n, n), hermitian=True)` with an `X >> 0` constraint solves correctly, `cp.Parameter((n, n), hermitian=True)` accepts Hermitian values, and `Constant.is_psd()` handles complex input fine (`is_psd_within_tol` is documented for Hermitian matrices). The `hermitian` branch of `_project_structural` already conjugates.

Making `PSD=True` itself accept complex values is a bigger change than two lines: `is_psd()` leaves are real in CVXPY's type system, so `Complex2Real` never sees them, and a stored complex value would flow silently into matrix stuffing and produce wrong results rather than an error. That would need `is_complex()`, the `Complex2Real` reduction and the canonicalization path to be taught about complex PSD leaves together.

So I'm going to close this one. If you have a concrete case that fails and isn't covered by `hermitian=True`, please do post it and we'll reopen — and thanks again for taking the time to send a fix.

Separately, and not something you need to act on: `cp.Parameter((2, 2), PSD=True, complex=True)` constructs fine and then accepts *any* value (I fed it `[[-5, 3], [1, -7]]`) while `is_psd()` still returns `True`. That's the same `num_attributes > 1` short-circuit, and I'll open a separate issue for it.

Issue #3489, opened 2026-08-19
Tracking: atoms that reject N-dimensional input


nicoloangileri:
Follow-up to the discussion in #3476, where @PTNobel wrote:

> I thought we had finished around N-d support to atoms? With the exception of like bmat or something. I would definitely support an issue with a list of remaining incompatibilities.

Coverage is indeed broad — `bmat` included, so that one is not an exception. Below are the atoms I could still make reject `ndim > 2` where the operation looks well defined. Line numbers are against `ae0fbff0` on master.

- [x] **`tv`** — [`atoms/total_variation.py:63`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/total_variation.py#L63): `raise ValueError("tv cannot have input arrays with more than 2 dimensions.")`. 0-D errors, 1-D is the L1 norm of successive differences, 2-D is isotropic TV, and anything higher is rejected. 3-D TV is a standard use case (volumetric and video denoising), and the 2-D branch generalises fairly directly.
- [ ] **`cvar`** — [`atoms/cvar.py:50`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/cvar.py#L50): `if len(x.shape) != 1: raise ValueError(...)`. Worth noting that the body is just `sum_largest(x, k) / k`, and `sum_largest` already accepts N-D input, so the restriction lives in the wrapper rather than in the machinery underneath. An `axis` argument looks like the cheapest of these four.
- [x] **`geo_mean`** — [`atoms/geo_mean.py:228`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/geo_mean.py#L228): requires `x.is_vector()`, so `raise ValueError('x must be a row or column vector.')` for anything else.
- [x] **`kron`** — [`atoms/affine/kron.py:46`](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/atoms/affine/kron.py#L46): `self.args[0].ndim != 2 or self.args[1].ndim != 2`. Arguably the most debatable of the four, since a batched Kronecker product needs a semantics decision first.

[...]

PR #3493, opened 2026-08-20
Accept any shape in geo_mean, and fix row vectors along the way

nicoloangileri:
## Description

Follow-up to #3489, where @PTNobel said to start with `geo_mean` and `tv`. This is the `geo_mean` half.

`geo_mean` required a vector and computed `n` as `max(x.shape)`. Anything past one dimension is now flattened in row-major order, so a row vector, a column vector and an N-D array holding the same entries all behave the same way. Row-major matches `numeric()`, which already flattened that way, and is the order `p` is expected in — there is a test pinning that down with asymmetric weights.

**It also fixes a row vector reaching the solver at all**, which I did not expect to find. Both the docstring and the error message promise "a row or column vector", but on master:

    x = cp.Variable((1, 4), nonneg=True)
    g = cp.geo_mean(x)                       # builds fine, shape ()
    g.value                                  # evaluates fine
    cp.Problem(cp.Maximize(g), [cp.sum(x) <= 4]).solve()
    # IndexError: Index 1 is out of bounds for axis 0 with size 1.

`geo_mean_canon` indexes `x[i]` along axis 0, which has size 1 for a row vector. Flattening removes the asymmetry rather than special-casing it, so `(4,)`, `(4, 1)`, `(1, 4)`, `(2, 2)`, `(2, 3)` and `(2, 3, 4)` now all solve to the same optimum.

The existing assertion in `test_geo_mean` that a `(3, 3)` input raises `ValueError` is replaced by the behaviour that supersedes it.

**On the test suite.** Running the full suite on this machine dies with a bus error partway through `nlp_tests`, but it does so on master too, so I compared around it: excluding `nlp_tests`, this branch is 1 failed / 2553 passed against master's 1 failed / 2551 passed, with identical failure lists — the two extra passes are the tests added here. `nlp_tests` alone fails identically on both (10 and 10). The one shared failure is `test_mip_vars.py::TestMIPVariable::test_miqp_warning`.

Refs #3489

## Type of change
- [x] Bug fix

- [x] Write unittests. — `test_geo_mean_shapes` (any shape, row-major order for `p`, 3-D input) and `test_geo_mean_row_vector` (each shape solves to the same optimum)
- [x] Run the unittests and check that they're passing. — see the note above; no regressions against master
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally. `geo_mean` adds a `reshape` node only for inputs past one dimension, which previously raised, so existing models are unaffected; the CI benchmark job will confirm.

### DIFF
--- a/cvxpy/atoms/geo_mean.py
+++ b/cvxpy/atoms/geo_mean.py
@@ -17,10 +17,11 @@ limitations under the License.
 from collections.abc import Sequence
 
 import numpy as np
-import scipy.sparse as sp
 
 from cvxpy.atoms.affine.promote import promote
-from cvxpy.atoms.atom import Atom
+from cvxpy.atoms.affine.reshape import reshape
+from cvxpy.atoms.affine.transpose import moveaxis
+from cvxpy.atoms.axis_atom import AxisAtom
 from cvxpy.atoms.errormsg import SECOND_ARG_SHOULD_NOT_BE_EXPRESSION_ERROR_MESSAGE
 from cvxpy.constraints.constraint import Constraint
 from cvxpy.expressions import cvxtypes
@@ -35,7 +36,8 @@ from cvxpy.utilities.power_tools import (
 )
 
 
-def geo_mean(x, p=None, max_denom=1024, approx=True) -> Expression:
+def geo_mean(x, p=None, max_denom=1024, approx=True, axis=None,
+             keepdims: bool = False) -> Expression:
     """Factory function for (weighted) geometric mean.
 
     Parameters
@@ -49,18 +51,25 @@ def geo_mean(x, p=None, max_denom=1024, approx=True) -> Expression:
     approx : bool
         When True (default), uses SOC approximation. When False,
         uses power cone (exact).
+    axis : int or tuple of int, optional
+        The axes to reduce along. ``None`` reduces every entry into a single
+        geometric mean.
+    keepdims : bool
+        Keep the reduced axes with length one.
 
     Returns
     -------
-    GeoMean or GeoMeanApprox
+    Expression
+        A scalar when ``axis`` is None, otherwise one geometric mean per
+        slice along ``axis``.
     """
     if approx:
-        return GeoMeanApprox(x, p=p, max_denom=max_denom)
-    else:
-        return GeoMean(x, p=p, max_denom=max_denom)
+        return GeoMeanApprox(x, p=p, max_denom=max_denom, axis=axis,
+                             keepdims=keepdims)
+    return GeoMean(x, p=p, max_denom=max_denom, axis=axis, keepdims=keepdims)
 
 
-class GeoMean(Atom):
+class GeoMean(AxisAtom):
     """ The (weighted) geometric mean of vector ``x``, with optional powers given by ``p``:
 
     .. math::
@@ -178,7 +187,10 @@ class GeoMean(Atom):
     Parameters
     ----------
     x : Variable
-        A column or row vector whose elements we will take the geometric mean of.
+        An expression of any shape whose elements we will take the geometric
+        mean of. Anything past one dimension is flattened in row-major order,
+        so a row vector, a column vector and an N-D array holding the same
+        entries all give the same result.
 
     p : Sequence (list, tuple, ...) of ``int``, ``float``, or ``Fraction`` objects
         A vector of weights for the weighted geometric mean
@@ -205,35 +217,37 @@ class GeoMean(Atom):
     """
 
     def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
-                 max_denom: int = 1024) -> None:
+                 max_denom: int = 1024, axis=None, keepdims: bool = False) -> None:
         Expression = cvxtypes.expression()
         if p is not None and isinstance(p, Expression):
             raise TypeError(SECOND_ARG_SHOULD_NOT_BE_EXPRESSION_ERROR_MESSAGE)
-        elif p is not None and hasattr(p, '__getitem__'):
+
+        if isinstance(x, list):
+            x = np.array(x)
+        x = Expression.cast(x)
+        if x.ndim == 0:
+            x = Expression.cast(promote(x, shape=(1,)))
+
+        super(GeoMean, self).__init__(x, axis=axis, keepdims=keepdims)
+
+        n = self._reduced_size()
+
+        if p is not None and hasattr(p, '__getitem__'):
             p = np.array(p)
-            idxs = p > 0
-            if isinstance(x, list):
-                x = np.array(x)
-            if x.ndim == 0:
-                x = Expression.cast(promote(x, shape=(1,)))[idxs]
-            else:
-                x = Expression.cast(x)[idxs]
-            p = p[idxs]
-        super(GeoMean, self).__init__(x)
-
-        x = self.args[0]
-        if x.is_vector():
-            n = 1 if x.ndim == 0 else max(x.shape)
+            if len(p) != n:
+                raise ValueError('x and p must have the same number of elements.')
+            # Entries with zero weight take no part. They are dropped from the
+            # canonicalization rather than from the argument, so that the
+            # argument keeps the shape the axis bookkeeping is based on.
+            self._keep = p > 0
+            p = p[self._keep]
         else:
-            raise ValueError('x must be a row or column vector.')
+            self._keep = np.ones(n, dtype=bool)
+            if p is None:
+                p = [1] * n
 
-        if p is None:
-            p = [1]*n
         self.p = p
         self.max_denom = max_denom
-
-        if len(p) != n:
-            raise ValueError('x and p must have the same number of elements.')
 
         if any(v < 0 for v in p) or sum(p) <= 0:
             raise ValueError('powers must be nonnegative and not all zero.')
@@ -242,20 +256,68 @@ class GeoMean(Atom):
         self.w = tuple(p_arr / p_arr.sum())
         self.approx_error = 0.0
 
-    # Returns the (weighted) geometric mean of the elements of x.
-    def numeric(self, values) -> float:
-        values = np.array(values[0]).flatten()
-        val = 1.0
-        for x, p in zip(values, self.w):
-            val *= x**float(p)
-        return val
+    def _reduced_size(self) -> int:
+        """How many entries each geometric mean is taken over."""
+        shape = self.args[0].shape
+        if self.axis is None:
+            return int(np.prod(shape))
+        axes = (self.axis,) if isinstance(self.axis, int) else tuple(self.axis)
+        return int(np.prod([shape[a] for a in axes]))
+
+    def _aligned_arg(self, arg=None):
+        """The argument with the reduced entries along axis 0.
+
+        Axes being reduced are moved to the front and flattened into one in
+        Fortran order, and entries whose weight is zero are dropped. Everything downstream — the
+        domain and the canonicalizers — works on this view, so none of them
+        has to know about the axis. The canonicalizers pass their own already
+        canonicalized argument, which has the same shape.
+        """
+        x = self.args[0] if arg is None else arg
+        if self.axis is None:
+            aligned = reshape(x, (x.size,), order='F')
+        else:
+            axes = (self.axis,) if isinstance(self.axis, int) else tuple(self.axis)
+            moved = moveaxis(x, axes, range(len(axes)))
+            aligned = reshape(moved, (self._reduced_size(), -1), order='F')
+        if not self._keep.all():
+            aligned = aligned[self._keep]
+        return aligned
+
+    def numeric(self, values):
+        """The (weighted) geometric mean, taken along the reduced axes."""
+        x = np.asarray(values[0], dtype=float)
+        w = np.array([float(w_i) for w_i in self.w])
+        if self.axis is None:
+            kept = x.ravel(order='F')[self._keep]
+            out = np.prod(kept ** w)
+        else:
+            axes = (self.axis,) if isinstance(self.axis, int) else tuple(self.axis)
+            moved = np.moveaxis(x, axes, range(len(axes)))
+            rest = moved.shape[len(axes):]
+            flat = moved.reshape((self._reduced_size(), -1), order='F')
+            kept = flat[self._keep]
+            out = np.prod(kept ** w[:, None], axis=0).reshape(rest, order='F')
+        return np.reshape(out, self.shape, order='F') if self.keepdims else out
 
     def _domain(self) -> list[Constraint]:
         """Returns constraints describing the domain of the node.
         """
-        # No special case when only one non-zero weight.
-        selection = np.array([w_i > 0 for w_i in self.w])
-        return [self.args[0][selection > 0] >= 0]
+        # Entries with zero weight are already gone from the aligned view, so
+        # they carry no implicit nonnegativity.
+        return [self._aligned_arg() >= 0]
+
+    def _column_grad(self, value):
+        """Gradient with respect to one column of reduced entries."""
+        column = np.asarray(value, dtype=float).ravel()
+        kept = column[self._keep]
+        w = np.array([float(w_i) for w_i in self.w])
+        if np.any(kept <= 0):
+            return None
+        gm = np.prod(kept ** w)
+        D = np.zeros(column.size)
+        D[self._keep] = w / kept * gm
+        return D[:, None]
 
     def _grad(self, values):
         """Gives the (sub/super)gradient of the atom w.r.t. each argument.
@@ -268,15 +330,7 @@ class GeoMean(Atom):
         Returns:
             A list of SciPy CSC sparse matrices or None.
         """
-        x = np.array(values[0])
-        # No special case when only one non-zero weight.
-        w_arr = np.array([float(w_i) for w_i in self.w])
-        # Outside domain.
-        if np.any(x[w_arr > 0] <= 0):
-            return [None]
-        else:
-            D = w_arr/x.ravel(order='F')*self.numeric(values)
-            return [sp.csc_array([D]).T]
+        return self._axis_grad(values)
 
     def name(self) -> str:
         weights = ', '.join(str(v) for v in self.w)
@@ -288,11 +342,6 @@ class GeoMean(Atom):
         weights = ', '.join(str(v) for v in self.w)
         return f"{type(self).__name__}({self.args[0].format_labeled()}, ({weights}))"
 
-
-    def shape_from_args(self) -> tuple[int, ...]:
-        """Returns the (row, col) shape of the expression.
-        """
-        return tuple()
 
     def sign_from_args(self) -> tuple[bool, bool]:
         """Returns sign (is positive, is negative) of the expression.
@@ -337,7 +386,7 @@ class GeoMean(Atom):
         return False
 
     def get_data(self):
-        return [self.p, self.max_denom]
+        return [self.p, self.max_denom, self.axis, self.keepdims]
 
     def copy(self, args=None, id_objects=None):
         """Returns a shallow copy of the GeoMean atom.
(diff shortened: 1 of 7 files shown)


nicoloangileri:
Note on CI: `test_optional_solvers` is red on ubuntu-22.04 and macos-14, as it is on master and on every recent PR — the same two jobs, failing at the same steps (ubuntu at both `test_conic_solvers` and `test_qp_solvers`, macos at `test_conic_solvers` only). #3483 was approved and merged with exactly this pattern.

It also can't come from here: that workflow runs the conic and QP solver suites, and this PR touches `cvxpy/atoms/geo_mean.py`, `test_atoms.py` and `test_problem.py` — none of which it executes.

PTNobel:
Quick aside; I think we should aim for the same API as cp.mean with axis and keepdims

nicoloangileri:
Good call — and it turned out the flattening was already the `axis=None` case of that API, so this only had to add the rest.

`axis` moves to the front, the remaining axes are flattened, one geometric mean is taken per column, and the result is reshaped back. Shapes agree with `cp.mean` everywhere I compared them — 1-D, 2-D and 3-D, both values of `keepdims`, and `axis=None` — and values agree with `exp(mean(log(x), axis))`. Weights apply to each slice along the reduced axis.

Only axes 0 and 1 are accepted, matching what `cp.mean`, `cp.std` and `cp.var` support today. I used `ValueError` with `cp.std`'s wording; `cp.mean` raises `UserWarning` for the same case, which looks like an oversight on its side rather than something to copy — happy to match it instead if you'd rather they be identical.

It's a separate commit on top of the flatten one, so it's easy to drop if you'd rather this landed in two pieces. Rebased onto master while I was here.

nicoloangileri:
Following your point on #3489 — matching `cp.mean`'s "0 or 1" limit here would have carried the same gap forward, so I've dropped it. Any axis reduces now, negative ones included.

It turned out to be one line: the implementation already moves the axis to the front and flattens the rest, which works for any of them, and `moveaxis` takes negative axes. Only the guard was stopping it. Checked against `exp(mean(log(x), axis))` for every axis of a 3-D array, positive and negative, with and without `keepdims`; the 2-D cases still agree with `cp.mean`, so nothing that worked before changes.

An axis the expression doesn't have now raises with numpy's wording, since "other than 0 or 1" no longer describes anything.

PTNobel (commented):
High level review: we should support tuples of axes as well.

I'm a little hesitant about the hstack, it feels like a performance foot gun. Doesn't mean we can't do it, but I'm going to have to think a bit about if there's a better approach

PTNobel:
Does having the geomean classes inherit from AxisAtom help anything?

nicoloangileri:
All three point the same way, and you're right about the hstack — it's a foot gun, and the batching to replace it already exists here.

**On `gm_constrs`.** `gm` is already vectorised: it takes `length = t.size` and emits one SOC over all entries, and the docstring says "x and t can either be scalar or matrix variables". Measured it with a (3, 5) input, reducing along axis 0:

    gm_constrs(t_of_shape_5, [X[i] for i in range(3)], w)
    -> 3 SOC constraints, each with t of shape (5,)

Three constraints for five columns, and the count doesn't grow with the number of columns  it's set by the weight tree. Values matched `exp(mean(log(X), axis=0))`. What this PR does instead is build five separate `GeoMean` atoms, each with its own tree. So the current approach is strictly worse and there's nothing to invent.

The exact path batches too: `PowConeND` already takes `axis` and validates `z.size == W.shape[1 - axis]`, one cone per column.

**On `AxisAtom`.** Yes, it helps, and it answers the tuple question at the same time — `normalize_axis` handles tuples, negative axes, and collapses "all axes" to `None`. That's the piece I'd otherwise have to write by hand and get subtly wrong.

**So the shape of it** would be: `GeoMean` and `GeoMeanApprox` become `AxisAtom`s carrying `axis`/`keepdims`, `shape_from_args` returns the reduced shape, and the canonicalizers do the batching the way `pnorm_canon` and `log_sum_exp_canon` already do no stacking anywhere. It touches `__init__`, `shape_from_args`, `get_data`, and the three canonicalizers, so it's a real change rather than a tweak.

Happy to do it. Two questions on how you'd like it: in this PR or as a follow-up, and would you rather I drop the axis commits here meanwhile? The first commit is independent of all of this — it's the flatten plus the row-vector fix, where `geo_mean(Variable((1, 4)))` built and evaluated fine but raised `IndexError` during canonicalization. That part could land on its own while the axis design settles.

PTNobel:
Let's do one PR with a correct implementation

nicoloangileri:
Rewritten as one implementation, as you asked. All three points turned out to be the same change, so `GeoMean` and `GeoMeanApprox` now inherit from `AxisAtom`.

Tuples of axes come with it — `normalize_axis` handles tuples, negative indices and reducing over every axis. Checked against `exp(mean(log(x), axis))` for 2-D and 3-D inputs, every axis and every pair, with and without `keepdims`.

The stacking is gone. You were right that there was a better approach and it was already here: `gm` takes `t.size` and emits one SOC over all entries, and `PowConeND` takes one cone per column, so both canonicalizers now cover every output with a single set of constraints. Measured through `geo_mean_approx_canon`:

    2 outputs   -> 3 SOC constraints
    20 outputs  -> 3 SOC constraints
    200 outputs -> 3 SOC constraints

There's a test asserting the count doesn't move with the number of outputs.

One thing I ran into that wasn't in the review: zero-weight entries were being indexed out of the argument itself, which put a stray `[True, True, True]` into the printed form of every `geo_mean` even with no zero weights, and wouldn't have survived an axis. They're now dropped in the aligned view the canonicalizers work on, so the argument keeps the shape the axis bookkeeping needs. That's why two expectations in `test_labels.py` lose that index.

The row-vector fix from the first version is still in: `geo_mean(Variable((1, 4)))` built and evaluated fine on master, then raised `IndexError` during canonicalization.

Against master, excluding `nlp_tests` (bus error on this machine, on master too): 1 failed / 2566 passed here vs 1 failed / 2558 passed, same single pre-existing failure. `ruff` clean across the tree. One caveat: I couldn't exercise the `dnlp2smooth` path locally — no NLP solver installed — so that canonicalizer is updated for consistency but untested on my side.

cvxpy/atoms/geo_mean.py
```diff
@@ -49,18 +51,27 @@ def geo_mean(x, p=None, max_denom=1024, approx=True) -> Expression:
     approx : bool
         When True (default), uses SOC approximation. When False,
         uses power cone (exact).
+    axis : int, optional
+        The axis to reduce along, following ``cp.mean``'s spelling. ``None``
+        reduces every entry into a single geometric mean. Any axis of the
+        expression is accepted, negative ones included; ``cp.mean``,
```
PTNobel: Why are we taking a dig at other atoms in these docs?

In any case, we just merged the fix to this limitation.

cvxpy/atoms/geo_mean.py
```diff
@@ -49,18 +51,27 @@ def geo_mean(x, p=None, max_denom=1024, approx=True) -> Expression:
...
+        ``cp.std`` and ``cp.var`` currently stop at 1.
+    keepdims : bool
+        Keep the reduced axis with length one, matching ``cp.mean``.
```
PTNobel: Let's kill this line keepdims has a consistent meaning without referencing mean

cvxpy/atoms/geo_mean.py
```diff
@@ -205,34 +219,39 @@ class GeoMean(Atom):
...
+            if p is None:
+                p = [1] * n
 
-        if p is None:
-            p = [1]*n
         self.p = p
         self.max_denom = max_denom
 
-        if len(p) != n:
+        if len(p) != int(self._keep.sum()):
```
PTNobel: This is true by construction and the error message seems unrelated to the test condition

cvxpy/atoms/geo_mean.py
```diff
@@ -242,20 +261,67 @@ def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
...
         self.approx_error = 0.0
 
-    # Returns the (weighted) geometric mean of the elements of x.
-    def numeric(self, values) -> float:
-        values = np.array(values[0]).flatten()
-        val = 1.0
-        for x, p in zip(values, self.w):
-            val *= x**float(p)
-        return val
+    def reduced_size(self) -> int:
```
PTNobel: Should this be a private method

cvxpy/atoms/geo_mean.py
```diff
@@ -242,20 +261,67 @@ def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
...
+        """How many entries each geometric mean is taken over."""
+        shape = self.args[0].shape
+        if self.axis is None:
+            return int(np.prod(shape)) if shape else 1
+        axes = (self.axis,) if isinstance(self.axis, int) else tuple(self.axis)
+        return int(np.prod([shape[a] for a in axes]))
+
+    def aligned_arg(self, arg=None):
```
PTNobel: Why is this public

cvxpy/atoms/geo_mean.py
```diff
@@ -242,20 +261,67 @@ def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
...
-        values = np.array(values[0]).flatten()
-        val = 1.0
-        for x, p in zip(values, self.w):
-            val *= x**float(p)
-        return val
+    def reduced_size(self) -> int:
+        """How many entries each geometric mean is taken over."""
+        shape = self.args[0].shape
+        if self.axis is None:
+            return int(np.prod(shape)) if shape else 1
```
PTNobel: What is the if shape catching? np.prod(()) == 1.0 which gets cast to 1

cvxpy/atoms/geo_mean.py
```diff
@@ -242,20 +261,67 @@ def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
...
+
+        Axes being reduced are moved to the front and flattened into one, and
+        entries whose weight is zero are dropped. Everything downstream — the
+        domain and the canonicalizers — works on this view, so none of them
+        has to know about the axis. The canonicalizers pass their own already
+        canonicalized argument, which has the same shape.
+        """
+        x = self.args[0] if arg is None else arg
+        if self.axis is None:
+            aligned = reshape(x, (x.size,), order='C')
```
PTNobel: We normally use F order in the CVXPY codebase. Why do something different here?

cvxpy/atoms/geo_mean.py
```diff
@@ -242,20 +261,67 @@ def __init__(self, x, p: Sequence[float] | np.ndarray | None = None,
...
+        else:
+            axes = (self.axis,) if isinstance(self.axis, int) else tuple(self.axis)
+            moved = moveaxis(x, axes, range(len(axes)))
+            aligned = reshape(moved, (self.reduced_size(), -1), order='C')
```
PTNobel: Ditto in C not F

cvxpy/atoms/geo_mean.py
```diff
@@ -268,15 +334,7 @@ def _grad(self, values):
...
-        x = np.array(values[0])
-        # No special case when only one non-zero weight.
-        w_arr = np.array([float(w_i) for w_i in self.w])
-        # Outside domain.
-        if np.any(x[w_arr > 0] <= 0):
-            return [None]
-        else:
-            D = w_arr/x.ravel(order='F')*self.numeric(values)
-            return [sp.csc_array([D]).T]
+        return self._axis_grad(values)
```
PTNobel, line 333: I'm surprised this isn't the default impl of _grad

cvxpy/atoms/geo_mean.py
PTNobel, line 470: Can't we just inherit from GeoMean

cvxpy/reductions/dcp2cone/canonicalizers/geo_mean_canon.py
```diff
@@ -28,42 +28,45 @@
...
     w = expr.w
+    # One row per reduced entry, one column per output. PowConeND takes one
+    # cone per column, so every output is covered by a single constraint.
+    W = expr.aligned_arg(args[0])
 
     # Single non-zero weight: geo_mean is just that element (affine).
     if len(w) == 1:
-        return x, []
+        out = W[0]
+        return (reshape(out, expr.shape, order='C')
```
PTNobel: Why C?

cvxpy/reductions/dcp2cone/canonicalizers/geo_mean_canon.py
```diff
@@ -28,42 +28,45 @@
...
+    t = Variable(W.shape[1:])
     alpha = np.array([float(wi) for wi in w]).reshape(-1, 1)
-    return t, [PowConeND(W, t, alpha, axis=0)]
+    if W.ndim == 2:
+        alpha = np.tile(alpha, (1, W.shape[1]))
+    else:
+        alpha = alpha.reshape(-1)
+    constrs = [PowConeND(W, t, alpha, axis=0)]
+    if t.shape != expr.shape:
+        return reshape(t, expr.shape, order='C'), constrs
```
PTNobel: Why C?

cvxpy/reductions/dcp2cone/canonicalizers/geo_mean_canon.py
```diff
@@ -28,42 +28,45 @@
...
+    x = expr.aligned_arg(args[0])
 
     # Single non-zero weight: geo_mean is just that element (affine).
     if len(w) == 1:
-        return x, []
-
-    shape = expr.shape
-    t = Variable(shape)
+        out = x[0]
+        return (reshape(out, expr.shape, order='C')
```
PTNobel: Why C?

cvxpy/reductions/dcp2cone/canonicalizers/geo_mean_canon.py
```diff
@@ -85,4 +88,6 @@ def geo_mean_approx_canon(expr, args, solver_context: SolverInfo | None = None):
                 stacklevel=6,
             )
 
+    if t.shape != expr.shape:
+        return reshape(t, expr.shape, order='C'), constrs
```
PTNobel: Why C?

cvxpy/reductions/dgp2dcp/canonicalizers/geo_mean_canon.py
```diff
@@ -1,5 +1,14 @@
...
+    x = expr.aligned_arg(args[0])
     out = 0.0
-    for x_i, p_i in zip(args[0], expr.p):
-        out += p_i * x_i
-    return (1 / sum(expr.p)) * out, []
+    for i, p_i in enumerate(expr.p):
+        out += p_i * x[i]
+    out = (1 / sum(expr.p)) * out
+    if out.shape != expr.shape:
+        out = reshape(out, expr.shape, order='C')
```
PTNobel: Why C?

cvxpy/reductions/dnlp2smooth/canonicalizers/geo_mean_canon.py
```diff
@@ -28,9 +29,16 @@ def geo_mean_canon(expr, args):
...
     log_expr = log(args[0])
     var, constr = log_canon(log_expr, expr.args)
-    return t, [log(t) == sum(multiply(weights, var))] + constr
+    # Reduced entries along axis 0, so the weights broadcast against whatever
+    # shape is left over.
+    aligned = expr.aligned_arg(var)
+    weights = weights.reshape((-1,) + (1,) * (aligned.ndim - 1))
+    reduced = sum(multiply(weights, aligned), axis=0)
+    if reduced.shape != expr.shape:
+        reduced = reshape(reduced, expr.shape, order='C')
```
PTNobel: Why C?

cvxpy/tests/test_atoms.py
```diff
@@ -215,6 +215,126 @@ def test_geo_mean(self) -> None:
             ):
             cp.geo_mean(self.x, self.y)
 
+    def test_geo_mean_shapes(self) -> None:
+        """Any shape is accepted, and the entries are taken in row-major order."""
+        from fractions import Fraction
```
PTNobel: Imports go at the top of files

cvxpy/tests/test_atoms.py
```diff
@@ -215,6 +215,126 @@ def test_geo_mean(self) -> None:
...
+        """axis and keepdims follow cp.mean, in shape and in value."""
+        def gmean(a, axis=None):
+            return np.exp(np.mean(np.log(a), axis=axis))
+
+        M = np.arange(1.0, 7.0).reshape(2, 3)
+        T = np.arange(1.0, 25.0).reshape(2, 3, 4)
+
+        for arr in (np.array([1.0, 2.0, 3.0, 4.0]), M, T):
+            const = cp.Constant(arr)
+            axes = (None, 0) if arr.ndim == 1 else (None, 0, 1)
```
PTNobel: Let's just merge in master and use this for all the axes to test

cvxpy/tests/test_atoms.py
```diff
@@ -215,6 +215,126 @@ def test_geo_mean(self) -> None:
...
+                        )
+                    assert np.allclose(np.array(got.value), expected)
+
+    def test_geo_mean_axis_with_weights(self) -> None:
+        """Weights apply to each slice along the reduced axis."""
+        M = np.arange(1.0, 7.0).reshape(2, 3)
+        got = cp.geo_mean(cp.Constant(M), [3, 1], axis=0)
+        assert np.allclose(np.array(got.value), (M[0] ** 3 * M[1]) ** (1 / 4))
+
+    def test_geo_mean_axis_beyond_the_second(self) -> None:
```
PTNobel: See previous comment

cvxpy/tests/test_atoms.py
```diff
@@ -215,6 +215,126 @@ def test_geo_mean(self) -> None:
...
+        assert cp.geo_mean(const, axis=-1, keepdims=True).shape == (2, 3, 1)
+
+    def test_geo_mean_axis_tuple(self) -> None:
+        """A tuple of axes reduces over all of them at once."""
+        def gmean(a, axis):
+            return np.exp(np.mean(np.log(a), axis=axis))
+
+        T = np.arange(1.0, 25.0).reshape(2, 3, 4)
+        const = cp.Constant(T)
+        for axes in ((0, 1), (1, 2), (0, 2)):
```
PTNobel: Can't these be included in the list in the other test

cvxpy/tests/test_atoms.py
```diff
@@ -215,6 +215,126 @@ def test_geo_mean(self) -> None:
...
+
+    def test_geo_mean_axis_matches_mean(self) -> None:
+        """axis and keepdims follow cp.mean, in shape and in value."""
+        def gmean(a, axis=None):
+            return np.exp(np.mean(np.log(a), axis=axis))
+
+        M = np.arange(1.0, 7.0).reshape(2, 3)
+        T = np.arange(1.0, 25.0).reshape(2, 3, 4)
+
+        for arr in (np.array([1.0, 2.0, 3.0, 4.0]), M, T):
```
PTNobel: Can we use a pytest mark parametrize over axis for better readability

cvxpy/tests/test_labels.py
PTNobel, line 1: Love these changes!

PTNobel (commented):
This is awesome! A handful of nits. Sorry about the same q 6 times or whatever, just answer it once. It makes it easier for me to review the revision if I repeat the question

nicoloangileri:
Thanks all applied. Answering the repeated one once, as asked.

**Why C?** No good reason: I matched what `numeric()` already did (`.flatten()` is C order) without checking the convention around it. You're right, and it wasn't cosmetic  the order decides which entry each weight belongs to. Everything is F now: `_aligned_arg`, all three canonicalizers, and `numeric`, which had to move with them or the computed value and the solved one would disagree on weighted inputs. Re-checked that they still agree.

**The dead check** right on both counts. After the masking, `len(p)` equals the mask's sum by construction, and the message was about something else. Gone.

**`_reduced_size` and `_aligned_arg`** are private now. `np.prod(())` is indeed 1, so the guard is gone. `GeoMeanApprox.copy` now calls `super().copy()` and only adds its own fields. The docstring no longer mentions other atoms and it would have been wrong by this morning anyway, since #3496 landed.

**On `_grad`:** I was surprised too. Several `AxisAtom` subclasses define `_grad` as exactly `return self._axis_grad(values)`. Making that the base implementation looks right, but it changes behaviour for every subclass that doesn't currently do it, so I left it alone here. Happy to open it separately if you want.

**Tests:** the axis cases are one parametrized test over eleven axes now, imports moved to the top, and it's rebased on master so it uses what #3496 added.

Against master: 1 failed / 2586 passed here vs 1 failed / 2558 passed, same single pre-existing failure, `ruff` clean. Still no NLP solver on my side, so `dnlp2smooth` remains updated-but-unexercised.

PTNobel (approved):
Great work! Thank you

PR #3494, opened 2026-08-20
Read a 3-D value in tv as channels

nicoloangileri:
## Description

The other half of @PTNobel's "let's start with geo_mean and TV" in #3489, after #3493.

**A correction to what I wrote in that issue first.** I described 3-D `tv` as volumetric and video denoising. That was imprecise, and looking at the atom properly changes what the right generalisation is. `tv` already has a third dimension, reached through `*args`: `tv(R, G, B)` is *colour* total variation, where the gradients of every channel at a pixel are stacked into one vector before its L2 norm is taken. It is not the sum of the per-channel variations — on a random 4x5 example, `tv(R, G, B)` is 12.93 while `tv(R) + tv(G) + tv(B)` is 20.47 — and the differences stay along the two spatial axes.

So the generalisation that matches the atom is channels, not volume. An `(m, n, k)` array now means the same as passing its `k` slices:

    tv(A) == tv(A[:, :, 0], ..., A[:, :, k - 1])

verified to machine precision. The two spellings also compose: `tv(A, X)` equals `tv(A[:, :, 0], A[:, :, 1], A[:, :, 2], X)`.

Vectors and matrices are untouched. Four dimensions still raise: a variation over a volume would take differences along the third axis as well, which is a different operator, and I would rather not guess which one is wanted. The message now says which shapes are accepted. **If volumetric TV is something you'd want, say so and I'll open it separately** — it is a genuinely different atom rather than an extension of this one.

**Tests.** Against master, excluding `nlp_tests` (which dies with a bus error on this machine, on master as well): 1 failed / 2561 passed here against 1 failed / 2559 passed on master, the two extra passes being the tests added. Same single pre-existing failure, `test_mip_vars.py::TestMIPVariable::test_miqp_warning`.

Refs #3489

## Type of change
- [ ] Bug fix

- [x] Write unittests. — `test_tv_three_dimensional` (the two spellings agree, channels stay coupled, composition with further arguments, 4-D still refused) and `test_tv_unchanged_for_vectors_and_matrices`
- [x] Run the unittests and check that they're passing. — see above; no regressions
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally. The 2-D and 1-D paths are unchanged and the 3-D one previously raised, so existing models are unaffected; the CI benchmark job will confirm.

### DIFF
--- a/cvxpy/atoms/total_variation.py
+++ b/cvxpy/atoms/total_variation.py
@@ -27,10 +27,17 @@ def tv(value, *args) -> Expression:
     Uses L1 norm of discrete gradients for vectors and
     L2 norm of discrete gradients for matrices.
 
+    Channels are coupled: the gradients of every channel at a pixel are
+    stacked into one vector before its L2 norm is taken, so this is colour
+    total variation and not the sum of the per-channel ones. Differences are
+    taken along the two spatial axes only.
+
     Parameters
     ----------
     value : Expression or numeric constant
-        The value to take the total variation of.
+        The value to take the total variation of. A three-dimensional value
+        of shape ``(m, n, k)`` is read as ``k`` channels of an ``m`` by ``n``
+        image, equivalent to passing its slices as separate arguments.
     *args : Matrix constants/expressions
         Additional matrices extending the third dimension of value.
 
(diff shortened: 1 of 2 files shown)


cvxpy/atoms/total_variation.py
```diff
@@ -27,10 +27,17 @@ def tv(value, *args) -> Expression:
     Uses L1 norm of discrete gradients for vectors and
     L2 norm of discrete gradients for matrices.
 
+    Channels are coupled: the gradients of every channel at a pixel are
+    stacked into one vector before its L2 norm is taken, so this is colour
+    total variation and not the sum of the per-channel ones. Differences are
+    taken along the two spatial axes only.
```
PTNobel, line 33: I don't understand this

cvxpy/atoms/total_variation.py
```diff
@@ -27,10 +27,17 @@ def tv(value, *args) -> Expression:
...
+
     Parameters
     ----------
     value : Expression or numeric constant
-        The value to take the total variation of.
+        The value to take the total variation of. A three-dimensional value
```
PTNobel, line 38: This seems like the wrong place for this discussion

PTNobel (commented):
Looking at the PR and what the proposed API is, I'm hesitant. For the color channel application the third dim should always be low dimensional enough that vectorization doesn't matter right

nicoloangileri:
You're right on the performance point, and I should be plain that I can't argue against it: this doesn't vectorise anything. The implementation builds the same list of slices you'd write by hand  `values = [value[:, :, i] for i in range(k)]` so there's no gain at k=3 and none at k=300 either.

What's left is only ergonomics, and it's a modest case. Since 1.6 an N-D `Variable` is an ordinary thing to have, and `cp.Variable((m, n, 3))` is how you'd naturally hold an image; `tv` is then the one atom in the picture that won't take it, so you slice it apart at the call site. That's `tv(*[X[:, :, i] for i in range(X.shape[2])])`  fine, just inconsistent with where #3489 is heading.

That's the whole argument. If you don't think it's worth the API surface, I'd rather close this than have it merged on a weak case  say so and I'll close it and mark `tv` as won't-do on #3489.

The two docstring comments are fair either way: the coupling paragraph is over-explaining, and the equivalence note doesn't belong under Parameters. I'll fix both if this goes ahead.

PTNobel:
Let's close this i think. If someone is trying to use tv on 3D and it would simplify their code, they're likely to find this or the original issue and we can talk to them about what API is best. I don't think it's clear to me this is right and what we should commit to.

Thank you for looking into it! Exploring APIs that don't get merged are a really important apart of good design

nicoloangileri:
Makes sense, and thanks I'd rather have it closed on that reasoning than merged on a weak one. If someone turns up with a real 3-D case the branch is still on my fork.

PR #3441, opened 2026-07-06
Dual feasibility checks

WilliamZhang20:
## Description
Please include a short summary of the change.
Issue link (if applicable): #434, #3418

Implement functionality for constraint residuals for dual post-solve feasibility checks.

## Type of change
- [x] New feature (backwards compatible)
- [ ] New feature (breaking API changes)
- [ ] Bug fix
- [ ] Other (Documentation, CI, ...)

## [Contribution checklist](https://www.cvxpy.org/contributing/index.html#contribution-checklist)
- [ ] Add our license to new files.
- [x] Check that your code adheres to our coding style.
- [ ] Write unittests.
- [ ] Run the unittests and check that they’re passing.
- [ ] Run the benchmarks to make sure your change doesn’t introduce a regression.

PR #3495, opened 2026-08-20
Add constraint-level complementarity checks

curious7-web:
## Description

This follows the post-solver feasibility design note in #3362, which frames primal violation, dual feasibility, and complementarity as first-pass checks that can be evaluated from unpacked primal/dual values in the original problem space.

With primal residual/violation consistency merged in #3434 and dual feasibility merged in #3441, this draft starts the next layer: constraint-level complementarity checks where well-defined.

Initial scope:
- add a constraint-level `complementarity_residual`
- add scalar `complementarity_violation()` reduction
- implement complementarity checks for `Inequality`, `NonPos`, `NonNeg`, and `PSD`
- use matrix-product complementarity for PSD constraints with operator-norm scalar reduction
- add tests for active/inactive complementarity behavior, PSD matrix-product behavior, batched PSD constraints, and missing values

Out of scope for this first draft:
- SOC / ExpCone / PowCone complementarity
- problem-level aggregation
- stationarity / duality gap

Issue link (if applicable): Related to #3362 and #3418. Builds on #3434 and #3441.

## Type of change
- [ ] Bug fix

- [x] Add our license to new files. No new files added.
- [x] Write unittests.

### DIFF
--- a/cvxpy/constraints/constraint.py
+++ b/cvxpy/constraints/constraint.py
@@ -333,6 +333,49 @@ class Constraint(u.Canonical):
             return 0.0
         return float(np.linalg.norm(residual_arr.ravel(), ord=np.inf))
 
+    @property
+    def complementarity_residual(self):
+        """The shape-preserving complementarity residual.
+
+        This residual measures complementary slackness for constraints where
+        it is well-defined. The reduction to a scalar happens in
+        ``complementarity_violation``.
+
+        Returns
+        -------
+        NumPy.ndarray or None
+        """
+        raise NotImplementedError(
+            f"{type(self).__name__} does not implement "
+            "complementarity_residual."
+        )
+
+    def complementarity_violation(self):
+        """Scalar complementarity violation.
+
+        The violation is the infinity norm of ``complementarity_residual``,
+        matching the scalar violation convention for primal and dual residuals.
+
+        Returns
+        -------
+        float
+
+        Raises
+        ------
+        ValueError
+            If the primal or dual variable has no value.
+        """
+        residual = self.complementarity_residual
+        if residual is None:
+            raise ValueError(
+                "Cannot compute complementarity violation: missing primal "
+                "or dual value."
+            )
+        residual_arr = np.asarray(residual)
+        if residual_arr.size == 0:
+            return 0.0
+        return float(np.linalg.norm(residual_arr.ravel(), ord=np.inf))
+
     def is_dual_feasible(self, tolerance: float = 1e-8) -> bool:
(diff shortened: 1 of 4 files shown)


curious7-web:
Marked this ready for review.

I rebased on current master and re-ran the local constraint/problem tests. The scope is still constraint-level complementarity checks for NonPos, NonNeg, Inequality, and PSD, with SOC/ExpCone/PowCone, problem-level aggregation, stationarity, and duality gap left out of scope for later.

The optional-solver failures look unrelated to this PR and match the CI issue being discussed separately.

PTNobel:
Did a quick scan, two things: 

1. What is a complementarity *residual*? Complementarity is about an inner product, there is no norm in its definition whose argument would be a residual.
2. I don't think we should merge this without the asymmetric cones. Frankly, I don't understand why the generic implementation can't handle all the cones, but I'm not familiar with how we store all the data structures so I might be wrong on that.

PTNobel:
Also the docs make no sense to me but perhaps I'm being dense; I can't imagine that they could be good without some equations.

PTNobel:
It has been pointed out to me that complementarity residuals do make sense in the Jordan algebra. I do not know how to generalize that to asymmetric cones, so I definitely don't want to merge any APIs till we know how we'll handle both

curious7-web:
Yeah fair point. I was following the residual/violation pattern from the primal/dual checks, but I agree the complementarity definition needs to be pinned down more carefully before this becomes an API.

I’ll move this back to draft for now and write out the equations/design more clearly, especially what “residual” should mean for symmetric cones vs the asymmetric cones.

rileyjmurray:
I saw "makes sense for Jordan algebras" when talking about an API and immediately became concerned about asymmetric cones. Just chiming in to say that I'd _love_ for us to not tie an API abstraction to symmetric cones.

curious7-web:
Makes sense. I don’t want the API tied to symmetric cones either.

I’ll keep this as a draft and rethink the abstraction. My current guess is that the scalar inner-product quantity is the common piece, and any residual-like object should probably stay cone-specific/internal unless there’s a clean story for asymmetric cones too.

PTNobel:
I thought about it more, and the Jordan algebra residual would be a bad choice to use regardless of all else, because for the SOC it computes a ton of extraneous information.

curious7-web:
That helps, thanks. I think I was trying too hard to fit this into the residual/violation pattern from the earlier checks.

I’ll back up from the public complementarity_residual API for now and think about whether the only thing worth exposing is the scalar complementarity measure. I’ll leave this draft parked until I have a cleaner proposal.

PR #3499, opened 2026-08-27
Extract direct cone constraints for Moreau

PTNobel:
## Description

Moreau supports placing cones directly on subvectors of `x`, avoiding redundant identity rows in `A x + s = b`. This PR adds an opt-in `ExtractDirectCones` reduction and advertises Moreau's supported direct cone families through `DIR_CONE_KINDS`.

The reduction recognizes structurally constant NonNeg, SOC, SvecPSD, ExpCone, PowCone3D, and PowConeND blocks after matrix stuffing. Registered extractors emit typed `DirectCone` metadata in `ParamConeProg.dir_cones`, and a shared linear scan removes eligible rows without sorting. Parameter-dependent blocks stay on the slack side, preserving DPP correctness.

The Moreau interface maps unscaled symmetric variables into scaled svec coordinates and recovers direct cone duals from the full quadratic KKT residual. Slack-side PSD constraints remain supported, with mixed PSD/EXP/POW3D ordering covered by an interface regression. CI uses the normal `MOREAU` optional dependency, and the solver support table now lists SDP support.

The interface uses `moreau.DirectConeSpec` and `Cones.dir_cones`. The corresponding API rename is [moreau-project/moreau#3](https://github.com/moreau-project/moreau/pull/3) and must be included in the planned Moreau >=0.4.0 release. The cone-ordering fix was merged in [moreau-project/moreau#1](https://github.com/moreau-project/moreau/pull/1).

Validation on the current revision: 71 extraction, parameter-program, and DPP tests pass, plus 14 subtests; all 28 Moreau tests and 30 CLARABEL tests pass. Pre-commit passes. Moreau integration was verified against a CPU wheel built from the companion branch, whose current version metadata is 0.3.3; this is not validation of a published >=0.4.0 package. The companion branch also passes 111 focused CPU tests. CUDA build and runtime validation are outstanding there.

The earlier broader cone-to-cone run had seven solver-availability/capability failures, all reproduced on the pre-change commit; one known ECOS_BB test was skipped.

## Type of change
- [ ] Bug fix

- [x] Write unittests.
- [x] Run the unittests and check that they are passing.
- [ ] Run the benchmarks to make sure your change does not introduce a regression.


tools/bench_moreau_extract.py
PTNobel, line 1: We should drop this before merge lol

.github/workflows/test_optional_solvers.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         run: uv sync --all-extras --no-extra uno --dev
       - name: Install Moreau
```
PTNobel: Can't we handle this normally now that we have it is an optional dep

cvxpy/reductions/cone2cone/extract_identity_cones.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def invert(self, solution, inverse_data):
+        """Restore extracted-block duals into the Solution.
+
+        The solver is responsible for computing x_cone dual variables
+        (it has the materialised q, A, z) and for placing them in
+        ``solution.dual_vars`` keyed by the *original* constraint id.
+        That keeps the reduction's invert simple — it only forwards.
+        """
+        if isinstance(inverse_data, _NoopInverse):
```
PTNobel: Why is this here

cvxpy/reductions/cone2cone/extract_direct_cones.py
```diff
@@ -0,0 +1,460 @@
...
+            return False
+        if problem.is_mixed_integer():
+            return False
+        if problem.x_cones:
+            return False
+        if not self._kinds:
+            return False
+        return True
+
+    def apply(self, problem):
```
PTNobel, line 142: This function is way too complex

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
```diff
@@ -20,12 +20,26 @@
...
+from cvxpy.constraints import SOC, ExpCone, PowCone3D, PowConeND, SvecPSD
 from cvxpy.reductions.solution import Solution, failure_solution
 from cvxpy.reductions.solvers import utilities
 from cvxpy.reductions.solvers.conic_solvers.conic_solver import ConicSolver
 from cvxpy.reductions.solvers.solver_inverse_data import SolverInverseData
 from cvxpy.utilities.citations import CITATION_DICT
+from cvxpy.utilities.psd_utils import TriangleKind
+
+
+def _moreau_supports_x_cones() -> bool:
```
PTNobel: We can assume this unconditionally since we're bumping the min to 0.4.0

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
```diff
@@ -152,6 +194,11 @@ def invert(self, solution, inverse_data):
             dual_vars = {}
             dual_vars.update(eq_dual_vars)
             dual_vars.update(ineq_dual_vars)
+            # x_cone duals (computed in solve_via_data from the KKT
+            # residual q + A.T z) are keyed by their original
+            # constraint id and slot in directly.
+            xcone_duals = getattr(solution, 'x_cone_duals', None) or {}
```
PTNobel: Access unconditionally

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
```diff
@@ -251,6 +298,21 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
         # Convert cone dimensions
         cones = dims_to_solver_cones(data[ConicSolver.DIMS])
 
+        # Direct-x cones from ExtractIdentityCones: subvectors of the
+        # primal variable that ride the XConeSpec path instead of
+        # being slack-side cones.
+        x_cones_meta = data.get('x_cones', []) or []
```
PTNobel: What is the or []

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
```diff
@@ -116,6 +135,29 @@ def supports_quad_obj(self) -> bool:
         """Moreau supports quadratic objective with conic constraints."""
         return True
 
+    def x_cone_kinds(self) -> frozenset:
```
PTNobel: This should just be a constant for consistency with the rest of the codebase

cvxpy/tests/test_conic_solvers.py
PTNobel, line 1: These tests feel insane and should be way shorter

.github/workflows/test_optional_solvers.yml
rileyjmurray: Agreed

cvxpy/reductions/cone2cone/extract_identity_cones.py
```diff
@@ -0,0 +1,460 @@
...
+        # Only kinds we know how to detect identity-pattern for.
+        self._kinds = kinds & {
+            'nonneg', 'soc', 'psd_triangle', 'exp', 'power', 'gen_power',
+        }
+
+    def accepts(self, problem) -> bool:
+        if not isinstance(problem, ParamConeProg):
+            return False
+        if problem.is_mixed_integer():
+            return False
```
rileyjmurray: This restriction seems unnecessary.

cvxpy/reductions/cone2cone/extract_direct_cones.py
rileyjmurray: 350 lines. Yikes! Helper functions are definitely in order. My weak preference would be to define a free function for each type of cone we want to identify, then register those functions in this class. The implementation of `apply` would iterate over the registered extractors. The API of the free function can include whatever metadata is necessary so the overall apply function only does one pass over the data.

cvxpy/reductions/cone2cone/extract_identity_cones.py
```diff
@@ -0,0 +1,460 @@
...
+            return problem, _NoopInverse()
+        sort_order = np.lexsort((row_of_k[candidate_k],
+                                 block_id_of_k[candidate_k]))
+        sorted_ks = candidate_k[sort_order]
+        sorted_block_ids = block_id_of_k[sorted_ks]
+        sorted_rows = row_of_k[sorted_ks]
+        sorted_cols = col_of_k[sorted_ks]
+        boundaries = np.searchsorted(
+            sorted_block_ids, np.arange(len(candidates) + 1)
+        )
```
rileyjmurray: Superlinear complexity operations in this kind of task make me uncomfortable. The current justification references a magic number `300+` for some reason.

cvxpy/reductions/cone2cone/extract_direct_cones.py
```diff
@@ -0,0 +1,460 @@
...
+        )
+
+        inv = {
+            'extracted': extracted,
+            'kept_M_idx': kept_M_idx,
+            'orig_constr_ids': [c.id for c in constraints],
+        }
+        return new_pcp, inv
+
+    def invert(self, solution, inverse_data):
```
rileyjmurray, line 195: All codepaths of this function just return `solution`. Remove the branching. Also, it seems all codepaths of `apply` return `_NoopInverse` as the second entry in a tuple. So it seems that `_NoopInverse` can be removed entirely.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
```diff
@@ -207,6 +208,16 @@ def __init__(self, q, x, A,
...
         self.formatted = formatted
 
+        # Direct-x cones: a list of (kind, indices, original_constr_id)
+        # tuples naming subvectors of ``x`` that are constrained to lie
+        # in a cone of the named kind ('nonneg', 'soc', ...).  Set by
+        # reductions that extract identity-pattern cone blocks (see
+        # cvxpy/reductions/cone2cone/extract_identity_cones.py); the
+        # solver consumes these alongside the slack-side cone_dims.
+        # ``None`` (default) means no x_cones — the slack-side cone
+        # interpretation is the entire story.
```
rileyjmurray: This verbose description could be obviated by a more informative type annotation.

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
rileyjmurray: Agreed, especially since Moreau 0.4.0 means we can eliminate branching and just set the constant to `{'nonneg', 'soc', 'psd_triangle', 'exp', 'power', 'gen_power'}`.

cvxpy/tests/test_conic_solvers.py
rileyjmurray: I'd prefer that direct cone tests go in another file. We shouldn't tie these tests to Moreau just because it's the only solver currently supporting this reduction.

I'd be fine with updating the standard test problems to make sure there are detectable direct cones.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -908,10 +914,214 @@ def test_moreau_variable_soc_dims(self) -> None:
...
+        ]
+        all_soc_sizes = list(soc_dims) + x_cone_socs
+        self.assertTrue(any(d > 3 for d in all_soc_sizes))
+
+    def _skip_if_no_xcones(self) -> None:
+        from cvxpy.reductions.solvers.conic_solvers.moreau_conif import (
+            _moreau_supports_x_cones,
+        )
+        if not _moreau_supports_x_cones():
+            self.skipTest("installed Moreau build lacks XConeSpec")
```
rileyjmurray: This can be removed entirely. (Moreau 0.4.0 lower bound.)

rileyjmurray (changes requested):
Left a bunch of comments.

One comment not attached to edited file: we need to update the table of supported cones for Moreau that appears in https://www.cvxpy.org/tutorial/solvers/index.html. (I trust the agent to figure out what needs to be edited in-repo for that to happen.)

cvxpy/tests/test_conic_solvers.py
rileyjmurray, line 1: Moreau's tests should be updated to include the standard test SDPs. See, e.g., the Clarabel tests.

PTNobel: Fair; we might want to bring Mosek onto this instead of dualize?

rileyjmurray: I'm sure the Mosek devs would love it if that were the default. We wouldn't want to take away the dualization option though.

tools/bench_moreau_extract.py
PTNobel: Removed `tools/bench_moreau_extract.py` from the PR.

Updated in eb128742a.

.github/workflows/test_optional_solvers.yml
PTNobel: Removed the separate Moreau install step. The existing `uv sync --all-extras --no-extra uno --dev` now installs it through the `MOREAU` extra, which requires Moreau >=0.4.0.

Updated in eb128742a.

cvxpy/reductions/cone2cone/extract_identity_cones.py
PTNobel: Removed `_NoopInverse` and the unused inverse metadata. `apply` now returns `None` for inverse data, and `invert` just returns the solution because the solver has already restored the original constraint IDs.

Updated in eb128742a.

cvxpy/reductions/cone2cone/extract_direct_cones.py
PTNobel: Split detection into one free extractor per cone family, registered in `ExtractIdentityCones.EXTRACTORS`. Shared helpers handle the single structural scan and sparse row remapping. `apply` now coordinates those helpers in about 50 lines.

Updated in eb128742a.

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
PTNobel: Removed `_moreau_supports_x_cones` and the runtime capability probe. The interface now assumes the Moreau >=0.4.0 API.

Updated in eb128742a.

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
PTNobel: Changed this to unconditional `solution.x_cone_duals` access; the solution wrapper receives that dictionary on every solve path.

Updated in eb128742a.

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
PTNobel: Removed the fallback. `ParamConeProg.x_cones` is always a list, `MOREAU.apply` always supplies `data['x_cones']`, and `solve_via_data` reads it directly.

Updated in eb128742a.

cvxpy/reductions/solvers/conic_solvers/moreau_conif.py
PTNobel: Replaced the method with the `X_CONE_KINDS` class constant. Moreau advertises all six families unconditionally; the base Solver constant is empty.

Updated in eb128742a.

cvxpy/tests/test_conic_solvers.py
PTNobel: Moved extraction coverage into `test_extract_identity_cones.py`, with shared cases for the six cone families, batched cones, structural rejection, DPP updates, and integer variables. These tests do not require Moreau; reconstructed problems use CLARABEL. `TestMoreau` retains the standard solver tests and focused interface regressions for PSD scaling, quadratic dual recovery, and mixed slack ordering. A future MOSEK integration can use this reduction while retaining the dualization option.

Updated in eb128742a.

cvxpy/reductions/cone2cone/extract_identity_cones.py
PTNobel: Removed the mixed-integer restriction. Extraction preserves the variable and its integer indices; the solver remains responsible for advertising MIP support. Added a solver-independent test for this.

Updated in eb128742a.

PTNobel:
Addressed the review in eb128742a and merged current master in 0f7450c28. Replies are posted on each review thread.

@rileyjmurray, the Moreau row in `doc/source/tutorial/solvers/index.rst` now advertises SDP support. The standard SDP tests caught a PSD scaling bug, which is fixed along with the missing quadratic term in direct-cone dual recovery.

Validation:
- 71 extraction, parameter-program, and DPP tests passed, plus 14 subtests.
- All 26 Moreau tests passed on the local 0.4.0-beta.1 CPU build; none skipped. All 30 CLARABEL tests also passed.
- Pre-commit passed, including workflow and pyproject validation.
- Seven failures from a broader cone-to-cone run reproduced on the pre-change commit (six COPT capability failures and one missing mixed-integer conic solver); the known ECOS_BB test was skipped.

The temporary benchmark script is removed. A small local comparison of extraction on 100/1,000/5,000 nonnegative blocks measured 3.93/26.47/187.27 ms before and 4.70/26.92/168.61 ms after (median of five runs, excluding stuffing). The full benchmark checklist remains unchecked.

The configured package index currently has Moreau only through 0.3.3. The declared >=0.4.0 dependency is retained as requested; release-based CI needs that version to be published. CPU validation here used the local beta build, not the unavailable final release.

CI confirms the release blocker: the pyright, SCIPY backend, derivative, and Ubuntu optional-solver jobs all stop during `uv` dependency resolution because `moreau>=0.4.0` is unavailable; they do not reach their checks. See the [pyright job log](https://github.com/cvxpy/cvxpy/actions/runs/33930437624/job/101207750679).

cvxpy/reductions/cone2cone/extract_identity_cones.py
PTNobel: Removed `lexsort`, `searchsorted`, and `np.unique`, along with the magic-number justification. The structural scan uses `bincount` and direct row-indexed lookups; integer-index sets handle within-cone uniqueness and overlap. The tensor is scanned once, with no sorting or whole-tensor scan per cone. The new tests also reject blocks with a repeated/missing row, which the old endpoint check could miss.

Updated in eb128742a.

cvxpy/reductions/cone2cone/extract_direct_cones.py
PTNobel: Removed both the sentinel and the unused success-path inverse dictionary. All `apply` paths return `None` as inverse data, and `invert` unconditionally forwards `solution`.

Updated in eb128742a.

cvxpy/reductions/dcp2cone/cone_matrix_stuffing.py
PTNobel: Added an `XCone` NamedTuple with named kind, indices, constraint ID, and extras fields, plus a typed extras dictionary. `x_cones` is now annotated as `list[XCone]`; the verbose and outdated tuple description is gone.

Updated in eb128742a.

cvxpy/tests/test_conic_solvers.py
PTNobel: Removed `_skip_if_no_xcones` and every call to it. Moreau tests now rely on the declared >=0.4.0 dependency.

Updated in eb128742a.

cvxpy/tests/test_conic_solvers.py
PTNobel: Added the standard SDP minimum, maximum, SDP2, and batched SDP tests. They exposed a real PSD coordinate-scaling bug: Moreau expects scaled svec variables, whereas CVXPY stores unscaled symmetric entries. The interface now transforms P, q, and A into solver coordinates and maps the primal solution back. All four standard SDP tests pass, including complementarity and dual-domain checks; a quadratic SDP regression also checks the recovered dual numerically.

Updated in eb128742a.

cvxpy/tests/test_conic_solvers.py
```diff
@@ -908,11 +908,78 @@ def test_moreau_variable_soc_dims(self) -> None:
...
+        X = cp.Variable((2, 2), symmetric=True)
+        target = np.array([[1., 2.], [2., 1.]])
+        constraint = X >> 0
+        problem = cp.Problem(
+            cp.Minimize(cp.sum_squares(X) - 2 * cp.sum(cp.multiply(target, X))),
+            [constraint],
+        )
+        self.assertAlmostEqual(problem.solve(solver=cp.MOREAU), -9., places=4)
+        self.assertItemsAlmostEqual(X.value, np.full((2, 2), 1.5), places=4)
+        self.assertItemsAlmostEqual(constraint.dual_value, np.array([[1, -1], [-1, 1]]), places=4)
```
rileyjmurray, line 951: Strictly speaking, these aren't conic problems in the way meant by the rest of the file.

rileyjmurray (commented):
Looks pretty good! I have one minor comment on test placement that you can ignore. 

Although, before we pass the point of no return I'd like to discuss naming choices.

Do you have objections to `DirectCone` instead of `XCone`? Variable names like `x_cones` could become `dir_cones`.

Similarly, might you be up for renaming `extract_identity_cones.py` to `extract_direct_cones.py`, or even `extract_direct_cone_constraints.py`?

cvxpy/tests/test_conic_solvers.py
PTNobel: Say more?

rileyjmurray: This test file was originally written only to test "standard" linear conic solver functionality. I might have made some test problem that involves a nonlinear objective, but that was with the expectation that those problems would be canonicalized into linear cone programs. 

To the extent that the distinction matters this is really a call to restructure some of our testing files. Could you open an issue making that a TODO? If tests are well-organized then agents know were to look when it comes time to write more.

PTNobel:
> Do you have objections to DirectCone instead of XCone? Variable names like x_cones could become dir_cones.

> Similarly, might you be up for renaming extract_identity_cones.py to extract_direct_cones.py, or even extract_direct_cone_constraints.py?

Good with both. Should I rename in Moreau too?

rileyjmurray:
> Good with both. Should I rename in Moreau too?

I think that'd be great!

rileyjmurray (approved):
Autobots, let's roll out!

PTNobel:
Let's release Moreau 0.4.0 and get CI passing

PR #3500, opened 2026-08-27
Give cvar the same axis behavior as var

nicoloangileri:
`cvar` now takes `axis` and `keepdims`, following #3489 item 1.

The restriction was only ever in the wrapper. `cvar` is `sum_largest(x, k) / k`,
and `sum_largest` has been an `AxisAtom` with `axis` and `keepdims` for a while,
so the 1-d check was refusing inputs the machinery underneath already handled.
Both arguments now pass straight through. An array with `axis=None` is one
distribution over all of its entries rather than a `ValueError` — the same thing
`var` does.

`k` is the part that needed care: it is `(1 - beta) * m`, with `m` the number of
samples reduced rather than `len(x)`. That is the same count `mean`, `std` and
`var` divide by, so `_axis_size` moved out of `stats.py` to sit next to
`normalize_axis` in `axis_atom.py`, where `cvar` can reach it without a third
copy of four lines. Say the word if you'd rather I duplicate it and leave
`stats.py` alone.

Checked against the one-dimensional atom applied slice by slice, over every
single axis and every pair of a (5, 4, 3) array, with and without `keepdims`:
shapes match the corresponding numpy reduction and values agree to 1e-9.
`beta = 0` keeps every sample and so must return the mean along the axis, which
pins the sample count down independently of `sum_largest`. The vector case is
unchanged and still matches the Rockafellar-Uryasev program. The axis form
canonicalizes through the same path as the vector one, with no backend fallback.

In the docs, `cvar` and `sum_largest` join the list of functions that can be
applied along an axis — `sum_largest` was already missing from it.

You mentioned wanting to talk to Steven before this is merged; that still
stands as far as I'm concerned, this is just the implementation to look at.
## Type of change
- [ ] Bug fix

- [x] Check that your code adheres to our coding style. — `ruff` 0.8.1 (the pre-commit pin) reports no findings, `pyright` none on the three changed modules
- [x] Write unittests. — `test_cvar_axis`: shapes against a numpy reduction, values against the 1-d atom slice by slice, tuple and negative axes, `beta = 0` against the mean, and a solve through the axis form
- [x] Run the unittests and check that they're passing. — 2586 passed against 2585 on the merge base, with `--collect-only` diffed: one test added, none lost
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally. `cvar` only gains keyword arguments and the vector path builds the same expression as before, so existing models are unaffected; the CI benchmark job will confirm.

Note: the docs build is not part of PR CI — `docs.yml` is `workflow_dispatch` only — so I ran it
locally instead. The full sphinx build produces the same 412 warnings as master, none of them from
the two lines this PR touches.

### DIFF
--- a/cvxpy/atoms/axis_atom.py
+++ b/cvxpy/atoms/axis_atom.py
@@ -40,6 +40,14 @@ def normalize_axis(
         return axes
 
 
+def axis_size(x, axis: None | int | tuple[int, ...] = None) -> int:
+    """Return the number of entries reduced by an axis argument."""
+    if axis is None:
+        return x.size
+    axes = normalize_axis_tuple(axis, len(x.shape), "axis")
+    return int(np.prod([x.shape[a] for a in axes]))
+
+
 class AxisAtom(Atom):
     """
     An abstract base class for atoms that can be applied along an axis.
--- a/cvxpy/atoms/cvar.py
+++ b/cvxpy/atoms/cvar.py
@@ -14,11 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from cvxpy.atoms.axis_atom import axis_size
 from cvxpy.atoms.sum_largest import sum_largest
 from cvxpy.expressions.expression import Expression
 
 
-def cvar(x, beta) -> Expression:
+def cvar(x, beta, axis=None, keepdims=False) -> Expression:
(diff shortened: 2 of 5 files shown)


PTNobel:
Thanks for opening this PR. Once we get the next patch out, I'll review it

PR #3502, opened 2026-08-28
Support tuple axes in std

nicoloangileri:
Refs #3489, item 2  the `std` half, following your "happy to do different behavior on std and norm even if that adds complexity to std". It does not touch `norm`.

A tuple of axes pools the entries of every axis it names into one sample set per remaining position. `norm` takes a single axis, so the pooled axes are moved to the front with `moveaxis`, flattened into one in Fortran order, reduced with `norm(..., 2, axis=0)`, and the result folded back into the shape the pooled axes left behind. That is the same alignment `geo_mean` has used since #3493, order included.

`std` is the one atom here that cannot borrow `var`'s route: `var` is `sum_squares` over the axes and takes tuples already, but the square root of a convex expression is not DCP, which is why `std` goes through `norm` in the first place.

Checked against numpy over `(0,1)`, `(1,2)`, `(0,2)`, the one-element `(0,)`, the full `(0,1,2)` and the negative `(-3,-1)`, each crossed with `keepdims` and `ddof` in {0, 1}: shapes and values all agree. The `int` and `None` paths are untouched and still match. The tuple form is DCP and solves.

Note this **drops an existing assertion** in `test_stats` that `cp.std(a, axis=(0, 1))` raises `ValueError`  it was pinning the gap this closes.

No new backend fallback: `std(X, axis=0)` on a three-dimensional variable already falls back to SCIPY, because the CPP backend does not take expressions past two dimensions, so the reshape adds nothing there.

## Type of change
- [ ] Bug fix

- [x] Check that your code adheres to our coding style. — `ruff` 0.8.1 (the pre-commit pin) reports no findings, `pyright` none
- [x] Write unittests. — extends `test_stats` to tuple axes, `keepdims` and `ddof`, plus a DCP and solve check
- [x] Run the unittests and check that they're passing. — 2585 passed, same as the merge base
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. — not run locally. Only the tuple branch is new; `axis=None` and integer axes build exactly the expression they built before, so existing models are unaffected. The CI benchmark job will confirm.

### DIFF
--- a/cvxpy/atoms/stats.py
+++ b/cvxpy/atoms/stats.py
@@ -19,7 +19,9 @@ import numbers
 import numpy as np
 from numpy.lib.array_utils import normalize_axis_tuple
 
+from cvxpy.atoms.affine.reshape import reshape
 from cvxpy.atoms.affine.sum import sum as cvxpy_sum
+from cvxpy.atoms.affine.transpose import moveaxis
 from cvxpy.atoms.norm import norm
 from cvxpy.atoms.sum_squares import sum_squares
 from cvxpy.expressions.expression import Expression
@@ -48,11 +50,24 @@ def std(x, axis=None, keepdims=False, ddof=0) -> Expression:
     """
     if axis is None:
         return norm((x - mean(x)).flatten(order='F'), 2) / np.sqrt(x.size - ddof)
-    elif isinstance(axis, numbers.Integral):
-        return norm(x - mean(x, axis, True), 2, axis=axis, keepdims=keepdims) \
-                / np.sqrt(_axis_size(x, axis) - ddof)
-    else:
-        raise ValueError("cp.std doesn't yet support tuple axis values.")
+
+    centered = x - mean(x, axis, True)
+    scale = np.sqrt(_axis_size(x, axis) - ddof)
+    if isinstance(axis, numbers.Integral):
+        return norm(centered, 2, axis=axis, keepdims=keepdims) / scale
+
+    # A tuple of axes pools the entries of every axis it names, which is one
+    # vector per remaining position. norm takes a single axis, so the pooled
+    # axes are moved to the front and flattened into one, in Fortran order,
+    # and the result is folded back into the shape those axes left behind.
+    axes = normalize_axis_tuple(axis, x.ndim, "axis")
+    kept = tuple(d for d in range(x.ndim) if d not in axes)
+    moved = moveaxis(centered, axes, range(len(axes)))
+    pooled = norm(reshape(moved, (_axis_size(x, axis), -1), order='F'), 2, axis=0) / scale
+    out_shape = tuple(x.shape[d] for d in kept)
+    if keepdims:
+        out_shape = tuple(1 if d in axes else x.shape[d] for d in range(x.ndim))
+    return reshape(pooled, out_shape, order='F')
 
 
 def var(x, axis=None, keepdims=False, ddof=0) -> Expression:
--- a/cvxpy/tests/test_atoms.py
+++ b/cvxpy/tests/test_atoms.py
@@ -2072,15 +2072,22 @@ class TestAtoms(BaseTest):
                 assert np.allclose(a.mean(axis=axis, keepdims=keepdims), expr_mean.value)
                 assert np.allclose(a.var(axis=axis, keepdims=keepdims), expr_var.value)
 
-        for axis in [0, 1, 2]:
+        for axis in [0, 1, 2, (0, 1), (1, 2), (0, 2), (0,), (0, 1, 2), (-3, -1)]:
             for keepdims in [True, False]:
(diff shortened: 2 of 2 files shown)


curious7-web:
Hey, Parth and I had looked at this same moveaxis/reshape route when #3496 was being reviewed. I’ve since looked through the #3489 discussion too, and this PR is already cleanly scoped to the std side. I’m going to explore native tuple-axis support in norm/pnorm separately, so no reason for this to wait on that.

PTNobel (approved):
Looks great! Thank you

PTNobel:
I think we need a second review because this is an API change

nicoloangileri:
Thanks  and good that you're taking norm/pnorm, that was the half I
deliberately left alone.

One thing I measured while scoping it, in case it saves you the detour:
numpy reads a tuple axis on norm as a *matrix* norm over those two axes,
not as the flattened vector norm. On a 3x4: 'fro' gives 2.0518 (the
flattened 2-norm), but 2 gives 1.5749 (sigma_max), 'nuc' the sum of the
singular values, and 1 and inf the max column and row sums. Any other p
raises "Invalid norm order for matrices". A 1-tuple is treated like an int,
and three or more axes raises.

So the tempting path — making p=2 with a tuple the flattened vector norm —
would quietly disagree with numpy, which is why std here goes its own way
instead of sitting on norm. Being numpy-faithful, 'fro' is a reshape and
1/inf are moderate, but 2 and 'nuc' need batched sigma_max and a batched
nuclear norm over a stack of matrices: new canonicalization rather than a
wrapper change. PTNobel's take on #3489 was that different behaviour on std
and norm is fine.

Also worth knowing before you start: norm(X, 'fro', axis=0) used to return
the Frobenius norm of the whole array as a scalar, silently. That's fixed
in #3501  it raises now.

curious7-web:
Thanks, this is really useful. Divyansh and I are coordinating on the norm/pnorm follow-up now, and he’s already traced the p=2/SOC path and found an existing N-D integer-axis failure there too. I’ll point him to this note.... should save us some wandering on the semantics side.

nej1gotnochill:
Thanks a ton @nicoloangileri this clears the semantics alot I was initially thinking axis `p=2` case in terms of flattening into an SOC reduction but the 2 axis case actually maps to matrix norms so `p=2` would need the batched `sigma_max` path instead. I will trace the existing `sigma_max`  and nuclear norm canonicalization and see how the batched case could fit into it.

Also thanks to @curious7-web for helping me work through the canonicalization side and pointing me in the right direction here, really appreciate the guidance on the same !

nej1gotnochill:
Hello @nicoloangileri @curious7-web  I ran into a semantic conflict while preparing the tuple axis PR. PR #3501 currently has a test asserting that `norm(X , 'fro'/'nuc' , axis = (0,1))` raises a `Value Error` whereas the tuple axis semantics we have been implementing treat a 2 tuple of axes as matrix norm semantics. For a 2D input, (0,1) reduces the whole matrix to a scalar matrix norm. 

So my current implementation and the existing #3501 test are enforcing semantics for the same call. I did not want to modify the merged PR's tests or make an arbitrary restriction in my branch without checking with you both first. Which behaviour do you think should be the intended CVXPY semantics? Once these are settled I will align the PR and its tests accordingly. Thank you , have a nice day !

nicoloangileri:
Thanks for checking first. The `(0, 1)` case in that test was a guard, not a decision about the semantics: when #3501 went in, `norm` had no tuple-axis support, and before it `norm(X, 'fro', axis=...)` quietly returned the norm of the whole array, so raising for every axis was the safe choice. The comment above the loop is about single axes.

Once tuple axes are supported, I think we should follow NumPy:

- `'fro'`/`'nuc'` with an int axis or a 1-tuple: keep raising, as now.
- A 2-tuple: the matrix norm over those two axes, batched over the remaining ones. `(0, 1)` and `(1, 0)` give the same value.
- Three or more axes: raise.

So feel free to move `(0, 1)` out of the raising loop in `test_norm_exceptions` and into a positive check against `np.linalg.norm`, keeping `0` and `1` there. If batched `'nuc'` (e.g. `axis=(1, 2)` on a 3-D input) is out of scope for your PR, a clear `NotImplementedError` for that case seems better than restricting the tuple semantics.

@PTNobel, does this match what you had in mind?

nej1gotnochill:
Thank you Nicolo ! I had interpreted the (0,1) case in #3501 as an intentional semantic restriction but your clarification about it being a guard for the pre tuple axis behavior clears that up. 

I will keep the current tuple axis implementation following the NumPy semantics: 2 tuples represent matrix norms over those two axes while single axes continue to raise for `'fro' / 'nuc'` and 3+axes remain unsupported. I will also update the (0,1) test accordingly and keep the batched `'nuc'` limitation explicit rather than restricting the general tuple axis semantics. Thank you once again for the clarification , have a nice day !

PR #3505, opened 2026-09-02
Add dual variables to ConstantSolver for feasible constant problems

nej1gotnochill:
This adds dual variable values to `ConstantSolver` for feasible constant problems. The solver now returns zero valued duals keyed by constraint ID and matching each constraint's shape so `constraint.dual_value` is populated instead of remaining none. Infeasible constant problems retain the existing behavior. 

The change is minimal: 4 lines in constant_solver.py plus regression coverage for inequality , equality , vector and infeasible cases.

### DIFF
--- a/cvxpy/reductions/solvers/constant_solver.py
+++ b/cvxpy/reductions/solvers/constant_solver.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 import cvxpy.settings as s
 from cvxpy.reductions.solution import Solution
 from cvxpy.reductions.solvers.solver import Solver
@@ -32,7 +34,8 @@ class ConstantSolver(Solver):
 
     def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
         if all(c.value() for c in problem.constraints):
-            return Solution(s.OPTIMAL, problem.objective.value, {}, {}, {})
+            dual_vars = {c.id: np.zeros(c.shape) for c in problem.constraints}
+            return Solution(s.OPTIMAL, problem.objective.value, {}, dual_vars, {})
         else:
             return Solution(s.INFEASIBLE, None, {}, {}, {})
 
--- a/cvxpy/reductions/solvers/constant_solver.py
+++ b/cvxpy/reductions/solvers/constant_solver.py
@@ -1,8 +1,28 @@
 import numpy as np
 
 import cvxpy.settings as s
+from cvxpy.constraints.second_order import RSOC, SOC
 from cvxpy.reductions.solution import Solution
 from cvxpy.reductions.solvers.solver import Solver
+
+
+def _zero_dual(constr):
+    """Return a zero dual in the representation used in ``Solution.dual_vars``.
+
+    Mirrors the packed representations real solvers produce via
+    ``utilities.get_dual_values`` / ``extract_dual_value`` and the reshape step
+    in ``cone_matrix_stuffing.ConeMatrixStuffing.invert``: SOC duals stay flat with
+    size equal to the packed cone size, RSOC duals use the ``[dt, du, dX]``
+    list that ``RSOC.save_dual_value`` consumes, and all other constraints use a
+    shape-matching array.
+    """
+    if isinstance(constr, SOC):
+        return np.zeros(constr.size)
+    if isinstance(constr, RSOC):
+        n_cones = constr.args[1].size
+        n_x = constr.args[2].size // n_cones
+        return [np.zeros(n_cones), np.zeros(n_cones), np.zeros((n_cones, n_x))]
+    return np.zeros(constr.shape)
 
 
 class ConstantSolver(Solver):
@@ -34,7 +54,7 @@ class ConstantSolver(Solver):
 
     def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
         if all(c.value() for c in problem.constraints):
-            dual_vars = {c.id: np.zeros(c.shape) for c in problem.constraints}
+            dual_vars = {c.id: _zero_dual(c) for c in problem.constraints}
             return Solution(s.OPTIMAL, problem.objective.value, {}, dual_vars, {})
         else:
             return Solution(s.INFEASIBLE, None, {}, {}, {})
--- a/cvxpy/tests/test_problem.py
+++ b/cvxpy/tests/test_problem.py
@@ -2373,3 +2373,37 @@ class TestProblem(BaseTest):
         result = reduction.invert(sol, inv_data)
         assert result.dual_vars is None
 
+    def test_constant_solver_duals(self) -> None:
+        """Test that ConstantSolver returns zero dual values for feasible problems."""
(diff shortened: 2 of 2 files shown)


cvxpy/tests/test_problem.py
```diff
@@ -2373,3 +2373,37 @@ def test_canonicalization_invert_none_duals(self) -> None:
...
+        prob.solve()
+        self.assertEqual(prob.status, s.OPTIMAL)
+        self.assertEqual(prob.constraints[0].dual_value, 0.0)
+
+        # Multiple constraints
+        prob = Problem(cp.Minimize(1), [2 >= 1, 3 >= 2])
+        prob.solve()
+        self.assertEqual(prob.status, s.OPTIMAL)
+        for c in prob.constraints:
+            self.assertEqual(c.dual_value, 0.0)
```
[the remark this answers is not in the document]
PTNobel: We should definitely have some Parameters in here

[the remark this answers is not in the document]
nej1gotnochill: Sure , i will update the tests to use explicit `cp.Constant` inequalities and add some `Parameter` cases to make sure we are actually exercising the ConstantSolver path. I will also check the occurence around the line 2404.

cvxpy/reductions/solvers/constant_solver.py
```diff
@@ -32,7 +34,8 @@ def solve_via_data(self, data, warm_start: bool, verbose: bool, solver_opts, sol
 
     def solve(self, problem, warm_start: bool, verbose: bool, solver_opts):
         if all(c.value() for c in problem.constraints):
-            return Solution(s.OPTIMAL, problem.objective.value, {}, {}, {})
+            dual_vars = {c.id: np.zeros(c.shape) for c in problem.constraints}
+            return Solution(s.OPTIMAL, problem.objective.value, {}, dual_vars, {})
         else:
             return Solution(s.INFEASIBLE, None, {}, {}, {})
```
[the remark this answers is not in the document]
PTNobel: Uhh this doesn't feel right? How do other solvers handle dual variables? Isn't there a helper function on the Solver class?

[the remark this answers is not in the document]
nej1gotnochill: Ah yea , I see what you mean. I'll look at how the other solvers handle dual variables and check the existing Solver helper before changing this further. Thanks for catching that and correcting me !

cvxpy/reductions/solvers/constant_solver.py
```diff
@@ -1,8 +1,30 @@
...
+from cvxpy.constraints.second_order import RSOC, SOC
 from cvxpy.reductions.solution import Solution
 from cvxpy.reductions.solvers.solver import Solver
 
 
+def _zero_dual(constr):
+    """Return a zero dual in the representation used in ``Solution.dual_vars``.
+
+    Mirrors the packed representations real solvers produce via
+    ``utilities.get_dual_values`` / ``extract_dual_value`` and the reshape step
```
PTNobel, line 13: Can't we just use those helpers so we don't have to special case the different dual formats here?

cvxpy/tests/test_problem.py
```diff
@@ -2373,3 +2373,64 @@ def test_canonicalization_invert_none_duals(self) -> None:
...
+        self.assertEqual(prob.constraints[0].dual_value, 0.0)
+
+        # Equality constraint
+        prob = Problem(cp.Minimize(1), [cp.Constant(1) == cp.Constant(1)])
+        prob.solve()
+        self.assertEqual(prob.status, s.OPTIMAL)
+        self.assertEqual(prob.constraints[0].dual_value, 0.0)
+
+        # Satisfied multiple constraints
+        prob = Problem(cp.Minimize(1), [cp.Constant(2) >= cp.Constant(1), cp.Constant(3) >= cp.Constant(2)])
```
PTNobel: Please install and run the pre commit

cvxpy/reductions/solvers/constant_solver.py
nej1gotnochill: I looked into those helpers. They rely on a solver result vector plus solver specific `parse_func ` function which ConstantSolver doesnt have so we would have to construct a synthetic result vector just to represent zeros. I also checked the SOC / RSOC representations which is why I kept the small local helper rather than using `np.zeros(c.shape)` for everything. That being said if you think there is cleaner way to reuse the existing helpers or structure it then I would definitely appreciate your feedback.

cvxpy/tests/test_problem.py
nej1gotnochill: Surely I will install and run the pre commit checks and addresss any issue they flag.

nej1gotnochill:
I have now installed pre-commit and ran it on the changed test file. Ruff passes and the other hooks were skipped because they don't apply to the Python file,
Screenshot 2026-09-05 165941

cvxpy/reductions/solvers/constant_solver.py
PTNobel: Hmm okay. It is possible that since this is happening further up the chain then I'm used to that the normal approach isn't available

PTNobel: I'm traveling for Janmashtami, will look into it later

nej1gotnochill: > I'm traveling for Janmashtami, will look into it later

Sure no worries , have a safe journey and Happy Janamasthmi !

nej1gotnochill: > Hmm okay. It is possible that since this is happening further up the chain then I'm used to that the normal approach isn't available

Yeah I feel that's the case. I think thats what I was running into as well since ConstantSolver is handling this further up the reduction chain , it doesnt have the usual solver result vector / parse step that the existing helpers expect. I will leave the current approach as it is for now and see if there is a better abstraction you have in mind. Thank you , have a nice day !

nej1gotnochill:
Hello Parth, just following up on this when you get a chance. I have pushed the pre commit fixes and everything is passing on my side. I also left the ConstantSolver dual helper approach as discussed , would appreciate your thoughts on whether you would like me to change anything else here. Thank you , have a nice day!

PTNobel:
Sorry I'm away till the weekend; probably don't have time to sit at a personal computer till after that

nej1gotnochill:
> Sorry I'm away till the weekend; probably don't have time to sit at a personal computer till after that

No worries , Enjoy your time away and whenever you get a chance after the weekend is completely fine. Thank you and have a nice day !

nej1gotnochill:
Hi Parth , if you feel any additions or iterastions are needed do let me know as any form of feedback will help me contribute more meaningfully to CVXPY !

cvxpy/reductions/solvers/constant_solver.py
PTNobel, line 1: Implementation looks good

PTNobel (approved):

PR #3506, opened 2026-09-02
cvxcore: Regenerate files using SWIG 4.5.0

stertingen:
## Description
Update generated cvxcore files to investigate freethreading support. According to official docs, SWIG 4.4 added support for freethreaded python.
Issue link (if applicable): #3110

## Type of change
- [ ] Bug fix

- [x] Add our license to new files. --> no new files added
- [x] Check that your code adheres to our coding style. --> pre-commit run did not complain
- [ ] Write unittests. --> not applicable I guess
- [x] Run the unittests and check that they’re passing. --> passed with warnings

Besides regenerating the files, `numpy.i` was adjusted according to https://github.com/swig/swig/blob/v4.5.0/CHANGES#L772-L794


PTNobel:
Isn't numpy.i vendored? Can we just get a new version from upstream?

Also, generally, as a matter of policy for very large autogenerated diffs, we insist on running the code generation instead to simplify auditing.

Would you consider posting instructions on how you generated this code so we can recreate?

https://numpy.org/doc/2.5/reference/swig.interface-file.html

stertingen:
> Isn't numpy.i vendored? Can we just get a new version from upstream?
> 
> Also, generally, as a matter of policy for very large autogenerated diffs, we insist on running the code generation instead to simplify auditing.
> 
> Would you consider posting instructions on how you generated this code so we can recreate?
> 
> https://numpy.org/doc/2.5/reference/swig.interface-file.html

I was just about to explain my recent changes to this PR; I copied the upstream's `numpy.i` as recommended by the official Numpy docs you referred to. (https://github.com/numpy/numpy/blob/main/tools/swig/numpy.i)

stertingen:
Going down the rabbit hole on enabling threads for cvxcore, there is more work to do than just updating swig:

- [ ] Check how python threads interact with OpenMP threads and OpenMP APIs (Building cvxcore with OpenMP is currently disabled and could be re-enabled in the future)
- [ ] Check how python threads interact with Eigen (https://libeigen.gitlab.io/eigen/docs-3.4/TopicMultiThreading.html)
- [ ] Generate files with `-nogil` to declare freethreading support (*after* above points have been checked!)

While freethreading support was the driving force to re-generate cvxcore files, there is more work to do which is beyond the scope of this PR.
Since updating SWIG and `numpy.i` does not hurt, I remove the 'draft' state from this PR, although there is more 
to do for actual freethreading support.

PTNobel:
I will say, I think we're likely to replace cvxcore with diff engine for 1.10 or 1.11.

stertingen:
> I will say, I think we're likely to replace cvxcore with diff engine for 1.10 or 1.11.

Well, that makes ambitions for multithreading cvxcore pretty low I guess.  Will it be removed completely or just deprecated?

Transurgeon:
> > I will say, I think we're likely to replace cvxcore with diff engine for 1.10 or 1.11.
> 
> Well, that makes ambitions for multithreading cvxcore pretty low I guess. Will it be removed completely or just deprecated?

I think it makes sense to remove it completely at some point. Since it would be quite nasty to keep both a C++ and C bindings to build on it. But maybe as a first step we should only deprecate it.

Maybe there could be another way to handle this, where cvxcore is not imported by default?

PTNobel:
> Maybe there could be another way to handle this, where cvxcore is not imported by default?

Personally, I think we should just not import it on free threaded at all. It's already optional and that's what we do on Pyodide

stertingen:
> > Maybe there could be another way to handle this, where cvxcore is not imported by default?
> 
> Personally, I think we should just not import it on free threaded at all. It's already optional and that's what we do on Pyodide

Please take a look at #3522. Tests on my local machine did not result in the GIL being enabled while importing `cvxpy`, which would enable using the scipy backend for my use case.

Transurgeon:
I'll close this PR since we merged #3522

PR #3508, opened 2026-09-02
CI: isolate OR-Tools tests from HiGHS to prevent native symbol conflicts

nej1gotnochill:
`OR Tools >= 9.14` bundles its own HiGHS library , which can conflict with `highspy` when both are loaded in the same process. This separates the GLOP/PDLP tests into a separate pytest invocation so the OR Tools and HiGHS native libraries are not loaded in the same process.

Related: google/or-tools#5246

Feedback on the approach is highly appreciated.

### DIFF
--- a/.github/workflows/test_optional_solvers.yml
+++ b/.github/workflows/test_optional_solvers.yml
@@ -65,10 +65,10 @@ jobs:
       # symbol conflicts (google/or-tools#5246). Run OR-Tools tests
       # (GLOP/PDLP) in a separate process to avoid this.
       - name: Run test_conic_solvers
-        run: uv run pytest -rs -m "not knitro" -k "not (GLOP or PDLP)" cvxpy/tests/test_conic_solvers.py
+        run: uv run pytest -rs -m "not knitro and not ortools" cvxpy/tests/test_conic_solvers.py
       - name: Run test_conic_solvers (OR-Tools)
         if: always()
-        run: uv run pytest -rs -k "GLOP or PDLP" cvxpy/tests/test_conic_solvers.py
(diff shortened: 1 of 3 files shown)


PTNobel:
Please sign the CLA so I can review the code!

nej1gotnochill:
I have signed the CLA successfully , I hope you can review the code now.

PTNobel:
Looks like CI is failing with the same error, do you need to pytest.mark the OR Tools tests?

nej1gotnochill:
Yes that could be the issue I feel, I'll check the failing step and investigate whether explicitly marking the OR tools is the right way to separate them from HiGHS. If so I will update the markers and workflow accordingly and rerun CI. Thank you for pointing that out !

nej1gotnochill:
Updated this based on your suggestion , I have now added an `ortools` pytest marker to the GLOP / PDLP tests and separated them from the HiGHS tests in CI following the KNITRO pattern.

PTNobel:
Seems that test_offset_in_opt_val[PDLP] , test_offset_in_opt_val[GLOP] , TestAllSolvers.test_installed_solvers 

Are all still failing

nej1gotnochill:
Ah , got it. Thank you for pointing that out ! I will trace these failures and check how these tests are actually interacting with the OR Tools including the `test_installed_solvers` case. I will look into this further and then make the necessary changes based on what I find.

nej1gotnochill:
I have now pushed the additional isolation fixed based on your feedback. This now also marks the standalone GLOP/PDLP cases and keeps them out of TestAllSolvers in the main process.

Feedback on whether this approach fits the intended CI structure is very welcome, happy to adjust anything that could be cleaner.

PTNobel (approved):

PR #3522, opened 2026-09-07
cvxcore: Build in python namespace and remove hack to import _cxvcore

stertingen:
## Description
Build cvxcore binary in the correct namespace and remove the cvxcore import hack. SWIG assumes the binary to be in the `cvxcore.python` namespace, but it is built as a global module. (see also: https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#setuptools.Extension)
Removing that hack fixes the unconditional include of `_cvxcore` which enables the GIL when importing `cvxpy`.
Issue link (if applicable): #3110

## Type of change
- [x] Bug fix

- [ ] Write unittests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,7 +146,7 @@ jobs:
         if: ${{matrix.build-cvxpy-base && env.PIP_INSTALL == 'True'}}
         run: |
           pip install . pytest cplex
-          pytest cvxpy/tests/test_conic_solvers.py -k 'TestCPLEX'
+          pytest --pyargs cvxpy.tests.test_conic_solvers -k 'TestCPLEX'
           # ensure a fresh state for cibuildwheel
           rm -r build
 
--- a/cvxpy/cvxcore/python/__init__.py
+++ b/cvxpy/cvxcore/python/__init__.py
@@ -4,10 +4,3 @@ try:
 	import coptpy
 except ImportError:
 	pass
-
-# TODO(akshayka): This is a hack; the swig-auto-generated cvxcore.py
-# tries to import cvxcore as `from . import _cvxcore`
-try:
-	import _cvxcore
-except ImportError:
-	pass
--- a/setup/extensions.py
+++ b/setup/extensions.py
@@ -37,7 +37,7 @@ compiler_args = [
 #
 # TODO wheels should be compiled with openmp ...
 cvxcore = Extension(
-    '_cvxcore',
+    'cvxpy.cvxcore.python._cvxcore',
     sources=['cvxpy/cvxcore/src/cvxcore.cpp',
              'cvxpy/cvxcore/src/LinOpOperations.cpp',
              'cvxpy/cvxcore/src/Utils.cpp',


stertingen:
The CI unit test failed because `pytest` was executed with the local test file while the `_cvxcore` binary was only available for the package which was built and installed using `pip install .`.
Running `pytest --pyargs` lets pytest using the python module from the `cvxpy` package which contains the `_cvxcore` binary.

I suspect that the CI script has been wrong as it was testing the local python scripts instead of the installed ones.

Transurgeon:
> The CI unit test failed because `pytest` was executed with the local test file while the `_cvxcore` binary was only available for the package which was built and installed using `pip install .`. Running `pytest --pyargs` lets pytest using the python module from the `cvxpy` package which contains the `_cvxcore` binary.
> 
> I suspect that the CI script has been wrong as it was testing the local python scripts instead of the installed ones.

wow, really nice find! and thanks a bunch for your PR.
I think it's looking good and is probably the simplest way we can fix the issue.

Also just out of curiosity, for what reasons are you using the free-threaded builds of CVXPY (or I guess in the process of using it?). We spent some time thinking about providing these wheels and you are the first one I hear that is actively needing it. Thanks a bunch!

Transurgeon (approved):

stertingen:
> Also just out of curiosity, for what reasons are you using the free-threaded builds of CVXPY (or I guess in the process of using it?). We spent some time thinking about providing these wheels and you are the first one I hear that is actively needing it. Thanks a bunch!

I'm doing my PhD with convex optimization. I'm solving lots of problem instances (~3000) in parallel, currently using joblib/loky, sometimes running out of memory on my local machine. Using Freethreading would theoreticaly allow me to use all cores while sharing memory if I got it correctly :-)

Transurgeon:
> > Also just out of curiosity, for what reasons are you using the free-threaded builds of CVXPY (or I guess in the process of using it?). We spent some time thinking about providing these wheels and you are the first one I hear that is actively needing it. Thanks a bunch!
> 
> I'm doing my PhD with convex optimization. I'm solving lots of problem instances (~3000) in parallel, currently using joblib/loky, sometimes running out of memory on my local machine. Using Freethreading would theoreticaly allow me to use all cores while sharing memory if I got it correctly :-)

oh that's interesting! are you solving similar problems that could be parametrized? Im curious if this could fix your memory issue potentially? I presume having one problem instance will take much less memory.

.github/workflows/build.yml
```diff
@@ -146,7 +146,7 @@ jobs:
         if: ${{matrix.build-cvxpy-base && env.PIP_INSTALL == 'True'}}
         run: |
           pip install . pytest cplex
-          pytest cvxpy/tests/test_conic_solvers.py -k 'TestCPLEX'
+          pytest --pyargs cvxpy.tests.test_conic_solvers -k 'TestCPLEX'
```
Transurgeon, line 149: also I am wondering why we are testing cplex here? That seems a bit random?

PR #3524, opened 2026-09-08
Reassociate matmul chains that end in a constant factor

Transurgeon:
Each engine node's Jacobian carries one row per output, so a chain like (A @ X) @ C builds its Jacobian over the wide A @ X intermediate before contracting -- the blow-up behind issue #2205. Pushing the trailing constant toward the leaves lets the constants fold together and keeps the intermediates narrow:

  constant @ c   -> Constant(value)
  (A @ B) @ c    -> A @ (B @ c)
  (-E) @ c       -> -(E @ c)
  (E1 + E2) @ c  -> E1 @ c + E2 @ c   (vector c only)

Only parameter-free constants fold, so a parametric tail is never frozen into a Constant. Reassociation also bails out unless the left factor is 2-D: a 1-D operand contracts to an inner product under numpy's matmul rules, and (a @ b) @ c != a @ (b @ c) across that collapse -- the rewritten factor can even be shape-invalid. General matrix-chain reordering is deliberately not attempted.

_normalize_matmul is a pure expression rewrite, so the tests check the decision directly: the rewrite fires and preserves .value on each rule, is a no-op on both 1-D-collapse shapes and on a parametric tail, and DerivativeChecker confirms the engine differentiates the rewritten tree. Verified that disabling the rewrite and dropping the 1-D guard each fail the corresponding test.

## Description
Issue link (if applicable):

## Type of change
- [ ] Bug fix

- [ ] Write unittests.

### DIFF
--- a/cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
+++ b/cvxpy/reductions/solvers/nlp_solvers/diff_engine/converters.py
@@ -15,12 +15,16 @@ limitations under the License.
 
 Main entry point for converting CVXPY expressions to C diff engine expressions.
 """
+import operator
+from functools import reduce
+
 import numpy as np
 from scipy import sparse
 from sparsediffpy import _sparsediffengine as _diffengine
 
 import cvxpy as cp
 import cvxpy.settings as s
+from cvxpy.expressions.constants import Constant
 from cvxpy.reductions.solvers.nlp_solvers.diff_engine.helpers import (
     make_dense_left_matmul,
     make_dense_right_matmul,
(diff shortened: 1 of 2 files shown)


dance858:
Very cool optimization!

Just a random thought:

Do you think the matmul reassociation stuff should live in its own file? It's a pure expression to expression rewrite, while the rest of converters.py is about building C engine nodes, so it feels like a different layer

rileyjmurray:
Maybe put the reduction in a new folder like cvxpy/reductions/algebra/, and
include the reduction in the solving chain whenever the SparseDiffEngine
backend is used.

PR #3525, opened 2026-09-09
Cone restructuring: apply it as a chain reduction, and as a permutation instead of a matrix product

Transurgeon:
## Description

Two related changes to cone restructuring: make it an explicit chain step, and stop building the restructuring matrix at all.

### 1. Formatting becomes a chain reduction

`ConicSolver.format_constraints` reorders stuffed constraint rows into the layout a solver's cone API expects. Until now every solver interface called it itself, guarded on `problem.formatted` — 11 copies of the same two lines that every new interface had to remember, and something `affine2direct` could only document in prose that `Dualize.apply` depended on.

`ConeFormat` does it once as a chain step, so the ordering is a chain fact rather than a per-interface convention. All 11 guards are removed and an already-formatted program becomes a stated **precondition of `ConicSolver.apply`**. Which formatting to apply is dispatched on the program via `ParamConeProg.format_for(solver)`, so a program that knows a cheaper way to restructure itself can say so rather than be rebuilt as a stock one.

Behaviour is unchanged: seven interfaces passed `exp_cone_order=None` explicitly rather than `self.EXP_CONE_ORDER`, and all seven inherit `ConicSolver`'s default of `None`, so routing them through `format_for(solver)` formats them identically.

`ExtractDirectCones` still formats on its own. It has to — it scans rows in per-cone order to find its cones, which is earlier than this step runs — and it leaves `formatted` set, so `ConeFormat` is then a no-op.

### 2. Restructuring becomes a permutation, not a matrix product

R is **always a signed permutation**: every block is either ±I (Zero contributes −I; NonNeg, PSD and SvecPSD contribute I) or a pure interleaving of one constraint's arguments (SOC and the exponential/power cones). This holds for every cone family, including `PowConeND` and a reversed `exp_cone_order`.

So R never has to be built, let alone multiplied by. `restruct_permutation` returns `(new_row, sign)` indexed by the old row, and `format_constraints` applies it by shifting the tensor's stored row indices in place:

```python
perm = restruct_permutation(problem.constraints, exp_cone_order)
A = problem.A if perm is None else _permute_rows(problem.A, *perm)
return _reformatted(problem, A)
```

That collapses two code paths into one. The previous fast path caught Zero/NonNeg/PSD problems, where R is diagonal, and skipped a `reshape → block-diag-matmul → reshape` round trip that also left `A` as a COO matrix for every downstream consumer to pay for. Everything with an SOC, exponential or power cone still took the slow path. Now nothing does, so the saving extends to the cone types the fast path missed, and the diagonal special case disappears rather than being maintained alongside.

Expressing the remap as a *shift* rather than recomputing `new_row[i] + m*j` keeps the arithmetic in the index dtype, which scipy requires to match `indptr`; computing it directly needs int64 plus a cast back and measured ~60% slower on an LP. `R = I` returns the tensor untouched, without even a copy.

This deletes `diagonal_restruct_signs`, `build_restruct_blocks`, `build_restruct_mat_sparse`, and the `LinearOperator` / `IdentityOperator` / `NegativeIdentityOperator` / `as_block_diag_linear_operator` wrappers that existed only to block-diagonalize R. `conic_solver.py` is a net **−36 lines** despite gaining seven named helpers. `get_spacing_matrix` stays: it is a public `ConicSolver` classmethod an external plugin may call, so removing it belongs in its own change — and `_spacing_rows` reuses its exact index expression, so the two cannot drift.

## ⚠️ Breaking change for downstream code that slices `chain.reductions`

Adding a reduction shifts positions in the chain, and code that removes the solver by slicing now also keeps `ConeFormat`, so it receives a **formatted** program (Zero rows sign-flipped, SOC rows interleaved) where it used to get the stuffed one.

This is why `test_gcsopt` is red. gcsopt does exactly that ([`programs.py:222`](https://github.com/TobiaMarcucci/gcsopt/blob/main/gcsopt/programs.py#L222)):

```python
chain.reductions = chain.reductions[:-1]
cp_conic = chain.apply(cp_convex)[0]
```

The fix is one line, mirroring what `perspective_canon` does in-tree:

```python
chain.reductions = [r for r in chain.reductions
                    if not isinstance(r, (Solver, ConeFormat))]
```

**This PR should not merge before gcsopt is updated.** In-tree, `suppfunc.py` had the same fragility (`invdata[-2]`) and is fixed here to look the stuffing step up by type instead.

## Benchmarks

Cold `get_problem_data`, BLAS pinned to one thread, best of 3:

| | master | this PR |
|---|---|---|
| `CVaRBenchmark` (LP, R diagonal) | 7.18s | **4.26s** |
| `SDPSegfault1132Benchmark` (SOC, R permutation) | 9.57s | **8.03s** |

`format_constraints` in isolation, against the intermediate fast-path implementation so the two halves are separable:

| | fast path only | permutation |
|---|---|---|
| LP, 2.2M nnz (R diagonal) | 6.2ms | 6.3ms |
| SOC, 0.6M nnz (R permutation) | 8.8ms | **2.4ms** |

i.e. the cone types the old fast path missed now get the win, and the ones it caught are unchanged. `SDPSegfault1132` was explicitly *unchanged* by the fast path alone; it now improves.

## Tests

New `cvxpy/tests/test_cone_restructuring.py` (+283):

- `restruct_permutation` is a genuine permutation with ±1 signs, per cone family, and only Zero cones are negated.
- Hand-computed expected permutations for SOC and ExpCone, including a reversed `exp_cone_order`.
- Equivalence: remapping the tensor equals applying an independently materialized `R` to the concrete `[A | b]`. Materializing R now lives only in the tests, since no production code wants it.
- The remap preserves the tensor's index dtype (built explicitly against an int32 tensor — this was a real bug, caught by `test_cvxcore_issues.py`).
- `format_constraints` does not mutate its input, and does no work at all when R = I.
- `ConeFormat` sits before the solver, is absent for QP chains, dispatches through `format_for`, and is load-bearing: handed an unformatted program, a solver interface now emits the stuffed row order rather than silently repairing it.
- No conic interface re-derives the layout (guards against the 11 guards creeping back).

Equivalence against the deleted implementation was checked before removing it — byte-identical R for every cone family and both `exp_cone_order` conventions, and identical stuffed `A` for parameter-free and parametric problems.

## Type of change
- [ ] Bug fix
- [x] Other (performance, refactor)

- [x] Write unittests.

### DIFF
--- a/cvxpy/reductions/__init__.py
+++ b/cvxpy/reductions/__init__.py
@@ -19,6 +19,7 @@ from cvxpy.reductions.solution import Solution
 
 from cvxpy.reductions.chain import Chain
 from cvxpy.reductions.complex2real.complex2real import Complex2Real
+from cvxpy.reductions.cone_format import ConeFormat
 from cvxpy.reductions.cvx_attr2constr import CvxAttr2Constr
 from cvxpy.reductions.dcp2cone.cone_matrix_stuffing import ConeMatrixStuffing
 from cvxpy.reductions.dcp2cone.dcp2cone import Dcp2Cone
--- a/cvxpy/reductions/cone2cone/affine2direct.py
+++ b/cvxpy/reductions/cone2cone/affine2direct.py
@@ -55,8 +55,9 @@ class Dualize:
     Usage
     -----
     Dualize applies to ParamConeProg problems. It accesses (P-Opt) data by calling
-    ``c, d, A, b = problem.apply_parameters()``. It assumes the solver interface
-    has already executed its ``format_constraints`` function on the ParamConeProg problem.
+    ``c, d, A, b = problem.apply_parameters()``. It assumes the rows are already in the
+    solver's cone layout, which the ``ConeFormat`` reduction establishes before the solver
+    runs.
 
     A solver interface is responsible for calling both Dualize.apply and Dualize.invert.
     The call to Dualize.apply should be one of the first things that happens, and the
@@ -79,8 +80,8 @@ class Dualize:
     The problem has no integer or boolean constraints. This is necessary because strong
     duality does not hold for problems with discrete constraints.
 
-    Dualize.apply assumes "SOLVER.format_constraints()" has already been called. This
-    assumption allows flexibility in how a solver interface chooses to vectorize a
+    Dualize.apply assumes the ``ConeFormat`` reduction has already applied the solver's
+    cone layout. That assumption allows flexibility in how a solver chooses to vectorize a
     feasible set (e.g. how to order conic constraints, or how to vectorize the PSD cone).
 
     Additional notes
@@ -88,8 +89,8 @@ class Dualize:
 
     Dualize.invert is written in a way which is agnostic to how a solver formats constraints,
     but it also imposes specific requirements on the input. Providing correct input to
-    Dualize.invert requires consideration to the effect of ``SOLVER.format_constraints`` and
-    the output of ``problem.apply_parameters``.
+    Dualize.invert requires consideration to the effect of the solver's cone layout, applied
+    by ``ConeFormat``, and the output of ``problem.apply_parameters``.
     """
 
     @staticmethod
--- a/cvxpy/reductions/cone_format.py
+++ b/cvxpy/reductions/cone_format.py
@@ -19,13 +19,16 @@ from cvxpy.reductions.reduction import Reduction
 class ConeFormat(Reduction):
     """Apply a conic solver's cone row layout as an explicit chain step.
 
-    Solver interfaces have historically each opened by calling
-    ``format_constraints`` themselves, guarded on ``problem.formatted``. Doing
-    it here instead makes the ordering a chain fact rather than a convention
-    every interface has to remember -- ``Dualize`` in particular assumes
-    formatting has already happened. It also lets a program that owns a
-    re-extractable form restructure itself instead of being rebuilt as an
-    ordinary program, via ``ParamConeProg.format_for``.
+    This is the only place cone formatting happens. Interfaces used to each
+    open by calling ``format_constraints`` themselves, guarded on
+    ``problem.formatted`` -- eleven copies of a step every new interface had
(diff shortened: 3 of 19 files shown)


Transurgeon:
> Benchmarks that have improved:
> 
> 3.94±0.04s	3.93±0.01s	1	simple_QP_benchmarks.UnconstrainedQP.time_compile_problem
> 1.49±0.01s	1.49±0.01s	1	tv_inpainting.TvInpainting.time_compile_problem
> 2.87±0.01s	2.83±0.01s	0.99	gini_portfolio.Cajas.time_compile_problem
> 2.77±0.02s	2.75±0.02s	0.99	simple_QP_benchmarks.ParametrizedQPBenchmark.time_compile_problem
> 4.58±0.02s	4.50±0.01s	0.98	optimal_advertising.OptimalAdvertising.time_compile_problem
> 1.48±0.01s	1.42±0.07s	0.96	matrix_stuffing.ParamConeMatrixStuffing.time_compile_problem
> 1.65±0.03s	1.53±0.01s	0.93	semidefinite_programming.SemidefiniteProgramming.time_compile_problem
> 1.91±0.02s	1.65±0.01s	0.87	finance.FactorCovarianceModel.time_compile_problem
> 14.4±0.2s	12.1±0.06s	0.84	finance.CVaRBenchmark.time_compile_problem
> 10.6±0.01s	8.80±0.07s	0.83	simple_LP_benchmarks.SimpleLPBenchmark.time_compile_problem
> 1.85±0.03s	1.51±0.01s	0.82	simple_LP_benchmarks.SimpleScalarParametrizedLPBenchmark.time_compile_problem
> 766±8ms	615±4ms	0.8	simple_LP_benchmarks.SimpleFullyParametrizedLPBenchmark.time_compile_problem

wow the benchmarks actually already improve :).

SteveDiamond (approved):
This is great!

Transurgeon:
I opened two issues on GSCOPT, merging now.

PR #3534, opened 2026-09-16
Fix HiGHS reporting a solution when a limit is reached before any solution exists

Vittal-Mukunda:
## Description

When HiGHS stops on a time, node, or iteration limit before it has a primal solution, the conic interface still maps that to `USER_LIMIT`. That status is in `SOLUTION_PRESENT`, so `invert()` copies `solution.col_value` into the variables. HiGHS already reports that there is no solution (`solution.value_valid == False`, `info.primal_solution_status == 0`); neither flag is checked. The copied values are zeros and can violate the constraints.

HiGHS is the default MILP solver (#2988), so this happens without passing `solver=`:

```python
import numpy as np
import cvxpy as cp

rng = np.random.default_rng(0)
A = rng.integers(0, 100, size=(3, 20))
d = A.sum(axis=1) // 2   # no binary x satisfies A @ x == d
x = cp.Variable(20, boolean=True)
prob = cp.Problem(cp.Minimize(0), [A @ x == d])
prob.solve(time_limit=0.0)
print(prob.status)                    # user_limit
print(np.abs(A @ x.value - d).max())  # 532.0
print(prob.solver_stats.extra_stats.primal_solution_status)  # 0
```

`mip_max_nodes=1` does the same.

If `USER_LIMIT` is set and `value_valid` is false, `invert()` now reports `INFEASIBLE_INACCURATE` instead. That is how `gurobi_conif.py`, `copt_conif.py`, and `cuopt_conif.py` already treat a limit with no solution (#1270, #3443). After the change, the snippet above has `status == "infeasible_inaccurate"` and `x.value is None`. `prob.value` is `inf`, which is what `failure_solution` uses for that status. HiGHS extra stats are still on `solver_stats`.

Results where HiGHS does have a point (`value_valid` True) are unchanged. That includes a MIP limit with an incumbent, and an LP `time_limit=0` that returns an infeasible iterate. The QP interface is not modified: in the limit cases I tried, HiGHS's QP solver always returned primal values.

Issue link (if applicable): none. Same situation as #1177 / #1270 (Gurobi) and #3443 (cuOpt).

## Type of change
- [x] Bug fix

- [x] Add our license to new files. (no new files)
- [x] Check that your code adheres to our coding style. (ruff)
- [x] Write unittests. (`TestHIGHS.test_highs_mip_limit_no_incumbent`; fails on master)


PTNobel:
Seems fine code change wise. I'm confused about the API: 

1. Why is user_limit in solution present in the first place? 
2. This would still give the most recent HiGHS iterate right? Isn't it expected behavior if you set a limit and it gets hit that your solution may not be feasible?

Vittal-Mukunda:
The PR makes CVXPY check HiGHS's own `value_valid` flag. On master, when HiGHS stops on a limit and says its values aren't valid, CVXPY still returns them under `user_limit`, so they look like a real result.

1. `USER_LIMIT` in `SOLUTION_PRESENT` is fine. It's what lets you keep what the solver found when it stops early, like a MIP incumbent. The PR doesn't touch it.

2. Only when HiGHS marks the iterate as valid. Then you still get it under `user_limit`, and yes, it can be infeasible. The PR only changes results HiGHS marks invalid. I found two cases:
   - A MIP stops before finding any solution. The values are just zeros (the repro).
   - An LP stops on an iteration limit after presolve reduced it. The values are sometimes the latest iterate and sometimes wrong. With a constraint `x[:10] == 0.01`, those entries didn't come back as 0.01.

   Both now return `INFEASIBLE_INACCURATE` with `x.value = None`.

Issue #1743, opened 2022-04-10
Install from source using `pip install -e .` fails install numpy on windows


yasirroni:
**Describe the bug**
Following CVXPY [docs](https://www.cvxpy.org/install/index.html#install-from-source), I can't simply use `pip install -e`, even after installing `Desktop development with C++` on `Visual Studio Build Tools`.

**To Reproduce**

`pip install -e .`

**Expected behavior**
It should be able to just install

**Output**

```
Encountered error while trying to install package.

      numpy

      note: This is an issue with the package mentioned above, not pip.
      hint: See above for output from the failure.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× pip subprocess to install build dependencies did not run successfully.
│ exit code: 1
╰─> See above for output.
```

**Version**
 - OS: WIndows
 - CVXPY Version: '1.3.0.dev0+358.75cc007'

yasirroni:
Following `scikit-learn` [tutorial](https://scikit-learn.org/stable/developers/advanced_installation.html#compiler-windows), we should use:

```
SET DISTUTILS_USE_SDK=1
"C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x64
```

The exact path to call `vcvarsall.bat` and the trailing `x64` is depends on your system. See above tutorial for more details.

yasirroni:
I will close this issue. Hope others will find this helpful.

Re-open the issue if someone thinks it needed to be opened.

yasirroni:
Hope the docs is updated. :)

rileyjmurray:
Thanks for reporting this, @yasirroni! I'll re-open this so we remember to update the web documentation.

PR #3535, opened 2026-09-17
Document Windows build tools workaround for source install

Specter842:
## Description
Adds a note to the "Install from source" tab covering the Windows build failure reported in #1743: `pip install -e .` can fail compiling a dependency (e.g. NumPy) unless the Visual Studio Build Tools environment is activated first (`DISTUTILS_USE_SDK=1` + `vcvarsall.bat x64`), per the workaround confirmed in the issue thread.

Issue link: #1743

## Type of change
- [ ] Bug fix

## Contribution checklist
- [x] Add our license to new files. (no new files)
- [ ] Write unittests. (docs-only change)
- [ ] Run the unittests and check that they're passing. (docs-only change)
- [ ] Run the benchmarks to make sure your change doesn't introduce a regression. (docs-only change)

AI assistance was used for drafting this doc change and PR description.


PTNobel:
I'm pretty hesitant on this, the SciKit Learn tutorial this fix came from is gone and I don't see this advice in their new one. I have no idea if this is still the right approach and I certainly won't know if it changes.

Also, people have agents now to set up build environments who will know far more about Windows than we do.

I think my preference would be to find and link another libraries dev guide. Especially, since this issue in this case was a NumPy build, we could maybe link their docs.

